PR #10, opened 2026-07-24
Gmail mock fidelity: relative-date search operators + attachment size contract

khj809:
Three reported Gmail mock fidelity gaps, reviewed and fixed.

## 1. Batch already honors sub-request query params — test hardened
Reproducing the real `google-api-python-client` `BatchHttpRequest` serialization against both `/batch` and `/batch/gmail/v1` showed the mock **already** forwards each sub-request's query string (`?format=full`) to the dispatched handler. The gap didn't reproduce at HEAD.

The real issue was test coverage: `test_google_batch_dispatches_subrequests` used `format=minimal`, which passes **whether or not** the query is honored (a dropped query defaults to `full`, and the id still appears in the body). It could never have caught a query-forwarding regression.

- Switched that test to `format=full`.
- Added `test_google_batch_honors_subrequest_query_params` (parametrized over `/batch` and `/batch/gmail/v1`): asserts `full` → payload present **and** `minimal` → payload absent. Now a regression in query forwarding fails loudly.

## 2. `newer_than:` / `older_than:` no longer zero out the query
These relative-date operators weren't recognized, so they leaked into the FTS free-text term, matched nothing, and zeroed the **entire** result set — including any real free-text term beside them.

Fix: register them as operators and fold each into an absolute `after:`/`before:` bound anchored to *now* (`newer_than:5d` → `ts >= now-5d`; `older_than:5d` → `ts < now-5d`), reusing the existing SQL range push-down and `_gmail_op_match`. Units `d`/`m`/`y` (month/year approximated in days — documented).

## 3. Attachment `body.size` now matches the served bytes
`messages.get` reported a part's `body.size` from the corpus-declared `size` (e.g. 2048) while `attachments.get` served `len(content)` (e.g. 21) — so the two disagreed and a client that stats an attachment from part metadata (real Gmail's contract) got the wrong number.

Fix: `_att_content` is the single source of served bytes; its length feeds the part `body.size`, the `attachments.get` size, and the `raw` MIME leaf alike.

## Tests
Full suite: **306 passed, 2 skipped**. New/changed tests placed by layer — search-operator logic in `test_search.py`, endpoint contracts in `test_endpoints.py`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,7 +257,7 @@ One runnable script per source (GitHub, S3, Confluence, Jira, Slack, Notion, Gma
 | Prefix | Service | Endpoints |
 |---|---|---|
 | `/slack/api` | Slack | `conversations.list`, `conversations.history` (+`oldest`/`latest`/`inclusive`), `conversations.replies`, `conversations.members`, `users.list`, `users.info`, `auth.test`, `api.test` (auth-free connectivity check), `search.messages` |
-| `/gmail/v1` | Gmail | `users/{u}/messages` (+`q`: free text / `from:` `subject:` `after:` `before:` `label:` `has:attachment`), `messages/{id}` (`format=full\|metadata\|minimal`), `messages/{id}/attachments/{id}`, `threads` (+`q`), `threads/{id}`, `labels`, `profile` |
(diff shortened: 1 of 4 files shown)


selloriwoo (approved):
I retested the part that was causing the issue, and it’s working fine now. Thank you for the quick feedback.

PR #12, opened 2026-07-24
Gmail mock: order messages.list newest-first by internalDate (#11)

### LINKED ISSUE #11: Gmail mock: messages.list is not ordered by internalDate descending
Real Gmail returns `users.messages.list` results ordered by `internalDate`
descending (newest first). The mock returns them in an apparently arbitrary
order — requesting the "newest" 300 with `maxResults` paging yields messages
scattered across 2025–2029.

**Repro**
GET /gmail/v1/users/me/messages?maxResults=500 
→ inspect the internalDate of the first page: not monotonically decreasing.

**Why it matters**
Clients legitimately rely on the ordering:
- a capped listing ("newest N") silently becomes "random N" on the mock;
- our mailbox mirror sync streams in list order and finalizes a month once
  the stream passes it — correct against real Gmail, but the guarantee can't
  hold against the mock mid-sync (final state is still correct, we merge).

**Suggested fix**
Sort list results by internalDate descending (and keep pagination stable
across pages).

khj809:
Fixes #11.

## Problem (validated)
Real Gmail returns `users.messages.list` ordered by `internalDate` descending. The mock listed via `list_documents`' `ORDER BY doc_id` — hash order, effectively **random by date**. Confirmed on the live corpus: the first page of a `maxResults` listing was scattered across years, so a capped "newest N" crawl silently became "random N", and a mirror-sync that streams in list order couldn't rely on the ordering mid-sync.

## Fix
- **`list_gmail_in_range`**: add `doc_id` as a tie-break after `created_ts DESC`, giving a stable **total** order so offset pagination can't dupe/skip rows across pages.
- **`gmail_messages_list` (no-`q` path)**: list via `list_gmail_in_range` with an open-ended range (= whole mailbox) instead of `list_documents`, so the plain listing is newest-first by `internalDate`.
- **`_gmail_query` (no-free-text branch)**: use `list_gmail_in_range` for the open-ended case too (not only when `after:`/`before:` pins a range), so operator-only listings like `label:INBOX` are date-ordered as well. Free-text `q` stays FTS-relevance-ranked (as the mock intends for text search).

Threads (`threads.list`) are out of scope — a different scoping path (`author_email`), not mentioned in the issue.

## Tests
- `test_gmail_messages_list_ordered_by_internaldate_desc` — endpoint order equals `created_ts DESC, doc_id`, and `internalDate` is monotonically non-increasing on the page. (Fails on the old `doc_id` order.)
- `test_gmail_messages_list_pagination_stable_and_ordered` — pages are a stable partition of that order (no dupes/skips, page 2 continues from page 1's tail).

Full suite: **308 passed, 2 skipped**.

### DIFF
--- a/app/routers/google.py
+++ b/app/routers/google.py
@@ -378,11 +378,10 @@ def _gmail_query(conn, mailbox, ids, q: str) -> list:
         # _gmail_op_match; the remaining ops still filter the (now small) candidate set below.
         lo = max((d for v in ops.get("after", []) if (d := _gmail_date(v)) is not None), default=None)
         hi = min((d for v in ops.get("before", []) if (d := _gmail_date(v)) is not None), default=None)
-        if lo is not None or hi is not None:
-            cand = store.list_gmail_in_range(conn, mailbox, lo, hi, ids, limit=100_000)
-        else:
-            cand = store.list_documents(conn, "gmail", container=mailbox, visible_ids=ids,
-                                        limit=100_000)
+        # list_gmail_in_range for BOTH the date-pinned and the open-ended case (lo=hi=None): its
+        # created_ts DESC, doc_id order is the newest-first listing real Gmail returns — the plain
+        # list_documents path ordered by doc_id (hash), scattering the listing by date.
+        cand = store.list_gmail_in_range(conn, mailbox, lo, hi, ids, limit=100_000)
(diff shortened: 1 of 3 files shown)


selloriwoo:
I actually tried it out, and it works just fine.
Thank you

selloriwoo (approved):

Issue #2, opened 2026-07-20
Example: deterministic RAG eval — retrieval + ACL scored reproducibly, offline, in CI


khj809:
An example that uses Backlot as an evaluation substrate rather than a thing to read from: retrieve over a served corpus, score the retrieval against a gold set the corpus itself declares, and get the same numbers every run — offline, with no API key, as a job that can block a PR.

## Why

Every consumer example so far — `using-official-sdk`, `using-llamaindex-readers`, `using-mcp-with-agents`, `using-mirage` — shows a client **reading** the mock. They prove the surfaces work. None of them uses the property that makes this server different from a recorded fixture: what it serves is *derived*, so a query's answer set is the same on the first run, the tenth, page three, and a fresh import of the same corpus.

That is the missing piece for retrieval evaluation. A RAG eval against live SaaS measures the vendor's index and last week's data drift along with your retriever; against hand-written fixtures it measures whether your fixtures agree with your assumptions. Here the gold set is a fact about the corpus you loaded, so a score that moves means the retriever moved.

## Scope

**In:** retrieval scored against known-relevant documents, with non-judge metrics, reproducibly, in CI.

**Out**, deliberately — each of these is a separate topic and each one would blur the claim this example exists to make:

- LLM-judge scoring (answer faithfulness, relevance-as-graded-by-a-model). Not reproducible in the sense this example is selling; see the caveat below.
- Tracing and observability. Diffing spans across two runs is a different demonstration with a different dependency footprint.
- A tour of eval frameworks. One path, no framework required to run it.
- Grading the public dataset (#19). That harness runs over a downloaded multi-source corpus and is its own deliverable in its own repo; this example runs on a corpus small enough to read in one sitting, with nothing to fetch.

## What to build

`examples/deterministic-rag-eval/`:

1. **A corpus that carries its own answer key.** A small BYO-JSONL corpus written for the eval, passed to `backlot.mock_server(records=...)`, with a companion file of queries → the documents that answer them. It ships its own corpus rather than leaning on the bundled one on purpose: the bundled corpus is a demo, its contents are free to change, and an eval that hardcodes ids out of it is asserting on something no one promised to keep.
2. **A retrieval step through a real client.** Crawl a couple of sources with an official SDK or a reader, index locally (BM25 or any local embedding — no hosted model in the required path), then query.
[...]

PR #28, opened 2026-07-28
Drive: close the 7 measured divergences from real Google Drive (#23)

### LINKED ISSUE #23: Drive API: 7 divergences from real Google Drive (measured side-by-side)
# Drive API: 7 divergences from real Google Drive (measured side-by-side)

While building a Google Drive client against this mock we diffed it against real
Google Drive — identical requests, equivalent credentials, one pass on
2026-07-28 (`mockserver/drive/v3` vs
`https://www.googleapis.com/drive/v3`). Seven behaviours differ. Filing them
together because they came out of one measurement, share the repro setup.

The axis that matters most is **"can a consumer notice?"** Five of these return
`200` with wrong or unfiltered data, so a client can't tell anything is off; it
builds on the wrong assumption and only finds out against production Google.

| # | Divergence | Silent? | Root area |
|---|---|---|---|
| 1 | `q: sharedWithMe` clause ignored | ● yes | `q` parsing |
| 2 | `orderBy` accepted, never applied | ● yes | list params |
| 3 | invalid `fields` mask → `200` instead of `400` | ● yes | projection |
| 4 | `files.get` ignores `fields` (differs from `files.list`) | ● yes | projection |
| 5 | folders invisible to `mimeType='…folder'` (+ skip projection) | ● yes | synthesized folders |
| 6 | ERB rows all `application/octet-stream` | ○ visible | corpus import |
| 7 | native Docs/Sheets omit `size` | ○ visible | row building |

---

## 1. `files.list`: the `sharedWithMe` query clause is silently ignored

**What happens** — `sharedWithMe=true` and `sharedWithMe=false` return
byte-identical result sets, i.e. the clause has no effect and every request sees
the caller's whole visible corpus.

```bash
G "/files?q=sharedWithMe%3Dtrue%20and%20trashed%3Dfalse&pageSize=5&fields=files(id)"
G "/files?q=sharedWithMe%3Dfalse%20and%20trashed%3Dfalse&pageSize=5&fields=files(id)"
```

| | mock | real Drive |
|---|---|---|
| `sharedWithMe=true` | same 5 ids | 123 items |
| `sharedWithMe=false` | **same 5 ids** | 0 items |

**Expected** — `true` returns only items shared with the caller by someone else;
`false` excludes them. The two result sets must differ.

**Likely cause** — `_drive_q_match` (`app/routers/google.py:557`) handles
`trashed`, `'x' in parents`, `mimeType`, `name contains`, `modifiedTime >` and
`'x' in owners`; `sharedWithMe` isn't among them, and unrecognized clauses are
ignored by design (per the function's own docstring).

**Why it matters** — Drive's "Shared with me" is a first-class section of the
product: items shared with you carry no `parents`, so `sharedWithMe=true` is the
*only* way to enumerate them. A client that mirrors that section (we do) shows
the entire corpus instead — 3,955 entries where the real account has 123. The
failure is silent: HTTP 200, plausible-looking data.

**Suggested fix** — honour `sharedWithMe=true|false` using the ACL data the mock
already has (owner email vs. caller, plus `doc_grants`): "shared with me" =
visible to the caller *and* not owned by them.

---

## 2. `files.list`: `orderBy` is accepted but never applied

**What happens** — adding `orderBy` does not change the response at all.

```bash
G "/files?q=trashed%3Dfalse&pageSize=10&fields=files(id)"
G "/files?q=trashed%3Dfalse&pageSize=10&orderBy=name&fields=files(id)"   # identical ids
```

Verified for `orderBy=name` and `orderBy=modifiedTime desc`: on the mock the
returned order is unchanged and unsorted; on real Drive both sort correctly.

**Expected** — either sort, or reject the parameter. Real Drive supports
`createdTime`, `folder`, `modifiedByMeTime`, `modifiedTime`, `name`,
`name_natural`, `quotaBytesUsed`, `recency`, `sharedWithMeTime`, `starred`,
`viewedByMeTime`, each optionally with ` desc`.

**Likely cause** — `drive_files_list` (`app/routers/google.py:689`) never reads
the `orderBy` query parameter (it is not declared in `_P_DRIVE_LIST` either).

**Why it matters** — silent again: 200 with arbitrary order. Any client that
relies on server-side ordering — pagination by sort key, "10 most recently
modified", stable listings — appears to work against the mock and misbehaves in
production. (We ended up sorting client-side after discovering this.)

**Suggested fix** — implement at least `name`, `modifiedTime` and `createdTime`
with the ` desc` modifier; return `400` for values you don't support so the gap
is loud rather than silent.

---

## 3. `files.list`: an unknown field in the `fields` mask returns `200 {}` instead of `400`

**What happens** — a bogus field name is accepted and yields empty file objects.

```bash
G "/files?pageSize=1&fields=files(totallyBogusField)"
```

| | response |
|---|---|
| mock | `200 {"kind":"drive#fileList","files":[{}]}` |
| real Drive | `400 {"error":{"code":400,"message":"Invalid field selection …"}}` |

**Expected** — 400 with an "invalid field selection" message, as real Drive does.

**Likely cause** — `_drive_file_field_keys` (`app/routers/google.py:636`) parses
the requested keys and `drive_files_list` (`:737-740`) intersects them with the
keys it produced; unknown names simply match nothing and disappear.

**Why it matters** — this is the one that undermines the mock as a *verification*
tool rather than just returning odd data. A typo or a stale field name in a
consumer's mask passes every mock-backed test and then fails in production with
a 400. We hit the mirror image of this: because the mock never validates masks,
nothing in our test suite could have caught a bad mask, so we had to add a
hand-written guard instead.

**Suggested fix** — validate the parsed key set against the known
file-resource keys (the same set `_drive_file` produces, plus the documented
ones you don't synthesize) and return 400 on anything unknown.

---

## 4. `files.get` ignores the `fields` mask, so one file differs between `get` and `list`

**What happens** — requesting the same projection two ways gives different
objects; `files.get` adds keys that were not asked for (`iconLink`, `trashed`,
`ownedByMe`, `spaces`, `version`, …).

```bash
ID=$(G "/files?q=trashed%3Dfalse&pageSize=1&fields=files(id)" | python3 -c 'import json,sys;print(json.load(sys.stdin)["files"][0]["id"])')
G "/files?q=trashed%3Dfalse&pageSize=1&fields=files(id,name,mimeType,size,modifiedTime,webViewLink)"
G "/files/$ID?fields=id,name,mimeType,size,modifiedTime,webViewLink"
```

On real Drive the listing row and the `files.get` object are identical for the
same mask. On the mock they are not.

**Likely cause** — `drive_files_get` (`app/routers/google.py:749`) never calls
`_drive_file_field_keys`, unlike `drive_files_list` (`:737`).

**Why it matters** — a client that resolves a file by id (after a search, or from
a stored id) and one that reads it out of a listing should see the same file.
Differing shapes make caching, diffing and "did this change?" logic behave
differently depending on which call produced the row.

**Suggested fix** — apply the same projection in `drive_files_get`.

---

## 5. Folders are returned by `'root' in parents` but not by `mimeType='…folder'`

**What happens** — the same folders are visible through one query and invisible
through another.

```bash
G "/files?q=%27root%27%20in%20parents&pageSize=5&fields=files(id,name,mimeType)"
#   mock: 2 folders   real: folders present
G "/files?q=mimeType%3D%27application/vnd.google-apps.folder%27&pageSize=5&fields=files(id,name,mimeType)"
#   mock: 0 rows      real: folders present
```

Related, same root cause: the folder rows also **bypass the `fields`
projection** — asking for `files(id,name)` on `'root' in parents` returns 18
keys (`capabilities`, `parents`, `spaces`, `version`, …), while stored file rows
honour the mask correctly.

**Likely cause** — folders are synthesized on the fly from the `gdrive_folders`
table (`_drive_folder_obj` `:604`, `_drive_root_folders` `:657`) instead of
existing as rows, so the general query path (`_drive_q_match`, projection) never
sees them.

**Why it matters** — enumerating folders by mimeType is a standard way to walk or
index a Drive ("give me every folder, then their children"). On the mock that
returns nothing, so a crawler concludes the account has no folders even though
listing the root shows two.

**Suggested fix** — route synthesized folders through the same matching and
projection code as stored rows (or materialise them as rows), so any query that
should match a folder does.

---

## 6. ERB-imported Drive rows all become `application/octet-stream`

**What happens** — on the hosted instance, 99 of 100 sampled rows have
`mimeType: application/octet-stream`; real Drive rows carry specific types
(`…google-apps.spreadsheet`, `application/pdf`, `image/jpeg`,
`…google-apps.folder`, …).

```bash
G "/files?q=trashed%3Dfalse&pageSize=100&fields=files(name,mimeType)" \
  | python3 -c 'import json,sys,collections;print(collections.Counter(f["mimeType"] for f in json.load(sys.stdin)["files"]).most_common(3))'
```

**Likely cause** — the ERB importer inserts `subtype = raw.get("doc_type")`
(`app/importer/erb.py:610-613`) and never sets `mime_type`. ERB `doc_type`
values aren't the mock's native keys (`document`/`spreadsheet`/`presentation`/
`folder`), so `_native()` (`app/routers/google.py:872`) returns `None` and
`_drive_file` (`:883`) falls back to `application/octet-stream`. That also forces
the *binary* `webViewLink` shape (`drive.google.com/file/d/…/view`) for every
row, so no row ever looks like a Docs-editors file.

**Why it matters** — anything that branches on mimeType is untestable against
the hosted corpus: native-vs-binary handling, `export` vs `alt=media` (including
the documented 403 "use Export with Docs Editors files"), per-type link shapes,
icon/type display. The BYO sample corpus *does* set proper subtypes, so this is
specific to the ERB import path.

**Suggested fix** — map ERB `doc_type` onto the native subtypes where it makes
sense, and otherwise set `mime_type` from the title's extension so at least
binary rows get a real type.

---

## 7. Native Docs/Sheets/Slides omit `size`; real Drive populates it

**What happens** — a native Google document has no `size` field on the mock.

```bash
G "/files?q=mimeType%3D%27application/vnd.google-apps.document%27%20and%20trashed%3Dfalse&pageSize=1&fields=files(id,name,mimeType,size)"
```

| | result |
|---|---|
| mock | `{"id":…,"name":"Region-hybrid batch …","mimeType":"…google-apps.document"}` — **no `size`** |
| real Drive | Doc → `"size":"26210"`, Sheet → `"size":"3588735"` |

**Expected** — per Google's reference, `size` "is populated for files with
binary content stored in Google Drive **and for Docs Editors files**; it is not
populated for shortcuts or folders."

**Likely cause** — `_drive_file` (`app/routers/google.py:883`) only sets
`size`/`md5Checksum` in the binary branch; the native branch deliberately omits
them.

**Why it matters** — this one actively teaches implementors something false
about the Google API. We read the mock's behaviour, concluded "native Docs have
no byte size", and wrote that into both code comments and an
**agent-facing prompt** describing the mount; it shipped that way and was only
caught by diffing against real Drive. A mock that omits a field is a data gap; a
mock that omits a field Google always sends is a source of wrong beliefs.

**Suggested fix** — populate `size` on native rows too (`len(content)` is a
reasonable stand-in), keeping folders and shortcuts without it.

---

## Deliberately not filed

- **`webViewLink` lacks `?usp=drivesdk`** — `usp` is a Google attribution
  parameter with no effect on access or destination, and real Drive itself omits
  it on 19/100 rows (maps, folders, colab). The mock is within real behaviour.
- **Timestamp precision** — mock returns whole-day `…T00:00:00Z`, real returns
  millisecond precision. Both are valid RFC 3339; no consumer impact observed.
- **`pageSize=1001`** — real 400s, mock clamps to 1000. Harmless in practice.
- **`grant_type=authorization_code` unsupported** and **`/drive/v3/about`
  missing** — both real gaps (Gmail's `users.getProfile` equivalent *is*
  implemented), but out of scope for now: consumers can assemble credentials
  directly and skip the OAuth code exchange.

khj809:
Closes #23.

All seven divergences were reproduced against the reporter's own repros, each with a test that
fails without the fix. Five of them returned `200` with wrong or unfiltered data, which is the
class this PR is really about: a consumer could not tell anything was off.

| # | Divergence | Fix |
|---|---|---|
| 1 | `q: sharedWithMe` ignored | partitions on ownership; pushed into SQL as an owner filter |
| 2 | `orderBy` accepted, never applied | Drive's documented keys sort, across the whole result set |
| 3 | invalid `fields` mask → `200 {}` | masks parsed + validated → `400 Invalid field selection X` |
| 4 | `files.get` ignored `fields` | same projection as `files.list` |
| 5 | folders invisible to `mimeType='…folder'` | folders are files: one matcher, one sort, one projection |
| 6 | ERB rows all `application/octet-stream` | bench `doc_type` → native subtypes (+ extension fallback) |
| 7 | native Docs omit `size` | populated, per Google's reference |

### Notes on the judgment calls

**`sharedWithMe` (1).** "Shared with me" = visible to the caller **and** not owned by them. Since
that is exactly an owner predicate, it is pushed into SQL (`author_email = / <> ?`) rather than
materializing the corpus to drop half of it in Python. `ownedByMe` becomes per-caller in the same
change — it was hardcoded `false`, which would have made the new partition incoherent with the
field a client reads to make the same decision. The admin/service token has no email, so it owns
nothing and everything reads as shared for it; documented at `_drive_owned_by`.

**`orderBy` (2).** A sort spans the result set, so a sorted request materializes and orders it
(~0.8s/page on the 25k-row bench corpus) instead of ordering each page in isolation. **Unsorted
requests keep the existing SQL pagination** — measured 25ms for a crawl page — so no existing
client or crawler pays for this. `shared`, the one field that costs a query per page and that no
sort key reads, is deferred to the page (saves 1.35s of the 2.1s).

Three documented keys — `viewedByMeTime`, `modifiedByMeTime`, `sharedWithMeTime` — return **400**
instead of sorting. The mock models no per-caller view/share timestamps, so any ordering would be a
lie, and the whole point of #2 is that a silently unapplied sort is worse than a loud failure. This
is the one place the mock is deliberately *stricter* than real Drive; it is the trade the issue
asks for ("return 400 for values you don't support so the gap is loud rather than silent").

**`fields` validation (3/4).** Validated against the whole **documented** File field set, not the
keys the mock synthesizes — real Drive accepts a documented field it has no value for and simply
omits it, so `fields=files(id,thumbnailLink)` must stay a 200. Both mask syntaxes are now honored
(`files(id,name)` and `files/id`), plus `*`. Top-level names are validated but not projected: the
mock always returns `kind`/`incompleteSearch` because its typed response model — the one the
OpenAPI schema is generated from — declares them.

**Folders (5).** Routed through the same matcher and projection as stored rows, which has three
consequences worth calling out: `permissions.list` now answers for a folder id (it 404'd, because
folders are not stored as rows), an unfiltered `files.list` includes folders **as it does on real
Drive**, and the folder stream is skipped for `fullText contains` (a folder's only text is its
name, which `name contains` already covers). Rows come first in the unsorted order so a client
reading `files[0]` out of an unfiltered listing still gets a file — the official-SDK test does
exactly that. Cost of the folder stream: 1–23ms per request (3 folders in the bench corpus).

### Deliberately not included

**The error envelope.** Real Drive returns `{"error": {"code", "message"}}`; this mock returns
`{"detail": ...}` for every non-Atlassian router, and `test_endpoints.py` documents that as
intentional. The three new 400s follow the repo convention rather than making Drive's errors
inconsistent with Gmail's. Worth a separate, mock-wide decision.

### Verification

`pytest` — **551 passed, 2 skipped** (from 535; 16 new tests). The two skips are the pre-existing
data-dependent/network-gated ones. The optional suites that make real HTTP calls both run:
`test_sdk.py` (google-api-python-client against a live uvicorn) and `test_mcp.py`.

Divergence 6 is an **import-time** fix, so the hosted corpus needs its rows updated in place before
they carry real mimeTypes — measured on the served DB: 22 of 25,108 Drive rows were native. The
other six take effect on restart.

Co-authored-by: Claude Opus 5 (1M context) <noreply@anthropic.com>


khj809:
### Divergence 6 applied in place to both served corpora (2026-07-28)

(6) is import-time only, so already-built DBs needed their rows remapped. `gdrive_files.subtype`
turned out to hold the **raw bench `doc_type` string**, so no re-import was necessary — the mapping
runs straight off the stored column, through the same `erb._drive_type` this PR adds.

Applied to the local build artifact and to the live box (`/home/ubuntu/data/mock.sqlite`), **3.5s
each**, 25,108 rows, no restart needed — the live RO connection picked the change up on its next
read. A `(doc_id, old subtype, old mime_type)` rollback TSV sits next to each DB, plus a full
`mock.sqlite.predrivemime` snapshot on the box.

```
before: [('application/octet-stream', 25086), ('application/vnd.google-apps.document', 22)]
after:  [('application/vnd.google-apps.document', 23078), ('application/vnd.google-apps.spreadsheet', 1875),
         ('application/vnd.google-apps.presentation', 153), ('application/pdf', 2)]
```

The reporter's own repro against the live deployment now returns real types instead of 99/100
`application/octet-stream`:

```console
$ curl -s -H "Authorization: Bearer $T" \
    "https://enterprise-mock.brekkylab.com/drive/v3/files?q=trashed%3Dfalse&pageSize=100&fields=files(name,mimeType)" \
  | python3 -c 'import json,sys,collections;print(collections.Counter(f["mimeType"] for f in json.load(sys.stdin)["files"]).most_common(3))'
[('application/vnd.google-apps.document', 96), ('application/vnd.google-apps.spreadsheet', 3),
 ('application/vnd.google-apps.presentation', 1)]
```

`examples/using-official-sdk/gdrive.py --url https://enterprise-mock.brekkylab.com` also passes
now, admin and impersonated-user alike — its `files.export` step **could not have worked before**,
since every row was binary and export 403s on those.

### State of the box

The box still runs `main`, so only (6) is live there; a 13-check sweep of the issue's repros against
the public URL is 3/13 (the 10 code-side ones need this PR). The same sweep against the updated
local corpus with this branch's code is **13/13**:

```
[PASS] #1 sharedWithMe true != false      [PASS] #5 mimeType='folder' == 'root' in parents
[PASS] #2 orderBy=name sorts              [PASS] #5 folder rows honor the projection
[PASS] #2 orderBy=modifiedTime desc       [PASS] #5 permissions.list on a folder id
[PASS] #2 unknown orderBy -> 400          [PASS] #6 ERB rows carry real mimeTypes
[PASS] #3 bogus fields -> 400             [PASS] #7 native Doc reports size
[PASS] #3 documented field -> 200         [PASS] #7 folder omits size
[PASS] #4 files.get == files.list
```

Deploying the code is a separate step, gated on this review.

selloriwoo:
`sharedWithMeTime` is never populated, which contradicts #1.

`_drive_file` (:1189) sets `shared` and `ownedByMe` but never `sharedWithMeTime` — the name appears
only in the validated field set and in `_DRIVE_ORDER_UNMODELLED`. So `q: sharedWithMe=true` returns
rows whose own fields say they aren't shared. Real Drive omits `parents` entirely on shared items, so
this field is the only signal a client has to classify one.

The fix sits next to `ownedByMe`: populate it when `_drive_owned_by(author, me)` is False, using the
row's created/modified time — gated on `me` being known, since that helper returns None for the admin token.

khj809:
Fixed in c805c5d.

Confirmed the field was never emitted — `sharedWithMeTime` appeared only in `_DRIVE_FILE_FIELDS`
and `_DRIVE_ORDER_UNMODELLED`. One nuance on the framing: those rows did report `shared: true` and
`ownedByMe: false`, so they weren't claiming *not* to be shared — what was missing was the
timestamp, which on real Drive is the canonical classifier. The gap and the fix are as you describe.

It also invalidated something you didn't mention, which is the part that mattered most:
`orderBy=sharedWithMeTime` was returning **400** on the grounds that the mock "models no per-caller
share timestamps". That justification was wrong for this key specifically — the mock models the
relation the key sorts on (owner vs caller); it just wasn't emitting it. So the key now **sorts**
rather than 400s. `viewedByMeTime`/`modifiedByMeTime` still 400: nothing here is ever viewed or
modified by anyone in particular, so those have no counterpart at all. Your catch is exactly the
line between the two.

Implementation notes:

- **`createdTime`, not `modifiedTime`**, as the stand-in. A share time that moved every time the
  document was edited would reorder `orderBy=sharedWithMeTime` for an unrelated reason; creation is
  stable and never later than a real share would be.
- **Gated on a known caller**, as you say. Nothing was shared *with* the admin/service token, so it
  gets no timestamp — and `orderBy` on the field still 200s there, every key tying, which is what
  real Drive does over nulls.
- **Synthesized folders answer the same way.** They have no owner, so they are never owned by the
  caller, and they come out of the very same `sharedWithMe` filter — a folder in the shared set
  without the field would have reintroduced the row-vs-folder split that divergence 5 was about.
  `files.get` on a folder id agrees with the listing.

Three tests, each failing before the change: the true/false partition carries/omits the field
(including a folder, and `files.get` agreeing with the listing), the admin token omits it, and
`orderBy=sharedWithMeTime desc` sorts. Suite: 554 passed.

### On `parents` — not changing it, deliberately

You're right that real Drive omits `parents` on shared items. Mirroring that here would break the
mock's folder tree rather than sharpen it: this corpus is an org-wide shared drive, so for a scoped
caller almost nothing is self-owned — measured on the bench corpus for `aditi.mehra@…`, 25,040 of
25,108 visible rows (99.7%) are owned by someone else. Dropping `parents` on those would leave
`'<folder-id>' in parents` **returning** files whose emitted `parents` is empty, so a client that
lists a folder's children and then reads the field to build a tree gets nothing — the query and the
row would contradict each other, which is the same class of bug as the seven above.

Real Drive is self-consistent there because a shared item genuinely is not in your My Drive; the
mock's files genuinely are in a visible folder. Happy to file it separately if you think the tree
should be per-caller instead, but it's a corpus-model change, not a field fix.

`sharingUser` is the other field real Drive populates on shared items, and I left it out on purpose:
it names *who performed the share*, which the mock has no record of. Filling it with the owner would
be inventing a fact — the "teaches implementors something false" failure mode from (7) in your issue.

selloriwoo (approved):

Issue #19, opened 2026-07-26
Build and publish a public dataset: a corpus covering every source, with a graded question set


khj809:
Build and publish a public, versioned dataset — a corpus that covers every source Backlot serves, with a question set graded against its own read policy — that a user pulls and serves in one command, plus the pipeline that produces it and the harness that grades it.

## Why

The corpus that ships with the package is 167 documents. That exists so `import -> serve -> query` works with nothing fetched; it is a smoke test, not something anyone can evaluate a connector or an agent against. Everything past that is "bring your own", and nobody's own corpus covers fourteen surfaces at once.

So the two things Backlot is for — crawl a source to exhaustion and diff, or point an agent at the whole stack and see what it can answer — both stall on the same missing piece: a corpus that is large enough to be interesting, covers every served source, and comes with answers you can grade against.

## What the dataset is

Four parts, versioned and published together:

1. **The corpus.** BYO-JSONL, sharded per source, covering **every** source ticked in #89 — eleven record types across the fourteen served surfaces. Around 10,000 documents in total, against the 167 that ship inside the package. Not a token document per source: enough of each that crawling it exercises what that vendor's API actually does — pagination, threads and replies, containers and nesting, MIME and subtype variety, states and transitions, the fields that are usually empty and occasionally not.
2. **A question set.** Each question carries the caller who asks it, the documents that answer it, and an answer key. Single-document lookups, multi-hop questions, and questions that cross sources — plus questions whose correct answer *depends on who is asking*, which is the thing no public corpus lets you test today.
3. **The read policy.** The per-document ACL the questions are graded against, so "did retrieval find it" and "did the connector leak it" stay two separate measurements instead of one blurred score.
4. **The roster.** The principals in the exact shape `backlot import --roster` reads. Without a roster the importer infers group membership from authorship and over-serves, so grading against an inferred roster measures the inference, not the connector. The dataset carries the one it was generated against.

## Quality bar

This is the whole value of the artifact — a corpus that is merely *large* is a load test, not an evaluation.

- **One world, not eleven.** A person, team, customer, project or incident named in one source resolves in the others: the thread that reports an outage, the ticket it opens, the document the fix lands in and the meeting where it is reviewed are the same event seen from four systems. Multi-hop questions are only answerable if this holds, and cross-source retrieval is only tested if it holds.
[...]

Issue #18, opened 2026-07-26
Unified corpus: convert all 9 ERB sources to BYO-JSONL + scale the augmented layer to all 1,695 scenes


khj809:
Build the unified corpus: convert full ERB into BYO-JSONL and scale the augmented layer to every
scene.

Subtask of the unified-dataset epic. Blocked by the new-source issues — converting ERB's Linear and
Fireflies sections needs their BYO schemas and ERB→schema mappings, which land there — and by the
BYO-lossless subtask (conversion has nothing to write into otherwise).

## Two layers, two jobs

ERB ships **9** source directories, so all nine get converted. The augmented layer covers only what ERB
omits — **GitHub source-code files, Notion, and S3**; HubSpot, Linear, and Fireflies are bench sources
and are converted, not generated.

## Measured starting point

| | Docs | Raw | gzip |
|---|---|---|---|
| ERB `generated_data` (9 sources) | **511,962** individual JSON files | ~3.65 GB | ~1.0 GB (the upstream tarball is 1.0 GB and already contains all 9 sources) |
| Augmented today | **592** — github `file` 392, notion 98, s3 102 | 3.5 MB | ~1 MB |

Per source, measured from the release tarball (entry counts include each directory's own entry and its
`agents.md`; `documents` excludes both; raw sizes are the local extraction):

| source | tarball entries | documents | raw |
|---|---|---|---|
| slack | 285,644 | 285,605 | 1.5 GB |
| gmail | 121,448 | 121,390 | 1.2 GB |
| linear | 35,315 | 35,308 | 295 MB |
| google_drive | 25,142 | 25,108 | 244 MB |
| hubspot | 15,020 | 15,017 | 105 MB |
| fireflies | 10,182 | 10,173 | 148 MB |
| github | 8,078 | 8,052 | 68 MB |
| jira | 6,126 | 6,120 | 51 MB |
| confluence | 5,313 | 5,189 | 64 MB |
| **total** | **512,268** | **511,962** | **~3.65 GB** |

The augmented layer is currently 0.12% of the corpus and was generated for only ~50 of the **1,695**
available scenes (`experiments/erb-augment/data/scenes.jsonl`). A scene is one `(repo, ticket)` pair —
a PR's real context — which is what keeps generated GitHub/Notion/S3 artifacts mutually consistent and
grounded in real ERB people, ticket ids, and repos.

## Two pieces of work

[...]

Issue #17, opened 2026-07-26
BYO-JSONL lossless round-trip for ERB imports


khj809:
Make BYO-JSONL able to represent an ERB import **losslessly**, so ERB can be redistributed in the same
schema as the augmented corpus.

Subtask of the unified-dataset epic. Independent of the three new-source issues — no code overlap, so
it can run in parallel — and it is worth shipping on its own merits: every gap closed here is
expressiveness that any BYO author gains.

## Why

The unified dataset ships ERB pre-converted into BYO-JSONL. That only works if BYO-JSONL can hold
everything `app/importer/erb.py` puts in the DB. Comparing `byo.py:111` `_service_columns()` against
`erb.py`'s `load_*()` functions, it currently cannot:

| Source | Written by `erb.py`, unrepresentable in BYO |
|---|---|
| confluence | `confidentiality`, `owner_team`, `reviewers` |
| google_drive | `collaborators` |
| jira | `severity`, `squad` |
| all | `owner_display` — BYO derives a display name from the email instead (`byo.py:84`) |
| gmail | **multi-message threads** — `replies` is Slack-only, so N messages sharing a `thread_id` with distinct `thread_seq` cannot be expressed |

Two more behaviours live in `erb.py` rather than in the data, and must become expressible so a
converted artifact reproduces them without re-running `erb.py`:

- the Slack far-future timestamp remap (`build_slack_ts_remap`, cutoff year 2035) — rank-based and
  order-preserving, so it cannot be recomputed from a single record in isolation; it has to be baked
  into the converted timestamps
- principal resolution from the employee directory (`app.importer.principals`) — names → real emails.
  The converted artifact must carry resolved emails, and the directory must ship alongside it

## Approach

Extend the per-service BYO mapping and the JSON Schemas to cover the columns above, then prove
equivalence rather than assert it. Gmail's multi-message thread is the one that needs a shape
decision, not just a column: either a Fireflies-style child-row array (see the Fireflies issue, which
faces the same "`replies` is Slack-only" problem) or explicit per-message records carrying
`thread` + an explicit sequence. Pick one and apply it consistently across both issues.

## Done when

`ERB → BYO-JSONL → DB` produces a database **equivalent to `erb.py` importing ERB directly** — same
rows, same column values, same `doc_acl`, same `tokens.yaml`. Implement that comparison as a test so
it keeps holding; a diff of the two databases is the acceptance criterion, not a spot check.

Also:

[...]

PR #31, opened 2026-07-30
Shard the BYO export, stream it back, and make both import paths agree

nuri-yoo:
Closes part of #18.

The BYO export held ERB losslessly as of #17, but as a single file. At bench scale that is one 955 MB artifact with no way to check a download, and no way to load it without holding it in memory. This shards the export, gives it a manifest, teaches the loader to stream it, and closes two places where the two import paths disagreed.

## Sharding and the manifest

`--export-byo --shard-records N` writes `data/<source>/part-NNNNN.jsonl.gz` plus `manifest.json`, which records every shard's path, record count, byte size and SHA-256. Shards are gzipped with `mtime=0` so the same input always produces the same checksums; gzip's default header would stamp the current time into every shard and change its digest run to run.

`python -m app.importer.byo <dir>` loads the whole thing in one command, checking every digest in the manifest — shards and `roster.yaml` alike — before it reads a record. A damaged or swapped download then fails up front instead of half-loading a database. A directory carries its own roster, which the importer picks up without `--roster` being passed again, and that is exactly why the roster's digest is verified with the shards rather than after them: it decides who holds a token and what they can see.

The case worth naming is a shard that is short but validly terminated, which is what a resumed or re-uploaded download looks like. Its gzip stream reads cleanly to its end, so nothing but the digest can tell that records are missing. A full sha256 pass over the 1.0 GB bench artifact takes 0.67 s, ahead of an import that writes 14 GB.

The loader streams. `corpus_records` iterates a plain file, a `.jsonl.gz`, or a shard directory, and splits on `\n` alone: U+2028 and the vertical tab are ordinary characters inside a JSON string, and treating either as a record boundary tears records in half. Org inference is its own pass over the corpus and streams the same way, since it needs only the addresses.

A downloaded manifest is untrusted input, so a shard path has to stay inside the artifact directory, and a directory with no manifest says what to pass instead of raising through. `--shard-records` rejects anything below 1: at 0 every record becomes its own shard, which for the bench is 600k files, the direction sharding exists to prevent. The manifest's `documents` counts what was written, and what it did not write is named in `layers.converted` rather than totalled, so it cannot contradict its own per-source sum.

## Two documents sharing a doc_id

Four bench documents share a `dataset_doc_uuid` with another: three across sources (a drive file that is also a confluence page, plus a hubspot and a jira one) and two jira issues sharing one. The two import paths resolved that differently, so the full-corpus round trip did not produce an equivalent database. Measured over all 511,961 documents, `jira_issues` had the same row count but a different digest, and `doc_acl` differed by three rows.

`import_structured` collected grants from a dict keyed by doc_id, so when two documents shared an id only the last one's grants were written and the earlier document's container group was never granted. Grants are now written inside the per-document loop, where every document is seen.

The BYO loader skipped a record whose `(source, doc_id)` it had already seen, keeping the earlier of two, while a direct import overwrites the row and keeps the later one. The loader writes both and lets `INSERT OR REPLACE` decide, which is the same answer and matches what a JSONL corpus means when a later line repeats an id.

`test_round_trip_survives_two_documents_sharing_a_doc_id` covers both. Without either change it fails with "via byo kept the earlier of the two jira records".

## One document with no content, recorded by name

The bench ships a slack thread whose `messages` is the empty string. A direct import accepts it; the BYO schema rejects it, so the exported artifact did not pass the mock's own validator. It is dropped in `select_records`, where both paths see the same decision, taking slack from 285,605 to 285,604.

Review asked whether the artifact should record that drop, not just the log. It does now. Each exclusion is collected by source, id, and its path within the source, and lands in the manifest under `layers.converted` beside `source_documents`, so the layer states its own arithmetic: source_documents == documents + excluded + failed. The path is the part that pays — a dsid is not a location, so recording only the id leaves the same 3.65 GB scan to do. `failed` is a list of the same shape, because identity matters more for a failure than for a policy drop.

An exclusion the code does not declare stops the run. `generated_data` has had exactly one commit since 2026-04-14, so the same bench has to yield the same set, and a numeric threshold would have nothing in the data behind it: content lengths run 0, 3, 5, 9 and then thirty-one more under fifty, with no gap to put a cutoff in. The three just above zero are the same damage as the one at zero — the five-character one is a thread with six participants and timestamps 69 seconds apart whose entire body is `(end)` — so the converter drops where the damage reached empty and ships where it stopped short, and `reason` records the mechanical rule instead of a diagnosis it cannot support. `--allow-excluded N` gets past the gate once someone has read them.

`source_documents` only means something against a known snapshot, and neither a ref nor a tag pins this one: `main` has moved past the commit that added `generated_data`, and the only tag (v1.0.0, 2026-03-29) predates it. `fetch_generated_data` records the tarball's digest at fetch time and the export carries it into the layer.

## Measured on the full bench

| | |
|---|---|
| Export | 511,961 documents to 581,294 records, 17 shards, 939 MB, 188 s |
| Validation | `OK: 600581 records valid` (with the augmented layer merged in) |
| Round trip | 34 tables compared, 0 differ; `tokens.yaml` identical, 167 users |
| Import | one command, 6,397,932 rows including child rows, 13.98 GB |

The round-trip check at bench scale is a streaming table-by-table comparison rather than the in-memory one in `tests/test_importer_erb.py`, which would materialise tens of millions of rows.

Serving the imported corpus, both a redistributed and a generated document come back over the real API surface (github contents, notion by its synthesized UUID, confluence by its synthesized numeric id), an unsigned S3 request is refused, a request with no token is 401, and a caller the document is not granted to gets 404.

The numbers above were measured before the manifest carried `layers.converted`, so the artifact published for #19 is re-exported from this branch to pick the field up; the shard digests are reproducible from the same input, so only the manifest changes.

Tests: 667 passed, 16 skipped. Coverage includes the shared-doc_id round trip, per-source id namespacing, a manifest whose shard is short, swapped byte-for-byte, missing, or named outside the artifact, a swapped roster, an import refusing an artifact that fails any of those, a directory with no manifest, a single `.jsonl.gz` corpus, an undeclared exclusion stopping the export and naming the file, a declared one recorded by identity with the layer's arithmetic closing, and the snapshot marker reaching the manifest.

### DIFF
--- a/app/importer/byo.py
+++ b/app/importer/byo.py
@@ -464,7 +464,7 @@ def load(path: Path, settings: Settings | None = None, reset: bool = True,
     memberships: set[tuple[str, str]] = set()      # (group_id, email)
     grants: list[tuple[str, str, str]] = []        # (doc_id, principal_type, principal_id)
     counts: dict[str, int] = {}
-    seen: set[str] = set()
+    seen: set[tuple[str, str]] = set()   # (source_type, doc_id)
     fts_ids: dict[str, list[str]] = {}
     # HubSpot associations are resolved after the whole corpus is read: a link may name a target
     # that appears on a later line, and an omitted `to_type` is filled in from the target's own
@@ -527,9 +527,13 @@ def load(path: Path, settings: Settings | None = None, reset: bool = True,
             rec = {**rec, "content": synth.fireflies_transcript_text(sentences)}
 
         doc_id = _doc_id(rec)
-        if doc_id in seen:
-            continue
-        seen.add(doc_id)
+        # Recorded, not deduplicated: `seen` answers "is this document in the corpus" for the
+        # cross-reference resolution further down. Two records sharing a (source, doc_id) are both
+        # written, and the row-level `INSERT OR REPLACE` leaves the later one — which is what a
+        # direct ERB import of the same documents produces. The bench has four such pairs (three
+        # across sources, one within jira); skipping the repeat instead would keep the earlier
+        # document and diverge.
+        seen.add((src, doc_id))
         gcol = store.grouping_col(src)
         container = str(rec.get(gcol) or src)   # channel / mailbox / folder / repo / project / space
         # An explicit `"group": null` means the container owns NO ACL group — which is a real
@@ -750,9 +754,7 @@ def load(path: Path, settings: Settings | None = None, reset: bool = True,
             register(rep_author, rep.get("author_name"))
             rep_id = rep.get("doc_id") or (
                 "dsid_" + hashlib.sha256((doc_id + str(i) + rep["content"]).encode()).hexdigest()[:32])
-            if rep_id in seen:
-                continue
-            seen.add(rep_id)
+            seen.add((src, rep_id))
             # A reply is a full message (reactions/files/subtype/edited carry through);
             # its time is the root's + its position so the thread stays ordered (created is now
             # always set, so a reply ts is never NULL).
@@ -775,9 +777,7 @@ def load(path: Path, settings: Settings | None = None, reset: bool = True,
             m_author = msg.get("author_email")
             register(m_author, msg.get("author_name"))
             msg_id = msg.get("doc_id") or f"{doc_id}::m{i}"
-            if msg_id in seen:
-                continue
-            seen.add(msg_id)
+            seen.add((src, msg_id))
             # Its own `created` when given, else the root's clock + an hour per position — the
             # spread a real reply chain has, and never NULL.
             insert(msg_id, m_author, msg.get("title") or title, msg["content"], i,
@@ -851,7 +851,7 @@ def load(path: Path, settings: Settings | None = None, reset: bool = True,
     # does not exist is an error rather than a dangling relation, matching the hubspot rule.
     for from_doc, a, created_ts in lin_links:
         to_doc = a["to"]
-        if to_doc not in seen and not conn.execute(
+        if ("linear", to_doc) not in seen and not conn.execute(
                 "SELECT 1 FROM linear_issues WHERE doc_id = ?", (to_doc,)).fetchone():
             raise SystemExit(
                 f"linear relation {from_doc} -> {to_doc}: target not found in this corpus or the "
--- a/app/importer/byo.py
+++ b/app/importer/byo.py
@@ -66,10 +66,12 @@ Usage:  python -m app.importer.byo path/to/corpus.jsonl [--append | --dry-run] [
 from __future__ import annotations
 
 import argparse
+import gzip
 import hashlib
 import json
 import re
 import sys
+from collections.abc import Iterator
 from pathlib import Path  # noqa: F401  (kept for typing/backcompat)
 
 import yaml
@@ -311,9 +313,68 @@ def _emails(rec: dict):
             yield cv
 
 
-def _infer_org(records: list[dict], settings: Settings) -> tuple[str, str]:
+def _infer_org(records, settings: Settings) -> tuple[str, str]:
     """Derive (org_name, org_domain) from the corpus's dominant author email domain."""
     return infer_org((e for rec in records for e in _emails(rec)), settings)
+
+
+def corpus_records(source) -> Iterator[tuple[int, str]]:
+    """``(lineno, line)`` over a plain JSONL file, a ``.jsonl.gz``, or a sharded artifact directory.
+
+    A directory is read through its ``manifest.json``, shard by shard, because the whole point of
+    sharding is a corpus too large to hold at once: reading the 581k-record ERB conversion with
+    ``read_text()`` would materialize several gigabytes of ``str``. Line numbers run across the
+    whole artifact so an error message still names one place.
+
+    Records are split on ``\\n`` alone (``newline="\\n"``), for the reason ``jsonl_lines`` documents:
+    U+2028 and the vertical tab are ordinary characters inside a JSON string, and Python's universal
+    newlines would otherwise tear a valid record in half.
+    """
+    source = Path(source)
+    if source.is_dir():
+        manifest = json.loads((source / "manifest.json").read_text())
+        paths = [source / s["path"] for src in sorted(manifest["sources"])
+                 for s in manifest["sources"][src]["shards"]]
+    else:
+        paths = [source]
+    n = 0
+    for p in paths:
+        opener = gzip.open if p.suffix == ".gz" else open
+        with opener(p, "rt", encoding="utf-8", newline="\n") as fh:
+            for line in fh:
+                n += 1
+                yield n, line.rstrip("\n")
+
+
+def verify_manifest(source) -> list[str]:
+    """Problems found checking a sharded artifact against its manifest ([] == intact).
+
+    Size and digest, per shard: a truncated or swapped download has to fail before it half-loads a
+    database, which is the same reason ``--dry-run`` exists for a hand-written corpus.
+    """
+    source = Path(source)
+    mf = source / "manifest.json"
+    if not mf.exists():
+        return [f"{mf} not found"]
+    manifest = json.loads(mf.read_text())
+    problems = []
+    for src in sorted(manifest.get("sources", {})):
+        for shard in manifest["sources"][src]["shards"]:
+            p = source / shard["path"]
+            if not p.exists():
+                problems.append(f"{shard['path']}: missing")
+                continue
+            if p.stat().st_size != shard["bytes"]:
+                problems.append(f"{shard['path']}: {p.stat().st_size} bytes, manifest says "
+                                f"{shard['bytes']}")
+                continue
+            h = hashlib.sha256()
+            with open(p, "rb") as fh:
+                for chunk in iter(lambda: fh.read(1 << 20), b""):
+                    h.update(chunk)
+            if h.hexdigest() != shard["sha256"]:
+                problems.append(f"{shard['path']}: sha256 mismatch")
+    return problems
 
 
 def load_roster(path) -> dict:
@@ -364,17 +425,24 @@ def load(path: Path, settings: Settings | None = None, reset: bool = True,
         settings.db_path.unlink()
     conn = store.connect_rw(settings.db_path)
 
-    lines = jsonl_lines(Path(path).read_text())
     # infer the org from the corpus (dominant email domain) before building any grants,
-    # since public docs are granted to the org principal — see _infer_org.
+    # since public docs are granted to the org principal — see _infer_org. Read in its own pass,
+    # keeping only the emails: a sharded corpus is streamed rather than held in memory.
     _scan = []
-    for _ln in lines:
+    for _no, _ln in corpus_records(path):
         _ln = _ln.strip()
         if _ln:
             try:
-                _scan.append(json.loads(_ln))
+                _rec = json.loads(_ln)
             except json.JSONDecodeError:
-                pass  # malformed lines are reported precisely in the main loop below
+                continue  # malformed lines are reported precisely in the main loop below
+            # Keep only what `_emails` reads, and drop the child rows' bodies: the addresses are
+            # kilobytes where the corpus is gigabytes.
+            _scan.append({
+                **{k: _rec[k] for k in ("author_email", "host_email", "readers") if k in _rec},
+                **{c: [{"author_email": r.get("author_email")} for r in (_rec.get(c) or [])
+                       if isinstance(r, dict)]
+                   for c in ("comments", "sentences", "messages", "replies") if c in _rec}})
     org_name, org_domain = _infer_org(_scan, settings)
     roster_data = load_roster(roster) if roster else None
     closed = roster_data is not None
@@ -407,7 +475,7 @@ def load(path: Path, settings: Settings | None = None, reset: bool = True,
     # since a target may appear on a later line — the same second pass the ERB importer runs.
     lin_links: list[tuple[str, dict, int]] = []
 
-    for lineno, line in enumerate(lines, 1):
+    for lineno, line in corpus_records(path):
         line = line.strip()
         if not line:
             continue
@@ -852,7 +920,8 @@ def load(path: Path, settings: Settings | None = None, reset: bool = True,
 
 def main(argv: list[str]) -> int:
     ap = argparse.ArgumentParser(description="Import (load) a BYO JSONL corpus into the mock DB.")
-    ap.add_argument("corpus", help="path to a JSONL corpus file")
+    ap.add_argument("corpus", help="a JSONL corpus file, a .jsonl.gz, or a directory holding "
+                                   "manifest.json + data/<source>/part-*.jsonl.gz shards")
     ap.add_argument("--append", action="store_true", help="add to the existing DB instead of resetting")
     ap.add_argument("--dry-run", action="store_true", help="validate the corpus only; don't touch the DB")
     ap.add_argument("--roster", type=Path, default=None,
@@ -862,9 +931,28 @@ def main(argv: list[str]) -> int:
     corpus = Path(args.corpus)
 
     if args.dry_run:
-        problems = validate_file(corpus)
+        if corpus.is_dir():
+            # A sharded artifact is checked against its manifest first: a truncated download is a
+            # different failure from a bad record, and saying so is cheaper than a schema error.
+            broken = verify_manifest(corpus)
+            if broken:
+                print(f"INVALID: {len(broken)} shard problem(s) in {corpus}", file=sys.stderr)
+                for b in broken:
+                    print(f"  {b}", file=sys.stderr)
+                return 1
+        problems, n = [], 0
+        for lineno, line in corpus_records(corpus):
+            line = line.strip()
+            if not line:
+                continue
+            n += 1
+            try:
+                rec = json.loads(line)
+            except json.JSONDecodeError as e:
+                problems.append((lineno, f"invalid JSON: {e}"))
+                continue
+            problems.extend((lineno, m) for m in record_errors(rec))
         if not problems:
-            n = sum(1 for line in jsonl_lines(corpus.read_text()) if line.strip())
             print(f"OK: {n} records valid.")
             return 0
         print(f"INVALID: {len(problems)} problem(s) in {corpus}", file=sys.stderr)
--- a/app/importer/byo.py
+++ b/app/importer/byo.py
@@ -961,7 +961,13 @@ def main(argv: list[str]) -> int:
         return 1
 
     settings = get_settings()
-    res = load(corpus, settings, reset=not args.append, roster=args.roster)
+    # A sharded artifact ships its own roster beside the manifest, so importing one takes the
+    # directory and nothing else; `--roster` still wins when it is given.
+    roster = args.roster
+    if roster is None and corpus.is_dir() and (corpus / "roster.yaml").exists():
+        roster = corpus / "roster.yaml"
+        print(f"using the artifact's own roster: {roster}", file=sys.stderr)
+    res = load(corpus, settings, reset=not args.append, roster=roster)
     print(f"Loaded {res['total']} documents into {settings.db_path}")
     for src, n in sorted(res["counts"].items()):
         print(f"  {src:14s} {n}")
--- a/app/importer/byo.py
+++ b/app/importer/byo.py
@@ -332,9 +332,20 @@ def corpus_records(source) -> Iterator[tuple[int, str]]:
     """
     source = Path(source)
     if source.is_dir():
-        manifest = json.loads((source / "manifest.json").read_text())
-        paths = [source / s["path"] for src in sorted(manifest["sources"])
-                 for s in manifest["sources"][src]["shards"]]
+        mf = source / "manifest.json"
+        if not mf.exists():
+            raise SystemExit(f"{mf} not found — pass a JSONL file, a .jsonl.gz, or a sharded "
+                             f"artifact directory")
+        manifest = json.loads(mf.read_text())
+        # The artifact is downloaded, which makes its manifest untrusted input: a shard path has to
+        # stay inside the directory it came with.
+        paths = []
+        for src in sorted(manifest["sources"]):
+            for s in manifest["sources"][src]["shards"]:
+                p = (source / s["path"]).resolve()
+                if not p.is_relative_to(source.resolve()):
+                    raise SystemExit(f"manifest names a shard outside {source}: {s['path']}")
+                paths.append(p)
     else:
         paths = [source]
     n = 0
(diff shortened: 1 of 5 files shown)


app/importer/byo.py
```diff
@@ -873,7 +961,13 @@ def main(argv: list[str]) -> int:
...
 
     settings = get_settings()
-    res = load(corpus, settings, reset=not args.append, roster=args.roster)
+    # A sharded artifact ships its own roster beside the manifest, so importing one takes the
+    # directory and nothing else; `--roster` still wins when it is given.
+    roster = args.roster
+    if roster is None and corpus.is_dir() and (corpus / "roster.yaml").exists():
+        roster = corpus / "roster.yaml"
+        print(f"using the artifact's own roster: {roster}", file=sys.stderr)
+    res = load(corpus, settings, reset=not args.append, roster=roster)
```
khj809, line 1015: **[blocker] `verify_manifest` is never called on the path that actually imports.**

Its only call site is the `--dry-run` branch above (byo.py:937). Both the README and the PR body claim the opposite:

> `python -m app.importer.byo out/` … **checks the manifest before reading a record, so a truncated download fails as a bad checksum instead of a quietly short corpus.**

I rewrote a 3-record shard as 2 records — not a truncation, a **properly terminated gzip stream**, which is what a resumed or re-uploaded download looks like:

    verify_manifest would catch it: ['data/slack/part-00000.jsonl.gz: 151 bytes, manifest says 158']
    load() reported total=2 counts={'slack': 2}   # the manifest says 3 documents / 3 records

That is exactly the quietly short corpus the README says cannot happen. A swap that keeps the byte count gets through too — I replaced a shard's content with `"Attacker payload."` and `load()` put that value in the DB. Only a mid-stream truncation is caught, and only incidentally, by gzip's `EOFError`.

Cost is not an argument here: I measured a full sha256 pass over a 1.0 GB sharded artifact at **0.67 s**, in front of an import that writes 14 GB.

```suggestion
    settings = get_settings()
    if corpus.is_dir():
        # The check --dry-run makes, on the path that actually writes a DB: a swapped or a short
        # shard has to fail before it half-loads one. A full sha256 over the 1.0 GB ERB artifact is
        # under a second, against an import that writes 14 GB.
        broken = verify_manifest(corpus)
        if broken:
            print(f"INVALID: {len(broken)} shard problem(s) in {corpus}", file=sys.stderr)
            for b in broken:
                print(f"  {b}", file=sys.stderr)
            return 1
    # A sharded artifact ships its own roster beside the manifest, so importing one takes the
    # directory and nothing else; `--roster` still wins when it is given.
    roster = args.roster
    if roster is None and corpus.is_dir() and (corpus / "roster.yaml").exists():
        roster = corpus / "roster.yaml"
        print(f"using the artifact's own roster: {roster}", file=sys.stderr)
    res = load(corpus, settings, reset=not args.append, roster=roster)
```

The block is now identical to the one in the `--dry-run` branch, so a small `_verify_or_die(corpus)` helper would be worth it. If you'd rather keep verification off the load path, then at minimum compare the streamed record count against `manifest["records"]` at the end — today a short read is reported as success.

khj809 (commented):
Reviewed the diff against #18. The suite passes (`pytest tests/ -q`).

The design calls are mostly good. `mtime=0` plus a second-run digest comparison, the `newline="\n"` reasoning, the argument for turning `seen` from a dedup into a record, and `test_round_trip_survives_two_documents_sharing_a_doc_id` carrying its own failure messages — the two round-trip fixes are solid in diagnosis, fix, and test. I also confirmed that moving `grants_for` into the loop is safe: `doc_acl` has no FK, `_NullConn.execute` is a no-op so the `dump_tokens` path is unaffected, and `fts_add_docs` is delete-then-insert, so duplicate ids in `fts_ids` are idempotent.

**Three things to fix before merge:**
1. `verify_manifest` is never called on the import path — I reproduced the "quietly short corpus" that the README and PR body both say is prevented. A full sha256 is well under a second, so there's no performance case for skipping it.
2. The roster digest is written to the manifest and never verified — while `main()` picks that roster up automatically, and the roster decides tokens and ACL.
3. Sharded mode prints a `corpus.jsonl` it did not write.

The rest (`--shard-records 0`, the manifest `documents` mismatch, the manifest-less directory, the uncovered sha256 branch, the `_scan` generator) have suggestions inline and are fine as follow-ups.

app/importer/byo.py
```diff
@@ -311,11 +313,70 @@ def _emails(rec: dict):
...
+                problems.append(f"{shard['path']}: {p.stat().st_size} bytes, manifest says "
+                                f"{shard['bytes']}")
+                continue
+            h = hashlib.sha256()
+            with open(p, "rb") as fh:
+                for chunk in iter(lambda: fh.read(1 << 20), b""):
+                    h.update(chunk)
+            if h.hexdigest() != shard["sha256"]:
+                problems.append(f"{shard['path']}: sha256 mismatch")
+    return problems
```
khj809, line 400: **[blocker] The roster digest is recorded but never checked.**

`_ByoWriter.close` writes `manifest["roster"]["sha256"]`, and `main()` picks that roster up automatically when no `--roster` is passed. But this function only walks `sources`, so the roster is never a candidate — I replaced it wholesale and `verify_manifest` still reported only the shard problem.

The roster is the CLOSED principal set: it decides who gets a token and what they can see. A swapped roster is a bigger problem than a swapped shard, and the manifest already carries the digest, so it's a few lines (973 KB, so no need to chunk it):

```suggestion
                problems.append(f"{shard['path']}: sha256 mismatch")
    # The roster is checked too, and not as an afterthought: it is the CLOSED principal set, so it
    # decides who gets a token and what they can see. `main()` picks it up automatically when a
    # directory is imported, which is precisely why its digest cannot go unverified.
    r = manifest.get("roster")
    if r:
        rp = source / r["path"]
        if not rp.exists():
            problems.append(f"{r['path']}: missing")
        elif rp.stat().st_size != r["bytes"]:
            problems.append(f"{r['path']}: {rp.stat().st_size} bytes, manifest says {r['bytes']}")
        elif hashlib.sha256(rp.read_bytes()).hexdigest() != r["sha256"]:
            problems.append(f"{r['path']}: sha256 mismatch")
    return problems
```

app/importer/erb.py
```diff
@@ -2259,7 +2365,8 @@ def main(argv: list[str]) -> int:
             question_ids.update(json.loads(line).get("expected_doc_ids", []))
 
     if args.export_byo:
-        counts = export_byo(settings, gen_dir, args.export_byo, question_ids=question_ids)
+        counts = export_byo(settings, gen_dir, args.export_byo, question_ids=question_ids,
+                            shard_records=args.shard_records)
         print(f"Converted {sum(counts.values())} documents -> {args.export_byo}/corpus.jsonl")
```
khj809: **[blocker] Sharded mode reports a file it did not write.**

With `--shard-records` there is no `corpus.jsonl` — the new test asserts exactly that (`assert not (out / "corpus.jsonl").exists()`). This line still prints that path as the success message, so the one thing a caller reads to find their output is wrong in the mode this PR adds.

```suggestion
        dest = (f"{args.export_byo}/data/<source>/part-*.jsonl.gz + manifest.json"
                if args.shard_records else f"{args.export_byo}/corpus.jsonl")
        print(f"Converted {sum(counts.values())} documents -> {dest}")
```

`--export-byo`'s help string (`"write a BYO-JSONL artifact (corpus.jsonl + roster.yaml)"`) and `export_byo`'s docstring summary line are now half-true for the same reason.

app/importer/erb.py
```diff
@@ -2238,6 +2341,9 @@ def main(argv: list[str]) -> int:
     ap.add_argument("--export-byo", type=Path, default=None, metavar="DIR",
                     help="write a BYO-JSONL artifact (corpus.jsonl + roster.yaml) into DIR instead "
                          "of building the DB; `app.importer.byo` loads it to an equivalent DB")
+    ap.add_argument("--shard-records", type=int, default=None, metavar="N",
+                    help="with --export-byo: write data/<source>/part-*.jsonl.gz shards of N "
+                         "records each plus manifest.json, instead of one corpus.jsonl")
     args = ap.parse_args(argv)
```
khj809, line 2436: **[minor] `--shard-records 0` (or negative) isn't rejected.**

`n >= self.shard_records` in `write()` is then always true, so every record gets its own shard. Confirmed:

    shards written with --shard-records 0: ['part-00000.jsonl.gz', 'part-00001.jsonl.gz', 'part-00002.jsonl.gz']

A stray `0` on the 531k-document corpus produces 600k files, which violates the `<100k files per repo` limit #18 cites — in the exact direction sharding was meant to fix.

```suggestion
    ap.add_argument("--shard-records", type=int, default=None, metavar="N",
                    help="with --export-byo: write data/<source>/part-*.jsonl.gz shards of N "
                         "records each plus manifest.json, instead of one corpus.jsonl")
    args = ap.parse_args(argv)
    if args.shard_records is not None and args.shard_records < 1:
        # 0 makes `n >= shard_records` always true: one shard per record, 600k files for the bench.
        ap.error("--shard-records must be at least 1")
```

app/importer/erb.py
```diff
@@ -2070,21 +2156,22 @@ def export_byo(settings, gen_dir, out_dir, *, question_ids=None) -> dict:
...
+        except Exception as e:  # one bad doc must not sink the conversion (as in load_structured)
+            failures.append((dsid, src, repr(e)))
+        if i % 25000 == 0:
+            print(f"  converted {i}/{len(records)} ({len(failures)} skipped)",
+                  file=sys.stderr, flush=True)
     if failures:
         print(f"  WARNING: skipped {len(failures)} docs. First few: {failures[:5]}",
               file=sys.stderr, flush=True)
     P.write_roster(out_dir / "roster.yaml", settings)
+    writer.close(counts=counts, documents=len(records))
```
khj809: **[minor] With any conversion failure the manifest's top-level `documents` contradicts its own per-source sum.**

`len(records)` counts documents the `except` above skipped; per-source `documents` is `counts[src]`, successes only. The two agree only as long as nothing is skipped; the moment one doc fails the manifest disagrees with itself — and that field is what a consumer reads to answer "did I get everything".

```suggestion
    writer.close(counts=counts, documents=sum(counts.values()))
```

If you want the skipped count recorded, a separate `skipped=len(failures)` field is the honest way to say it.

app/importer/byo.py
```diff
@@ -311,11 +313,70 @@ def _emails(rec: dict):
...
+
+    Records are split on ``\\n`` alone (``newline="\\n"``), for the reason ``jsonl_lines`` documents:
+    U+2028 and the vertical tab are ordinary characters inside a JSON string, and Python's universal
+    newlines would otherwise tear a valid record in half.
+    """
+    source = Path(source)
+    if source.is_dir():
+        manifest = json.loads((source / "manifest.json").read_text())
+        paths = [source / s["path"] for src in sorted(manifest["sources"])
+                 for s in manifest["sources"][src]["shards"]]
```
khj809: **[minor] A directory with no manifest dies as a raw traceback, and shard paths are trusted as given.**

`verify_manifest` says `"… not found"` for the same situation; this raises through:

    corpus_records on a dir with no manifest: FileNotFoundError [Errno 2] No such file or directory: '…/manifest.json'

And `source / s["path"]` opens whatever path the manifest names. The reason this format exists is that the artifact is *downloaded*, which makes its manifest untrusted data — rejecting absolute paths and `..` components is one line, and cheaper to add now than after the dataset is published. (Impact is low, it's a read either way, but so is the fix.)

```suggestion
    if source.is_dir():
        mf = source / "manifest.json"
        if not mf.exists():
            raise SystemExit(f"{mf} not found — pass a JSONL file, a .jsonl.gz, or a sharded "
                             f"artifact directory")
        manifest = json.loads(mf.read_text())
        # A downloaded manifest is untrusted data, so a shard path has to stay inside the artifact.
        paths = []
        for src in sorted(manifest["sources"]):
            for shard in manifest["sources"][src]["shards"]:
                p = (source / shard["path"]).resolve()
                if not p.is_relative_to(source.resolve()):
                    raise SystemExit(f"manifest names a shard outside {source}: {shard['path']}")
                paths.append(p)
```

tests/test_importer_byo.py
```diff
@@ -926,3 +926,82 @@ def test_byo_empty_readers_means_nobody(tmp_path):
...
+
+
+def test_verify_manifest_catches_a_tampered_shard(tmp_path):
+    """The digest is the whole point: a truncated or swapped download must fail before it loads."""
+    out = _shard_artifact(tmp_path)
+    assert byo.verify_manifest(out) == []
+    shard = next(out.glob("data/*/part-00000.jsonl.gz"))
+    shard.write_bytes(shard.read_bytes() + b"\x00")
+    problems = byo.verify_manifest(out)
+    assert len(problems) == 1 and "bytes" in problems[0]
```
khj809, line 985: **[test] This test never reaches the sha256 branch.**

Appending a byte trips the size check and `continue`s, so the digest comparison (byo.py:375-376) doesn't run anywhere in the suite — even though it is the reason the function exists, and the same-size content swap from comment A is exactly that branch.

XOR-ing the last byte preserves the length and changes only the digest. Verified — `['data/slack/part-00000.jsonl.gz: sha256 mismatch']`:

```suggestion
def test_verify_manifest_catches_a_tampered_shard(tmp_path):
    """The digest is the whole point: a truncated or swapped download must fail before it loads."""
    out = _shard_artifact(tmp_path)
    assert byo.verify_manifest(out) == []
    shard = next(out.glob("data/*/part-00000.jsonl.gz"))
    good = shard.read_bytes()
    shard.write_bytes(good + b"\x00")
    problems = byo.verify_manifest(out)
    assert len(problems) == 1 and "bytes" in problems[0]

    # A swap that keeps the byte count is what the digest is FOR — the size check above waves it
    # through, so without this case the sha256 branch never runs.
    shard.write_bytes(good[:-1] + bytes([good[-1] ^ 0xFF]))
    assert shard.stat().st_size == len(good)
    problems = byo.verify_manifest(out)
    assert len(problems) == 1 and "sha256 mismatch" in problems[0]
```

Two more gaps in the same vein:
- The **single `.jsonl.gz` input** this PR newly documents in the README (`python -m app.importer.byo corpus.jsonl.gz`) has no test — every `.gz` in the tests is reached through a shard directory.
- Nothing asserts that `load()` refuses a broken manifest. Fixing comment A gives you that test for free.

app/importer/byo.py
```diff
@@ -364,17 +425,24 @@ def load(path: Path, settings: Settings | None = None, reset: bool = True,
...
-                pass  # malformed lines are reported precisely in the main loop below
+                continue  # malformed lines are reported precisely in the main loop below
+            # Keep only what `_emails` reads, and drop the child rows' bodies: the addresses are
+            # kilobytes where the corpus is gigabytes.
+            _scan.append({
+                **{k: _rec[k] for k in ("author_email", "host_email", "readers") if k in _rec},
+                **{c: [{"author_email": r.get("author_email")} for r in (_rec.get(c) or [])
+                       if isinstance(r, dict)]
+                   for c in ("comments", "sentences", "messages", "replies") if c in _rec}})
     org_name, org_domain = _infer_org(_scan, settings)
```
khj809: **[nit] This list can be a generator — the comment's intent is only half-achieved.**

`infer_org` consumes `emails` exactly once, in the single `for` loop at `app/config.py:124`. This PR also dropped `_infer_org`'s `list[dict]` annotation, which reads like the same idea. So there's nothing to materialize.

As written it says "a sharded corpus is streamed rather than held in memory" and then builds 600k dicts, plus one `{"author_email": …}` dict per child row — millions of them at bench scale. The pass count is unchanged; only the memory becomes constant:

```suggestion
    # infer the org from the corpus (dominant email domain) before building any grants,
    # since public docs are granted to the org principal — see _infer_org. Its own pass, and a
    # streaming one: `infer_org` consumes the addresses exactly once (app/config.py), so a
    # generator means nothing but the domain counter is ever held.
    def _parsed():
        for _no, _ln in corpus_records(path):
            _ln = _ln.strip()
            if not _ln:
                continue
            try:
                yield json.loads(_ln)
            except json.JSONDecodeError:
                continue  # malformed lines are reported precisely in the main loop below
    org_name, org_domain = _infer_org(_parsed(), settings)
```

nuri-yoo:
All eight taken, pushed as `f0cf8bd`. The suite passes.

The first two are the same defect twice: a digest written and never compared. `verify_manifest` was reachable only from `--dry-run`, and the roster's digest was recorded in the manifest and read by nothing — while importing a directory picks that roster up on its own. Worse than the missing calls, the README and this body both described the check as happening on import, so the claim was false when I wrote it; I had run `--dry-run` and the import separately, seen both pass, and read that as one path. Both now go through one `_verify_or_die` shared with `--dry-run`, so they cannot drift apart again. Your repro on the fix:

```
shard now holds 2 records; manifest says 3
exit=1
INVALID: 1 shard problem(s) in /tmp/vfix/artifact
  data/slack/part-00000.jsonl.gz: 134 bytes, manifest says 137
DB created: False
```

The rest are in as written: the success message names what sharded mode actually wrote (help text and docstring too), `--shard-records` rejects anything below 1, `documents` is the sum of successes with `skipped` separate, a manifest-less directory says what to pass, shard paths must resolve inside the artifact, and org inference yields instead of building a list.

The test point was the sharpest one — appending a byte returns early on the size check, so the digest comparison never ran in the suite. It flips a byte in place now, and the two gaps you named beside it are covered along with the roster digest, a manifest-less directory, and a shard named outside the artifact.

Three places where I did not take the suggestion verbatim have replies inline.

app/importer/byo.py
nuri-yoo: Took the helper you suggested at the end rather than the inline block, and moved `--dry-run` onto it too. Duplicating the block was how the two paths came apart in the first place, so having one of them is the point; `_verify_or_die` returns early for a non-directory, which keeps both call sites to one line.

I did not add the record-count comparison you offered as a fallback: with verification in front of the load it would only fire on a manifest that is internally inconsistent, which is a different bug and one the export side should not be able to produce.

tests/test_importer_byo.py
nuri-yoo: The byte-flip case is in as written.

The roster case needed a different shape. `_shard_artifact` writes no `roster.yaml` and no `manifest["roster"]`, so a test that swaps the roster would skip rather than assert. Rather than change the shared helper and touch the other tests using it, the roster test writes the roster and its manifest entry itself, asserts clean, then swaps — which also gives it a positive case the helper alone would not have.

app/importer/erb.py
nuri-yoo: Took `sum(counts.values())`, and also the separate field you called the honest way to record skips: `close(..., skipped=len(failures))`.

Dropping the number entirely would have lost the only place the artifact says a conversion failure happened, since the warning goes to stderr and is not part of what a consumer receives.

app/importer/erb.py
```diff
@@ -2151,9 +2257,22 @@ def select_records(gen_dir: Path, question_ids: set[str] | None = None):
...
+        # API cannot produce — so it is dropped here, where every consumer of the corpus sees the
+        # same decision. Dropping it in only one of them is what made a converted artifact fail the
+        # BYO schema (`content: '' should be non-empty`) while a direct import accepted it.
+        if not (derive_title_content(raw)[1] or "").strip():
+            skipped_empty += 1
+            continue
+        yield src, dsid, raw
+    if skipped_empty:
+        print(f"  skipped {skipped_empty} document(s) with empty content", file=sys.stderr,
+              flush=True)
```
khj809: **[follow-up] The dropped document is recorded as a count, not an identity.**

No argument with the drop itself — there is nothing to serve, and dropping it in one path only is what made the artifact fail its own schema. But this is the only place the loss is recorded.

Running the rule over the whole bench: [1 document](https://github.com/onyx-dot-app/EnterpriseRAG-Bench/blob/main/generated_data/sources/slack/general/2960012345-office-owl-spotting-and-two-word-quip.json) of 511,962, and that dsid appears in 0 lines of the shipped slack shards — the whole of 285,605 → 285,604.

```json
{
  "channel": "general",
  "thread_ts": "2960012345",
  "first_message_ts": "2960012345",
  "last_message_ts": "2960012390",
  "participants": ["Ava", "Sam", "Priya", "Diego", "Maya", "OpsBot"],
  "messages": "",
  "title_field_name": "channel",
  "content_field_names": ["messages"],
  "dataset_doc_uuid": "dsid_33cbedf0709949fd9416c8c864a86cf2"
}
```

Damaged input rather than empty by design: six participants and two timestamps 55 seconds apart, asserting messages that aren't there.

Identifying it meant rescanning 3.65 GB of raw bench data — neither the artifact nor the logs led back to it. At one document that's a curiosity; if an ERB refresh makes it forty, nobody reading the diff can tell which ones went.

```suggestion
    skipped_empty = []
    for src, dsid, raw in iter_records(gen_dir / "sources"):
        if question_ids is not None and dsid not in question_ids:
            continue
        # The bench ships one document with no content at all (a slack thread whose `messages` is
        # ""). There is nothing to serve for it, and a thread with zero messages is a state the real
        # API cannot produce — so it is dropped here, where every consumer of the corpus sees the
        # same decision. Dropping it in only one of them is what made a converted artifact fail the
        # BYO schema (`content: '' should be non-empty`) while a direct import accepted it.
        if not (derive_title_content(raw)[1] or "").strip():
            # By id, not by count: this is the only place the loss is recorded, and identifying the
            # current single case from a bare number meant rescanning 3.65 GB of raw bench data.
            skipped_empty.append((src, dsid))
            continue
        yield src, dsid, raw
    if skipped_empty:
        print(f"  skipped {len(skipped_empty)} document(s) with empty content: "
              f"{', '.join(f'{s}/{d}' for s, d in skipped_empty[:20])}"
              f"{' …' if len(skipped_empty) > 20 else ''}", file=sys.stderr, flush=True)
```

A question rather than a request: should the artifact record the drop too? `manifest["skipped"]` is `len(failures)`, so it reads 0 here, and a consumer comparing against the bench's published 511,962 sees 511,961 with nothing explaining the difference. The plumbing is the awkward part — `select_records` is a generator and `export_byo` never learns what it dropped — so that shape is your call.

nuri-yoo: Recorded by identity, with the path next to the id. `_erb_path` is the part that pays: a dsid is not a location, so recording only the id leaves the same 3.65 GB scan to do, while the path makes it a `cat`. Both go in the manifest under `layers.converted`, beside `source_documents`, so the layer states its own arithmetic — source_documents == documents + excluded + failed. `failed` is a list of the same shape now; identity matters more for a failure than for a policy drop, not less. One correction in your favour: the shipped artifact does not carry `skipped` at all. It was built before that field existed, so the gap is wider than a count.

Three things came out of checking the claims, all worth putting here.

**The damage is not one document.** Reading it as damaged input rather than empty by design holds up, and it applies past the one that reached zero. Content lengths across the bench run 0, 3, 5, 9, then thirty-one more under fifty — no gap anywhere. I opened the four smallest: all four have metadata asserting substantially more than the body holds. The five-character one is a slack thread with six participants and timestamps 69 seconds apart whose entire body is `(end)`. So we drop where the damage reached empty and ship where it stopped short, and `reason` now records the mechanical rule ("content empty after strip") rather than a diagnosis it cannot support.

**A numeric threshold for the forty case would have had nothing behind it.** That was my first answer to your last paragraph and the distribution above kills it — there is no gap to put a cutoff in. `generated_data` has had exactly one commit since 2026-04-14, so the same bench has to yield the same set; the question is not how many are tolerable but whether the input changed. Known exclusions are declared by id, anything undeclared stops the run and names it, and `--allow-excluded N` gets past it once someone has read them. That also settles capping the log line: nothing caps the list, because the gate bounds how long it can get. The summary moved out of the generator tail while I was there — it only ran if the caller exhausted the iterator.

**`source_documents` on its own is disclosure, not verification.** It only means something against a known snapshot, and neither a ref nor a tag pins this one: `main` has moved past the commit that added generated_data, and the one tag (v1.0.0, 2026-03-29) predates it. `fetch_generated_data` records the tarball digest at fetch time and the export carries it into the layer.

Separate question, not this PR: should the converter report near-empty documents? Dropping them is a data-policy change and out of scope here, but serving `(end)` as a valid document is its own problem.

738f01a.

khj809: The empty message case above is invalid in terms of schema level, so it should be definitely dropped and reported. However, we cannot guarantee that the messages under specific numeric threshold are "near-empty" (even though the content is just `(end)`), and as you mentioned it's out of scope of this PR to make such policies.
I think it's user's responsibility to make the policies and filter out useless documents based on their policies, so let's just try to keep every data (except the empty one).

khj809 (approved):
LGTM. Thanks!

PR #44, opened 2026-08-07
Rename to Backlot, and make it installable from PyPI

khj809:
Renames the project and its Python module from `enterprise-mock`/`app` to **`backlot`**, and turns it into something you install rather than clone: a wheel with a CLI, two import surfaces, and a corpus inside it.

```bash
pip install backlot
backlot import --bundled     # the corpus that ships in the wheel
backlot serve                # http://127.0.0.1:8000
```

```python
import backlot
with backlot.mock_server() as m:          # no arguments — same bundled corpus, free port
    WebClient(token=m.token, base_url=f"{m.base_url}/slack/api/")
```

`backlot.integrations.{llamaindex,mirage}` carries the redirects for official clients that hardcode their vendor host — the ones a caller cannot express as an argument — so an existing pipeline needs two lines instead of a copied file.

## The surface this adds

| | |
|---|---|
| **Name** | `app/` → `backlot/`, `MOCK_` → `BACKLOT_`, and the last strings in served output (the OAuth service-account email, the Notion integration bot — whose derived id moves with its seed). The ASGI object stays `app` (`uvicorn backlot.main:app`), and `/_mock/*` paths are unchanged. |
| **CLI** | `backlot serve` and `backlot import [--type byo\|enterpriserag-bench]`. `import` parses only `--type` and hands the rest of argv to that importer's own `main`, so no flag or message is declared twice and `python -m backlot.importer.{byo,erb}` still takes the same ones. `python -m backlot` is the same CLI for an unactivated venv. |
| **Import surfaces** | `backlot.mock_server()` / `serve_or_connect()` for a server on a free port; `backlot.integrations` for the clients that hardcode their host (a client taking a base URL gets `f"{mock.base_url}/<prefix>"` at the call site, not a helper). |
| **Bundled corpus** | `backlot/data/hello.jsonl` — 136 documents covering every source, several containers each, cross-referenced as one company's incident, deal and hire. `--bundled` and `mock_server()` both load it; the default Docker build bakes it. |
| **Packaging** | `schemas/` and the GraphQL SDL live inside the package; path defaults resolve against cwd rather than the install location; `tests/test_packaging.py` fails if a shipped non-`.py` file has no `package-data` glob. |
| **Docker** | `docker build .` produces a server with the bundled corpus baked in — seconds, no network. `--target serve` is the corpus-less image for a deployment mounting its own `/app/data`. |
| **Deployment config** | `docker-compose.yml` reads a host `.env` (see `.env.example`) instead of carrying anyone's values. SQLite read tuning defaults to what a pip-installed mock wants rather than to any one deployment's box: `sqlite_mmap_mb` 256 (was 12288), `sqlite_cache_mb` 64 (was 256), `sqlite_busy_ms` 5000 (was 30000). **A corpus above a few GB should raise `BACKLOT_SQLITE_*` in that `.env`** — the busy timeout especially, since it is what waits out another connection's long write commit. |
| **`/health`** | Reports `source_documents` beside `documents`, because rows are not documents: parsing turns one Slack transcript into many message rows. A failed cache warm-up reports `status: "degraded"` and the reason, rather than `ok` beside counts that will never arrive. Still HTTP 200: the affected routes fall back to per-request queries and keep answering correctly, so failing a load balancer's probe would take down a working server. **A probe asserting `status == "ok"` needs to accept `degraded`.** |

37 commits · 153 files · **930 passed, 1 skipped** · ruff clean

## Structure

**One dataset stays inside one importer.** `backlot/config.py` no longer carries `raw_dir` / `dataset_repo` / `employee_yaml` — those belong to `backlot.importer.erb`, which keeps them on its own `BenchSettings` under the same env prefix. `atlassian_site` is gone as well, on a different ground: the site host is derivable, so `routers/atlassian.py` reads it off the request's `Host` header instead of taking it as configuration — one fewer setting whose value has to agree with the URL the caller already used. `backlot/importer/byo.py` imports nothing from `erb`: the two helpers it needed are Linear/Fireflies vocabulary, so they sit in `synth.py` beside the code that consumes them, and `synth.parse_transcript_text` is the declared inverse of `synth.fireflies_transcript_text` — the pair that *defines* a transcript's `content` is now one module's fixed point. That parser swap changes what a BYO transcript serves, which is worth stating plainly: it reads only the conventions the record format documents, so a header-shaped line (`Date: 2025-02-20`) names a speaker where `erb`'s own parser suppresses 38 such labels. Deliberate — suppressing the speaker also strands the line, since `content` is re-derived from the sentences, so bodies opening `Host:` or `Notes:` lost that line entirely from the served text and from FTS. Pinned by two tests and documented in `schemas/README.md`; the ERB path is untouched, gating on declared attendees that a BYO record does not have. The served OpenAPI title is the server's name, not a corpus's.

**Per-vendor error envelopes live together.** `backlot/errors/` holds one module per vendor (`atlassian.py`, `google.py`) plus a dispatch, so `main.py`'s two exception handlers ask the package instead of carrying a branch each, and a third envelope is a module plus one registry entry.

**Three corpora, three jobs, each enforced by a test.** `examples/bring-your-own-corpus/sample_corpus.jsonl` is the *field reference*: every field the schemas declare, populated at least once — pinned by a test that derives the field list from `backlot/schemas/*.json`, so a new schema field fails the suite until the example carries it. `backlot/data/hello.jsonl` is the *demo corpus*: volume and containers, pinned by a test driven off `store.SOURCE_TABLE` rather than a hand-written list. `tests/conftest.py::SAMPLE` stays the fixture. The split is stated once in `backlot/schemas/README.md`.

**One word per concept.** `source` for what Backlot serves — the codebase's own term (`source_type`, `store.SOURCE_TABLE`, `/_mock/openapi/<source>`). `vendor` for the real company and its artifacts: a vendor SDK, a vendor MCP server, "the real vendor APIs". `provider` was a third synonym for both and is gone from 16 files. The rule is in `CONTRIBUTING.md`.

## Documentation

`README.md` is rewritten: `pip install` quickstart, "why this exists" as the situations it is for, a Configuration table with every setting and its default, `Auth & tokens` as one paragraph plus the two `/_mock/*` responses, and the four client-integration sections grouped under `Example Usages`. Per-source behaviour is documented at the point it is implemented — router docstrings, the two `.graphql` headers, `linear_filters.py`, the SDL field descriptions — so the README does not restate it. Dataset-specific material lives with the dataset in `examples/import-enterpriserag-bench/`.

Corrections this carries: the Docker quickstart described the default build as "a small corpus" when it baked a ~500k-document dataset; `--all` / `--augment` are documented flags the bench importer does not have (and `--all` abbreviation-matches `--allow-excluded`, so `--build-arg BUILD_ARGS=--all` fails); `CONTRIBUTING.md` claimed `backlot/schemas/` validates *response* shapes, when it is the ingest contract and `response_model` pins responses; the Mirage snippet imported `point_mirage_at` from `examples/using-mirage/_mirage.py`, neither of which exists; GitHub `contents`/`trees`/`blobs`/`branches`/`commits` and Confluence `child/comment`, CQL `search` and `space/{key}` were served but absent from the endpoint table.

## Fidelity fixes

| | |
|---|---|
| **Drive `files.list` returned trashed files** | Real Drive excludes them unless `trashed = true` asks. Every `q`-bearing path honoured that, so the one call that returned trash was the plainest: no `q` at all. Fixed on both the listing and the count, since a count including dropped rows makes `nextPageToken` promise a page that does not exist. |
| **Gmail ids were not Gmail-shaped** | 16 lowercase hex below 2^63, reversible to the corpus document. Measured against the live API: `7fffffffffffffff` resolves, `8000000000000000` is refused. |
| **Google's error envelope** | `{"error": {...}}` per API family — `errors[]` always on Drive/Gmail and never on the editors, `status` on auth failures only for Drive. A missing header is 403 `PERMISSION_DENIED` on Drive and 401 `UNAUTHENTICATED` on Gmail. |
| **Two upstream client bugs** | `@linear/sdk` constructs `IssueSharedAccess` unconditionally, and `LinearReader` crashes on any unassigned issue — both against real `api.linear.app` too. Documented and pinned by tests. |

## Verification

A green suite is not the acceptance criterion for anything in this PR that concerns packaging or served output.

**Packaging** claims are checked by hand when packaging changes, not on every PR: build a wheel, install it into a venv outside the checkout, and run `backlot --version` / `import backlot.main` / `backlot import --bundled` from a directory the source tree cannot shadow. `pip install -e` leaves the source tree importable, so the suite structurally cannot answer a question about the wheel — `tests/test_cli.py` says so where a reader will hit it. `tests/test_packaging.py` keeps the `package-data` declaration exhaustive, though that table is no longer what decides the outcome: setuptools defaults `include_package_data` to true for a pyproject build, so with the globs deleted the wheel still ships all 15 assets.

CI is two jobs: `test` (pytest, then the TypeScript Linear example against a mock it starts itself) and `lint` (`ruff format --check`, `ruff check`).

**Served output** is verified by diffing it. The error-envelope refactor was checked with 12 probes covering all three Google families, both anonymous-request answers, a bad token, Atlassian 401/404/422 and a non-vendor path, run against `main` in a throwaway worktree and against this branch: byte-identical. The new corpus was checked by serving it and making one real call per source, including S3 through a SigV4-signing boto3 client and both GraphQL endpoints.

## Known items

- **`/health`'s `documents` counts only the root tables**, excluding the six served comment tables, so it is *lower* than the servable row count. The field names and the comment beside them are accurate; whether to publish a third number is a launch-copy decision.
- **`pip install backlot` requires a release.** The name is unclaimed on PyPI; the README documents the installed spelling.


backlot/config.py
```diff
@@ -0,0 +1,130 @@
...
+
+from functools import lru_cache
+from pathlib import Path
+
+from pydantic import model_validator
+from pydantic_settings import BaseSettings, SettingsConfigDict
+
+
+class Settings(BaseSettings):
+    model_config = SettingsConfigDict(env_prefix="BACKLOT_", env_file=".env", extra="ignore")
```
nuri-yoo, line 21: **1.** The `MOCK_` -> `BACKLOT_` migration is silent, and the settings it silently drops are the security ones. A leftover `MOCK_*` var is not an unknown key pydantic could warn about — it never matches the prefix, so it is simply never read. Verified against this branch's Settings class:

```
$ MOCK_ADMIN_TOKEN=super-secret MOCK_EXPOSE_TOKENS=false python -c "..."
admin_token   = admin-service-token      # the published default, ACL-bypassing
expose_tokens = True                     # /_mock/users serves every token again
warnings      = 0
```

The repo itself is fully migrated (`git grep -E '\bMOCK_[A-Z]'` is empty), so the blast radius is exactly the deployments we do not control: an operator's `.env`, systemd unit, or k8s manifest written against main. Suggestion: read the old prefix for one release and warn, or fail startup when any `MOCK_*` var is present — one `os.environ` scan in `lifespan`. At minimum the PR body and release notes should say "existing MOCK_* configuration is ignored, silently", because the two settings that revert are `admin_token` and `expose_tokens`.

README.md
```diff
@@ -1,190 +1,189 @@
...
 ```
 
-The image ships with a small corpus and generated ACLs already built in (no accounts, no data
-download at runtime), so it's ready to crawl immediately.
+That puts the `backlot` command on PATH. A server needs a corpus, and one is bundled with the
+package — 136 documents covering every source it serves — so there is nothing to fetch or write:
 
-## Why this exists
+```bash
+backlot import --bundled    # the bundled corpus -> data/mock.sqlite + data/tokens.yaml
```
nuri-yoo, line 24: **2.** The quickstart's second command does not exist. `cli.py` has no `--bundled` (grep: 0 hits); `import` takes a required positional `corpus`, so this exits 2 with a usage error. This README is the PyPI project page (`readme = "README.md"`), so the first thing a new user runs after `pip install` fails. The PR body's opening quickstart and the bundled-corpus table row make the same claim. Also `HELLO_CORPUS` is not exported (`__all__` is `MockServer, mock_server, serve_or_connect, url_from_argv`), so there is no public spelling of "import the bundled corpus" at all. Any one of the three fixes works: add the flag, export `HELLO_CORPUS` and document the one-liner around it, or rewrite the quickstart around `mock_server()`. The flag seems truest to the README as written.

backlot/integrations/mirage.py
```diff
@@ -0,0 +1,82 @@
...
+    Patches the source module — imported first so it exists — AND every already-imported
+    ``mirage.core.*`` that copied a same-named constant by value, which is what makes the redirect
+    order-independent. Idempotent.
+    """
+    importlib.import_module(source_module)
+    for mod in list(sys.modules.values()):
+        if not getattr(mod, "__name__", "").startswith("mirage.core."):
+            continue
+        for const, value in overrides.items():
+            if hasattr(mod, const):
```
nuri-yoo, line 53: **3.** The patchers fail silently, which is the one failure mode the package says it does not have. `backlot/integrations/__init__.py:9` promises: "Every patcher is idempotent and fails loudly if its seam disappears in a client upgrade, rather than silently letting a run that was supposed to hit the mock reach the real vendor." `_rebind_mirage_constants` is `if hasattr(mod, const): setattr(...)` with no counter and no error. Reproduced against a stand-in mirage whose constants were renamed:

```
point_google_at / point_github_at: returned, no exception
gmail constant still https://gmail.googleapis.com/gmail/v1
github constant still https://api.github.com
```

That run then talks to the real vendors with whatever credentials the caller loaded. `llamaindex.patch_notion_at` already has the right shape (count what you patched, raise RuntimeError on zero), so this is four lines. Related: there is no tests/test_mirage.py, and `point_google_at` / `point_github_at` appear in no test file, so nothing would have caught this.

backlot/importer/byo.py
```diff
@@ -638,7 +635,12 @@ def add(self, rec: dict, where: str = "record") -> None:
                     for s in given
                 ]
             else:
-                sentences = parse_fireflies_transcript(rec.get("content") or "")
+                # `synth.parse_transcript_text` and not an importer's own parser: it is the declared
+                # inverse of the `fireflies_transcript_text` two lines below, so the pair that
+                # defines `content` is one module's fixed point rather than two modules agreeing by
+                # convention. A record format documents the forms it accepts; salvaging some other
+                # dataset's transcript layout is that importer's job, not this loader's.
+                sentences = synth.parse_transcript_text(rec.get("content") or "")
```
nuri-yoo, line 643: **4.** Question: the BYO transcript parser swap changes served content, and the regressed case has no test. Base called `erb.parse_fireflies_transcript`; head calls `synth.parse_transcript_text`, and the result overwrites the record (`content = fireflies_transcript_text(sentences)`, line 651) and feeds `fireflies_speaker_stats` -> `analytics` (lines 848-849). The new parser dropped `_FF_NOT_SPEAKER` (38 labels), so auto-notes header lines mint speakers. Ran both parsers on the same input:

```
"Date: 2025-02-20\nDuration: ~52 min\nMaya: let's start with the numbers."
main:  [Maya]                                   # label lines excluded
head:  [Date, Duration, Maya]                   # two fabricated speakers
```

Those names then carry talk-time and word-count shares in the served analytics. The `[S00:12]` clock tolerance is gone too, while two real fixes arrived (the `https://` mis-split, prose before the first speaker no longer discarded). The ERB path is untouched — `erb.py` keeps its own parser. If the narrowing is intended ("only the conventions the record format documents"), the label-line case deserves a test pinning the intended output, because it is exactly the shape real auto-notes exports lead with; if not intended, porting `_FF_NOT_SPEAKER` restores it. Either way the PR body describes this as moving two helpers, and the parser swap with output changes is a third thing worth a sentence.

Dockerfile
```diff
@@ -18,40 +19,39 @@
...
 
 ENV PATH="/opt/venv/bin:$PATH" \
-    MOCK_DATA_DIR=/app/data \
+    BACKLOT_DATA_DIR=/app/data \
     PYTHONUNBUFFERED=1
 
 WORKDIR /app
 RUN python -m venv /opt/venv
 # pyproject declares `readme`, so README.md has to be present for the install to resolve.
 COPY pyproject.toml README.md ./
```
nuri-yoo: **5.** The image's dist has no license text, and the build will eventually fail outright. `pyproject.toml:12` declares `license-files = ["LICENSE"]`, but the build context copies only `pyproject.toml README.md` and `backlot/`. Building from exactly that context:

```
SetuptoolsDeprecationWarning: Cannot find any files for the given pattern.
        Pattern 'LICENSE' did not match any files.
```

and the resulting dist-info has no `licenses/LICENSE` (the full-tree wheel has it). So the image redistributes MIT code without the license text, and since `build-system.requires = ["setuptools>=77"]` has no ceiling, the day that deprecation becomes an error `docker build` breaks. Pre-existing on main, but this PR edits the adjacent COPY line and CI has no Docker job to catch it. One word: `COPY pyproject.toml README.md LICENSE ./`.

backlot/main.py
```diff
@@ -274,14 +276,34 @@ async def parse_slack_form(request: Request, call_next):
...
+    # `documents` because faithful parsing turns one Slack transcript into many message rows.
+    # Publishing only the larger of the two reads as inflation, which is why both are reported.
     counts = getattr(app.state, "doc_counts", None)
-    body = {"status": "ok"}
+    warm_error = getattr(app.state, "warm_error", None)
+    # `degraded`, not a non-200: the corpus is still served correctly — every consumer of the warm
+    # caches falls back to a per-request query — so failing the healthcheck would take down a server
+    # that works. What must not happen is `ok` alongside counts that will never arrive.
+    body = {
+        "status": "degraded" if warm_error else "ok",
```
nuri-yoo, line 293: **6.** The body says this was not done, and it was — which is good, but the body and a probe contract disagree with the code. PR body, Known items: "`_warm_caches` is a bare try/finally ... Not addressed here." Head wraps the warm-up, records `warm_error`, and `/health` returns `status: "degraded"` with the reason — tested at tests/test_cross_vendor.py:221. The change is right (base even called `connect_ro` outside the try). Two follow-ups: drop the Known-items paragraph, and document `degraded` in the README health section, since any probe asserting `body["status"] == "ok"` flips on the first warm-up failure while the HTTP status stays 200.

backlot/config.py
```diff
@@ -0,0 +1,130 @@
...
+    # --- paths --- (default supplied by _resolve_path_defaults above)
+    data_dir: Path = Path("data")
+
+    # --- identity / org ---
+    # The org name/domain are derived at import time from the dominant email domain in whatever
+    # was loaded, via infer_org() below. These are only the last-resort fallback for data that
+    # carries no emails; BACKLOT_ORG_NAME / BACKLOT_ORG_DOMAIN override the derivation entirely.
+    org_name: str = "example"
+    org_domain: str = "example.com"
+    # No `atlassian_site` here on purpose: the host in a Jira/Confluence `self` URL comes from the
```
nuri-yoo, line 47: **7.** Body sync: the `atlassian_site` removal is documented here in code but not in the body's removal list, which names `raw_dir` / `dataset_repo` / `employee_yaml` only. Worth adding, and it is one more entry in the same silently-ignored-env-var class as comment 1 (`BACKLOT_ATLASSIAN_SITE` from a main-era .env now does nothing).

backlot/config.py
```diff
@@ -0,0 +1,130 @@
...
+    # Sized for the corpus most people serve — their own, or the bundled one, which is under a
+    # megabyte. A multi-GB corpus wants all three raised, and a deployment that serves one says so
+    # explicitly (see the `environment:` block in docker-compose.yml) rather than every laptop
+    # inheriting numbers picked for the biggest DB anyone has run here.
+    #
+    # Memory-map the DB so reads come from the OS page cache instead of a syscall each — the main
+    # lever against the "slow first request after idle" cold-read hit. SQLite maps
+    # min(this, db size), so a small DB costs only its own size in address space; raise it to at
+    # or above the DB size to map a big one fully.
+    sqlite_mmap_mb: int = 256
```
nuri-yoo, line 76: **8.** Body sync: the SQLite defaults dropped (`sqlite_mmap_mb` 12288 -> 256, `sqlite_cache_mb` 256 -> 64, `sqlite_busy_ms` 30000 -> 5000). README table and .env.example carry the new values; the PR body says nothing about sqlite. New installs are fine and the small defaults suit a pip-installed mock. The interaction with comment 1 is the sharp edge: a deployment that tuned `MOCK_SQLITE_*` for a large corpus gets the ignored override and the lower default at once, and the busy window that rode out in-place FTS rebuild commits shrinks 30s -> 5s. A sentence in the body ("retune BACKLOT_SQLITE_* for corpora above a few GB") covers it.

nuri-yoo (commented):
Took this for a full spin — wheel built from the branch, installed on 3.11 and 3.14, import/serve/REST/GraphQL from outside the source tree. It works, and the packaging is solid.

The inline comments are mostly small, but a few of them are really release decisions we should settle together rather than code fixes: what we owe existing deployments on the `MOCK_` -> `BACKLOT_` rename (right now their security settings silently revert), what the official story for the bundled corpus is (the README teaches a flag that does not exist), and whether the Docker image is a distribution channel we stand behind (its dist currently ships without the license text). The mirage one (3) I would treat as a plain bug — the package promises loud failure and it fails silently.

Details and repro output are in each comment; 4 and 6 are questions, 7 and 8 just body/doc sync.

backlot/config.py
khj809: This PR does not consider any backward compatibilities deliberately. Since this library is not officially released, I think it's okay to ignore the previous env vars entirely.

README.md
khj809: `--bundled` is defined by `build_parser` in byo.py, and cli.py uses it.
Also `backlot import --bundled` does not exit abnormally in my environment. Could you check again after pulling the latest commit?

`HELLO_CORPUS` is just a path pointing to the bundled jsonl, and it does not meant to be used by itself publicly.

backlot/config.py
nuri-yoo: Fair. Dropping it.

README.md
nuri-yoo: I only looked at `cli.py`. Ran it, imports cleanly. Withdrawing this and the `HELLO_CORPUS` note.

Dockerfile
khj809: Added the LICENSE file in [3a268ba](https://github.com/brekkylab/backlot/pull/44/commits/3a268ba20d74fce22d4c37b3823f6640a90a9c46).

backlot/main.py
nuri-yoo: Confirmed, the code's right. My note was the two things around it: the body's Known items still says "Not addressed here", and `degraded` isn't in the README.

Dockerfile
nuri-yoo: Verified on `3a268ba` — rebuilt from that exact context and the wheel now carries `dist-info/licenses/LICENSE`, with the setuptools pattern warning gone. Thanks.

backlot/main.py
khj809: @nuri-yoo Yes, I misread the comment so removed mine.. Currently trying to rewrite the PR description.

backlot/config.py
khj809: Added in body.

backlot/config.py
khj809: Added in body.
Note that the previous default values were empirically decided based on the ec2 server where backlot.brekkylab.com is deployed.

backlot/main.py
khj809: Removed from known items.

backlot/integrations/mirage.py
khj809: Fixed in 09b9a8f. The sweep now tracks which names landed and raises listing the ones that did not, same shape as `patch_notion_at`. Also rebinds `DRIVE_UPLOAD_BASE`, which mirage ships and the shim never named — read-only mock, so pointing it at Backlot turns a write that would have reached real Drive into a 404.

Three tests, since there were none: the raise, the success path against the constants mirage actually ships, and one that fails when mirage grows a host constant the shim does not name. Both behavioural ones confirmed red beforehand.

One narrowing: a moved *module* already raised `ImportError` at `importlib.import_module`, so only a renamed constant was silent.

backlot/importer/byo.py
khj809: Addressed in 4445301, keeping the current reading and pinning it rather than porting `_FF_NOT_SPEAKER` back.

Your repro is exact. What tipped the decision is that the deny list also strands the line, not just the speaker: `content` is re-derived from the sentences, so a suppressed line is held by nothing. Ran the erb parser over BYO bodies (no attendees, so it takes its ungated branch) and compared `fireflies_transcript_text(parse(text))` against the input:

```
opens "Host: welcome everyone"   erb: content lost -> "Maya: numbers first."   synth: fixed point
opens "(recording starts)"       erb: content lost                             synth: fixed point
opens "Date: 2025-02-20"         erb: content lost                             synth: fixed point
```

So on the BYO path the deny list was dropping the first line from the served body and from FTS, and 34 of its 38 labels are words a person could be labelled by — `Host:` most obviously. A BYO record has no attendee list to gate on, which is what makes `Host: welcome everyone` and `Date: 2025-02-20` indistinguishable here; ERB keeps its own parser and its gate, so that path is unchanged.

Two tests in `test_synth.py`: the label-line reading, asserting the fabricated names do reach `fireflies_speaker_stats` so the consequence is visible, and the fixed point a ported-back deny list would break. `schemas/README.md` states the rule and the escape hatch (supply `sentences`). PR body now describes the parser swap as its own change, per your last line.

Not restored: the `[S00:12]` clock tolerance. `[A-Za-z]?` before the clock lives only in erb's `_FF_TS` and no BYO schema or doc ever documented that form — happy to add it if you think a corpus would write it.

backlot/integrations/mirage.py
nuri-yoo: Verified on `09b9a8f` — reran the renamed-constant stand-in I built for this and both entry points now raise, listing every name that missed. `DRIVE_UPLOAD_BASE` is a good catch.

backlot/importer/byo.py
nuri-yoo: You're right, and the round-trip is what settles it — I checked: on the BYO path the erb parser drops the opening line from `content` entirely (`Host: welcome everyone` → gone), while the current one is a fixed point. I was weighing fabricated speakers without weighing what the deny list costs, and `Host:` being in that list makes the point by itself.

No need for `[S00:12]` — it's an erb artifact, and `sentences` is the right way to carry timing.

backlot/main.py
nuri-yoo: Thanks — confirmed on all three.

backlot/config.py
nuri-yoo: Thanks — confirmed on all three.

backlot/config.py
nuri-yoo: Thanks — confirmed on all three.

nuri-yoo (approved):
Approving — everything I raised is closed.

One thing for the merge rather than the diff: you picked the old sqlite defaults off that ec2 box, and this PR lowers them — mmap 12288 → 256. `/health` there reports 6.4M documents. Whether it sets `MOCK_SQLITE_*` today or leans on the old defaults, after the rename it lands on 256 either way, so it wants `BACKLOT_SQLITE_*` set explicitly at redeploy.

khj809:
> One thing for the merge rather than the diff: you picked the old sqlite defaults off that ec2 box, and this PR lowers them — mmap 12288 → 256. `/health` there reports 6.4M documents. Whether it sets `MOCK_SQLITE_*` today or leans on the old defaults, after the rename it lands on 256 either way, so it wants `BACKLOT_SQLITE_*` set explicitly at redeploy.

You're right, and it's already redeployed with `BACKLOT_SQLITE_MMAP_MB=2048`, `BACKLOT_SQLITE_CACHE_MB=256`, and `BACKLOT_SQLITE_BUSY_MS=30000`. 
While investigating the configuration, I found that the sqlite used in there is compiled with `MAX_MMAP_SIZE=0x7fff0000` (~2GB), so any larger `BACKLOT_SQLITE_MMAP_MB` is silently clamped to this same value. That's why I reduced it from 12288 to 2048.

PR #45, opened 2026-08-07
byo roster: a person may hold more than one group

nuri-yoo:
## What

A roster person entry accepts an optional `groups` list on top of the department (or a contact's `group`). `load_roster` slugifies each name once and keeps first occurrence, so a repeated or unslugged name never doubles a membership row, and the closed-roster path unions every membership into `principals` / `group_members`. An entry without the list produces exactly the rows it produced before.

```yaml
departments:
  Engineering:
    - {name: Bo Ryu, email: bo.ryu@redwoodinference.com,
       groups: [proj-checkout-rework, res-emea-support]}
```

An email appearing in more than one entry holds the union of all of them, and a `contacts` entry never upgrades or demotes a `departments` account. A scalar `groups:` is one group, not a character sequence.

## Why

One department slot per person cannot say "also in a squad, a compliance register, and a region-scoped grant". A corpus with layered ACL groups therefore had to import without a roster, and the no-roster path reconstructs membership from activity: reply and mail-message authors inherit the group of every container they wrote in. Measured on a 2,800-record corpus with squad and region groups, 9.6% of the reconstructed memberships (384 of 3,992) named a group the person does not belong to, every one in the over-grant direction — someone who answered once in another team's public channel could read that team's group-scoped documents. With this change such a corpus ships a roster that states the real memberships, and the reconstruction path never runs.

## Regression safety

- No `groups` list in the file means byte-identical output: both pre-existing roster tests pass unchanged, and the ERB `--export-byo` round-trip still writes scalar `group` entries, which remain valid input.
- The only consumer of `load_roster`'s per-user shape was the closed-roster block in `byo.py` itself; the one test asserting that internal shape (`test_importer_erb.py`) is updated to the list form with the same semantics.

## Docs

`backlot/schemas/README.md` documents the list beside the roster format it belongs to: what it adds on top of the department membership, the union rule for a repeated email, and that a `contacts` entry never changes an account's token status.

## Tests

- `test_byo_roster_person_may_hold_many_groups`: department + extra groups, a contact with both `group` and `groups`, dedupe of a repeated department name, and that tokens are unaffected.
- `test_byo_roster_duplicate_entries_union_their_groups`: one person under two departments and again as a contact holds the union; a scalar `groups:` is one group; the contact entry never demotes the account.

- `test_byo_roster_a_stated_name_beats_a_derived_one`: derived-then-stated, stated-then-derived, and two stated names.
- `test_byo_roster_a_list_under_the_singular_group_key_is_read_not_crashed`: every entry kept, and the scalar path unchanged.

Full suite: `872 passed, 22 skipped`.

## Review follow-ups

Names do not union, so first-seen-wins was wrong for them: `name` always has a fallback derived from the address and therefore never looks absent, so an entry stating "Tomás Rré" lost to an earlier entry stating nothing and the corpus served "Tomas Rre". A stated name now wins over a derived one whichever is seen first; between two stated names the first still wins.

A list under the singular `group:` reached `slugify` and raised `AttributeError: 'list' object has no attribute 'lower'`. Its first entry is the primary membership and the rest are kept as extra ones.

The `if g` filter in `_merge` is gone — both operands come from `_groups`, which drops them already. The duplicate-entry test asserts a set plus a length rather than a list, since the rendered order rode on `yaml.safe_dump`'s key sorting.

The closed-mode gap — a record naming a group no roster entry holds gets its `doc_acl` row while the group never becomes a principal — is left for its own change: it predates this branch and wants a warning pass over the whole corpus.

## Base

Stacked on #44 (`rename/backlot`) so the diff is written against the `backlot/` layout; when #44 merges this retargets to `main`. Rebased onto the current `rename/backlot`, which is where `ruff` is pinned in the `dev` extra — the earlier red `wheel-install` run here was an unpinned `pip install ruff` picking up 0.16, whose formatter defaults changed. The last commit is `ruff format` over the two files this branch touches, under that pin.

### DIFF
--- a/backlot/importer/byo.py
+++ b/backlot/importer/byo.py
@@ -535,28 +535,35 @@ def load_roster(path) -> dict:
     def _groups(entry: dict, primary: str | None) -> list[str]:
         # The department (or `group`) membership first, then the entry's extra `groups`, each
         # slugified once — dict.fromkeys keeps first occurrence so a repeat never doubles a row.
-        listed = [slugify(g) for g in (entry.get("groups") or []) if g]
+        # A scalar `groups:` is one group, not a character sequence (the sibling `group:` field
+        # is a scalar, so the shape is a natural slip), and a bare number slugifies as text.
+        raw = entry.get("groups")
+        raw = [raw] if isinstance(raw, str) else (raw or [])
+        listed = [slugify(str(g)) for g in raw if g]
         return [g for g in dict.fromkeys(([primary] if primary else []) + listed) if g]
 
     users: dict[str, dict] = {}
+
+    def _merge(email: str, name: str, groups: list[str], token: bool) -> None:
+        # A person may appear more than once — two departments, or a department entry plus a
+        # contact carrying extra register memberships. Membership is the UNION: replacing the
+        # entry silently dropped the earlier groups, and a `readers: [group:...]` clause then
+        # wrongly denied the person the feature was written for. The first-seen name stays;
+        # a contact never upgrades an account, but it never demotes one either.
+        cur = users.get(email)
+        if cur is None:
+            users[email] = {"name": name, "groups": groups, "token": token}
+            return
+        cur["groups"] = [g for g in dict.fromkeys(cur["groups"] + groups) if g]
+        cur["token"] = cur["token"] or token
+
     for dept, people in (data.get("departments") or {}).items():
         for p in people or []:
-            users[p["email"]] = {
-                "name": p.get("name") or _display_name(p["email"]),
-                "groups": _groups(p, slugify(dept) or None),
-                "token": True,
-            }
+            _merge(p["email"], p.get("name") or _display_name(p["email"]),
+                   _groups(p, slugify(dept) or None), True)
     for p in data.get("contacts") or []:
-        # A contact never upgrades an account: `departments` is the authenticating roster, so a
-        # duplicated email keeps its token rather than losing it to listing order.
-        users.setdefault(
-            p["email"],
-            {
-                "name": p.get("name") or _display_name(p["email"]),
-                "groups": _groups(p, slugify(p["group"]) if p.get("group") else None),
-                "token": False,
-            },
-        )
+        _merge(p["email"], p.get("name") or _display_name(p["email"]),
+               _groups(p, slugify(p["group"]) if p.get("group") else None), False)
     return {"org": data.get("org"), "org_domain": data.get("org_domain"), "users": users}
 
 
--- a/backlot/importer/byo.py
+++ b/backlot/importer/byo.py
@@ -559,11 +559,19 @@ def load_roster(path) -> dict:
 
     for dept, people in (data.get("departments") or {}).items():
         for p in people or []:
-            _merge(p["email"], p.get("name") or _display_name(p["email"]),
-                   _groups(p, slugify(dept) or None), True)
+            _merge(
+                p["email"],
+                p.get("name") or _display_name(p["email"]),
+                _groups(p, slugify(dept) or None),
+                True,
+            )
     for p in data.get("contacts") or []:
-        _merge(p["email"], p.get("name") or _display_name(p["email"]),
-               _groups(p, slugify(p["group"]) if p.get("group") else None), False)
+        _merge(
+            p["email"],
+            p.get("name") or _display_name(p["email"]),
+            _groups(p, slugify(p["group"]) if p.get("group") else None),
+            False,
+        )
     return {"org": data.get("org"), "org_domain": data.get("org_domain"), "users": users}
 
 
--- a/backlot/importer/byo.py
+++ b/backlot/importer/byo.py
@@ -510,6 +510,8 @@ def load_roster(path) -> dict:
         departments:                      # authenticating users -> a bearer token each
           Engineering:
             - {name: Ava Chen, email: ava.chen@redwoodinference.com}
+            - {name: Bo Ryu, email: bo.ryu@redwoodinference.com,
+               groups: [proj-checkout-rework, res-emea-support]}
         contacts:                         # principals with NO token (display-only)
           - {name: Zoe Newperson, email: zoe.newperson@redwoodinference.com, group: engineering}
 
@@ -519,17 +521,29 @@ def load_roster(path) -> dict:
     ``contacts`` are people a corpus names who are not accounts — they own and read documents but
     cannot authenticate, the distinction ``tokens.yaml`` draws.
 
+    A person may belong to more than one group — a squad, a compliance register, a region-scoped
+    grant — which one department slot cannot say. An entry's ``groups`` list adds those memberships
+    on top of the department (or ``group``) one; it never replaces it, so a directory that only
+    knows departments and a roster that also states squads produce the same department rows.
+
     With a roster, `principals`, `group_members` and `tokens.yaml` come from it ALONE: a record's
     `author_email` / `readers` become references into it, and an address absent from it (a Slack
     handle, an outside sender) stays a plain address instead of becoming an account with a token.
     """
     data = yaml.safe_load(Path(path).read_text()) or {}
+
+    def _groups(entry: dict, primary: str | None) -> list[str]:
+        # The department (or `group`) membership first, then the entry's extra `groups`, each
+        # slugified once — dict.fromkeys keeps first occurrence so a repeat never doubles a row.
+        listed = [slugify(g) for g in (entry.get("groups") or []) if g]
+        return [g for g in dict.fromkeys(([primary] if primary else []) + listed) if g]
+
     users: dict[str, dict] = {}
     for dept, people in (data.get("departments") or {}).items():
         for p in people or []:
             users[p["email"]] = {
                 "name": p.get("name") or _display_name(p["email"]),
-                "group": slugify(dept) or None,
+                "groups": _groups(p, slugify(dept) or None),
                 "token": True,
             }
     for p in data.get("contacts") or []:
@@ -539,7 +553,7 @@ def load_roster(path) -> dict:
             p["email"],
             {
                 "name": p.get("name") or _display_name(p["email"]),
-                "group": (slugify(p["group"]) if p.get("group") else None),
+                "groups": _groups(p, slugify(p["group"]) if p.get("group") else None),
                 "token": False,
             },
         )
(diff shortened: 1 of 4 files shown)


backlot/importer/byo.py
```diff
@@ -519,29 +521,56 @@ def load_roster(path) -> dict:
...
+
+    def _merge(email: str, name: str, groups: list[str], token: bool) -> None:
+        # A person may appear more than once — two departments, or a department entry plus a
+        # contact carrying extra register memberships. Membership is the UNION: replacing the
+        # entry silently dropped the earlier groups, and a `readers: [group:...]` clause then
+        # wrongly denied the person the feature was written for. The first-seen name stays;
+        # a contact never upgrades an account, but it never demotes one either.
+        cur = users.get(email)
+        if cur is None:
+            users[email] = {"name": name, "groups": groups, "token": token}
```
khj809: The old departments loop assigned `users[p["email"]] = {...}`, so a later entry's *stated* name won. First-seen-wins now lets a guess beat a stated name:

```yaml
Engineering: [{email: tomas.rre@redwoodinference.com}]              # -> _display_name
Security:    [{name: Tomás Rré, email: tomas.rre@redwoodinference.com}]
```

gives `name: "Tomas Rre"`; base gave `"Tomás Rré"`. Unreachable through `--export-byo` (`erb.py:382` writes each email once, always with `name`), so hand-written rosters only — but the union rule is right for groups and wrong for names, since `name` always has a fallback value and so never looks absent. Worth preferring an explicitly stated name over a derived one.

backlot/importer/byo.py
```diff
@@ -519,29 +521,56 @@ def load_roster(path) -> dict:
...
-        users.setdefault(
+        _merge(
             p["email"],
-            {
-                "name": p.get("name") or _display_name(p["email"]),
-                "group": (slugify(p["group"]) if p.get("group") else None),
-                "token": False,
-            },
+            p.get("name") or _display_name(p["email"]),
+            _groups(p, slugify(p["group"]) if p.get("group") else None),
```
khj809: `_groups` handles a scalar under plural `groups:`, and the reasoning given ("the sibling `group:` field is a scalar, so the shape is a natural slip") cuts both ways — but a list under singular `group:` reaches `slugify` and raises `AttributeError: 'list' object has no attribute 'lower'`:

```yaml
contacts: [{name: A, email: a@x.com, group: [engineering, security]}]
```

Adding the plural field makes this slip likelier, so it seems worth normalizing the singular path the same way (or raising a roster error that names the file and key).

backlot/importer/byo.py
```diff
@@ -1263,8 +1292,8 @@ def _scanned():
         # The roster IS the principal set: users, the groups they belong to, and the memberships
         # between them. Nothing the records referenced adds to it.
         users = {e: u["name"] for e, u in roster_data["users"].items()}
```
khj809, line 1324: *Follow-up, not this diff.* Closed mode reassigns `groups` from the roster alone, so a record naming a group no roster entry holds still gets its `doc_acl` row while the group never becomes a principal:

```
doc_acl        [('c1', 'group', 'proj-typo-rework')]
principals     ['engineering']      # <- not one of them
group_members  []
```

The document is readable by nobody and the import says nothing. Pre-existing, but this PR makes "declare it in the roster" the remedy, which makes the silent mismatch the likeliest way to misconfigure the new field. A closed-mode warning listing record-referenced groups absent from the roster would finish the feature.

khj809 (commented):
Reviewed the diff and verified the claims locally: the per-user shape has no consumer outside `byo.py`, a roster without `groups:` produces identical output, the suite is green (`924 passed, 1 skipped` here — the skip count differs from the body's `22` because more optional tests run in my env; nothing fails), and `ruff format --check` / `ruff check` are both clean under the `~=0.15.22` pin.

The design holds up — closed mode reassigns `groups`/`memberships` from the roster alone, so before this there really was no way to state a layered membership. Three notes inline: two small ones on the diff, one follow-up.

Nits, take or leave: `_merge`'s `if g` filter is dead — both inputs already come from `_groups`, which filters (inside `_groups` it is load-bearing, it catches a slug collapsing to `""`). And `test_byo_roster_duplicate_entries_union_their_groups` asserts exact list order, which rides on `yaml.safe_dump`'s key sorting; downstream it becomes a set, so the order carries no meaning.

backlot/importer/byo.py
nuri-yoo: Fixed in `bf96c5b`. A stated name now wins over a derived one whichever is seen first; between two stated names the first still wins, as memberships do. `test_byo_roster_a_stated_name_beats_a_derived_one` covers all three orders.

backlot/importer/byo.py
nuri-yoo: Fixed in `bf96c5b`. The first entry becomes the primary membership and the rest are kept as extra ones — a silent loss seemed no better than the crash. Scalar path unchanged. `test_byo_roster_a_list_under_the_singular_group_key_is_read_not_crashed`.

backlot/importer/byo.py
nuri-yoo: Agreed, and left for its own change as you suggested — it predates this branch and wants a warning pass over the whole corpus, not a line here. Noted in the PR body so it does not get lost.

nuri-yoo:
Both nits taken, in `bf96c5b`. The `if g` filter in `_merge` is gone (both operands come from `_groups`, which drops them already), and the duplicate-entry test asserts a set plus a length instead of a list — the order rode on `yaml.safe_dump`'s key sorting, and downstream it becomes a set anyway.

Also rebased onto `rename/backlot` at `eec35a5` (no conflicts). Suite `872 passed, 22 skipped`; CI green.

backlot/importer/byo.py
```diff
@@ -519,30 +521,91 @@ def load_roster(path) -> dict:
...
+        raw = entry.get("groups")
+        raw = [raw] if isinstance(raw, str) else (raw or [])
+        # A list written under the singular `group:` keeps every entry: `_primary` took the
+        # first as the primary membership, and dropping the rest silently would trade one
+        # crash for one quiet loss.
+        singular = entry.get("group")
+        if isinstance(singular, (list, tuple)):
+            raw = list(raw) + list(singular)
+        listed = [slugify(str(g)) for g in raw if g]
+        return [g for g in dict.fromkeys(([primary] if primary else []) + listed) if g]
```
khj809: Two shape gaps are left in `_primary` / `_groups`, and one helper closes both.

**`group:` on a department entry now depends on its YAML shape.** `_groups` reads `entry.get("group")` for any entry, but only when it is a list — so the same key on the same entry behaves two ways:

```yaml
Engineering:
  - {email: a@x.com, group: [squad-x, squad-y]}   # -> engineering, squad-x, squad-y
  - {email: b@x.com, group: squad-x}              # -> engineering          <- dropped, silently
```

Before this commit both were dropped, consistently. The scalar is the more natural spelling, and it is the one that now vanishes without a word — the outcome this commit's own message argues against ("trading the crash for a silent loss would be no improvement").

**A non-string scalar under `groups:` raises.** The guard is `isinstance(raw, str)`, so anything else falls through to `raw or []` and gets iterated:

```yaml
groups: 2024      # TypeError: 'int' object is not iterable
groups: true      # TypeError: 'bool' object is not iterable
```

That contradicts the code's own intent twice over: `groups:` accepts a string scalar, and `test_byo_roster_duplicate_entries_union_their_groups` already asserts `2024` slugifies to `"2024"` inside a list.

Same root cause — shape normalization lives in two places with different rules. Reading each field once, in one place, makes the shapes uniform and comes out 4 lines shorter:

```suggestion
    def _slugs(raw) -> list[str]:
        """One roster field — ``group:`` or ``groups:`` — in any shape, as group ids.

        The two fields differ only in which membership they NAME, never in how they are
        written, so one reader serves both and neither shape can be a slip. A sequence is
        each of its entries; anything else is a single group, so a scalar (a string, or a
        bare ``2024`` that YAML hands over as an int) is one group rather than a character
        sequence or a ``TypeError``. Slugified once here, so no caller does it twice."""
        items = raw if isinstance(raw, (list, tuple)) else [raw]
        return [s for s in (slugify(str(g)) for g in items if g) if s]

    def _primary(raw) -> str | None:
        # The membership an entry's own `group:` names. A list means its first entry; the rest
        # are not dropped, `_groups` reads the whole field again as extra memberships.
        return next(iter(_slugs(raw)), None)

    def _groups(entry: dict, primary: str | None) -> list[str]:
        # The primary membership first — a department entry's is its department, a contact's is
        # its own `group:` — then everything either field names. dict.fromkeys keeps first
        # occurrence, so a group repeated across the two fields never doubles a row, and a
        # `group:` on a department entry is read rather than silently dropped.
        listed = _slugs(entry.get("group")) + _slugs(entry.get("groups"))
        return list(dict.fromkeys(([primary] if primary else []) + listed))
```

Verified on this branch: `929 passed, 1 skipped`, `ruff format --check` and `ruff check` clean. Every case the branch already guarantees is byte-identical — contact `group: [a, b]`, scalar `groups:`, mixed `[str, int]`, a `groups:` entry repeating the department, a slug collapsing to `""`, and an entry with neither field. What changes is only the two above: a department's scalar `group:` becomes an extra membership, and a bare-number `groups:` slugifies instead of raising.

Worth extending `test_byo_roster_a_list_under_the_singular_group_key_is_read_not_crashed` with a department entry carrying a scalar `group:` and a `groups: 2024` — both are unasserted today.

khj809 (commented):
The two fixes check out — I probed both. A stated name now wins whichever entry it sits in (and a contact's stated name upgrades a department entry's derived one without touching its token), the `_stated` key is popped so the returned shape stays three keys, and a list under `group:` keeps every entry rather than erroring — better than what I suggested. Suite and lint green here too (`929 passed, 1 skipped` in my env).

One thing the fix introduced, plus a latent crash next to it — both in one inline suggestion below.

backlot/importer/byo.py
nuri-yoo: addressed in 4f0e4bf. Took the suggestion as written.

Confirmed each independently against the old reader: a department's scalar group: gave ['engineering'], and groups: 2024 raised. The new test test_byo_roster_group_and_groups_read_the_same_in_every_shape covers both, plus a group named across both fields collapsing to one row. 876 passed, 24 skipped, ruff clean.

khj809 (approved):
LGTM
Please just resolve the conflicts.

PR #46, opened 2026-08-07
byo: store a provided jira key / github number, serve it verbatim

nuri-yoo:
## What

A jira record accepts an optional `key` ("PAY-101") and a github issue/PR record an optional `number`. A provided value is stored verbatim and the routers and reverse indexes read the stored column, so the exact string a document cites is a valid lookup. This is the treatment Linear's `identifier` already gets, for the same reason: an id the API serves has to be resolvable, and an id a document cites has to exist on the served surface.

When the corpus writes none, the column stays NULL and the id is derived — but derived in one place. `_build_index` carries `doc_id -> number` and `doc_id -> key` alongside the reverse maps, and the routers read those rather than deriving the id a second time, so what a document advertises and what fetches it cannot disagree. Unlike Linear's identifier, neither id needs storing to be resolvable; keeping synthesized values out of the column is what makes "provided" mean exactly "non-NULL", which the behaviors below depend on.

## Why

A corpus that writes its own tracker keys into document text ("see PAY-101", a PR number cited from a retro) currently dangles at serve time: the Jira router synthesizes `<initials+hash>-<n>` keys per request, so the exact string a document cites is never a valid lookup, and cross-source citation chains break at the API even though they are internally consistent in the corpus. Measured on a 2,800-record pilot corpus with 1,092 citation edges, none of the jira/github citations resolve on the served surface today, while the linear ones all do — the asymmetry is just that Linear honors a provided identifier and jira/github do not.

## Behavior notes

- github `file` rows keep `number` NULL in the table, not only in the reverse index — the schema says a file row's number is ignored, so a stored number can never shadow a real issue or PR. The github covering index turns partial over the same predicate its one consumer filters by.
- A provided key's prefix becomes the project's key everywhere: the index aliases each container to that prefix, and `fields.project.key`, JQL `project = PAY`, the project picker and the synthesized keys of its keyless issues all speak it. Without this, an issue served `PAY-7` beside `fields.project.key: PAYMENTS<hash>`, `project = PAY` returned zero rows and the picker never listed `PAY` — the corpus cited a key its own API could not navigate.
- A project may therefore carry a **mix** of provided and absent keys, which is the normal case for a corpus that cites keys in document text: only the cited issues spell one out, and the rest answer at `PAY-<n>`. A key's prefix is a fact about the container, so writing a synthesized one into a keyless row at load stamped it with the wrong prefix — the row cannot know a sibling provides `PAY-7` — and the project's spelling then came down to whichever row sorted first by `doc_id`. Deriving it at build time instead means the container decides once, with every one of its rows visible, and under `--append` a later shard can still introduce the first provided key in a project whose earlier rows are already written.
- **A provided id is read from the field that declares it.** `meta` is documented free-form and seeds the extras, so `meta: {"number": 3}` claimed issue 3 exactly as a top-level `number` would — a spelling no schema describes, that shadows a real issue, and that the uniqueness rule below would then refuse an import over. Both ids are popped from that seed: ordinary `meta` content, never promoted to a served column.
- **A provided id has one holder.** Two records claiming the same github number in a repository, or the same jira key, stop the import and name the record already holding it: whichever the index picked, the other would be unreachable at the only id it advertises, and the loader is the one place that sees every row. The same number in a different repository is a different id and loads.
- **A provided prefix is held 1:1 with its project** (review, 2nd round). Real Jira project keys are unique, and an issue key's prefix is its project's key; the index can only give a tie to one side with `setdefault`. Two projects each providing `PAY-` keys left `project = PAY`, the picker and the role endpoint silently serving only the first, and `PAY-1` beside `BILL-2` in one project served an issue key whose prefix was not its `fields.project.key`. Both directions are refused at load, naming the holder like a duplicate full key, and the claim is seeded from the DB so it holds across `--append`.
- **A derived id that is already taken moves.** A provided id claims its spelling first, and a row whose derived id collides with it goes deterministically to the next free number in that repository, or sequence number in that project — and answers there, because the router reads the same map. Left as an alias that `setdefault` dropped, such a row advertised a number that fetched the provider and was reachable at nothing. A provider still answers at its own derived spelling too where that is free, handed out last so it never takes the number a displaced row wanted.
- `connect_rw`'s self-heal list gains both columns, so an existing DB opened for writing is altered in place instead of failing on the first INSERT. A DB from before the columns existed is *served* through `connect_ro`, where no migration runs: the index build retries column-less and the routers read through `synth.stored()`, which treats a missing column as an expected state rather than an error. Such a DB serves exactly what the previous version served.

## Example corpus

`test_example_corpus_populates_every_field_the_schemas_declare` derives its field list from the schemas, so the two new properties have to appear in `examples/bring-your-own-corpus/sample_corpus.jsonl` — that file's whole job is to show every field a record may carry. The gateway issue and its PR now carry `number` 412 and 418, and all three payments issues carry a `PAY-` key. Every issue rather than one: the corpus's own `issuelinks` already cites `PAY-42`, so `PAY` is the spelling its documents speak, and an example whose ids match the text around them is the point of the file.

`backlot/schemas/jira.schema.json` and `github.schema.json` each state where an omitted id comes from, since that is what tells an author they can write only the ids their documents cite.

## Regression safety

A corpus that provides neither field keeps NULL columns and synthesized ids: `synth.stored()` falls through to the value the router synthesized per request, with no provided key anywhere a project's prefix stays the synthesized one, and the ERB conversion path sets neither field and goes through the same loader. "Byte-identical to the previous version" holds for every row no collision reaches, which is narrower than this paragraph first claimed (review, 2nd round): two rows whose *derived* ids collide previously both advertised one number, with one of them unreachable at it; now the later row moves to the next free number — and can displace a row whose own derived id never collided, since the next free number may be that row's spelling. At realistic scale this is common, not marginal: within a 200-issue project, the odds of at least one derived-derived pair are ~89% for jira (9,000-value key space) and ~20% for a github repo (90,000). The contract chosen instead — a served id is a function of the container's whole row set, stable across restarts of the same DB, free to move when `--append` changes the set, self-consistent at every moment — is stated in `_free_number`'s docstring and pinned by `test_byo_a_derived_id_may_move_across_append_but_always_fetches_its_advertiser`. Full suite: `891 passed, 24 skipped`.

## Tests

- `test_byo_provided_tracker_ids_are_stored_and_served`: provided key/number stored and served verbatim through `GET /atlassian/rest/api/3/issue/PAY-7` and `GET /github/repos/{owner}/{repo}/pulls/4242`, a row that provided neither storing NULL and still resolving, file row staying NULL.
- `test_byo_provided_key_prefix_is_the_project_key`: the issue payload, JQL and the project picker all speak the provided prefix.
- `test_byo_one_provided_key_sets_the_prefix_for_its_keyless_siblings`: a project mixing one provided key with a keyless issue that sorts FIRST by `doc_id` — the arrangement that produced the wrong prefix. The keyless issue serves and resolves at `PAY-<n>`, both issues come back from `project = PAY`, and the picker lists only `PAY`. It and the assertion above fail with 404 without the loader change.
- `test_byo_meta_cannot_smuggle_a_tracker_id`: `meta: {"number": …}` / `meta: {"key": …}` leave both served columns NULL.
- `test_byo_two_records_cannot_claim_one_tracker_id`: a duplicate provided number, and a duplicate jira key, each stop the import naming the record that holds it; the same number in another repository loads.
- `test_byo_colliding_provided_id_claims_the_spelling_and_the_other_row_moves`: a provided id equal to another row's derived one. The provider keeps it, the displaced row advertises a different number, and fetching that number returns the displaced row rather than the provider.
- `test_byo_a_displaced_jira_key_moves_and_stays_reachable`: the same contract on the jira side — the displaced issue moves to another sequence number in its own project and answers there.
- `test_byo_a_provided_number_wins_whichever_doc_id_sorts_first`: the same collision with the OTHER doc_id order — the row whose hash was taken sorts first. Fails with 404 without the loader change.
- `test_byo_server_boots_read_only_on_a_pre_column_db`: a DB with both columns and their indexes dropped boots and serves synthesized ids through the read-only path.

## Base

Stacked on #44 (`rename/backlot`) so the diff is written against the `backlot/` layout; when #44 merges this retargets to `main`. Rebased onto `eec35a5`; the one conflict was main.py's import block, where this branch's `sqlite3` met that branch's `logging`. Independent of #45 — either can land first. Rebased onto the current `rename/backlot`, which is where `ruff` is pinned in the `dev` extra — the earlier red `wheel-install` run here was an unpinned `pip install ruff` picking up 0.16, whose formatter defaults changed.


backlot/routers/github.py
```diff
@@ -301,7 +306,7 @@ async def pull_reviews(owner: str, repo: str, number: int, request: Request):
     if row is None:
         raise HTTPException(status_code=404, detail="Not Found")
     ab = _api_base(request)
-    number = synth.github_number(row["doc_id"])
+    number = synth.stored(row, "number") or synth.github_number(row["doc_id"])
```
khj809: **`pull_reviews` derives the number instead of using the resolved one.**

Every other handler moved to `_issue_number(index, row)`, but this one still derives. When a PR provides no number, its derived number is already held by a row that *did* provide it, and `_build_index` displaces this row, the response body links to the displacing PR: `GET /repos/o/r/pulls/501/reviews` resolves the right row but emits `pull_request_url` and `html_url` pointing at `/pulls/500` — a different document. The path number can also be an alias rather than the canonical id, so reading it back is not equivalent to what `_pr_obj` serves for the same row.

```suggestion
    number = _issue_number(request.app.state.index, row)
```

backlot/importer/byo.py
```diff
@@ -941,6 +958,49 @@ def insert(
...
+            )
+            if provided_id is not None:
+                scope = container if src == "github" else ""
+                claim = (src, scope, str(provided_id))
+                if claim in self.tracker_ids:
+                    label = "number" if src == "github" else "key"
+                    raise SystemExit(
+                        f"{where}: {label} {provided_id!r} is already claimed by "
+                        f"{self.tracker_ids[claim]!r}" + (f" in repo {scope!r}" if scope else "")
+                    )
```
khj809: **A repeated `(source_type, doc_id)` collides with itself.**

The loader explicitly supports two records sharing a `(source, doc_id)` — the comment at `byo.py:662` says both are written and `INSERT OR REPLACE` leaves the later one, and notes one real corpus has four such pairs, **one of them within jira**. If such a repeated jira record carries a provided `key`, the second `add()` finds *its own* `did` under `claim` and aborts the whole import with `key 'PAY-7' is already claimed by 'jira-x'`, naming the same doc_id it is currently inserting. The claim is only violated when a *different* doc_id holds it.

```suggestion
                claimed = self.tracker_ids.get(claim)
                if claimed is not None and claimed != did:
                    label = "number" if src == "github" else "key"
                    raise SystemExit(
                        f"{where}: {label} {provided_id!r} is already claimed by "
                        f"{claimed!r}" + (f" in repo {scope!r}" if scope else "")
                    )
```

backlot/importer/byo.py
```diff
@@ -643,6 +645,12 @@ def __init__(
         # Linear relations name a target by doc_id and are resolved after the whole corpus is read,
         # since a target may appear on a later line.
         self.lin_links = []
+        # Tracker ids the corpus provided, so a second record claiming one is refused here.
+        # Two records providing the same github number or jira key used to load without a
+        # word: one of them then owned the id in the reverse index and the other was
+        # unreachable at the only id it advertised. The loader is the one place that sees
+        # every row, so it is the only place the claim can be checked at all.
+        self.tracker_ids = {}  # (source_type, container, id) -> doc_id
```
khj809, line 653: **The uniqueness check does not survive `--append`.**

`tracker_ids` lives on `_Loader`, which `load_records` builds fresh per invocation, and the `reset=False` path never seeds it from the existing DB. Two shards appended in separate runs can each provide `PAY-7` (or number 412 in the same repo) and both load without a word. `_build_index`'s `setdefault` then hands the key to whichever doc_id sorts first, while `_issue_key`/`_issue_number` return `synth.stored(row, ...)` for the loser — so it advertises an id that fetches somebody else and is reachable at nothing. That is exactly the failure this PR sets out to remove, just moved behind `--append`.

The HubSpot association resolver already has the shape for this (fall back to a `SELECT` against the stored row when the in-memory map only sees this shard). Suggest a method on `_Loader`:

```python
    def seed_tracker_ids(self) -> None:
        """Re-read the ids already in the DB, so a claim holds ACROSS runs too.

        A fresh _Loader per `load_records` only sees the shard it is loading; without this, two
        shards appended separately could each provide PAY-7 and neither would be told. Same
        fall-back-to-the-stored-row reasoning as the HubSpot association resolver below.
        """
        for src, col in (("github", "number"), ("jira", "key")):
            scope = store.grouping_col(src) if src == "github" else "''"
            for row in self.conn.execute(
                f"SELECT doc_id, {col} AS v, {scope} AS scope FROM {store.table(src)} "
                f"WHERE {col} IS NOT NULL"
            ):
                self.tracker_ids[(src, row["scope"], str(row["v"]))] = row["doc_id"]
```

called where the loader is constructed (`byo.py:1371`, outside this diff):

```python
    loader = _Loader(conn, org, org_domain, closed=closed, validate=validate)
    if not reset:
        loader.seed_tracker_ids()
```

`test_byo_two_records_cannot_claim_one_tracker_id` is single-run only — worth a sibling that loads two shards with `reset=False` and asserts the second is refused.

khj809 (commented):
Three issues on the tracker-id materialization. (1) is a wrong link in a response body, (2) aborts an import on a corpus shape the loader explicitly supports, and (3) lets the new uniqueness check lapse under `--append` — the failure this PR sets out to remove.

backlot/routers/github.py
nuri-yoo: Confirmed and fixed in `f851f80`, suggestion taken as written.

Reproduced first: a PR providing no number, whose derived number a second PR in the same repo provides, is displaced by the index to 60015. `GET /repos/acme/core/pulls/60015/reviews` returned that PR's own reviews while citing the other one — `pull_request_url` `.../pulls/60014`, `html_url` `.../pull/60014#pullrequestreview-38995` — and `/pulls/60014` is the thief. Both links now carry the resolved number.

`test_byo_a_pull_requests_reviews_link_to_the_number_the_index_resolved` pins it; it fails on the previous revision.

backlot/importer/byo.py
nuri-yoo: Confirmed and fixed in `f851f80`, suggestion taken as written.

Reproduced: the same jira record twice, both carrying `key: PAY-7`, aborted with `line 2: key 'PAY-7' is already claimed by 'jira-x'` — the doc_id it was inserting. `test_byo_a_repeated_document_may_restate_its_own_tracker_id` covers it.

One residual worth naming, unchanged by this fix and present before it: if a repeat restates a *different* id (`jira-x` as `PAY-7`, then `jira-x` as `PAY-8`), the row ends as `PAY-8` but the `PAY-7` claim stays in the map, so a later record providing `PAY-7` is refused and told `jira-x` holds it when it no longer does. Reaching it needs the corpus to contradict itself about one document, which the duplicate case documented at `byo.py:721` does not produce — a direct import of the same documents writes the same id. Left alone, since refusing a self-contradicting corpus is defensible and only the message is off. Happy to have the map drop a superseded claim if you would rather it did.

backlot/importer/byo.py
nuri-yoo: Confirmed and fixed in `f851f80`, `seed_tracker_ids` taken as suggested.

Reproduced: two shards loaded in separate runs, each providing `PAY-7`, both loaded without a word. Served, `jira-b` advertised `PAY-7` and fetching it returned `jira-a`.

Two things I checked on the seeding. The github scope resolves through `grouping_col`, so the same number in another repo still loads while the same number in the same repo is refused. And this fix and the repeated-doc_id one above depend on each other: seeding without that one makes re-appending an already-loaded shard abort on its own claim.

`test_byo_a_tracker_id_claim_survives_append` is the sibling you asked for — two shards with `reset=False` and the second refused, then the first shard re-appended to show a re-run still works, then the cross-repo case.

backlot/main.py
```diff
@@ -61,19 +106,81 @@ def _build_index(conn) -> dict:
...
+        f"FROM {store.table('jira')} ORDER BY doc_id",
+    )
+    for r in j_rows:
+        if r["key"]:
+            idx["jira"].setdefault(r["key"], r["doc_id"])
+            idx["jira_key"][r["doc_id"]] = r["key"]
+            prefix = str(r["key"]).rsplit("-", 1)[0]
+            if prefix:
+                idx["jira_project_keys"].setdefault(r["container"], prefix)
+                idx["jira_project_containers"].setdefault(prefix.upper(), r["container"])
```
khj809, line 173: **A provided prefix is not held 1:1 with a container, in either direction.** Both are corpus shapes the loader can see and reject; the index can only pick one with `setdefault`.

*Two containers, one prefix.* The loader only rejects duplicate *full* keys, so two projects can each provide `PAY-` keys. Both get `jira_project_keys[c] = "PAY"` and both advertise `key: "PAY"`, but `jira_project_containers["PAY"]` holds only the first by doc_id — so `project = PAY` JQL and `/rest/api/3/project/PAY/role/{id}` silently return only the first project's issues, while the second's issues all serve `PAY-<n>` and `fields.project.key: PAY`. Real Jira guarantees project keys are unique.

*One container, two prefixes.* `PAY-1` and `BILL-2` in project `payments` is accepted. `setdefault` keeps whichever sorts first, so `_jira_issue` serves `key: "PAY-1"` with `fields.project.key: "BILL"` — the issue key's prefix is not its project's key, which the comment just below and `atlassian.py:126` both assert as an invariant.

backlot/main.py
```diff
@@ -35,6 +36,38 @@
 )
 
 
+# Numbers a derived id may take before the search gives up. A repository with more
+# documents than this has run out of the space `synth.github_number` draws from, which is
+# a corpus-scale problem and not something a silent duplicate should paper over.
+_PROBE_LIMIT = 90_000
+
+
+def _free_number(taken: dict, container: str, start: int) -> int:
```
khj809, line 45: **A served id now depends on the set of rows in the table, not on the row's own doc_id** — so ids move across `--append`.

Rows `a` (derived 5) and `c` (derived 5) build to `{a:5, c:6}`. Append `b` (derived 6) and it rebuilds to `{a:5, b:6, c:7}`: `c` moves from 6 to 7, so a number an agent recorded yesterday fetches a different document today.

This also qualifies the PR's "byte-identical to the previous version" claim for corpora that provide no ids at all — `c` previously served 5 (shadowed), now it serves 6. Jira is the sharper case: `synth.jira_key` draws from only 9,000 values, so a ~200-issue project has an ~89% chance of at least one displacement.

No suggested diff — displacement is the price of not shadowing, and which one wins is a design call. But it is worth saying in the docstring which was chosen, since "it stays put whenever nothing collides" reads as a stability guarantee that `--append` does not keep.

khj809 (commented):
Two more, both about the index rather than the loader. Neither blocks the three fixes in f851f80.

backlot/main.py
nuri-yoo: Confirmed in both directions and fixed in `0eff060`.

Reproduced first. Two projects, one prefix: `payments` PAY-1 + `billing` PAY-2 loaded without a word, then `project = PAY` returned only PAY-1, the picker listed two projects under one key, and `/rest/api/3/project/PAY/role/10002` served only the first project's actors. One project, two prefixes: BILL-2 + PAY-1 in `payments` both served `fields.project.key: BILL`, and `project = PAY` and `project = BILL` returned the same two issues.

Both are refused at load now, next to the full-key claim: the loader holds `container -> prefix` and `prefix -> container`, a violation names the holder, and `seed_tracker_ids` rebuilds both maps from the stored keys so the claim holds across `--append` — a later shard bringing `BILL-` keys into a project already answering at `PAY` is the same violation whether the earlier keys arrived this run or a previous one.

`test_byo_two_projects_cannot_share_a_provided_key_prefix` and `test_byo_one_project_cannot_provide_two_key_prefixes` pin both directions, single-run and across `reset=False`; both fail on the previous revision with DID NOT RAISE.

backlot/main.py
nuri-yoo: Right on both counts, and the exception set is wider than the example: displacement cascades. With `gh-000000` and `gh-014031` both deriving 85973 and `gh-054074` deriving 85974, the displaced row reaches 85974 first and `gh-054074` ends up serving 85975 — a row that never collided with anything moved. So "byte-identical" was already wrong for any corpus a collision reaches, no append required; your jira estimate checks out (~89% for a 200-issue project, ~20% for a same-size github repo).

No code change; docs and a pin in `0eff060`. `_free_number`'s docstring now states the choice — a served id is a function of the container's whole row set, stable across restarts of the same DB, free to move when `--append` changes the set, and the invariant kept is self-consistency at every moment, since the routers read the same map. The stays-put sentence is gone. `test_byo_a_derived_id_may_move_across_append_but_always_fetches_its_advertiser` pins exactly your a/b/c shape plus the cascade; it passes unchanged on the previous revision, so it documents the choice rather than a fix. The PR body's regression-safety paragraph is rewritten to match.

If stability across `--append` ever becomes a requirement, the door is a stored resolved column, separate from the provided one — deliberately not taken here for the reason in `_Loader.add`'s comment: a synthesized jira key written at load freezes a prefix that a later shard may still be the first to provide.

khj809 (commented):
LGTM
It seems `main.py` became too long, need to split index-related logics and `/_mock/*` routes into separate files. I'll handle it later.

khj809 (approved):

PR #47, opened 2026-08-11
byo validation: an error names the record and the condition that applies to it

nuri-yoo:
Loading a 23,040-record corpus was refused with dozens of lines reading

```
line 773: <root>: 'path' is a required property
```

and nothing else. Two things a corpus author needs were missing from every one of them.

## Which record

A line number is not an identifier. A sharded artifact numbers each shard from one, so "line 773" is ambiguous across the shards of a single corpus, and it is not what the author's own build wrote down. The id is. Messages now open with the record's `doc_id`, or its `title` where it carries no id, and keep `<root>` only for a record that gave nothing to be called by.

## Which rule

`path` is required of source files and of nothing else, but the rule is an `if`/`then`, so jsonschema reports it against the whole document with a bare message — the condition sits in the schema for the reader to go and find. It is recoverable: the failing branch's schema path runs through a `then` (or an `else`), and the `if` beside it is the predicate. So the line now ends with the condition:

```
gh-1: 'path' is a required property when subtype is "file"
```

A condition shape the renderer does not recognise adds no clause rather than a guessed one, and an unconditional rule reads exactly as it did before.

## Second commit: the two shapes that were guessed at rather than skipped

Neither is reachable from the schemas shipped today — the one conditional in `backlot/schemas/` is a single-field `if`/`then` with no `else` — so both are about what the next conditional someone writes will produce.

An `else` was negated field by field. `else` is in force when the whole predicate fails, so a two-field condition rules out the pair together: a record with `a == "x"` and `b != "y"` takes the `else` branch, and the old rendering (`a is not "x" and b is not "y"`) told that record's author the rule did not apply to it. It now reads `unless a is "x" and b is "y"`, which is the branch's actual entry condition.

`then` was read as a keyword wherever it appeared in the schema path, including under `properties`, where it is a field name a corpus author picked. A schema carrying fields named `if` and `then` produced a whole clause out of two ordinary fields — `properties/then/type` rendered as `" when mode is \"fast\""`. The branch keyword is now honoured only where a keyword can stand, which rules out the five keywords whose value is a map of author-chosen names (`properties`, `patternProperties`, `dependentSchemas`, `$defs`, `definitions`).

Alongside: a label is put on one line and says when it has been cut. `--dry-run` prints one problem per line and a title is free text that may hold a newline; a label silently truncated at 60 characters reads as a shorter id, which the author then searches the corpus for and never finds.

## Third commit: bracketing the field path

Putting the record's name in front of the field path left nothing between them but a space, and a label is free text. A record titled `subtype` whose `subtype` field was wrong read `subtype subtype: 9 is not one of [...]`, which names the field twice and marks neither of them as the record. The path is now bracketed, so the label ends where the bracket opens — `subtype [subtype]: ...`. A record-wide error still has no path and so gains no brackets, which keeps the common line as short as it was.

The path's own spelling is left alone. `associations/0`, `comments/0` and `messages/0` are pinned by three tests that predate this branch, so the slash form is this suite's convention and the delimiter was the only thing missing. Deferring to jsonschema's own `err.json_path` (`$.comments[0].body`, where `$` would close the label by itself) was the alternative, and it rewrites a spelling the suite already settled on for no gain the bracket does not give.

## Scope

`backlot/validation.py` and `tests/test_schema.py` only. `record_errors` keeps its signature and its return type, and the two call sites in `backlot/importer/byo.py` (fail-fast on load, and `--dry-run`) join and print the strings without parsing them, so the format change reaches the reader and stops there.

## Known, and deliberately left alone

Where conditionals nest, only the outermost condition is reported. All the enclosing conditions hold when the rule fires, so the clause is incomplete rather than wrong.

A condition over a wide `enum` renders every member, so a 30-value enum would append a 267-character clause. No shipped schema has one — the sole conditional keys off a `const` — and a clause that long stops saving the reader the trip to the schema it exists to save. Worth a cap the first time a schema needs it.

## Verification

`pytest`: 877 passed, 24 skipped. `ruff format --check` and `ruff check` clean on both touched files at the pinned 0.15.22.

### DIFF
--- a/backlot/validation.py
+++ b/backlot/validation.py
@@ -60,6 +60,23 @@ def _subschema(schema: dict, parts: list):
     return node
 
 
+# Keywords whose value is a MAP of subschemas, keyed by names the schema author chose. A
+# `then` sitting directly under one of these is such a name — a field a corpus may carry — and
+# the `if` beside it is another one, so reading the pair as a conditional would invent a clause
+# out of two ordinary fields.
+_SUBSCHEMA_MAPS = frozenset(
+    {"properties", "patternProperties", "dependentSchemas", "$defs", "definitions"}
+)
+
+
+def _branch_index(parts: list) -> int | None:
+    """Index of the first ``then``/``else`` that stands where a keyword can stand."""
+    for i, p in enumerate(parts):
+        if p in ("then", "else") and (i == 0 or parts[i - 1] not in _SUBSCHEMA_MAPS):
+            return i
+    return None
+
+
 def _when_clause(schema: dict, schema_path) -> str:
     """`" when subtype is \\"file\\""` for a rule a condition put in force, else ``""``.
 
@@ -70,16 +87,19 @@ def _when_clause(schema: dict, schema_path) -> str:
     ``else``), and the ``if`` beside it is the predicate — so it is stated instead of left for the
     reader to find in the schema. Shapes this does not recognise add no clause rather than a
     guessed one.
+
+    An ``else`` reads "unless", not a field-by-field negation: it is in force when the whole
+    predicate fails, so a two-field condition rules out the pair together, and negating each
+    field separately would claim something the schema does not say.
     """
     parts = list(schema_path)
-    branch = next((i for i, p in enumerate(parts) if p in ("then", "else")), None)
+    branch = _branch_index(parts)
     if branch is None:
         return ""
     owner = _subschema(schema, parts[:branch])
     cond = owner.get("if") if isinstance(owner, dict) else None
     if not isinstance(cond, dict):
         return ""
-    negated = "not " if parts[branch] == "else" else ""
     clauses = []
     for field, spec in (cond.get("properties") or {}).items():
         if not isinstance(spec, dict):
@@ -91,8 +111,9 @@ def _when_clause(schema: dict, schema_path) -> str:
         else:
             continue
         shown = " or ".join(json.dumps(v, ensure_ascii=False) for v in allowed)
-        clauses.append(f"{field} is {negated}{shown}")
-    return " when " + " and ".join(clauses) if clauses else ""
+        clauses.append(f"{field} is {shown}")
+    lead = " unless " if parts[branch] == "else" else " when "
+    return lead + " and ".join(clauses) if clauses else ""
 
 
 def _record_label(rec: dict) -> str:
@@ -100,11 +121,16 @@ def _record_label(rec: dict) -> str:
 
     A line number is not that on its own — a sharded artifact numbers each shard from one, and
     the id is what the author's own build wrote down and can grep for.
+
+    One line, and truncation says so. ``--dry-run`` prints one problem per line and a title is
+    free text that may hold a newline; a silently cut label, meanwhile, looks like a shorter id
+    that the author would search for and never find.
     """
     for key in ("doc_id", "title"):
         value = rec.get(key)
         if isinstance(value, str) and value.strip():
-            return value.strip()[:60]
+            label = " ".join(value.split())
+            return label if len(label) <= 60 else label[:59] + "…"
     return ""
 
 
--- a/backlot/validation.py
+++ b/backlot/validation.py
@@ -92,6 +92,9 @@ def _when_clause(schema: dict, schema_path, failing) -> str:
     predicate fails, so a two-field condition rules out the pair together, and negating each
     field separately would claim something the schema does not say.
 
+    A predicate field is reported as ``(or absent)`` unless the ``if`` requires it, because
+    ``properties`` alone constrains a value without demanding the field.
+
     ``failing`` is the subschema that actually reported the error, and the walk is only trusted
     when it arrives there. No shipped schema uses ``$ref`` yet, so this guard changes nothing
     today; it is what keeps the clause honest once one does.
@@ -123,7 +126,16 @@ def _when_clause(schema: dict, schema_path, failing) -> str:
         else:
             continue
         shown = " or ".join(json.dumps(v, ensure_ascii=False) for v in allowed)
-        clauses.append(f"{field} is {shown}")
+        req = cond.get("required")
+        if isinstance(req, list) and field in req:
+            clauses.append(f"{field} is {shown}")
+        else:
+            # `properties` constrains a field's value, it does not require the field. An `if` with
+            # no sibling `required` therefore SUCCEEDS for a record that omits the field, so `then`
+            # is in force there too -- and saying only `when subtype is "file"` sends the author
+            # looking through a record for a field they never set. `unless` needs it for the mirror
+            # reason: an absent field satisfies the predicate, so it does not lift the rule.
+            clauses.append(f"{field} is {shown} (or absent)")
     lead = " unless " if parts[branch] == "else" else " when "
     return lead + " and ".join(clauses) if clauses else ""
 
--- a/backlot/validation.py
+++ b/backlot/validation.py
@@ -116,7 +116,12 @@ def _when_clause(schema: dict, schema_path, failing) -> str:
     if not isinstance(cond, dict):
         return ""
     clauses = []
-    for field, spec in (cond.get("properties") or {}).items():
+    # Guarded like `owner`, `cond` and `spec` are: `_load_schemas` only json.loads the files, and
+    # `check_schema` runs in a test rather than at import, so a hand-edited schema reaches here
+    # unvalidated. A non-object `properties` must not turn "report this bad record" into a
+    # traceback on every record -- least of all on `--dry-run`, whose whole job is to report.
+    props = cond.get("properties")
+    for field, spec in (props if isinstance(props, dict) else {}).items():
         if not isinstance(spec, dict):
             continue
         if "const" in spec:
--- a/backlot/validation.py
+++ b/backlot/validation.py
@@ -77,7 +77,7 @@ def _branch_index(parts: list) -> int | None:
     return None
 
 
-def _when_clause(schema: dict, schema_path) -> str:
+def _when_clause(schema: dict, schema_path, failing) -> str:
     """`" when subtype is \\"file\\""` for a rule a condition put in force, else ``""``.
 
     A conditional requirement reports at the document root with a bare message: a 23,000-record
@@ -91,10 +91,22 @@ def _when_clause(schema: dict, schema_path) -> str:
     An ``else`` reads "unless", not a field-by-field negation: it is in force when the whole
     predicate fails, so a two-field condition rules out the pair together, and negating each
     field separately would claim something the schema does not say.
+
+    ``failing`` is the subschema that actually reported the error, and the walk is only trusted
+    when it arrives there. No shipped schema uses ``$ref`` yet, so this guard changes nothing
+    today; it is what keeps the clause honest once one does.
     """
     parts = list(schema_path)
     branch = _branch_index(parts)
     if branch is None:
+        return ""
+    # A schema path is reported WITHOUT the `$ref` hops it passed through, so a path from inside a
+    # referenced subschema reads as a root-relative one and this walk lands on whatever conditional
+    # happens to sit at those keys. The last element is the failing keyword, so its parent is the
+    # schema that failed; anything else means the walk went somewhere other than the real branch.
+    # `is not`, not `!=`: two structurally identical branches under different conditionals must not
+    # count as a match, and jsonschema hands the subschema object through rather than copying it.
+    if _subschema(schema, parts[:-1]) is not failing:
         return ""
     owner = _subschema(schema, parts[:branch])
     cond = owner.get("if") if isinstance(owner, dict) else None
@@ -155,7 +167,9 @@ def record_errors(rec: dict) -> list[str]:
         # not close: a record titled `subtype` with a bad `subtype` field read
         # "subtype subtype: ...", naming the field twice and marking neither.
         head = f"{label} [{loc}]" if label and loc else (label or loc or "<root>")
-        msgs.append(f"{head}: {err.message}{_when_clause(SERVICE_SCHEMAS[st], err.schema_path)}")
+        msgs.append(
+            f"{head}: {err.message}{_when_clause(SERVICE_SCHEMAS[st], err.schema_path, err.schema)}"
+        )
     return msgs
 
 
--- a/backlot/validation.py
+++ b/backlot/validation.py
@@ -130,7 +130,10 @@ def _when_clause(schema: dict, schema_path, failing) -> str:
             allowed = spec["enum"]
         else:
             continue
-        shown = " or ".join(json.dumps(v, ensure_ascii=False) for v in allowed)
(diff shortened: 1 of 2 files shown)


backlot/validation.py
```diff
@@ -49,17 +49,113 @@ def _validator(source_type: str) -> Draft202012Validator:
...
+
+    An ``else`` reads "unless", not a field-by-field negation: it is in force when the whole
+    predicate fails, so a two-field condition rules out the pair together, and negating each
+    field separately would claim something the schema does not say.
+    """
+    parts = list(schema_path)
+    branch = _branch_index(parts)
+    if branch is None:
+        return ""
+    owner = _subschema(schema, parts[:branch])
```
khj809, line 114: **A schema path arrives without its `$ref` hops, so this walk can land on a different conditional and name a field the record never mentions.**

`_when_clause` walks `SERVICE_SCHEMAS[st]` by `parts[:branch]` on the assumption that `err.schema_path` is root-relative. jsonschema 4.26 omits the `$ref` keyword from the reported path, so a path from inside a referenced subschema reads as a root path. When the truncated path collides with a conditional the root happens to have at those same keys, the clause names the wrong condition:

```python
ROOT = {
    "type": "object",
    "$ref": "#/$defs/Sub",
    "allOf": [{"if": {"properties": {"mode": {"const": "fast"}}, "required": ["mode"]},
               "then": {"required": ["speed"]}}],
    "$defs": {"Sub": {"allOf": [{"if": {"properties": {"kind": {"const": "blob"}}, "required": ["kind"]},
                                 "then": {"required": ["sha"]}}]}},
}
# validating {"kind": "blob"}:
#   err.schema_path -> ['allOf', 0, 'then', 'required']   # the $ref hop is gone
#   err.schema      -> {'required': ['sha']}
# today:  'sha' is a required property when mode is "fast"      <- no 'mode' in the record at all
# fixed:  'sha' is a required property
```

`absolute_schema_path` is identical here, so it is not a way out. No shipped schema uses `$ref` yet, but this is the same forward-looking class 3bd2a77 took on — and unlike the two shapes that commit fixed, this one emits an actively wrong clause rather than none, which is the case the docstring says it wants to avoid.

The path's last element is the failing keyword, so its parent must be the schema that actually failed. Comparing there (rather than at `parts[:branch + 1]`) also keeps the clause for an error nested deeper inside `then` — verified: `then/properties/path/type` still renders `9 is not of type 'string' when subtype is "file"`.

```suggestion
    parts = list(schema_path)
    branch = _branch_index(parts)
    if branch is None:
        return ""
    # A schema path is reported WITHOUT the `$ref` hops it passed through, so a path from inside
    # a referenced subschema reads as a root-relative one and this walk lands on whatever
    # conditional happens to sit at those keys. The last element is the failing keyword, so its
    # parent is the schema that failed; anything else means the walk went somewhere else.
    if _subschema(schema, parts[:-1]) is not failing:
        return ""
    owner = _subschema(schema, parts[:branch])
```

GitHub applies only one suggestion block per comment, so the two companions:

```diff
-def _when_clause(schema: dict, schema_path) -> str:
+def _when_clause(schema: dict, schema_path, failing) -> str:
```

```diff
-        msgs.append(f"{head}: {err.message}{_when_clause(SERVICE_SCHEMAS[st], err.schema_path)}")
+        msgs.append(
+            f"{head}: {err.message}"
+            f"{_when_clause(SERVICE_SCHEMAS[st], err.schema_path, err.schema)}"
+        )
```

`is not`, not `!=`, is deliberate: two structurally identical branches (`{"required": ["path"]}` under two different conditionals) must not count as a match, and jsonschema hands the subschema object through rather than copying it. A schema `_subschema` cannot reach at all returns `None`, which fails the check and suppresses the clause — the safe direction.

Test-side: the four direct `_when_clause` calls in `tests/test_schema.py` need the third argument. Three take the literal subschema; `test_an_else_branch_rules_out_the_condition_whole` must pass the object itself (`two_field["allOf"][0]["else"]`) for the identity check.

backlot/validation.py
```diff
@@ -49,17 +49,113 @@ def _validator(source_type: str) -> Draft202012Validator:
...
+        if not isinstance(spec, dict):
+            continue
+        if "const" in spec:
+            allowed = [spec["const"]]
+        elif isinstance(spec.get("enum"), list) and spec["enum"]:
+            allowed = spec["enum"]
+        else:
+            continue
+        shown = " or ".join(json.dumps(v, ensure_ascii=False) for v in allowed)
+        clauses.append(f"{field} is {shown}")
```
khj809: **`when <field> is <const>` is rendered from `if.properties` alone, but `properties` does not imply presence — so the clause is wrong for a record that omits the field.**

Drop the `required` beside the `if` and `then` still fires for a record with no `subtype`:

```python
{"if": {"properties": {"subtype": {"const": "file"}}}, "then": {"required": ["path"]}}
# validating {}:
# today:  'path' is a required property when subtype is "file"            <- record has no subtype
# fixed:  'path' is a required property when subtype is "file" (or absent)
```

The author opens the record, finds no `subtype`, and concludes the diagnostic is broken. `github.schema.json:237` happens to carry `required: ["subtype"]`, which is why the headline example in the PR body is correct — but the `required`-less form is the more common way an `if` gets written, and the renderer cannot tell the two apart.

```suggestion
        req = cond.get("required")
        if isinstance(req, list) and field in req:
            clauses.append(f"{field} is {shown}")
        else:
            # `properties` alone does not require presence: an `if` with no sibling `required`
            # succeeds for a record that omits the field, so `then` is in force there too.
            # Saying otherwise sends the author looking for a field they never set.
            clauses.append(f"{field} is {shown} (or absent)")
```

This applies to `unless` as well, not only `when`: `else` is in force only when the predicate *fails*, and an absent field satisfies a bare `properties` predicate vacuously — so an absent field means the rule is *not* lifted, which ` unless a is "x"` on its own does not say either. Verified: ` unless a is "x" (or absent) and b is "y" (or absent)`.

Alternative, if the extra words are not wanted: gate the whole clause on presence and emit nothing for the unguarded shape, matching "shapes this does not recognise add no clause." That loses a useful diagnostic on the more common schema shape, which is why the wording fix is suggested here.

backlot/validation.py
```diff
@@ -49,17 +49,113 @@ def _validator(source_type: str) -> Draft202012Validator:
...
+    parts = list(schema_path)
+    branch = _branch_index(parts)
+    if branch is None:
+        return ""
+    owner = _subschema(schema, parts[:branch])
+    cond = owner.get("if") if isinstance(owner, dict) else None
+    if not isinstance(cond, dict):
+        return ""
+    clauses = []
+    for field, spec in (cond.get("properties") or {}).items():
```
khj809: **`cond.get("properties")` is the one level with no `isinstance` guard, and a non-object value raises out of the diagnostics path.**

`owner`, `cond` and `spec` are all `isinstance`-checked; this one is not.

```python
_when_clause({"if": {"properties": ["a"]}, "then": {}}, ["then", "required"])
# AttributeError: 'list' object has no attribute 'items'
```

`_load_schemas` only `json.loads`es the files — `Draft202012Validator.check_schema` runs in a test, not at import — so one hand-edited schema turns "report this bad record" into a traceback on *every* record, including on the `--dry-run` path whose whole job is to report problems.

```suggestion
    props = cond.get("properties")
    for field, spec in (props if isinstance(props, dict) else {}).items():
```

backlot/validation.py
```diff
@@ -49,17 +49,113 @@ def _validator(source_type: str) -> Draft202012Validator:
...
+        if not isinstance(spec, dict):
+            continue
+        if "const" in spec:
+            allowed = [spec["const"]]
+        elif isinstance(spec.get("enum"), list) and spec["enum"]:
+            allowed = spec["enum"]
+        else:
+            continue
+        shown = " or ".join(json.dumps(v, ensure_ascii=False) for v in allowed)
```
khj809: **`or` (within a field) and `and` (between fields) are mixed with no grouping, so a multi-value condition reads as the wrong grouping.**

```python
if: {properties: {a: {enum: ["x", "y"]}, b: {const: "z"}}}
# today:  when a is "x" or "y" and b is "z"        -> parses as x or (y and z)
# fixed:  when a is one of ["x", "y"] and b is "z"
```

Same "no guessed clause" concern as the rest of the function — the rendering says something the schema does not. Reachable as soon as a schema pairs an `enum` predicate with a second field. A single-value `enum` is unaffected: still ` when a is "x"`.

```suggestion
        dumped = [json.dumps(v, ensure_ascii=False) for v in allowed]
        # Not `a or b`: joined by " and " to the next field it reads as `a or (b and c)`.
        shown = dumped[0] if len(dumped) == 1 else "one of [" + ", ".join(dumped) + "]"
```

backlot/validation.py
```diff
@@ -49,17 +49,113 @@ def _validator(source_type: str) -> Draft202012Validator:
...
     for err in sorted(_validator(st).iter_errors(rec), key=lambda e: list(e.path)):
-        loc = "/".join(str(p) for p in err.path) or "<root>"
-        msgs.append(f"{loc}: {err.message}")
+        loc = "/".join(str(p) for p in err.path)
+        # The record's own name says "the whole record" better than a placeholder does, so
+        # `<root>` is only there for a record that gave nothing to be called by. The path is
+        # bracketed rather than set off by a space, because a label is free text and a space does
+        # not close: a record titled `subtype` with a bad `subtype` field read
+        # "subtype subtype: ...", naming the field twice and marking neither.
+        head = f"{label} [{loc}]" if label and loc else (label or loc or "<root>")
```
khj809: **The bracket only appears when both label and path exist, so the one-token head is still ambiguous between a label and a field path.**

d51bc1a closed the two-token collision and left two unbracketed shapes that collide with each other. From one `--dry-run`:

```
subtype: 9 is not one of ['issue', 'pull_request', 'file']   # anonymous record -- `subtype` is the FIELD PATH
subtype: 'content' is a required property                    # record titled `subtype` -- `subtype` is the LABEL
```

That is the confusion the commit set out to remove, moved from the two-token case to the one-token case. The PR body justifies the bare record-wide line on length grounds but does not note the residual collision. Marking the labelless case keeps the length argument intact — a labelled record-wide error stays a bare label — and makes an unbracketed head *always* a label:

```suggestion
        if label:
            head = f"{label} [{loc}]" if loc else label
        else:
            head = f"<root> [{loc}]" if loc else "<root>"
```

After: `<root> [subtype]: 9 is not one of [...]` vs `subtype: 'content' is a required property`. The headline example and the record-wide forms are untouched (`gh-1: 'path' is a required property when subtype is "file"`, `d: 'title' is a required property`, `<root>: 'title' is a required property`).

Test-side: `test_a_label_cannot_be_mistaken_for_the_field_path` pins the old bare-path form for a nameless record and needs `"<root> [subtype]: "`.

khj809 (commented):
Read through the three commits. The diagnostics are a real improvement and I like that the docstrings carry the corpus that motivated each one; the notes below are all about the clause renderer's edges.

Verification: full suite green on the branch as-is (with `PYTHONPATH` pinned to the branch — the editable install otherwise shadows it). Every finding below was reproduced by running the new helpers, and all five suggestions applied together keep the suite green (73 passed) with three test expectations updated where the contract deliberately changes — noted per comment.

Two are worth fixing before merge (the `$ref` path collision and the presence-implication), the other three are small.

backlot/validation.py
nuri-yoo: addressed in (87cabd1)

Confirmed against jsonschema 4.26.0 with your schema: `schema_path` comes back as `['allOf', 0, 'then', 'required']`, `absolute_schema_path` is identical, and the clause named `mode` for a `{"kind": "blob"}` record. Applied the identity guard and both companions.

One correction to the parenthetical, for whoever reads the thread later: the `then/properties/path/type` case you verified is a synthetic shape, not the shipped github schema. There `path` is defined under the root `properties`, so the path is `['properties', 'path', 'type']`, never runs through `then`, and carried no clause before this commit either. Your point about comparing at `parts[:-1]` rather than `parts[:branch + 1]` is right, so I pinned it with the synthetic form beside the collision.

Also took the one-line form at the call site instead of the two-line f-string in the companion diff -- `ruff format --check` rejoins it, and CI's lint job runs that. The joined line is exactly 100 characters, so it fits `line-length = 100`.

backlot/validation.py
nuri-yoo: addressed in (8e357fa)

Right, and the `then` side is where it misleads most: with a bare `properties` predicate a record of `{}` takes the `then` branch and was told `when subtype is "file"` with neither field present anywhere.

Walked the `else` logic too, since your note about the mirror case is easy to read the wrong way round. `else` fires only when a field is present-and-wrong -- an absent field satisfies the predicate vacuously and takes `then` -- so an absent field means the rule is not lifted, which the bare `unless` did not say. Both leads carry the words now.

Verified the shipped output is untouched: `github.schema.json` carries `required: ["subtype"]`, so the headline clause stays `when subtype is "file"`. The test pins the guarded and unguarded forms next to each other so the distinction cannot quietly go away.

backlot/validation.py
nuri-yoo: addressed in (4161970)

Reproduced exactly as written -- `AttributeError: 'list' object has no attribute 'items'`. Guarded, with a test covering a list, a string and an int in that position.

The reason it matters is the one you gave: `_load_schemas` only `json.loads`es, `check_schema` runs in a test rather than at import, so this is reachable from a hand-edited schema, and the blast radius is every record on the `--dry-run` path that exists to report problems.

backlot/validation.py
nuri-yoo: addressed in (4b3a063)

Applied. Confirmed `' when a is "x" or "y" and b is "z"'` on the branch as-is, and `one of ["x", "y"]` after.

Worth noting it composes with (8e357fa): an unguarded multi-value predicate now reads `when a is one of ["x", "y"] (or absent) and b is "z" (or absent)`. Wordy, but each clause is separately true, and the test uses a `required`-guarded condition so it pins the grouping on its own.

backlot/validation.py
nuri-yoo: addressed in (8ae7c9d)

Fair -- the collision moved rather than went. Applied, and verified the forms you listed are untouched: `gh-1: 'path' is a required property when subtype is "file"`, `d: 'title' is a required property`, `<root>: 'title' is a required property`.

Measured what actually changes today across all eleven schemas: of 134 real messages, 9 change, and all 9 are this commit. The other four findings emit nothing different against the shipped schemas -- no `$ref` anywhere, the one conditional carries `required`, no `enum` in any predicate, no malformed `properties` -- so this is the only one of the five that a user sees before someone writes a new schema. Flipping your before-merge ordering on that basis, not on the substance.

Recorded two residuals in the commit message rather than chasing them: a record titled `foo [bar]` still reads as label `foo` at path `bar`, and one titled `<root>` reads as nameless. Both need a corpus author to name a record after this function's own notation.

khj809 (approved):
Verified all five findings were addressed correctly. LGTM

PR #48, opened 2026-08-11
Rebuild the CLI on Typer, with every command and option declared in one place

khj809:
Rebuilds the `backlot` CLI on [Typer](https://typer.tiangolo.com/) so that **`cli.py` is the one place the command line is defined**, and cuts the options that were not carrying their weight.

Before, `cli.py` declared only `serve`'s flags and `--type`; the eleven options `backlot import` accepted lived in `build_parser` inside `byo.py` (4) and `erb.py` (7), with raw argv forwarded to each importer's own `main`. Reading `cli.py` told you almost nothing, and the bench options stayed invisible in help until you already knew to pass the bench type.

Now every option is a Typer parameter in `cli.py`; the importers expose plain functions and parse nothing.

```
backlot serve                      # unchanged
backlot import <corpus.jsonl>      # BYO: --bundled --append --dry-run --roster
backlot import --type enterpriserag-bench   # the bench: no options at all
backlot export DIR                 # the bench as a BYO artifact  [--shard-records N]
backlot status                     # what the data dir holds
```

`README`, `CONTRIBUTING` and the Dockerfile's `backlot import --bundled` work verbatim, and `python -m backlot.importer.{byo,erb}` stay valid as shims that re-enter the CLI.

## Options removed, with the code behind them

- **`--slice-questions`** — and the `question_ids` plumbing through `select_records`, `_resolve_roster`, `export_byo` and `import_structured`. A subset leaves containers sparse, so every ACL and pagination result becomes a property of the subset rather than of the corpus.
- **`--tokens-only`** — and `dump_tokens`. The import writes the roster anyway.
- **`--export-byo`** → `backlot export DIR`, which carries `--shard-records`. As a command it cannot be invoked without a destination, and `--shard-records` cannot be given without an export; both needed hand-written checks as flags.
- **`--allow-excluded`** — the `KNOWN_EMPTY_DOCS` gate is unconditional: what the list names is excluded silently, and an undeclared exclusion stops the run with nothing to wave it past.
- **`--no-download`** — `fetch_generated_data` already returns the cache when it is populated, so this only added a way to fail confusingly when it was not.
- **`--ref`** — off the CLI, still a keyword on `fetch_generated_data`: the same API-level-but-not-env treatment `atlassian_site` gets.

Net **−97 lines** for the cull, and `erb.run()` now takes no parameters.

## Refused, instead of ignored

- a BYO option under the bench type, and a corpus path there too
- `--shard-records 0` — one shard per record is 600k files for the bench

## Added

- **`--data-dir DIR`** on every command, overriding `BACKLOT_DATA_DIR`
- **`backlot status`** — data dir, per-source counts, roster size; exits 1 when there is no corpus
- **`serve` pre-flight** — an empty data dir fails at once with the fix named, rather than binding the port and dying inside the lifespan
- bare `backlot` prints the command list; `--install-completion`

`typer>=0.12` joins the core dependencies: the console script is the headline interface, and rich draws the help panels.

## Testing

`946 passed, 1 skipped`, green in CI; `test_cli.py` goes 19 → 37 tests.

Routing is asserted from the **keyword arguments** each importer receives, bound against its real signature — which caught a real `export_byo=` vs `export_byo_dir=` mismatch that a `**kw` spy accepted happily. The export round trip is verified end to end against a hand-built `generated_data` tree: flat and sharded export, both artifacts imported back, `status` agreeing on the count. Packaging is verified from an installed wheel run outside the checkout, since the suite structurally cannot answer that.

Three gotchas, each recorded in the commit that hit it: typer 0.27 vendors its own click (so `typer.BadParameter is not click.BadParameter`, and `main()` translates click's `SystemExit` rather than catching its exceptions); typer renders a docstring newline as a literal line break; and CI's `FORCE_COLOR` plus an 80-column terminal makes rich split option names into styled fragments *and* wrap messages so the panel frame lands between words — so output assertions go through a `plain()` helper and the tests run at 80 columns rather than a comfortable width.

## Open for review

- Removing `--ref` and `--no-download` was my judgement call under "cut anything unnecessary", not something asked for by name. Both are recoverable in one line if you disagree.
- `--type` no longer accepts `erb`; the bench is spelled out, so an invocation says what it downloads. `-t` survives as a short form for `--type`, since `-t enterpriserag-bench` still names the thing — drop it too if you would rather have only the long flag.
- The bench's `ERB_E2E` test now imports the whole corpus rather than a slice, so it costs ~1GB and minutes when enabled. It is opt-in and never runs in CI.
- `status`, `--data-dir` and `--install-completion` are documented only in `--help`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,8 +31,12 @@ curl -s localhost:8000/health
 ```
 
 For a real corpus use `backlot import --type enterpriserag-bench` (the bench) or
-`backlot import mycorpus.jsonl` (your own) — see the README. Every `backlot import` flag is the
-importer module's own, so `python -m backlot.importer.{byo,erb}` still takes exactly the same ones.
+`backlot import mycorpus.jsonl` (your own) — see the README.
+
+Every command and every option lives in `backlot/cli.py`, declared as Typer parameters; the
+importers expose a plain `run(**kwargs)` and parse nothing. So `backlot import --help` is the
+complete list, and adding a flag means touching one file. `python -m backlot.importer.{byo,erb}`
+still work — they re-enter the same CLI, so they cannot take a different set.
 
 ## Terminology
 
--- a/README.md
+++ b/README.md
@@ -26,6 +26,11 @@ backlot serve               # http://127.0.0.1:8000
 curl -s localhost:8000/health
 ```
 
+`backlot status` says what the data dir currently holds, `--data-dir DIR` points any command at a
+different one (so several corpora can sit side by side), and `backlot --install-completion` adds
+shell completion. `backlot <command> --help` lists a command's options — all of them are declared in
+[`backlot/cli.py`](backlot/cli.py).
+
 Or skip the CLI entirely and let a test spin one up on a free port, serving that same corpus
 (`pip install "backlot[examples]"` for the vendor SDKs the first snippet uses):
 
--- a/README.md
+++ b/README.md
@@ -26,11 +26,6 @@ backlot serve               # http://127.0.0.1:8000
 curl -s localhost:8000/health
 ```
 
-`backlot status` says what the data dir currently holds, `--data-dir DIR` points any command at a
-different one (so several corpora can sit side by side), and `backlot --install-completion` adds
-shell completion. `backlot <command> --help` lists a command's options — all of them are declared in
-[`backlot/cli.py`](backlot/cli.py).
-
 Or skip the CLI entirely and let a test spin one up on a free port, serving that same corpus
 (`pip install "backlot[examples]"` for the vendor SDKs the first snippet uses):
 
--- a/README.md
+++ b/README.md
@@ -134,7 +134,7 @@ enterprise documents across nine of the supported sources. One command downloads
 ACL-derives it:
 
 ```bash
-backlot import --type enterpriserag-bench    # -t erb for short
+backlot import --type enterpriserag-bench
 ```
 
 What that dataset does and does not carry, and how to redistribute it as BYO-JSONL, is in
--- a/backlot/cli.py
+++ b/backlot/cli.py
@@ -2,7 +2,7 @@
 
     backlot serve                       # uvicorn backlot.main:app, with uvicorn's own defaults
     backlot import <corpus.jsonl>       # backlot.importer.byo   (--type byo, the default)
-    backlot import --type erb           # backlot.importer.erb, no options of its own
+    backlot import --type enterpriserag-bench   # backlot.importer.erb, no options
     backlot export out/                 # the bench as a BYO artifact instead of a database
     backlot status                      # what the data dir currently holds
 
@@ -13,7 +13,7 @@ taking keyword arguments, and the flags that drive them are the parameters below
 
 ``import`` keeps both corpus types behind one command (``--type``) because the bench importer has no
 options left to separate: writing an artifact became ``export``, and the rest went. So every option
-on ``import`` is BYO's, and one given under ``--type erb`` is refused rather than ignored — see
+on ``import`` is BYO's, and one given under the bench type is refused rather than ignored — see
 ``_reject_byo_flags_under_bench``.
 """
 
@@ -30,10 +30,12 @@ import typer
 # The importers and uvicorn are imported inside each command, not here: `serve` must not pay for
 # `backlot.importer.erb` (2,400 lines) and `import` must not pull in uvicorn.
 
-# --type value -> the importer it selects. `erb` is accepted beside the full bench name because that
-# is what the module, the tests and every existing doc call it.
-BYO, BENCH, BENCH_ALIAS = "byo", "enterpriserag-bench", "erb"
-IMPORTER_TYPES = (BYO, BENCH, BENCH_ALIAS)
+# --type value -> the importer it selects. The bench is spelled out in full and has no short alias:
+# `erb` is what this codebase calls it among itself, and a caller reading `--type erb` learns nothing
+# about what is being downloaded. The cost of the abbreviation lands on the person who did not write
+# the code, so it is not offered.
+BYO, BENCH = "byo", "enterpriserag-bench"
+IMPORTER_TYPES = (BYO, BENCH)
 
 _BYO_PANEL = "BYO corpus options (--type byo)"
 
@@ -173,7 +175,7 @@ def serve(
 
 # The BYO options, and the default that means "not given". Kept as data so the check below names the
 # offending flags in the user's own spelling. There is no bench counterpart: importing the bench
-# takes no options at all, which is why every one of these is refused under `--type erb`.
+# takes no options at all, which is why every one of these is refused under the bench type.
 _BYO_ONLY: dict[str, object] = {
     "--bundled": False,
     "--append": False,
@@ -183,7 +185,7 @@ _BYO_ONLY: dict[str, object] = {
 
 
 def _reject_byo_flags_under_bench(supplied: dict[str, object]) -> None:
-    """Fail when a BYO option is given with `--type erb`, which has no options of its own.
+    """Fail when a BYO option is given with the bench type, which has no options of its own.
 
     Named explicitly because "unrecognized argument" is not what happened — the flag is real, just
     not for this importer — and because the alternative is ignoring it in silence.
@@ -216,8 +218,8 @@ def import_(
             "-t",
             metavar="{byo,enterpriserag-bench}",
             help="what kind of corpus to import: `byo` reads a BYO-JSONL corpus, a `.jsonl.gz`, or "
-            "a sharded artifact directory; `enterpriserag-bench` (alias `erb`) downloads and "
-            "imports EnterpriseRAG-Bench",
+            "a sharded artifact directory; `enterpriserag-bench` downloads and imports "
+            "EnterpriseRAG-Bench",
             rich_help_panel="Common",
         ),
     ] = BYO,
@@ -259,7 +261,7 @@ def import_(
 ) -> None:
     """Build the data dir from a corpus.
 
-    Two importers behind one command, chosen with --type. Importing the bench takes no options of its own, so every option below is BYO's, and passing one with `--type erb` is an error rather than silently ignored.
+    Two importers behind one command, chosen with --type. Importing the bench takes no options of its own, so every option below is BYO's, and passing one with `--type enterpriserag-bench` is an error rather than silently ignored.
     """  # noqa: E501 — see the note on `serve`: a paragraph must be one source line.
     if corpus_type not in IMPORTER_TYPES:
         raise typer.BadParameter(
--- a/backlot/cli.py
+++ b/backlot/cli.py
@@ -267,8 +267,8 @@ def import_(
         raise typer.BadParameter(
             f"{corpus_type!r} is not one of {', '.join(IMPORTER_TYPES)}", param_hint="'--type'"
         )
-    _use_data_dir(data_dir)
-
+    # _use_data_dir AFTER the checks, in both branches: it writes the environment and clears the
+    # settings cache, and a call that is about to be rejected should leave neither behind.
     if corpus_type != BYO:
         _reject_byo_flags_under_bench(
             {
@@ -283,6 +283,7 @@ def import_(
                 f"--type {BENCH} downloads its own corpus; a path is only for --type {BYO}",
                 param_hint="'CORPUS'",
             )
+        _use_data_dir(data_dir)
         from backlot.importer import erb
 
         raise typer.Exit(erb.run())
@@ -300,6 +301,7 @@ def import_(
 
         corpus = HELLO_CORPUS
 
+    _use_data_dir(data_dir)
     from backlot.importer import byo
 
     raise typer.Exit(byo.run(corpus, append=append, dry_run=dry_run, roster=roster))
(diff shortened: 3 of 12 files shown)


nuri-yoo (commented):
I cloned the head (`de8c1ac`), installed it into a fresh venv (Python 3.14.6, macOS; the dependency range resolves to typer 0.27.1 + click 8.4.2), and ran everything below. Outputs are quoted as captured. The refactor itself is clean and the option cull is well argued; what follows is what the rewrite changed underneath it, most of it on error paths.

## 1. `main()` turns every string-valued `SystemExit` into a `ValueError` traceback

`return int(e.code or 0)` (cli.py:437) assumes exit codes are ints. `byo.py` reports corpus-validation failures as `raise SystemExit("<message>")` in eleven places (byo.py 415, 426, 603, 620, 968, 973, 1000, 1031, 1082, 1150, 1180), and those are the most common failure this tool produces: a user handing it a bad corpus.

Repro, same malformed corpus at each revision:

```console
$ printf '{"source_type": "slack", "id": "x"}\nthis is not json\n' > bad_corpus.jsonl

# main (936fb10)
$ BACKLOT_DATA_DIR=/tmp/d python -m backlot import bad_corpus.jsonl
line 2: invalid JSON: Expecting value: line 1 column 1 (char 0)
$ echo $?
1

# PR head (de8c1ac), installed console script
$ BACKLOT_DATA_DIR=/tmp/d backlot import bad_corpus.jsonl

Traceback (most recent call last):
  File ".../.venv/bin/backlot", line 6, in <module>
    sys.exit(main())
             ~~~~^^
  File ".../backlot/cli.py", line 437, in main
    return int(e.code or 0)
ValueError: invalid literal for int() with base 10: 'line 2: invalid JSON: Expecting value: line 1 column 1 (char 0)'
$ echo $?
1
```

The exit code happens to stay 1 (the interpreter dies on the ValueError), but the diagnostic the user needs is now the payload of a stack trace instead of the program's last line. Mirroring Python's own `SystemExit` semantics fixes it: `None` becomes 0, an int is returned as-is, anything else is printed to stderr and returns 1.

## 2. `serve --log-level` lost its validation

main constrained it with argparse `choices`; the PR declares it as a bare `Optional[str]` (cli.py:114-118), so a typo travels into uvicorn's internals. Side by side, against a data dir with a corpus so the pre-flight passes:

```console
# main (936fb10)
$ backlot serve --log-level bogus
usage: backlot serve [-h] [--host HOST] [--port PORT] [--reload]
                     [--log-level {critical,error,warning,info,debug,trace}]
                     [--no-proxy-headers] [--forwarded-allow-ips IPS]
backlot serve: error: argument --log-level: invalid choice: 'bogus' (choose from 'critical', 'error', 'warning', 'info', 'debug', 'trace')
# exit 2

# PR head (de8c1ac)
$ backlot serve --log-level bogus
│ .../site-packages/uvicorn/config.py:415 in configure_logging                │
│   413 │   │   if self.log_level is not None:                                │
│   414 │   │   │   if isinstance(self.log_level, str):                       │
│ ❱ 415 │   │   │   │   log_level = LOG_LEVELS[self.log_level.lower()]        │
KeyError: 'bogus'
# exit 1
```

The help text still lists `critical|error|warning|info|debug|trace`, it just no longer enforces it. An Enum parameter restores the check and gets help rendering and completion with it.

## 3. `typer>=0.12` is not a working floor; the measured floor is 0.18

pyproject.toml:33 declares `typer>=0.12`, but CI only ever exercises the latest resolve. Measured in a clean venv per version, first with a minimal app to separate typer's own breakage from this PR's code:

```console
typer 0.12.0 -> BROKEN TypeError Secondary flag is not valid for non-boolean flag. | click 8.4.2
typer 0.12.5 -> BROKEN TypeError Secondary flag is not valid for non-boolean flag. | click 8.4.2
typer 0.15.0 -> BROKEN TypeError Parameter.make_metavar() missing 1 required positi | click 8.4.2
typer 0.16.0 -> OK    click 8.4.2
typer 0.19.0 -> OK    click 8.4.2
```

A four-line typer app (one command, one bool option, `--help`) fails identically at 0.12 and 0.15, so that band is a typer-vs-click-8.4 incompatibility, not this PR's fault. The band that is this PR's problem is 0.16/0.17, where the app runs but a required `Argument` silently arrives as `None`. Probe, using the PR's own `export` declaration verbatim:

```python
# argprobe.py — out_dir declared exactly as in cli.py
@app.command()
def export(
    out_dir: Annotated[Path, typer.Argument(metavar="DIR", help="dest")],
    shard_records: Annotated[Optional[int], typer.Option(metavar="N")] = None,
) -> None:
    print(f"BODY ENTERED out_dir={out_dir!r}")
app(args=[], prog_name="backlot export")
```

```console
typer 0.16.0 -> BODY ENTERED out_dir=None   exit 0
typer 0.17.0 -> BODY ENTERED out_dir=None   exit 0
typer 0.18.0 -> Missing argument 'DIR'.     exit 2
typer 0.19.0 -> Missing argument 'DIR'.     exit 2
typer 0.20.0 -> Missing argument 'DIR'.     exit 2
```

Against the real CLI at 0.16, with `erb.run_export` replaced by a spy: `backlot export` (no DIR) reaches the importer with `out_dir=None` and returns 0. With `_fetch_and_seed_roster` left real it goes further than that — the test suite at 0.16 shows the consequence:

```console
$ pip install 'typer==0.16.0' && pytest tests/test_cli.py
FAILED tests/test_cli.py::test_export_requires_a_destination - TypeError: arg...
----------------------------- Captured stderr call -----------------------------
downloading https://codeload.github.com/onyx-dot-app/EnterpriseRAG-Bench/tar.gz/refs/heads/main
1 failed, 36 passed in 177.54s (0:02:57)
```

That single test takes three minutes because `export` with a missing DIR starts the ~1GB bench download and only then dies inside pathlib on the `None`. At 0.18.0 the same suite is `37 passed in 1.35s`. Raising the bound to `typer>=0.18` closes all of it.

## 4. `run_export` downloads before it touches `out_dir`

`_fetch_and_seed_roster()` runs first (erb.py:2329); the first thing that touches the destination is `export_byo`'s `out_dir.mkdir`, after the fetch. So an unwritable or mistyped destination fails only after the download completes (cached case aside), and the ordering is also what let the 0.16/0.17 case in finding 3 reach the network at all. Creating the destination before fetching makes both fail in milliseconds.

## 5. The `serve` pre-flight only checks existence

A corrupt or truncated corpus passes `settings.db_path.exists()` (cli.py:151), binds the port, and dies inside the lifespan — the failure shape the pre-flight was added to remove. With 4KB of `/dev/urandom` at `data/mock.sqlite`:

```console
$ backlot serve
ERROR:    Traceback (most recent call last):
  ...
sqlite3.DatabaseError: file is not a database
ERROR:    Application startup failed. Exiting.
# exit 3
```

`status` against the same data dir already diagnoses it, in both corruption modes I tried:

```console
$ backlot status        # 4KB of random bytes
corpus:   .../mock.sqlite (0.0 MB) — unreadable: file is not a database
Not a Backlot corpus, or incomplete. Rebuild it with `backlot import`.
# exit 1

$ backlot status        # first 8KB of a real corpus
corpus:   .../mock.sqlite (0.0 MB) — unreadable: database disk image is malformed
Not a Backlot corpus, or incomplete. Rebuild it with `backlot import`.
# exit 1
```

The same readability check can gate `serve`.

## 6. The suite never crosses the cli-to-importer seam

The cli tests always patch `byo.run`/`erb.run` (per test_cli.py's own docstring, routing is asserted from the keyword arguments the spy receives), and the importer tests call `run`/`import_structured` directly. Nothing drives `cli.main` into a real importer, and finding 1 lives exactly on that seam; one test feeding a malformed corpus through `cli.main` would have caught it. The suite is green around the hole:

```console
$ FORCE_COLOR=1 COLUMNS=80 pytest
887 passed, 24 skipped, 1 warning in 21.22s
```

All 24 skips are optional-extra imports (`llama_index` x13, `googleapiclient` x3, `mirage` x2, `mcp`, `slack_sdk`, `botocore`) plus the opt-in `ERB_E2E`, consistent with CI's 946/1.

## Minor

- `status` prints a 4KB file as `0.0 MB` (visible in the finding-5 output); for a command meant to diagnose truncated downloads, the size is the signal.
- Usage says `[corpus]` while errors say `Invalid value for 'CORPUS'`.
- `python -m backlot.importer.erb --type byo --bundled` runs the BYO importer end to end (verified: it imports the bundled corpus). The shim prepends `--type enterpriserag-bench` and the user's later `--type` wins.
- The shims' errors say `Try 'backlot import -h'`; main passed a per-module `prog=` so errors named the spelling the user actually typed.
- Bare `backlot` prints the command list to stderr and exits 2 (click treats `no_args_is_help` as a usage error), so the PR body's "prints the command list" holds, but not on stdout with exit 0.
- The branch is behind main: merge-base is `936fb10`, main is at `641af93` (#45). GitHub reports CLEAN, but a rebase before merge would keep the history tidy.
- `test_the_snapshot_the_data_came_from_reaches_the_manifest` replaces `KNOWN_EMPTY_DOCS` wholesale where the neighbouring tests union into it. Harmless, just inconsistent.

## Checked and fine

- Repo-wide grep for every removed flag and symbol (`--slice-questions`, `--tokens-only`, `--export-byo`, `--allow-excluded`, `--no-download`, `--ref`, `question_ids`, `dump_tokens`, `build_parser`, `-t erb`): one comment mention and the rejection-test parametrization remain, nothing else.
- The Dockerfile's `backlot import --bundled` (line 51) and the `serve` CMD (line 42) run verbatim against this head.
- `read_meta` swallows only missing-table errors, so the `status` comment about pre-meta DBs is accurate.
- `fetch_generated_data(settings, *, ref="main")` keeps its default, and the cache short-circuit at erb.py:579 backs the `--no-download` removal.
- `BACKLOT_RAW_DIR` in the example README is the real name (`env_prefix="BACKLOT_"` plus the `raw_dir` field).
- `KNOWN_EMPTY_DOCS` is byte-identical to main; the tests declare their fixtures via monkeypatch instead of widening the production set.
- The `--data-dir` ordering fix holds in all three rejected branches — after each of `import --data-dir D --type enterpriserag-bench --dry-run`, `export DIR --shard-records 0 --data-dir D`, and `import --data-dir D` (no source), `BACKLOT_DATA_DIR` is still unset in the process.
- The positive `--data-dir` path works: `import --bundled --data-dir D` then `status --data-dir D` reports 167 rows from 136 source documents, matching an env-var run.
- All three module shims and `python -m backlot --version` work from the installed package.

Of these, 1 is the one that reaches every user with a bad corpus; 2 and 3 are a line each.

khj809:
Fixed in f77f75e. Five of six were real; thank you for driving it from a clean install, which is where all of them surfaced.

**1** — reproduced exactly, and the count is worse than the flag: eleven `SystemExit("<message>")` sites in `byo.py`. `_exit_code` now follows the interpreter’s own rules (None → 0, int as-is, anything else to stderr with status 1), so a bad corpus reads `line 2: invalid JSON: …` / exit 1, identical to main.

**2** — `LogLevel` enum. Rejected at parse time with exit 2, and a test asserts uvicorn is not reached at all.

**3** — raised to `typer>=0.27`, the current release, rather than the measured 0.18 floor. CI only exercises the latest resolve, so a lower bound would be a range nothing tests.

**4** — the destination is created before the fetch, and `export` names the failure rather than surfacing a pathlib traceback from mid-export.

**5** — `_unreadable_corpus` reads the SCHEMA, never the rows: a check that runs before the server binds cannot afford the `COUNT(*)` that `/health` deliberately defers to a background thread on the 6.4M-row deployment. Shared with `status`. I probed six damage shapes and each gets an accurate message; three are tested with the string sqlite actually produces:

```
random bytes               -> file is not a database
truncated real corpus      -> database disk image is malformed
valid db, wrong schema     -> not a Backlot corpus — 11 table(s) missing, e.g. confluence_pages
empty file                 -> (same missing-tables message)
```

Consequence worth naming: the `serve` fixture wrote an empty placeholder file, which no longer passes. That is the check working, so the fixture now writes a real schema instead of the check being loosened to admit it.

**6** — a test drives a malformed corpus through `cli.main` into the real importer and asserts `"Traceback" not in stderr`. That is the regression test for **1**, and it sits exactly on the seam you identified.

Minor, all taken: scaled size units in `status`; `[CORPUS]` metavar to match the error text; the module shims refuse a `--type` that would override the module they belong to (`python -m backlot.importer.erb --type byo --bundled` ran the BYO importer to completion — good catch); the erb tests union into `KNOWN_EMPTY_DOCS` consistently.

Two I did not take:

- **Bare `backlot` goes to stdout, not stderr.** Measured on typer 0.27.0: 2,671 bytes on stdout, stderr empty, exit 2. The exit code we agree on, and the PR body only claims the command list is printed. If you saw stderr on 0.27.1 I would like to know, since the suite asserts stdout.
- **No rebase yet.** Not merge-blocking, and rebasing now rewrites the SHAs this review cites. Happy to do it immediately before merge.

`952 passed, 1 skipped` under `FORCE_COLOR=1 COLUMNS=80`; CI green on f77f75e.

nuri-yoo (approved):
Approving — re-ran all six from a clean install of f77f75e, and the five fixes hold.

The stdout one was mine. I split the streams with `backlot 2>&1 >/dev/null`, and zsh's MULTIOS makes that second redirect an extra destination rather than a replacement, so stdout rode along in what I called stderr. To separate files: 2,671 bytes on stdout, stderr empty — your numbers, and the same on de8c1ac.

Two notes, neither blocking. #46 landed, so the branch is two behind; `MERGEABLE` only covers textual conflict, so I merged current main in locally — clean, `913 passed`. That merge also takes `byo.py` from eleven string `SystemExit` sites to fourteen, which is the case for having handled the payload generically rather than site by site.

PR #50, opened 2026-08-11
GitHub: honour media types, validate the owner, and give a pull a real changeset (#49)

### LINKED ISSUE #49: GitHub surface: divergences from the real REST API found by a filesystem client
# GitHub surface: divergences from the real REST API found by a filesystem client

Probed the hosted deployment (`https://enterprise-mock.brekkylab.com/github`, org `engineering`,
admin token) with a client that mounts a GitHub repository as a filesystem — it reads git trees
and blobs for code, and issues/pulls as rendered documents. Nine divergences, grouped by what
fixing them would take. Filed as one issue because the three in group B share one root cause.

Reproduced against repo `engineering/audit-log-exporter` unless noted. For D1 the same requests
were also sent to `api.github.com` so the difference is a comparison rather than an assumption
about what the real API does.

**Where each finding comes from**, so you know what to expect a repro for: D1, D5 and the empty
group-B renders were reproduced at runtime; D3, D4, D6, D7 and D8 come from reading
`backlot/routers/github.py`. D3, D4 and D6 could not be exercised at runtime *because* they are
missing — the client has to be configured around them (an explicit repo, a commit-shaped ref) to
get as far as anything else, which is itself the report.

| ID | Endpoint / behaviour | Group | Effect on a client |
|---|---|---|---|
| D1 | **No content endpoint honours `Accept: …raw`** (`git/blobs`, `contents`, `readme`) | A | **Silent content corruption** |
| D2 | `pulls/{n}` ignores `Accept: application/vnd.github.diff` | B | **Silently returns JSON as a diff** |
| D3 | `GET /user/repos` absent | A | A token cannot enumerate its own repos |
| D4 | `GET /repos/{o}/{r}/git/ref/heads/{ref}` absent | A | Cannot resolve a branch to a commit |
| D5 | `GET /repos/{o}/{r}/pulls/{n}/comments` absent | B | 404 breaks any pull render |
| D6 | `GET /repos/{o}/{r}/pulls/{n}/files` absent | B | No changed-file list |
| D7 | `owner` path segment never validated | A | Wrong owner returns 200 |
| D8 | A pull's `node_id` is built as an Issue | A | Type-parsing clients misread it |
| D9 | Trees have no history; `truncated` always `false` | C | Documentation, see below |

---

### D1. No content-serving endpoint honours `Accept: application/vnd.github.raw`

Every content endpoint returns the base64 JSON envelope regardless of the media type asked for.
The real API returns the raw bytes, so a client that asks for raw receives **a JSON document as the
file's content**, with a 200 and no way to detect it.

Established by sending the identical request to both targets — same header, same
`X-GitHub-Api-Version` — and comparing the body against the `size` the tree reported for the same
blob:

| Endpoint | real `api.github.com` | this mock |
|---|---|---|
| `git/blobs/{sha}` | `text/plain`, **20 B** = tree `size`, raw | `application/json`, **1672 B**, envelope |
| `contents/{path}` | `application/vnd.github.raw`, **20 B** = tree `size`, raw | `application/json`, **2539 B**, envelope |
| `readme` | `application/vnd.github.raw`, **246 B**, raw markdown | `application/json`, **1052 B**, envelope |

Real, `selloriwoo/rust-practice`, `src/garden.rs` (tree `size: 20`):

```
git/blobs/{sha}    200 text/plain; charset=utf-8                20B
                   pub mod vegetables;\n
contents/{path}    200 application/vnd.github.raw; charset=utf-8  20B
                   pub mod vegetables;\n
```

Mock, `engineering/audit-log-exporter`, `terraform/modules/audit_sink/outputs.tf` (tree
`size: 1046`):

```
git/blobs/{sha}    200 application/json                       1672B
                   {"sha":"3249bb81c5298e019d8ccc1232d361cbd2dce945","node_id":…
contents/{path}    200 application/json                       2539B
                   {"type":"file","name":"outputs.tf","path":"terraform/modules…
readme             200 application/json                       1052B
                   {"type":"file","name":"README.md","path":"README.md","encodi…
```

So this is not a gap in one handler — there is no media-type negotiation on the GitHub surface at
all. `/contents/{path}` matters most in practice: it is the endpoint most GitHub clients reach for
to read a file, so the blast radius is wider than the client that found this.

The only signal available to a caller is that the byte count disagrees with the tree's `size`. A
client that trusts the response body silently ingests JSON envelopes instead of source files.

The content is already in hand in each handler (`row["content"]`), so this is returning a `Response`
with the raw body and an appropriate `Content-Type` when the header asks for it. **Highest impact
here and among the smallest fixes — worth doing first even if nothing else in this issue is.**

(Minor, for accuracy: real GitHub answers `git/blobs` with `text/plain` and `contents`/`readme` with
`application/vnd.github.raw`. Both are raw bodies; the difference is real GitHub's own, not
something the mock needs to reproduce exactly.)

### D3. `GET /user/repos` is not implemented

This is the endpoint a *credential* uses to discover what it can reach, and it has no substitute
here. `/orgs/{org}/repos` exists but is org-scoped, which is a different question: a real
fine-grained token may span several orgs, or cover only personal repositories, so an org listing
is not the token's view of the world.

Consequence for a client: it cannot list an owner's repositories at all, and has to be configured
with an explicit repository name per mount.

The pieces look present — `github_repos` plus `auth.visible_ids`, and `/repos/{o}/{r}/collaborators`,
`/orgs/{org}/teams` and `/repos/{o}/{r}/teams` all resolve — so this seems to be a matter of
exposing the caller-visible set under the real route.

### D4. `GET /repos/{o}/{r}/git/ref/heads/{ref}` is not implemented

`/branches/{branch}` exists, but the git-refs route is the one that resolves a branch to a commit
SHA **when the branch name contains slashes** (`release/2026-03`), because it takes the ref as its
trailing path. Without it a client cannot pin a branch to a commit.

`_repo_commit_sha(repo)` already exists, and pulls store `head_ref` / `base_ref`, so the values
needed are there.

### D7. The `owner` path segment is never validated

`/repos/{any-owner}/{repo}` succeeds whenever `repo` exists — the handler looks up only the repo
(`store.get_container(conn, "github", repo)`) and echoes the supplied owner back into the response.
The real API 404s on a wrong owner. This lets a client's owner-handling bug pass unnoticed against
the mock and fail in production. `get_settings().org_name` is available to compare against.

### D8. A pull request's `node_id` is built as an Issue

`MDEyOklzc3VlMjY5MzM1` decodes to `012:Issue269335`. The real API uses a `PullRequest`-typed node
id. Cosmetic, but a client that parses the node id to learn a resource's type gets the wrong answer.

---

## Group B — one root cause: no per-pull changeset data

These three are grouped because they are the same gap, not three independent handlers. The schema
carries nothing about what a pull *changed*:

```
github_items:    doc_id, repo, author_email, title, content, kind, state, labels, assignees,
                 merged_at, head_ref, base_ref, reviews, reactions, created_ts, updated_ts,
                 closed_ts, closed_by, merged_by, milestone, requested_reviewers,
                 owner_display, path
github_comments: id, doc_id, seq, author_email, body, created_ts, reactions
```

No `files`, no `patch`, no `additions`/`deletions`; and `github_comments` has no `path`, `line` or
`diff_hunk`, so a line-anchored review comment cannot be represented. (`reviews` is present, which
is why `/pulls/{n}/reviews` works — but reviews are approve/request-changes events, a different
resource from review *comments*.)

### D6. `GET /repos/{o}/{r}/pulls/{n}/files` is not implemented

The changed-file list — `filename`, `status`, `additions`, `deletions`, `patch`. Absent from the
route table.

### D5. `GET /repos/{o}/{r}/pulls/{n}/comments` is not implemented

Returns 404. Any client that renders a pull by combining its metadata, conversation, review
comments and files fails here, before it reaches D6.

### D2. `pulls/{n}` ignores `Accept: application/vnd.github.diff`

Returns the pull's JSON instead of a unified diff, with a 200:

```
read 10572 bytes for a `.diff` request
first bytes: {"id":269335,"node_id":"MDEyOklzc3VlMjY5MzM1","number":20770,"title":"jurisdictional-sink-…
```

Same class as D1 — the caller cannot tell. A consumer that pipes this to `git apply` or a diff
viewer gets a document that merely occupies the position of a diff.

**Note on scope.** This group is not "add three handlers": there is no before/after file state to
diff against. It is a *data-generation* feature. It does not require a schema change — the repo's
current file set is available, so a deterministic synthesis seeded on `doc_id` could pick files,
emit well-formed hunks, and derive a review comment's `path`/`line` the same way. Whether the mock
should invent that content is a design call rather than a bug fix, which is why it is described
this way rather than as a missing route.

Worth noting what it would buy: a client's diff path, its handling of an omitted `patch` (binary or
oversized files), and its pagination over a large changed-file list would all become testable
against the mock. Today none of them are reachable.

---

### D9. Trees carry no history and `truncated` is always `false`

`git/trees/{ref}` resolves *any* ref — branch name, commit SHA, anything — to the repository's
current file set, and the docstring states this outright ("We keep no history"). Not reporting it
as a bug, but two consequences are worth documenting, because a client author will assume
otherwise:

- Pinning to a commit SHA gives no immutability guarantee. A client that pins a snapshot expecting
  content-addressed stability (which is the whole basis of caching against a blob SHA) gets a
  moving target.
- `truncated` never being `true` means a client's truncation-handling path is never exercised. The
  real API caps a recursive tree at 100k entries / 7 MB, and that fallback is exactly the kind of
  code a mock is valuable for testing. A way to force `truncated: true` would help.

---

## Checked and *not* broken

Recording these so they are not chased:

- **Tree scoping is correct.** `audit-log-exporter` and `benchmark-lab` return different root trees.
  (`audit-log-exporter`'s tree contains a `redwood-private-installer/` directory, which looks like
  cross-repo leakage but is corpus content.)
- **`/issues` including pulls is correct** and matches the real API. A probe saw an empty issue list
  for a repo with 18 pulls only because the client filters rows carrying the `pull_request` marker
  and that repo has no actual issues.
- **Tree entry shape is correct**: `sha` is 40-hex, `size` is the exact byte length, `mode` and
  `type` match the real API.
- **`/repos/{o}/{r}` and `/repos/{o}/{r}/issues/{n}/comments` work.** The first returns a usable
  `default_branch`; the second serves a pull's conversation comments correctly. Only the media type
  (D1) and the routes listed above are at issue.

---

## Suggested order

1. **D1** — largest effect, smallest change, data already there.
2. **D3, D4** — unblock repository discovery and branch pinning.
3. **D7, D8** — small correctness fixes that stop a mock run from hiding a client bug.
4. **Group B** — needs a scoping decision first.
5. **D9** — document the no-history contract; optionally add a way to force `truncated`.

khj809:
Closes #49 — nine divergences from the real GitHub REST API, found by a client that mounts a repo as a filesystem.

Three of them are **silent**: `Accept: …raw` returns a JSON envelope with a 200, `Accept: …diff` returns the pull's JSON, and a wrong `{owner}` returns a 200. A client cannot detect any of them.

## The nine

| | |
|---|---|
| **D1** | `Accept: application/vnd.github.raw` on `contents`/`readme`/`git/blobs` returns the bytes, in every spelling GitHub honours (including `…v3.raw+json`). There was no media-type negotiation on this surface at all, so it is added once and reused for D2 |
| **D2, D6** | A pull has a changeset: `/pulls/{n}/files` (paginated) and `…diff`/`…patch` on the pull itself |
| **D3** | `GET /user/repos` — the credential's own reach |
| **D4** | `GET /repos/{o}/{r}/git/ref/{ref}`, ref as a trailing path so `heads/release/2026-03` resolves |
| **D5** | `/pulls/{n}/comments` serves line-anchored review comments, and a comment resolves by its own id at `…/issues\|pulls/comments/{id}` |
| **D7** | `{owner}`/`{org}` validated against the served org, as a router-wide dependency so a later route cannot forget it. Repo existence is answered against the caller's view through one shared check, so a repo whose every document is hidden is not confirmable. Every builder reads the org from one place, so the absolute URLs a response carries are ones the routes accept |
| **D8** | A pull's `node_id` is PullRequest-typed; the issues stream keeps the Issue one — real GitHub models two nodes |
| **D9** | `truncated` follows the real 100k-entry / 7 MB caps; `get_tree`'s docstring states what keeping no history costs a sha-pinning client |

## A pull's changeset

`github_items` records nothing about what a pull changed. **Two columns**, no stored patches:

- **`github_items.changed_paths`** — a corpus states *which* files a pull touched. Omit it and they are chosen deterministically: well-formed, but unrelated to what the pull is about. Declared paths are all `modified`; a corpus naming a path says the pull *changed* a file the repo has, and `added` would claim it created one.
- **`github_comments.path` / `line` / `diff_hunk`** — a comment with a `path` is a review comment, one without is conversation. `/issues/{n}/comments` excludes anchored rows, and `comments` / `review_comments` each count their own half.

The **hunks are always derived from each file's own snapshot** — no invented lines, no file the repo lacks. The snapshot is the pull's head, expressed as one convention throughout, which is what makes the diff apply: a diff mixing a `removed` file with a `modified` hunk would claim the snapshot is base and head at once, so `removed` is not produced and deletions come from a "removed a duplicated block" hunk instead.

Three properties that follow from that, each with a test:

- A declared path is **ACL-resolved**, and one the caller cannot read behaves exactly like a typo — otherwise the response reveals which. `review_comments` counts what the list actually returns, so the two never disagree and the count cannot leak that a hidden file carries a comment; fetching that comment by its own id refuses it the same way.
- **`patch` may be omitted** from the JSON file object when it is too large to inline, as on real GitHub. That is a limit on the JSON representation only — the `.diff` media type carries the hunks either way, and a `diff --git` header with no body is what git calls garbage.
- A review comment's **`position` indexes into the diff hunk**, not into the file, and its `diff_hunk` is the pull's own hunk for that file when that hunk covers the commented line, else a context window around it.

### What the import says about a declared changeset

Both fields are **pull-only**: an issue has no changeset endpoint and no review-comment endpoint, so either one on an issue is data the mock would store and never serve. Refused — and refused by `--dry-run` too, so a corpus never passes the check that exists to spare the author a failed import and then fails the import. The rules live in Python rather than as a schema `if`/`then`, which reports them as `'subtype' is a required property` and names neither the field that put the rule in force nor what is unservable about the pairing.

A declared path whose `file` document is **nowhere in the repo** is the other half, and it is reported rather than refused. A corpus is routinely a *slice* of a repo — the property that already leaves a linear `parent` pointing outside it — so refusing would make a pull that states the files it really touched unimportable whenever the slice stopped short of one of them; under `--append` the file document may also land in a later shard, or already sit in a DB that `--dry-run` cannot see. So the record loads as written and the router drops the path, exactly as it drops one the caller cannot read. `--dry-run` names the line and the path, a load counts them and points at `--dry-run` — a report and a summary are for different readers — and neither touches the exit code, so a legitimate slice still passes a gate that shells out to `--dry-run`.

## Verification

- `1019 passed, 1 skipped` (full suite incl. the MCP/Docker tests), ruff clean. 37 of those are this branch's, against the 983 on `main`
- **`git apply --reverse --check` over 150 PR seeds / 543 changed files**, half declaring `changed_paths`, covering unicode, single-line and no-trailing-newline files — 0 rejections. Nothing weaker proves a diff is a diff: a hunk header off by one still passes a shape assertion
- **Perf**: a changeset is built per row of a `/pulls` page, so `store.list_repo_file_paths` plus a per-request `_RepoFiles` memo keeps it to one cheap listing and the files actually chosen. `/pulls?per_page=100` over a 3000-file repo is 27ms against a 16ms `/issues` baseline; worst case (100 pulls each declaring 20 files) is 28ms
- Both GitHub examples run. They discover owner and repo from `/user/repos` rather than naming an org, and the PyGithub one reads a pull through `get_files()` / `get_review_comments()` / `get_issue_comments()` / `get_git_ref()`
- The shipped BYO corpus carries the new fields (a test keeps it exhaustive), and its review comment's derived `diff_hunk` is byte-identical to the hunk the same pull's diff serves. Its one declared path resolves, so a correct corpus draws no report
- A path whose file arrives in a **later** shard: shard 1 reports it, then `--append` of shard 2 resolves it against the DB and does not re-report shard 1's records

## Notes for the reviewer

- **An existing corpus must be re-imported, not just served.** The server opens the DB read-only, so `connect_rw`'s ALTERs never run; a pull-and-restart against a DB built before this change raises `no such column: path`. Run `backlot import` first.
- **A comment's `id` is assigned at import** (`github_comments.served_id`), not hashed at serve time. Its `url` resolves through that id, and a hash into any fixed range collides by the birthday bound — ~4% at the 27k comments the ERB github source produces, certain by 500k — which would serve one comment's url with another's body. The seed is probed until free and a UNIQUE index enforces it, so a duplicate is an import error rather than a silent shadow; a re-import keeps the ids it already assigned. Verified at 200k comments: 0 collisions, uniform across the range.
- **Out of scope:** storing verbatim patches (`files` with `patch`/`status` per file), which is what `status: "removed"` and real patch text would need.


nuri-yoo:
Not the requested reviewer — I had an early look ahead of any review because the synthetic-corpus generator now targets this branch: its serving-contract gate runs `backlot import --dry-run` from this head, and its post-generation pass declares `changed_paths` on pulls whose repo shipped a `file` document for the same work topic. So this is a field report plus one observation, not a review.

**What I exercised on `020d00d`, end to end:** a 1,000-scene corpus (928 records after gating) imported clean through `--dry-run`, with 4 pulls declaring `changed_paths`; every declared path resolves and the served changesets match the declarations. As a cross-check, the same corpus against a pre-#50 checkout is refused at exactly `'changed_paths' was unexpected` — `additionalProperties` still guards the boundary from the other side. The re-import note in the PR body is right and worth keeping prominent.

**One observation, non-blocking:** import accepts a `changed_paths` entry that resolves to no `file` document anywhere in the repo, and nothing ever says so. At serve time that is deliberate and correct — "one the caller cannot read behaves exactly like a typo" is the right call, since distinguishing them per caller would reveal which paths exist. But import is not per-caller: a path that resolves for *no* caller is a corpus typo, and today its pull serves an empty changeset with no signal at any stage, including `--dry-run`. That sits oddly beside this PR's own guard one screen up ("changed_paths is for subtype='pull_request' only... Refused here rather than stored and silently unservable") — the shape is validated, the referent is not. A `--dry-run` diagnostic naming the pull and the unresolved path would close it; whether an unresolved path should refuse the import outright or warn is your call, and I am happy to send that as a follow-up PR after this merges if you would rather keep this one closed.

Tiny nit while reading: the `changed_paths` schema description says the hunks are synthesized from the file's snapshot, which is accurate, but nothing on the corpus side documents that a *declared-but-unresolved* path is dropped from the changeset — if the silent drop stays, that sentence may be worth adding to the field description.

khj809:
@nuri-yoo Sorry for not assigning you as a reviewer. I was thinking of requesting you a review after @selloriwoo confirms that all items in #49 were correctly addressed, because you're already handling too many things.

Anyway, your observation is right. I'm trying to fix the importer to validate if `changed_paths` includes any unresolvable files. Will request a review after this is pushed.

nuri-yoo (approved):
Approving. I read the full diff against the merge base, not only the newest commit, and exercised the corpus side end to end. The changeset design holds where I pushed on it: snapshot-as-head is what makes the diff applicable, `removed` being unavailable follows from that convention rather than being an omission, and resolving `review_comments` through the same pass that builds the list is what stops the count leaking that a hidden file carries a comment. Both points from my earlier note are addressed, field description included.

Field report on `ab5c5ae`, same 1,000-scene corpus as before (1,082 records, 41 pulls, 4 declaring `changed_paths`): identical to `020d00d` at every stage — `OK: 1082 records valid`, real import with a stated roster, and a served-ACL cross-check of 954 callers x 4,397 documents with 0 leaks and 0 question mismatches on both heads. No github report fires, since every declared path resolves. Moving the pairing rules into `--dry-run` adds no failure surface for a corpus that keeps pull-only fields on pulls; I checked those shapes directly and found no `changed_paths` on a non-pull and no review comment without a `path`.

One line worth adding to the re-import note: an existing database needs a clean import, not an `--append`. Appending onto one built before this change lands the ALTERs through `connect_rw` but leaves the comments already in it unassigned, and they then serve `"id": null` — the one path where that note's loud `no such column: path` goes quiet instead.

selloriwoo:
Verified all nine locally in a worktree, with `api.github.com` alongside for comparison, and
spot-checked the deployed instance

The two worth naming: **D1** matches real byte-for-byte *and* on `Content-Type`, including the split
real makes between `text/plain` for `git/blobs` and `application/vnd.github.raw` for
`contents`/`readme` — that one holds on the deployment too. **D2** passes
`git apply --reverse --check`, which a hunk header off by one fails where a shape assertion would
not. On the deployment, `pulls/*.json` renders end to end where it used to 404 and `/user/repos`
answers where it used to be absent.

Suite is green here too: **960 passed, 21 skipped, 0 failed** of 981 collected, at `ab5c5ae`. The
skips are all environmental — `uvx`, `npx` and Docker absent on my machine — so the MCP/Docker
coverage is on your run, not mine.

LGTM on the GitHub surface and the client-facing behaviour. The CLI, importer and atlassian changes
I did not exercise beyond what the suite covers.

selloriwoo (approved):

PR #53, opened 2026-08-13
slack: a reply may carry its own `created`, the treatment a gmail message gets

nuri-yoo:
## What

A slack reply accepts an optional `created` (epoch seconds or ISO 8601, like every other BYO time). Provided, it is the reply's stored second and therefore its served `ts`; omitted, the reply lands one second after the message before it — which places every reply of a clockless thread exactly where the old root+position rule always put it, so existing corpora load byte-identically. A gmail message's `created` has been honored all along; this closes the one child-clock gap in the BYO contract.

A thread's seconds are resolved together, before any of its rows are written, and ordering is judged only against clocks the corpus actually supplied. A root that carries no `created` of its own holds a second hashed from its `doc_id`, which is not a fact about the thread: such a root is re-grounded on the first reply that does carry a clock, so a corpus that dates its replies but not its roots serves a coherent thread rather than a root years ahead of its own reply. A thread that supplies no clock anywhere keeps the synthesized root and the root+position spacing.

Strictly increasing is enforced at import rather than assumed: a Slack `ts` is identity as well as clock, so two thread messages sharing a second would collide at the ts every endpoint resolves, and a reply at or before its parent is a shape real Slack cannot emit (every message in `conversations.replies` carries its own unique `ts`, ordered down the thread). Three things refuse the import, each naming the record, the reply and the value: a clock at or before a corpus-supplied one; a clock that leaves no free second for a clockless reply ahead of it, which says so rather than quoting the defaulted second as though the author had written it; and a `created` that is filled in but unreadable, which is a typo rather than a request for the default.

That last rule holds for a record's own clocks too, not just its replies'. Any time field whose absence puts a default in its place — a record's `created` (a hash of the doc_id), its `updated` (NULL), a comment's `created_ts`, a gmail message's `created` — refuses a value it cannot read, so a typo can never load as a time nobody wrote. Unreadable covers numbers with no usable second: `inf` and `nan`, which `json.loads` accepts as bare literals, and a finite second outside the years `datetime` can render, which otherwise imported clean and failed when a router dated the row.

## Why

A generated or exported corpus knows when its replies were written — an incident acknowledged in seconds, the real answer hours later — and the importer was flattening that to a metronome: root+1, root+2, root+3. Temporal reasoning over threads ("how long did the rollback take?") then grades against invented spacing, and the uniform one-second cadence is a synthesis fingerprint any corpus-level reader can see.

## Serving-side consequences (the actual work)

Two derivations leaned on the metronome and are re-grounded in stored rows:

- **`thread_ts`** was derived as the reply's own second minus its position. That held only while replies sat exactly their position in seconds after the root; with a real clock it splits a thread apart at the API (a reply's `thread_ts` stops matching its root's `ts`, so a search-hit → `conversations.replies` chain breaks). A reply now resolves its root's stored second; roots, and corpora without reply clocks, resolve to the same number either way.
- **`latest_reply`** was root + reply count. It now reads `MAX(created_ts)` over the thread's *visible* replies — which was also wrong before this change whenever a hidden reply sat mid-thread: the count shrank while the surviving replies kept their seconds, so the served ts pointed at nothing.

The `conversations.replies` ts fast path (resolve a public ts by its integer second over `idx_slack_channel_ts`) needs no change and is covered by a test against an off-metronome reply.

Costs: the root's second is resolved once per response, not once per reply. `conversations.replies` already holds the root row and passes that second down; `search.messages`, whose hits are commonly all in one thread and which has no root row in hand, memoizes it per thread for the page. `conversations.history` serves only roots, which answer from their own row, so it reads nothing extra. `latest_reply` is one indexed `MAX` per thread root, replacing arithmetic.

A stored `created_ts` of `0` is now read as the second it is. `1970-01-01T00:00:00Z` is a time a corpus can write, and the truthiness tests in `_msg_ts` and `_root_epoch` read it as absent — a thread rooted there served a synthesized root ts against replies serving real ones, and `thread_ts` values a client could not fetch the thread with. Both tests moved together, since either one alone still disagrees with the other.

## Verification

- Full suite at this head: 946 passed, 24 skipped (env-dependent Docker/MCP skips). Base at the merge-base `c4faef9`, same environment: 923 passed, same 24 skipped, so the twenty-three tests this branch adds account for the difference. Ruff lint and format clean repo-wide.
- New tests pin: a provided clock is stored and served; a clockless reply lands one second after the message before it (byte-compat with the old rule); a mixed thread stays ordered and collision-free; a backwards clock refuses the import; an unreadable clock refuses it too, and epoch seconds written as a string parse instead; a clockless root is grounded on its replies identically for every `doc_id`; a clockless thread keeps the synthesized root; a refusal against a defaulted second says it was defaulted; a reply's `thread_ts` equals its root's `ts` where position arithmetic would land early; a thread rooted at second `0` stays coherent; the root is read once per response on both endpoints that serve in-thread rows; `latest_reply` equals the last reply's real ts; the replies fast path resolves an off-metronome ts; `_epoch` answers None rather than raising for every number a corpus can write, by string and by JSON literal; a record's own `created` / `updated`, a comment's `created_ts` and a gmail message's `created` each refuse an unreadable value, while an absent one still takes its default.
- The sample BYO corpus's incident thread now clocks all three replies (`18:00:00 / :00:40 / :02:00 / :06:00`), so the record demonstrates the field instead of the cadence it replaces, and the BYO README states the rule.

## Review round

Six commits answer the six inline notes, one each, in reply order:

| Commit | Note |
| --- | --- |
| `99c442c` | `prev_cts` anchored ordering on a hash of the root's `doc_id` |
| `692ff3a` | a present-but-unreadable `created` took the default silently |
| `6db2db3` | the root was re-read once per reply |
| `9bebb8b` | the root `created` description still promised `created+position` |
| `ce8d277` | `if base:` read a stored `0` as missing |
| `8c56f21` | only one of the example thread's three replies carried a clock |

A second round closed the two things those fixes introduced:

| Commit | Note |
| --- | --- |
| `5656ffa` | the new string-to-float step let `'inf'` and `'1e400'` escape as an `OverflowError` traceback |
| `a0352ad` | a record's **own** `created` still took the synthesized second for a typo, while its replies refused |

`a0352ad` widened past the note: the same silent substitution was in a comment's `created_ts` and a gmail message's `created`, and it was never slack-only, so every record-level clock now refuses an unreadable value across source types. `5656ffa` likewise guards the numeric branch as well as the new string one — the corpus is read with a plain `json.loads`, which accepts bare `Infinity`, so the crash had a second door that predates this branch. It also drops a finite second `datetime` cannot render, which used to import clean and raise inside `synth.rfc3339` at serve time.

Still deliberately left: `google.py` and `atlassian.py` test the same `created_ts` column with the same truthiness `ce8d277` fixed for slack, which is older than this branch and belongs to other services; and the optional linear/jira auxiliaries in `_service_columns` keep plain `_epoch`, since an absent one stores NULL rather than a substituted default, so a blank and a typo stay distinguishable there.


backlot/importer/byo.py
```diff
@@ -1174,6 +1174,7 @@ def insert(
                 ),
             )
 
+        prev_cts = created
```
khj809: `prev_cts` starts at the root's *stored* second, which is `synth.epoch(doc_id)` whenever the record has no `created` (line 930). So a corpus that dates its replies for real but leaves the root's clock to the mock gets validated against a hash of its own `doc_id`. With `doc_id` `s-root-0`…`s-root-11` and one reply at `2026-05-01T03:00:00Z`, all twelve load — and serve a thread whose root sits in 2023/2024, up to two years before its own reply. A `doc_id` that happens to hash *after* the reply date refuses the import instead, naming a second (`the previous message is at 1732182639`) that appears nowhere in the corpus; rename the doc_id and the identical corpus loads.

Same root cause one line down: an explicit clock is also refused when it collides with a *defaulted* one (root 18:00:00, reply 1 clockless → silently takes 18:00:01, reply 2 written as 18:00:01 → dies), which kills an ordinary "short ack, then the real answer" thread.

Suggest ordering against corpus-supplied seconds only (track the last explicit clock), or shifting the default rather than refusing the value the author actually wrote — and if it does refuse, saying the previous second was synthesized.

backlot/importer/byo.py
```diff
@@ -1184,10 +1185,23 @@ def insert(
...
-            rep_cts = created + i
+            # A reply is a full message (reactions/files/subtype/edited carry through).
+            # Its time is its own `created` when the corpus wrote one — the gmail
+            # treatment — else one second after the message before it, which lands every
+            # reply of a clockless thread exactly where root+position always put it.
+            # Strictly increasing is enforced rather than assumed: a Slack ts is identity
+            # as well as clock, so two thread messages sharing a second would collide at
+            # the ts every endpoint resolves, and a reply at or before its parent is a
+            # shape real Slack cannot emit.
+            rep_cts = _epoch(rep.get("created"))
```
khj809: A present-but-unparseable `created` is indistinguishable from an absent one here, so a typo silently reinstates the metronome this PR exists to remove. All of these return `None` from `_epoch` and load green at `prev + 1`:

```
'1770746760'           -> None
'1770746760.000000'    -> None
'2026-05-01 03:00 PM'  -> None
'2026-13-01T00:00:00Z' -> None
```

The first two are the shape a Slack `ts` takes — and the shape of the `edited.ts` sitting on the very reply this PR adds to `sample_corpus.jsonl`, so it's the natural thing to write next to it. The schema advertises `"string"` for the field, and a backwards clock refuses loudly three lines below; a malformed one deserves the same treatment rather than the opposite.

backlot/routers/slack.py
```diff
@@ -426,12 +431,13 @@ async def conversations_replies(request: Request):
...
         return _err("thread_not_found")
     rc = sum(1 for x in rows if x["thread_seq"] > 0)
-    latest = _latest_reply(root, rc) if rc else None
+    latest = _latest_reply(conn, root, rc, ids) if rc else None
     ru = store.slack_reply_authors(conn, root["doc_id"], ids)
     ruids = [synth.slack_user_id(e) for e in ru[:5]]
     parent_uid = synth.slack_user_id(root["author_email"])
     messages = [
         _message(
+            conn,
```
khj809, line 443: This re-reads the root by primary key once per reply, though `root` is already resolved and in scope four lines up: each `_message(conn, x, …)` for a reply goes to `_root_epoch` → `store.slack_root_created_ts`. Measured 30 `SELECT created_ts …` for a 30-reply thread in one response (`conversations.history` does 0, as the description says). Passing the root's second (or the root row) into `_message` makes it zero instead of N.

backlot/schemas/slack.schema.json
```diff
@@ -115,6 +115,14 @@
             "type": "string",
             "description": "Optional stable id for the reply."
           },
+          "created": {
```
khj809, line 118: Contract text one field up is now stale: the **root** `created` description (line 95) still reads "Sets the message ts; thread replies follow at `created+position`" — this PR makes that false, and it contradicts the reply-level text you just updated. Since the schemas are the stated source of truth for the ingest contract, it's the sentence a corpus author reads first.

backlot/routers/slack.py
```diff
@@ -724,16 +730,25 @@ def _search_match(conn, row) -> dict:
...
+    """The thread root's base second. A root answers from its own row; a reply looks
+    the root up (a primary-key read). The old shortcut — the reply's own second with
+    its position backed out — assumed every reply sits exactly its position in
+    seconds after the root, which stopped holding once a corpus could write a
+    reply's own clock. Corpora without one resolve identically either way: the
+    root's stored second is the same number the subtraction produced."""
+    if not row["thread_id"] or row["thread_seq"] == 0:
+        return row["created_ts"] or synth.epoch(row["thread_id"] or row["doc_id"])
+    base = store.slack_root_created_ts(conn, row["thread_id"])
+    if base:
```
khj809: `if base:` vs `is not None`: `created: 0` really does store as 0 (line 930 in the importer tests `is None`), and then a reply falls into the position-arithmetic branch this PR retires. With stored `[0, 1, 5000]` the served values are root `1723123144.660197`, reply `thread_ts` `0.660197` and `4998.660197` — three numbers that agree with nothing.

Caveat on the fix: the root's own ts is wrong there because `_msg_ts` has the same truthiness check, which predates this PR — so changing only `_root_epoch` wouldn't make the thread coherent. Fine to leave as consistent-with-existing and handle `created_ts == 0` separately; flagging so it's a choice rather than an oversight. The comment on line 749 (`# a reply whose root row is absent`) is also reached when the root exists with a falsy `created_ts`.

examples/bring-your-own-corpus/sample_corpus.jsonl
```diff
@@ -2,7 +2,7 @@
...
+{"source_type": "slack", "content": "Anyone else seeing 502s from the gateway in us-east?", "channel": "incidents", "group": "engineering", "author_email": "bob@acme.com", "author_name": "Bob Ito", "author_groups": ["engineering"], "visibility": "public", "created": "2026-02-10T18:00:00Z", "reactions": [{"name": "eyes", "count": 2, "users": ["UC20FA2B1C0", "U025E855846"]}], "replies": [{"content": "Yeah, started ~5 min ago. Looking now.", "author_email": "ava@acme.com", "author_name": "Ava Ng"}, {"content": "It's the rate limiter drop — rolling back the refill change.", "author_email": "bob@acme.com", "author_name": "Bob Ito", "reactions": [{"name": "eyes", "count": 1, "users": ["UC20FA2B1C0"]}]}, {"content": "Rollback done, 502s clearing. Postmortem to follow.", "author_email": "ava@acme.com", "author_name": "Ava Ng", "created": "2026-02-10T18:06:00Z", "files": [{"id": "FD09F114C47", "name": "gateway-502-graph.png", "mimetype": "image/png", "title": "502 rate graph"}], "edited": {"user": "UC20FA2B1C0", "ts": "1770746880.000000"}}], "participants": ["bob", "ava"]}
```
khj809: Only the third of three replies gets a clock, so the showcase thread still reads 18:00:00 / :01 / :02 → 18:06 — the metronome is still the first thing a reader sees for the feature the record is meant to demonstrate. It also leaves reply 1 ("started ~5 min ago. Looking now.") one second after the root, at odds with its own text. Clocking all three (say :00, +40s, +2m, +6m) would make the example teach the field and read like a real incident.

khj809 (commented):
Reviewed on a worktree at 384baa1. The design holds up: `slack_fmt_ts` keys the micro-fraction on `thread_id` alone, so within a thread the integer second is the *only* discriminator — strict increase isn't belt-and-braces, it's required. `thread_ts` re-grounding and the `MAX(created_ts)` `latest_reply` are both right, `routers/slack.py` is the only site deriving either, and back-compat is airtight because `replies.items` was already `additionalProperties: false`. Full suite green here (986 passed, 1 skip in this env — more ran than the 926/24 the description reports), ruff check + format clean.

Six notes inline. The two on the importer are the ones I'd want resolved before merge; the rest are small.

backlot/importer/byo.py
nuri-yoo: Confirmed both halves and fixed in 99c442c. Reproduced first: with the twelve `s-root-N` doc_ids and one reply at `2026-05-01T03:00:00Z`, all twelve loaded with roots spread over 2023-01 to 2024-11 — up to 3.3 years, not two, since `synth.epoch` draws from `[2023-01-01, 2025-01-01)` and a 2026 reply always wins.

One correction to the refusal half, which makes it worse rather than better. A 2026 reply can never be refused: no synthesized second can reach past 2025-01-01. The `1732182639` in your note is from a reply dated inside the synthesizer's own window, and that is the case that matters — it is the window a corpus dating things "realistically" lands in. With one reply at `2024-06-01T00:00:00Z` over 40 doc_ids, 18 refused and 22 loaded, so the import is a coin flip on sha256 of the root's doc_id. `s-root-2` died naming `the previous message is at 1729182155`; renaming it to `s-root-0` loaded the identical file.

Took the first of your three suggestions, plus the re-grounding it implies. The thread's seconds are now resolved together in `_thread_seconds` before any of its rows are written, so a root whose clock was synthesized is re-grounded on the first reply that carries one, and ordering is judged only against corpus-supplied seconds. The 40-doc_id sweep now loads 40/40 with no reply at or before its root. Re-grounding cannot touch an existing corpus, for the reason you gave in the summary: `replies.items` was already closed, so no corpus can have a reply `created` yet.

The second half stays a refusal, and I want to flag that as a decision rather than let it read as one. Root `:00`, a clockless reply, then an explicit `:01` has no integer second available for the clockless one, and a Slack ts is identity as well as clock, so the two cannot share `:01`. There is nothing to assign. What I did instead is your last clause — the message now says the second it collided with was defaulted, and names two ways out:

```
line 1: reply 2: created must be after the message before it (got '2026-02-10T18:00:01Z'),
but reply 1 carries no created of its own and took the next free second, 1770746401.
Give reply 1 a created too, or move this one later.
```

Where there is room it loads as before: moving that reply to `:05` gives `18:00:00 / 18:00:01 / 18:05:00`. If you read the unsatisfiable case as still worth admitting some other way, say so and I will take another pass.

One thing I left alone deliberately: a malformed or backwards `created` on the **root** still falls back to the synthesized second silently. The reply site has an ordering contract of its own that makes a loud refusal consistent there; the root has none, and adding one changes how existing corpora load. Called out in the commit message.

backlot/importer/byo.py
nuri-yoo: Confirmed and fixed in 692ff3a. All four returned `None` from `_epoch` and loaded green at `prev + 1` exactly as you list.

Split the four, because two of them are not typos. `'1770746760'` and `'1770746760.000000'` are epoch seconds, the schema admits `string` for the field, and — your point about the `edited.ts` next door — that is the form the value beside them is already in. Those now parse: the reply lands at 18:06:00, where the author meant it, instead of 18:00:01. ISO is still attempted first, so an 8-digit basic-format date stays a 2026 date rather than becoming a 1970 second.

The remaining two are refused, naming the value and the two accepted forms:

```
line 1: reply 1: created is not a time this importer can read
(got '2026-05-01 03:00 PM'; write epoch seconds or ISO 8601)
```

`_epoch` still answers `None` for both a missing and an unreadable value — a sentinel there would be read as a truthy second by the several callers that test `is None` — so the reply site distinguishes them with `_time_given`, which asks only whether the corpus wrote anything in the field.

Worth recording from writing the test: `"created": true` never reaches this code. The schema's own `["integer", "number", "string"]` rejects it a layer up, with `replies/0/created: True is not of type ...`. I had a bool in the parametrized list and it failed for that reason.

backlot/routers/slack.py
nuri-yoo: Reproduced your 30 exactly, and found the same N+1 on a second endpoint you did not mention. Fixed in 6db2db3.

Counting calls into `store.slack_root_created_ts` for one response over a 30-reply thread:

```
                        before  after
conversations.replies      30      0
conversations.history       0      0
search.messages            30      1
```

`search.messages` reaches it through `_search_match` -> `_root_epoch` at line 733, and it is the worse of the two in shape: a page of hits is commonly all in one thread, and unlike `conversations.replies` it has no root row in hand to pass down. So the fix is two-part — `conversations.replies` resolves the root's second once and passes it to `_message` as `root_epoch`, and the search path takes a per-response memo keyed on `thread_id`.

Both still serve the same `thread_ts`. The new test in `tests/test_slack.py` asserts the read counts and that every message in the thread carries the root's ts, on both endpoints; against the previous `_message` it fails with `one root read per reply is back (30 reads)`.

backlot/schemas/slack.schema.json
nuri-yoo: Fixed in 9bebb8b. The root's description now reads:

> Creation time (epoch seconds or ISO 8601). Default: synthesized from doc_id. Sets the message ts. A thread reply times itself: its own `created` when it has one, else one second after the message before it.

Grepped the tree for other copies of the retired rule (`created+position`, `created + position`, `replies follow`) — the only other hit is a docstring in `tests/test_importer_byo.py` about the *comment* thread rule, which still holds.

backlot/routers/slack.py
nuri-yoo: Took the fix rather than the note, in ce8d277. Your caveat is right and it decided the shape: `_msg_ts` has the same truthiness check, it dates from `c9ba95b` (the initial public release) per `git log -L`, and changing only `_root_epoch` leaves the thread just as incoherent — so both moved together.

Reproduced with `created: 0` and replies at `1` and `5000` (my doc_id differs, so the synthesized constant differs from yours):

```
before   root ts 1687203164.267301   reply thread_ts 0.267301 and 4998.267301
after    root ts 0.267301            all three thread_ts 0.267301
```

The `4998` confirms your reading that the reply fell into the position arithmetic this branch retires. Also worth naming: before the fix a reply's `thread_ts` did not equal the root's `ts`, so a client could not fetch the thread with the `thread_ts` it had just been handed.

Every test on the column in `_root_epoch` and `_msg_ts` is now against `None`. Your point about the `# a reply whose root row is absent` comment was right too — `slack_root_created_ts` returns `None` both for an absent row and for a present one storing no clock, so the comment now says either.

`google.py:639` and `atlassian.py:880` test the same column the same way. Other services and older than this branch, so left alone here; noted in the commit message.

examples/bring-your-own-corpus/sample_corpus.jsonl
nuri-yoo: Took your timings in 8c56f21. The thread now imports as:

```
seq=0  18:00:00  Anyone else seeing 502s from the gateway in us-east?
seq=1  18:00:40  Yeah, started ~5 min ago. Looking now.
seq=2  18:02:00  It's the rate limiter drop — rolling back the refill change.
seq=3  18:06:00  Rollback done, 502s clearing. Postmortem to follow.
```

Both points held: it was `18:00:00 / :01 / :02 / 18:06`, and the reply reporting something five minutes old sat one second after the message it answers.

backlot/importer/byo.py
```diff
@@ -130,10 +138,85 @@ def _epoch(v):
     s = str(v).strip().replace("Z", "+00:00")
     try:
         return int(datetime.fromisoformat(s).timestamp())
+    except ValueError:
+        pass
+    # Epoch seconds written as a string — `"1770746760"`, or the `<sec>.<frac>` a Slack
+    # ts takes, which is the form an `edited.ts` next door is already in. ISO is tried
+    # first, so an 8-digit basic-format date stays a date rather than becoming a second.
+    try:
+        return int(float(s))
```
khj809: The new float fallback is guarded by `except ValueError` only, so `'inf'`, `'-inf'` and `'1e400'` escape as an uncaught `OverflowError` and come out of `load()` as a traceback — which is the outcome the `created is not a time this importer can read` refusal added in this same commit exists to prevent. `except (ValueError, OverflowError)` on the next line covers it. (`'nan'` is already fine: `int(nan)` raises `ValueError`.)

backlot/importer/byo.py
```diff
@@ -926,13 +1009,17 @@ def register(email: str | None, name: str | None = None) -> None:
         # created_ts must never be NULL (the server sorts/filters by it; a NULL would need a
         # runtime null-check). Fall back to the same deterministic synth.epoch the server would have
         # synthesized for a missing ts, so the served time is unchanged — just materialized now.
-        created = _epoch(rec.get("created"))
-        if created is None:
-            created = synth.epoch(doc_id)
+        created_written = _epoch(rec.get("created"))
```
khj809: The root's own clock keeps the old treatment — plain `_epoch`, so a filled-in-but-unreadable value still falls to `synth.epoch(doc_id)` without a word. A record with `"created": "2026-13-01T00:00:00Z"` loads green with its root at 2024-02-24, while the identical typo on a reply now refuses. `_time_given` is right there and makes it a one-liner, but it would reach every source type and every time field, so whether it belongs here or in a follow-up is your call — flagging only because this fix made the asymmetry sharper than it was before.

khj809 (commented):
Re-checked all six on a worktree at `8c56f21111b32fe3c327d8a0dddd2e28ae3af00a`. Everything I raised is closed, and I verified each rather than reading for it: the clockless-root re-grounding makes `doc_id` irrelevant (r0…r3 all resolve to 02:59:58 / :59 / 03:00:00), a typo'd reply clock refuses while `'1777604400'` and `'1777604400.000000'` now parse, `conversations.replies` went 30 root reads → **0** for a 30-reply thread, a thread rooted at second 0 now has its root ts, `thread_ts` and reply ts agree, and the clockless path is still byte-identical to root+position. Suite green, ruff clean. Nice call refusing the defaulted-second collision with a message that names the cause instead of shifting the default — minimal packing means those really are impossible threads.

Two small things the fixes themselves introduced, inline.

backlot/importer/byo.py
nuri-yoo: Confirmed, and it is a regression the previous round introduced rather than something pre-existing — worth stating plainly. Fixed in 5656ffa, with one extension I want to flag because your one-liner alone would not have closed it.

Reproduced end-to-end first. Each of these came out of `load()` as a traceback, not the refusal:

```
reply created='inf'       TRACEBACK: OverflowError: cannot convert float infinity to integer
reply created='-inf'      TRACEBACK: OverflowError
reply created='1e400'     TRACEBACK: OverflowError
reply created='Infinity'  TRACEBACK: OverflowError
reply created='nan'       refused cleanly
```

`'nan'` behaves as you said. Two spellings beyond your list reach it too: `float()` accepts `'Infinity'` and `'-Infinity'`, and `'1e309'` overflows the same way as `'1e400'`.

The extension. I checked base to attribute the regression and found the crash has a second door that predates this branch:

```
base _epoch('inf')  -> None                    <- the string door, opened in 692ff3a
base _epoch(inf)    -> RAISED OverflowError    <- the numeric door, always open
```

The corpus is read with a plain `json.loads`, which accepts bare `Infinity` and `NaN` as number literals, so a record can hand a float straight to the unguarded `int(v)` in the numeric branch. Verified at the previous head with a corpus line carrying `"created": Infinity` unquoted — traceback. So `except (ValueError, OverflowError)` on the float fallback would have fixed the door I opened and left the older one; both now go through one `_seconds` helper.

`_seconds` also drops a finite second `datetime` cannot render, which is the same failure moved later rather than avoided: `1770746760000` (milliseconds where seconds belong, year 58105) imported clean and then raised inside `synth.rfc3339` when a router dated the row, so the corpus typo surfaced as a server error long after the import reported success. Representability is asked of `datetime.fromtimestamp` — the call the routers date through — rather than compared against a bound written down in the importer.

Seconds a corpus legitimately writes are unaffected, `0` and pre-1970 included. Tests cover all eight unservable spellings by string and by JSON literal, plus the values that must still parse.

backlot/importer/byo.py
nuri-yoo: Taking it here rather than in a follow-up, in a0352ad. You are right that the previous round sharpened the asymmetry, and right that `_time_given` makes it small — but checking the blast radius turned up more of the same defect than the note describes, which is what decided it for me.

Confirmed your case first (my doc_id differs, so the date does):

```
s-typo-root   created_ts = 2024-08-12T11:57:45  = synth.epoch(doc_id)   <- record wrote "2026-13-01T00:00:00Z"
cf-typo       created_ts = 2024-05-18T04:14:26  = synth.epoch(doc_id)
              updated_ts = NULL                                        <- record wrote "2026-05-01 03:00 PM"
```

So it was never slack-only — every source type reads `created` through that line — and `updated` was quietly becoming NULL alongside it. Two more sites had the identical substitution the reply clock had, in sibling code: a comment's `created_ts` falls back to the second after the comment before it, and a gmail message's `created` to the root's clock plus an hour. Both took that default for a typo without a word.

The rule I settled on is narrower than "every time field", which is why it did not have to reach everywhere: **a time refuses an unreadable value wherever an absent one gets a default put in its place.** That covers a record's `created` (a hash of the doc_id), its `updated`, a comment's `created_ts`, and a gmail message's `created`. It deliberately leaves the linear and jira auxiliaries in `_service_columns` — `closed_at`, `archivedAt`, `completedAt` and the rest — on plain `_epoch`: an absent one stores NULL, so nothing is substituted and a blank never becomes indistinguishable from a typo. That function also has no `where` to name the record with, and threading one through for fields that are not being disguised did not seem worth it.

All five refusals verified by running them, including the numeric door from the note above:

```
root created (slack)   line 1: created is not a time this importer can read (got '2026-13-01T00:00:00Z'; ...)
root created (conf)    line 1: created is not a time this importer can read (...)
root updated           line 1: updated is not a time this importer can read (got '2026-05-01 03:00 PM'; ...)
comment created_ts     line 1: comment 1: created_ts is not a time this importer can read (got 'nope'; ...)
gmail message created  line 1: message 1: created is not a time this importer can read (...)
root created = Infinity line 1: created is not a time this importer can read (got inf; ...)
```

The risk in a change like this is refusing the blank case too, so there is a test next to the refusals asserting an absent `created` still keeps `epoch(doc_id)`, an absent `updated` still stores NULL, and an undated comment still follows the one before it. I also checked the refusal tests fail without the change (`DID NOT RAISE SystemExit`) so they pin the behaviour rather than passing incidentally.

`erb.py` has its own `to_epoch` and is untouched: it reads a machine-generated artifact rather than a file someone typed, so a silent fallback there is not hiding an author's mistake.

khj809 (approved):
LGTM

Issue #51, opened 2026-08-12
Retire doc_id: per-source id/served_id, ACL and FTS


khj809:
`doc_id` is an ERB-era identity that the code has outgrown. It was adopted as a corpus-wide document
key, it never actually was one, and two live defects follow directly from continuing to treat it as
one. Neither is patchable where it surfaces — both are consequences of the identity model, so the
fix is structural.

## `doc_id` was never corpus-wide

`tests/test_importer_byo.py::test_two_sources_may_share_a_doc_id` states it outright — *"Ids are per
service, not corpus-wide"* — and records that the bench has three documents appearing under two
sources with the same `dataset_doc_uuid`. The real key has always been `(source_type, doc_id)`.

Everything keyed on `doc_id` alone therefore conflates documents that are not the same document.

### Defect 1 — grants merge across sources, and a restricted document becomes readable

`doc_acl` is `PRIMARY KEY (doc_id, principal_type, principal_id)`. Two documents in different
sources sharing a `doc_id` share their ACL rows, and the union is what gets enforced.

```python
# a public confluence page and a drive file restricted to hana, sharing doc_id 'shared-1'
{"source_type": "confluence",   "doc_id": "shared-1", "visibility": "public", ...}
{"source_type": "google_drive", "doc_id": "shared-1", "readers": ["hana@acme.com"], ...}
```

```
doc_acl rows for shared-1:
  org    acme            <- from the confluence page
  user   hana@acme.com   <- from the drive file

bob@acme.com  sees drive docs: ['Secret sheet']   # in no granted group
hana@acme.com sees drive docs: ['Secret sheet']
```

`bob` reads a file restricted to `hana` alone, because the confluence page contributed `org:acme` to
the same rows. This is the corpus shape the bench already contains.

### Defect 2 — a fifth of a project's issues are unreachable by their own key

A served id is a hash of `doc_id` into a per-vendor range, resolved back through a reverse map built
at startup (`main._build_index`). The map keeps whichever `doc_id` hashed last; the rest are
shadowed. Measured with the real synth functions over the bench's own ids:

| source | container | documents | distinct served ids | unreachable |
|---|---|---|---|---|
| jira | `customer-support` | 4,584 | 3,125 | **1,459 (32%)** |
| jira | `internal-support` | 1,536 | 1,238 | **298 (19%)** |
| github | `redwood` | 4,297 | 4,201 | 96 (2%) |

[...]

PR #54, opened 2026-08-13
linear: a provided identifier's prefix is a fact about its team, and a superseded id claim is released

### LINKED ISSUE #52: Linear: a provided identifier resolves, but its prefix teaches the team nothing — one team serves two spellings and two teams can share one
#46 gave jira and github the rule that a provided tracker id is stored verbatim and served resolvable, and its body notes "This is the treatment Linear's `identifier` already gets." That holds for resolution — `issue(id: "ENG-7")` finds the row — but the second half of what #46 built for jira never reached Linear: a provided key's prefix becoming a fact about the container (commit `0eff060`, "a provided prefix is 1:1 with its project"). Linear still derives every container-level spelling from the team's *name*, so a corpus that writes its own team keys gets a served surface that contradicts both itself and the real API's data model.

Reproduced on main + #50 (`020d00d`), three records:

```json
{"source_type": "linear", "doc_id": "ln-a", "team": "payments-platform", "identifier": "ENG-7", ...}
{"source_type": "linear", "doc_id": "ln-b", "team": "payments-platform", ...}
{"source_type": "linear", "doc_id": "ln-c", "team": "growth", "identifier": "ENG-9", ...}
```

What loads and serves, all without a word at import:

1. `ln-a` resolves at `ENG-7` (good — that is #46's verbatim rule working), but its issue object serves `team { key: "PP" }`: `graphql/linear_resolvers.py` derives the key from the container name (`synth.linear_team_key("payments-platform")` → `PP`) in every place a team is rendered. The identifier's prefix is not its team's key — the same self-contradiction the #46 review flagged for jira ("the issue key's prefix is not its project's key") and `0eff060` refused at load.
2. `ln-b`, the keyless sibling, is materialized at import as `PP-711` — so one team serves two spellings at once (`ENG-7` beside `PP-711`). Real Linear cannot represent this: an issue's identifier *is* `team.key + "-" + number`, so every issue of a team carries the same prefix by construction.
3. `team(id: "ENG")` resolves nothing; the corpus's own prefix never becomes navigable at the team level (`linear_teams` maps `PP` → the team). The jira analogue — `project = PAY` returning zero rows for a corpus that cites PAY keys — is exactly what #46's prefix aliasing fixed.
4. `ln-c` loads a second team also answering ENG-prefixed identifiers. Real Linear enforces workspace-unique team keys (the identifier scheme depends on it: one `ENG-9` must resolve to one issue). jira got both directions of this 1:1 refused at load in `0eff060`, with `seed_tracker_ids` carrying the claim across `--append`; Linear has neither direction.

Why this is a semantics decision rather than a bug fix, and so an issue rather than a PR: Linear's identifier is materialized at load (`byo.py`, "an id that is served has to be resolvable... it is just written down now"), and the comment right below it explains why jira and github deliberately do the opposite — "Written here the prefix would be the synthesized one, since the record cannot know whether a sibling issue in the same project provides `PAY-7`." That reasoning indicts the Linear path too, but the cure #46 chose for jira (derive at index-build time, where the whole container is visible) conflicts with the property the materialization exists for: `issue(id:)` resolves through the stored column (`store.linear_issue_by_identifier`), not through the in-memory index, so Linear's identifier is resolvable *only* from its column. Moving Linear to build-time derivation, rewriting materialized siblings when a container's first provided prefix arrives, or refusing provided prefixes that disagree with the name-derived key are three different contracts, and `--append` (a later shard may be the first to provide a prefix for a team whose keyless rows are already materialized) cuts across all three — the same trade `_free_number`'s docstring records for derived ids ("a served id is a function of the container's whole row set").

Whichever way it lands, two pieces look uncontroversial and jira-shaped:

- Load-time 1:1 validation between provided prefixes and teams, both directions, seeded across `--append` — the `0eff060` treatment. Silent acceptance is the only clearly wrong state.
- The issue object should not contradict itself: if `identifier: "ENG-7"` is served, `team.key` for that issue should say `ENG` (or the load should have refused the record).

This is not only a contrived shape: a generated corpus hit it without trying. Its team containers are the tracker keys themselves, which keeps the derived prefix and the provided one aligned for every key of three letters or fewer — and the one team whose key ran to four (`GSSS`) shipped identifiers the served team would contradict, since `_key` truncates a single word to three letters (`GSSS` → `GSS`). The corpus side now caps its keys to stay a fixed point of that derivation, but a corpus has to read `synth._key` to learn that rule; load-time validation is where it would have been told.

For reference, the real API's model: a team's key is the prefix of every one of its issue identifiers, keys are unique within a workspace, and `issue(id:)` accepts the UUID or the human identifier — https://linear.app/developers/graphql, https://linear.app/docs/teams.

nuri-yoo:
Closes #52. Rebased onto `main` (through #53). Happy to fold this into `issue-51/per-source-identity` instead if that lands first — see the cross-reference on #52.

## What

Two things #46 built for jira never reached Linear, plus the claim residue named on the #46 thread.

A provided identifier resolved (that much worked) but taught the container nothing: its keyless siblings materialized with the name-derived prefix, so one team served two spellings at once; the issue object served `identifier: "ENG-7"` beside `team { key: "PP" }` — contradicting the invariant real Linear is built on (an identifier is derived *from* its team's key); `team(id: "ENG")` resolved nothing; and a second team providing ENG-prefixed identifiers loaded without a word. All four are reproduced in #52 against `main`.

- **The 1:1 both-directions refusal jira got in `0eff060` now covers linear teams**, seeded across `--append`, riding on the same claim machinery: a provided identifier claims its spelling and its prefix, and a violation refuses the import naming the holder.
- **A container whose rows do not agree on one prefix is settled at the end of the load**, where every row is visible. Stamping inside the record loop can only use the prefix known so far, so the two halves of a container would otherwise be decided by different rules; instead every materialized identifier in such a container is assigned in doc_id order under the container's key. What a keyless row ends up with is therefore a function of the row set, not of which side of the provided line it sat on.
- **The free-number search spans the workspace rather than the container.** A stated prefix can be the key some other team derives from its name — a pairing no claim covers, because nothing was written down twice — and an assignment landing on that team's row would leave one document unreachable at the only id it advertises.
- **A team bigger than the number space still imports.** The search ends the way `main._free_key` ends, handing the derived value back once every number is spoken for. A repeated identifier is a shape this mock already serves (one real corpus repeats 5,055 keys, and `issue(id:)` answers a repeat with the first row by doc_id), and the teams large enough to fill the space are exactly the ones worth having.
- **A materialized identifier is a hash rather than a claim, so seeding no longer records one.** Seeding it refused a later shard that stated that exact spelling — the corpus's own id losing to a synthesized one, the inversion #46 exists to prevent — and pointed the refusal at a holder that had only landed there. The stated id wins and the row sitting on it steps aside during the same load.
- **The serving side derives a team's key from its own rows** (a prefix that differs from the name-derived key is a corpus-provided fact and wins) in one startup map, read by the team object, `team(id:)` lookup, the teams-listing filter, and the compiled `IssueFilter`'s derived `team.key` expansion — including the branches nested under `and` / `or`, so one query written two ways cannot disagree. Ties resolve the way the github and jira indexes resolve them: what the corpus wrote claims its spelling first, and only then does a derived one take what is left. That holds for team keys too, so `team(id: "ENG")` answers with the team the corpus called ENG rather than the one whose name happens to shorten that way.
- **`identifier` carries a pattern now**, for the reason jira's `key` carries one: the prefix is a fact about the whole team, so a single mistyped identifier renames it for every one of its issues. `identifier: "7"` used to make the team answer at `7` and stamp its siblings `7-n`.
- **Riding along**: a repeated doc_id re-stating a *different* id (or none) now takes its old claim with it, this run or seeded across `--append` — a third record providing the superseded id was refused and told a holder holds it when the row had already moved on. Applies to github numbers, jira keys and linear identifiers alike.

**Deliberately not here:** unwinding the materialization itself (identifier as a provided-only column, resolution through the index). That is #51's identity work, and this PR leaves the identifier column's semantics exactly where that work expects to find them. One consequence is worth naming: a value an earlier run moved off its derived spelling is not recomputable, so the next run reads it as stated. That is what keeps a re-import or an append from renumbering an id a client may already hold, and it is the residue #51 removes.

## How provided and materialized values are told apart

They share a column, but a materialized value is recomputable — it is exactly what materialization writes for `(doc_id, the container's key)`, under the name-derived key for a row stamped before the prefix arrived and under the claimed one for a row stamped after. Both readings are checked, so a row is treated as the corpus's own spelling only when it matches neither; that spelling never moves. The container's key is itself only known once the rows have been read, which is why seeding and settling each make two passes. A corpus that provides precisely the value materialization would have written is indistinguishable — and needs nothing: both readings leave the same claims and the same served surface.

## Verification

- Full suite: **1006 passed, 24 skipped** (base at this checkout: 989, same skips). Ruff lint and format clean repo-wide. The jira/github claim refactor keeps every error string byte-identical — the existing wording-matching tests pass unmodified.
- New tests pin, and each one fails against the code without the change: identifiers that do not depend on where the provided line sits; a team larger than the number space importing at all; an assignment never landing on another team's identifier; a stated identifier taking a spelling a materialized row was sitting on; re-import and append leaving every existing identifier alone; five shapes of mistyped identifier refused; `team.key` agreeing flat and under `and` / `or`; a stated team key outranking one another team only derives from its name; a stated identifier outranking another team's derived one.
- Also pinned from the original work: the in-flow materialization, `--append` in all three shapes (stored keyless shard settled when the prefix first arrives, later keyless shard materializing off the seeded prefix, cross-run prefix refusal); both 1:1 directions; the superseded-claim release within a run and across `--append`; and the served surface end to end — issue object, `team(id: "ENG")`, teams filter, and compiled issue filter all speaking the provided prefix, with the keyless sibling's identifier carrying it too.
- Measured on a single team of 12,000 issues, past the 9,000 the identifiers draw from: settling costs 0.72s. Counting how much of a prefix's space is spoken for is what keeps that linear — walking the whole space once per row cost 3.1s on the same corpus.
- Found while writing it: a generated corpus hit this in the wild — one team of 82 whose four-letter tracker key (`GSSS`) the name-derivation truncates to `GSS` shipped identifiers its own served team contradicted.

### DIFF
--- a/backlot/graphql/linear_filters.py
+++ b/backlot/graphql/linear_filters.py
@@ -242,13 +242,15 @@ def _labels_predicate(spec: dict, *, every: bool) -> tuple[str, list]:
 # field name on IssueFilter -> how it compiles.
 #   ("col", column, kwargs)    a comparator straight onto a column
 #   ("nested", {sub: spec})    an object filter whose sub-fields map onto columns
-def compile_issue_filter(conn, flt: dict | None) -> tuple[str, list] | None:
+def compile_issue_filter(
+    conn, flt: dict | None, team_keys: dict | None = None
+) -> tuple[str, list] | None:
     """``IssueFilter`` -> ``(sql_fragment, params)``, or None when there is nothing to filter."""
-    sql, params = _issue_filter(conn, flt or {})
+    sql, params = _issue_filter(conn, flt or {}, team_keys)
     return (sql, params) if sql else None
 
 
-def _issue_filter(conn, flt: dict) -> tuple[str, list]:
+def _issue_filter(conn, flt: dict, team_keys: dict | None = None) -> tuple[str, list]:
     parts: list[str] = []
     params: list = []
 
@@ -261,7 +263,10 @@ def _issue_filter(conn, flt: dict) -> tuple[str, list]:
         if spec is None:
             continue
         if key in ("and", "or"):
-            subs = [_issue_filter(conn, s) for s in spec]
+            # `team_keys` rides down the recursion: without it a `team.key` nested under and/or
+            # compiled against the name-derived key while the same filter at top level compiled
+            # against the corpus's, so the two spellings of one query disagreed.
+            subs = [_issue_filter(conn, s, team_keys) for s in spec]
             frags = [f for f, _ in subs if f]
             for _, p in subs:
                 params.extend(p)
@@ -330,7 +335,13 @@ def _issue_filter(conn, flt: dict) -> tuple[str, list]:
                     spec,
                     {
                         "name": ("col", "team"),
-                        "key": ("derived", "team", synth.linear_team_key),
+                        # The key a team's own identifiers carry (corpus-provided
+                        # prefixes win, built at startup); name-derived otherwise.
+                        "key": (
+                            "derived",
+                            "team",
+                            lambda name: (team_keys or {}).get(name) or synth.linear_team_key(name),
+                        ),
                         "id": ("derived", "team", synth.linear_team_id),
                     },
                 )
--- a/backlot/graphql/linear_resolvers.py
+++ b/backlot/graphql/linear_resolvers.py
@@ -28,6 +28,15 @@ def _ctx(info):
     return info.context
 
 
+def _team_key(container: str, info) -> str:
+    """The key this team's own identifiers carry — a corpus-provided prefix when its
+    rows have one (built at startup from the stored column), else the name-derived
+    key. Deriving from the name unconditionally served `identifier: "ENG-7"` under
+    `team { key: "PP" }` — the object contradicting itself, since real Linear derives
+    an identifier FROM its team's key."""
+    return info.context.get("team_keys", {}).get(container) or synth.linear_team_key(container)
+
+
 def _org(info) -> str:
     """The workspace slug, which is what a Linear URL is keyed on."""
     return _ctx(info).get("org") or "org"
@@ -171,12 +180,12 @@ def _match_fields(spec: dict | None, fields: dict) -> bool:
     return True
 
 
-def _match_team(container: str, spec) -> bool:
+def _match_team(container: str, spec, info) -> bool:
     return _match_fields(
         spec,
         {
             "name": container,
-            "key": synth.linear_team_key(container),
+            "key": _team_key(container, info),
             "id": synth.linear_team_id(container),
         },
     )
@@ -518,7 +527,7 @@ def _team(container: str, info) -> dict:
     """A ``Team``. 42 of its fields are non-null in the SDK's fragment; the ones the mock cannot
     know take Linear's own product defaults (cycles off, 2-week duration, estimate scale
     ``notUsed``) rather than zero values that would read as configured."""
-    key = synth.linear_team_key(container)
+    key = _team_key(container, info)
     created = synth.rfc3339(synth.epoch("linear-team:" + container))
     return {
         "id": synth.linear_team_id(container),
@@ -585,7 +594,7 @@ def _issue(row, info) -> dict:
     sort orders, bot actors and shared access are declared because `@linear/sdk`'s fragment
     selects them, and resolve empty because a document corpus has nothing behind them."""
     identifier = row["identifier"] or synth.linear_identifier(
-        row["doc_id"], synth.linear_team_key(row["team"])
+        row["doc_id"], _team_key(row["team"], info)
     )
     title = row["title"] or ""
     created = row["created_ts"]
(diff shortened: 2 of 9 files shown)


nuri-yoo:
Absorbed into #58 and rewritten: #60.

#58 removed the startup index this branch built the team-key map on, and dropped the `doc_id` column its re-stamp ordering and its claim seeding both read. It also landed the identifier materialization and the `linear_teams.served_key` column independently, so part of this diff is now duplicate and the rest needed a different mechanism (the key is stored on the team row and read back from there, not held in a startup map).

#60 carries the behaviour forward against the post-#58 model, minus two pieces #58 made unreachable: releasing a superseded id claim, and a later shard taking a spelling an earlier run materialized. Both are explained there.

PR #56, opened 2026-08-13
GitHub: honour X-GitHub-Api-Version, and shape a pull like a pull (#55)

### LINKED ISSUE #55: GitHub: response bodies diverge from the real API — removed fields still served, and a pull shaped like an issue
# GitHub: response bodies diverge from the real API — removed fields still served, and a pull shaped like an issue

Two fidelity gaps in the JSON bodies, found by diffing key sets against `api.github.com`. Both are
about the *shape* of a response rather than a missing route, so nothing 404s — a client just reads
fields that real GitHub does not send, and misses ones it does.

Measured against the deployed instance today: `redwoodinference/audit-log-exporter` pull `20770`
versus `brekkylab/backlot` pull `54` on `api.github.com`. **Both requests carried
`X-GitHub-Api-Version: 2026-03-10`.**

## 1. Fields removed in `2026-03-10` are still served

Real honoured the pinned version; this did not.

| | real | deployed |
|---|---|---|
| `pull.assignee` | absent | **present** |
| `pull.merge_commit_sha` | absent | **present** |
| `issue.assignee` | absent | **present** |

`assignee` (singular) was removed in favour of the `assignees` array, and `merge_commit_sha` was
removed from every endpoint returning a pull.

The awkward part is not the fields themselves — it is that the version header is **accepted and then
ignored**. A client that pins a version has no way to tell it had no effect, which is worse than not
supporting the header at all. Picking one version to emulate and stating it in the docs would settle
it either way; varying the payload per version is the fuller answer but a bigger one.

## 2. A pull is shaped like an issue

Key-set diff on `GET /repos/{owner}/{repo}/pulls/{number}`:

```
real-only:  _links, auto_merge, commits_url, maintainer_can_modify,
            review_comment_url, review_comments_url, statuses_url
mock-only:  assignee, closed_by, events_url, labels_url, merge_commit_sha,
            pull_request, reactions, repository_url, state_reason, timeline_url
```

Every field on the second line belongs to an **issue**. `pull_request` is the clearest tell: that is
the marker GitHub puts on an *issue* to say it is really a pull, and a real pull object does not
carry it — it has no reason to point at itself.

What is missing costs a client its hypermedia. `_links` and `review_comments_url` exist so a caller
can *follow* to a pull's sub-resources; without them the caller has to assemble those URLs itself,
which is the thing those fields are there to avoid. The repository object has the same gap — the
whole `*_url` template family is absent — and that is what an SDK follows to complete an object
lazily, including the PyGithub example this repo ships.

**No stored data is needed.** `commits_url`, `statuses_url`, `review_comments_url` and `_links` are
all derivable from owner/repo/number; `auto_merge` is `null` and `maintainer_can_modify` a boolean
default. The issue-only fields just come off.

## Note on the overlap

`assignee` and `merge_commit_sha` appear in both sections — they are removed fields *and* part of
what makes the pull object issue-shaped. Fixing 2 handles them on the pull; 1 still covers the
issues endpoint and the version question behind it.

## Not at issue

`tree` and `pulls/{n}/files` returned **identical key sets** to real. `Link` is present and correct
on paginated listings. This is narrowly about the two bodies above.

khj809:
Closes #55 — two response bodies that diverge from the real API, and the version header that was supposed to govern them.

Nothing here 404s. A client reads fields real GitHub does not send, misses ones it does, and pins an API version that has no effect — all with a 200, which is why none of it showed up as a failure.

## The version header was the load-bearing half

`X-GitHub-Api-Version` was **accepted and ignored**. That is worse than not supporting it: a client that pins a version gets a payload from a different one and has no way to tell, so its own version handling looks tested when it is not.

`2026-03-10` turns out to be a real, currently-supported version — GitHub names it itself when you send a bogus one:

```
The version you specified in the "X-GitHub-API-Version" request header, "1999-01-01", is not a
supported version. The following versions are currently supported: "2026-03-10" (most recent) and
"2022-11-28".
```

Both are served now, and the difference on this surface is exactly three fields:

| | `2022-11-28` | `2026-03-10` |
|---|---|---|
| `issue.assignee` | present | **removed** (superseded by `assignees`) |
| `pull.assignee` | present | **removed** |
| `pull.merge_commit_sha` | present | **removed** |

- **Unpinned is `2022-11-28`**, which is what real defaults to — read off the API, not chosen here.
- **Every response reports its choice** in `X-GitHub-Api-Version-Selected`, including a 404. A rejected version gets no echo, because it selected nothing — also real's behaviour.
- **An unsupported value gets real's own 400, ahead of the credential and the owner check.** That ordering is real's and is verified rather than assumed: `api.github.com` 400s a bad version on a repo that does not exist while sending no credentials at all. It matters to the caller — a header typo reported as 401 sends them to their token, as 404 to their path.

Which versions carry a field is stated as **who has it**, not who dropped it (`_HAS_SINGULAR_ASSIGNEE`), so a third version has to answer the question instead of inheriting whichever side the condition was written on.

The echo is middleware rather than a router dependency for a specific reason: a dependency's headers are dropped whenever the route returns a `Response` itself, which the raw-content and diff media types do — so exactly the responses whose *shape* the header describes would have shipped without it.

## A pull is a pull, not an issue with extra keys

`_pr_obj` was `_issue_obj` plus additions, so every issue-only field leaked onto the pull:

```
mock-only on the pull:  assignee, closed_by, events_url, labels_url, merge_commit_sha,
                        pull_request, reactions, repository_url, state_reason, timeline_url
real-only on the pull:  _links, auto_merge, commits_url, maintainer_can_modify,
                        review_comment_url, review_comments_url, statuses_url
```

`pull_request` is the clearest of them: that marker exists to tell a caller an **issue** is really a pull, so a pull carrying one points at itself and tells a client the opposite of the truth.

Both objects now build from `_shared_obj` — the fields real serves on *both* — and each adds its own. `comments_url` is shared in value too, because a pull's conversation comments *are* the issue's and real points both at the same collection.

**The pull's key set is now an exact match for real's, in both versions** — zero mock-only, zero real-only.

## A repository's URL templates, on one rule

Real serves 42 `*_url` templates; this mock had one. The rule for what it serves now is **a template iff the resource**: 11 API templates plus the 4 git-protocol URLs, and the other 25 stay absent with a test asserting it.

A key set that lies about what can be fetched is a worse deal for a caller than a short one — and short is the one they can detect and work around. Following a link is supposed to reach something. (`git_refs_url` is absent for a subtler version of the same reason: this mock serves `/git/ref/{ref}`, singular, not real's plural `/git/refs{/sha}`.)

Two routes were added rather than advertised falsely, because a pull's `_links` names them:

- **`/pulls/{n}/commits`** — one commit, the head, which is what the pull object already claims with `commits: 1`. The sha, message and author are the pull's own; `commit.author` is a git author (name/email/date), not the GitHub user beside it, because real distinguishes them.
- **`/statuses/{sha}`** — empty, and empty is honest rather than a stub: a corpus records no CI, and `[]` is what real answers for a sha nobody reported a status on, so it is a shape a client meets in production. Repo-scoped, so a repo the caller cannot see does not answer for its shas.

## Verification

- `1025 passed, 1 skipped`, ruff clean. 6 of those are this branch's, against the 1020 on `main`
- **Key sets diffed against saved `api.github.com` bodies**, per version, asserting *no invented fields anywhere*: the pull matches real exactly in both versions; the issue and the repo have zero mock-only fields. Nothing weaker settles a shape question — the original reports were key-set diffs, and the fix has to be checked the same way
- **Values, not just keys**, on the fields both already had: every difference is accounted for by the `/github` mount prefix. A template with the wrong placeholder (`{sha}` where real says `{/sha}`) expands to a URL that 404s, so the templates are asserted by value
- **Every emitted URL still resolves.** `test_github_emitted_urls_are_fetchable` now walks the pull's new hypermedia and the repo's templates *expanded the way a client would*, since an unexpanded template would pass the check while helping nobody
- **PyGithub still drives the surface** (`test_sdk_read_coverage`): `get_pulls`, `get_reviews`, and the `pull_request` marker on the issues stream, which this change deliberately keeps on the issue
- Removing ten fields from the pull broke no other test, which is its own result: nothing in the repo depended on a pull being issue-shaped

## Notes for the reviewer

- **No stored data, no schema change, no re-import.** Every field added derives from owner/repo/number or from the row already loaded, so a deployed corpus serves this after a code deploy alone.
- **The issue object is still missing 6 fields real serves** — `type`, `sub_issues_summary`, `issue_dependencies_summary`, `pinned_comment`, `issue_field_values`, `performed_via_github_app`. Mostly null/empty defaults, so cheap, but #55 diffed only the pull and this is a different gap: *missing* fields rather than removed ones still being served. Left for its own change.
- **The github surface still answers errors as `{"detail": …}`**, not real's `{"message", "documentation_url", "status"}`. `backlot/errors/github.py` is added and owns the path, but renders only the body this router builds deliberately — the 400 above, which would otherwise have to ship in a shape real never sends. Converting every 404 on the surface is a real improvement and its own change, with its own test sweep; it is deliberately not bundled.
- **Deliberately not full key-set parity on the repository.** The alternative — all 42 templates, ~28 of them pointing at routes that do not exist — trades a short key set for dead links, and this repo already enforces the opposite with a test. Flagging it because it is the one place this PR chooses honesty over matching real's key count.


selloriwoo (approved):
Verified both against `api.github.com` at `3eefd10`, in a worktree over a fresh import — the
deployment is still on the previous build, so this is a local run.

**The version header matches real in every cell** I checked: `2026-03-10`, `2022-11-28`, and omitted,
across `pull.assignee` / `merge_commit_sha` / `issue.assignee` — plus the `…Version-Selected` echo,
which real does send and does report as `2022-11-28` when unpinned. That echo wasn't in the issue;
finding it is a better answer than the one I asked for. The 400-ahead-of-credentials ordering holds
too, with the `Authorization` header genuinely absent.

**The pull's key set is an exact match** — zero real-only, zero mock-only, at both versions.

LGTM.

PR #58, opened 2026-08-14
Key every table on the id it serves, and drop doc_id

### LINKED ISSUE #51: Retire doc_id: per-source id/served_id, ACL and FTS

khj809:
Closes #51.

Every table's primary key is now the identifier its API serves — `(repo, number)`, `(channel, ts)`, `(bucket, key)`, `key`, `id` — and the dataset's own `doc_id` is gone from the schema. It survives only as an input: it seeds a value at import and is discarded.

What follows from that:

- Values that were derived per request (a github number, a slack `ts`/`thread_ts`, a confluence `content_id`, a hubspot `record_id`, a linear `identifier`) are assigned once at import and read back, so nothing can advertise an id that fetches another row.
- The startup reverse index is gone; a lookup is a primary-key read.
- A DB that still has a `doc_id` column is refused by every opener rather than migrated — a served id is a function of the whole corpus, so it cannot be backfilled per row.
- `--append` names new documents. github, jira, confluence and hubspot records must state their id; an id an earlier import already answers at is refused, for those four and for an s3 `(bucket, key)` or a fireflies `transcript_id`. There is no update path: re-import to change a document that is already in.
- A repeated `doc_id` still names one document, and a shorter version of one drops the children it lost.
- Grants replace a row's own rather than accumulating, so a corpus can take a reader back.

`main` is merged in, with its API-version negotiation, changeset and reply-clock work intact.

**One capability is removed:** a cross-run reference is no longer resolvable, because the identifier naming it does not outlive its import. An optional import-time manifest (dataset id → served id) would restore external auditability without putting `doc_id` back in the schema; that is filed separately rather than done here.


nuri-yoo (commented):
Reviewed at head 3af6064. I ran the full suite (1,062 passed / 21 skipped), imported our 10k-document corpus, and swept the redesign with roughly 300 targeted executed checks: DDL/registry consistency for all 11 sources (PK == ID_COLUMNS == ACL leading columns == FTS columns, with matching affinities), ACL boundaries under deliberate id collisions, composite-key paths, admin/empty-set ACL semantics, cross-source id separation, erb parity, forward-reference resolution, serve round-trips for hubspot/s3/linear/fireflies, pagination disjointness, FTS rebuild idempotence, `--append` transactionality (a failed append leaves the DB byte-identical), and import-order invariance of served ids. All clean. Keying every table on the id it serves is the right resolution of #51, and explicitly provided ids are served verbatim.

The defects below concentrate in two seams: (a) the `--append` identity contract this PR itself establishes ("a clash fails loudly, a re-import is recognizable") is violated on every source outside the DEFERRED_ID four, and (b) serve-time fallout of the rekey (double-hashing, TEXT ordering, type changes) in paths the suite does not reach. Every item was reproduced by an executed script against a checkout of 3af6064; I am happy to share the scripts.

### Append identity

1. **github `subtype:"file"` rows cannot be appended at all.** The file-row branch sets `cols["number"] = None` before the DEFERRED_ID identity branch, so on `--append` into a DB that already has github rows, `_require_provided_id` fires for every file row — and providing `number` cannot help because it is discarded first (verified both variants). This contradicts the sharding intent stated in `tests/test_importer_byo.py` ("under --append the file may land in a later shard"): any sharded github corpus containing file rows is unloadable past the first shard. Since a file is addressed by `(repo, path)` by design, the append-time identity check for file rows should run on `(repo, path)` and let the deferred pass probe a number, rather than requiring one.

2. **slack `--append` silently duplicates re-imported messages (regression vs main).** A re-imported record's deterministic ts is already taken by its own previous row in `_slack_ts_taken`, so the salt walk hands it a fresh fraction and the message imports as a second row (observed `1704164645.131500` + `.886191`, same content; main keeps 1 row). The schema has no field to state a ts, so a corpus cannot opt out. This is exactly the "duplicated in silence" mode the `_require_provided_id` docstring says the guard exists to end. A schema-level `ts` (mirroring what this PR gave confluence `content_id` and hubspot `record_id`) would fix it structurally; short of that, the probe needs to recognize "taken by the same document" the way the DEFERRED_ID claim check does.

3. **fireflies: a provided `transcript_id` colliding with a previous run's row silently replaces that transcript and merges grants.** The insert is `ON CONFLICT(id) DO UPDATE` and never raises; `seed_tracker_ids` preloads cross-run claims only for DEFERRED_ID sources, and grants are `INSERT OR IGNORE`, never deleted. Observed: run 2's document steals run 1's id, run 1's content is gone, and `fireflies_acl` holds both readers — run 1's reader can now read run 2's document. The github control case raises `SystemExit` correctly. Cross-run claims need seeding for every source with a providable id, not just DEFERRED_ID.

4. **s3: an appended record restating an existing `(bucket, key)` silently replaces the object and unions grants.** Same path as 3; also `readers: []` re-statement cannot revoke (grants are append-only), so the widened grant set is unrecoverable from the corpus side.

5. **Duplicate `(source_type, doc_id)` records now import as TWO documents on probed sources** (observed on jira/github/confluence; gmail still dedupes via pure hashing). On main all sources deduped via the doc_id upsert. The retained comment near the upsert still says the row-level upsert leaves one row and cites a real corpus's jira duplicate pair — now false for keyless probed sources, where the second record gets a fresh probed id.

6. **confluence/jira `parent` attaches to the wrong row when two records share a doc_id.** `_parent_seeds` is keyed by dataset id and `resolve_parents` resolves through `self.keys[(src, did)]`, which is last-writer-wins. Observed: the record that declared the parent serves `parent_id = NULL`, and the record that declared none gets the parent. Same path applies to jira subtasks.

7. **The root upsert never deletes stale child rows, so a re-import (or in-run duplicate) leaves a root that disagrees with its own children.** Children are written `INSERT OR REPLACE` under deterministic sequence ids (`::s{j}` / `::c{j}` / `::a{j}`, gmail `::m{i}`); a version with fewer children overwrites 1..n and leaves the old tail in place. Observed cross-run: a fireflies transcript re-imported with 1 sentence serves root content `'A: one'` while `fireflies_sentences` still serves 3 rows including the removed one — breaking the importer's own invariant that content IS the sentence concatenation; a gmail thread and notion comments keep removed rows the same way. Observed in-run: two records sharing a doc_id serve merged children, falsifying the comment that the upsert result "is what a direct import of the same documents produces". Fix: delete the parent's child rows on root upsert (or refuse pure-source re-import the way probed sources do).

8. **Keyless linear issues can silently share a synthesized `identifier`.** The importer materializes `synth.linear_identifier` with no probe and no taken-set, in a 9,000-value per-team space — a collision surfaced after only 97 generated doc_ids, and both rows then serve `identifier='ENG-2686'` while `issue(id:)` resolves only one of them; the other is unreachable at the only human-facing id it advertises. The PR's own rationale probes hubspot because its 9×10^9 space still collides at scale; linear's 9×10^3 space reaches 50% collision odds around 112 keyless issues per team. The synthesized-identifier path needs the same probe shape as the others (per-team taken-set, preloaded on append).

### Serve-time

9. **notion: `_parent_field` double-hashes `parent_id`** — the router applies `synth.notion_id()` to a column that already holds the served id (the importer applies it at write time). Every page/block inside every database advertises a `parent.database_id` that 404s on GET. Fix: serve `row["parent_id"]` verbatim.

10. **confluence CQL search 500s on epoch-0 pages** — the fallback reads `r["key"]`, a jira column that `confluence_pages` does not have (IndexError). Should be the same spelling `_confluence_page` uses. The `or`-truthiness also routes a legal `0` timestamp into the fallback.

11. **github comment endpoints crash on 1970-boundary timestamps** — `synth.epoch(c["id"])` now receives an INT after the TEXT→INT rekey, and `_digest` calls `.encode()` on it (AttributeError → 500 on `/issues/{n}/comments` and `/pulls/{n}/comments` when a comment has `created_ts=0`).

12. **slack `latest_reply` is a lexicographic TEXT `MAX(ts)`** (base computed `MAX(created_ts)` as INT). Any thread whose reply seconds differ in digit count — the 1970-boundary era this PR series explicitly supports — serves the wrong reply's ts on both `conversations.history` and `conversations.replies` (observed `latest_reply: 9.989736` when the real latest is `10.617499`). Fix: select the reply ordered numerically (`ORDER BY created_ts DESC, thread_seq DESC LIMIT 1`).

13. **slack chronology is lexicographic wherever ordering goes through `_order_by("slack")`**, which orders by the TEXT `ts` id column: `conversations.history` top-level listing, `list_documents`, `children`, and the search LIKE fallback. Observed a second-10 message served before a second-9 message. The order stays total so pagination does not skip, but the chronology is wrong. A fix that types or zero-pads `ts` covers all sites at once; patching individual queries does not.

14. **gmail `threads.get` became case-sensitive while `messages.get` folds case (regression vs base).** `_gmail_resolve` now returns the id as spelled, the exact `thread_id = ?` lookup misses, and the route falls through to the case-folding document lookup — so an uppercase spelling of a 2-message thread answers **200 with a one-message thread** rather than either the full thread (base behavior) or a 404. `gmail_by_id`'s own docstring pins case-folding as the contract. Fix: `return served_id.lower()`.

15. **google epoch-0 or-truthiness: a legal `created_ts == 0` routes to the hash fallback** in `_gmail_ts`, `_drive_file`, and `_drive_facts` (which also feeds `q` time filters). A corpus-declared 1970-01-01T00:00:00Z serves as `internalDate: 1696389917000` / `createdTime: 2023-01-15…`. The truthiness predates this PR, but the PR reseeded the fallback from doc_id to the served id, silently changing the served value across the rekey. Fix: `is not None` tests, as the slack/atlassian rework already does.

16. **linear resolver fallback synthesizes a different identifier than the importer assigned** — the resolver seeds `synth.linear_identifier` on the stored UUID while the importer seeds on the dataset id (verified: `PLA-4821` vs `PLA-4442` for the same issue). Currently unreachable because the loader always materializes `identifier`, i.e. latent code contradicting the invariant — an assert like github's `_issue_number` would keep it from going live silently.

### Vendor shape of served ids

Beyond the crashes, I audited every source's served ids against the real vendor's format (56 shape checks; live import + client). Dual-spelling parity is clean across all four DEFERRED sources — a provided id and a probed id resolve at the same endpoints with the same status and JSON type, listings and search mix both, and a probed id can never land on a later provided one (verified in all four sources: the keyless victim walks to a free value). Six shape gaps:

17. **jira and confluence comment ids leak the internal `::cN` child spelling to the wire** — both routers serve the stored `c["id"]` verbatim, so clients see `"id": "PAY-7::c1"` (and `self: …/comment/PAY-7::c1`, `focusedCommentId=123456::c1`). Real Jira/Confluence comment ids are numeric strings; a client that parses or pattern-matches the id rejects it, and the internal child-id scheme leaks. This predates the PR (base leaked the dataset doc_id there — `{doc_id}::c1` — which this PR at least stopped), but notion and linear already wrap their stored child ids at serve time (`synth.notion_id(c["id"])` / `synth.linear_comment_id`), so the same one-line wrap would finish the job.

18. **hubspot v4 `toObjectId` is served as a JSON string; the real API sends a number.** The value comes straight off a TEXT column (and the paging cursor echoes it). The official python client models `to_object_id` as int, so typed SDKs reject or silently coerce. `int(...)` at the boundary, both places.

19. **notion and linear UUIDs do not pin the RFC-4122 version/variant bits** — `_uuid_from` slices a digest raw, so only ~7% of served ids pass an RFC-4122 regex and ~1.4% pass strict v4 (real Notion and Linear ids are v4). Any strict validator (zod `.uuid()`, `IsUUID(4)`) rejects ~93% of every served notion page/block/user/comment id and linear issue/team/comment/attachment id. Forcing the version nibble to `4` and the variant into `[89ab]` inside `_uuid_from` is deterministic and one place — and since this PR already renumbers every served id, this is the natural moment.

20. **gmail ids are zero-padded 16-hex** (`:016x`): a measurable fraction of ids get a leading `0`, real Gmail ids are unpadded int64-hex, and the real API resolves the unpadded spelling — the mock serves the padded one and 404s the unpadded (verified 200/404). `:x` fixes it.

21. **drive folder ids kept the old hex shape** — `"0A" + hex[:26]` (28 chars, hex body) matches neither a My-Drive folder id (file-shaped, 33-char base64url) nor a shared-drive id (`0A`+17). This PR fixed exactly this hex-tell for file ids, with a comment arguing hex "reads noticeably unlike a real id"; folders were left behind.

22. **A review comment whose `path` names no file document gets an id but is served nowhere, silently** — dropped from `/pulls/{n}/comments` and 404 at `/pulls/comments/{id}` by design (ACL non-disclosure), but unlike `changed_paths` (which gets a load-time count and a per-line `--dry-run` report) nothing reports the orphan at import. Including review-comment paths in the same unresolved-paths report would surface it.

### Identity vs schema (import-time)

23. **fireflies: the schema-declared top-level `transcript_id` is silently dropped, and only the undocumented `meta.transcript_id` works — which is then promoted to the primary key.** The top-level→extras forward list carries `number`/`key`/`content_id`/`record_id` but not `transcript_id`, so the field `fireflies.schema.json` declares never reaches the id assignment; meanwhile `meta` seeds extras and only `DEFERRED_ID.values()` are popped, so the meta spelling is promoted to the served id — exactly the "spelling no schema describes" promotion the adjacent comment says must never happen. Visible in the shipped example: `sample_corpus.jsonl`'s transcript provides `transcript_id` plus a matching `transcript_url`; at head it serves under a hashed id while the stored `transcript_url` still points at `/view/<provided id>` — a URL whose id `transcript(id:)` does not resolve. Fix: add `transcript_id` to the forward list and pop it from meta seeding like the DEFERRED_ID fields.

24. **jira top-level `changelog` is silently dropped the same way** (declared in the schema, promised by `schemas/README.md`, read only off extras, absent from the forward list — stored NULL while `meta.changelog` is stored). Pre-existing, but a systematic schema-vs-loader sweep found exactly these two fall-throughs, so they are worth fixing together.

### Import hygiene / operational

25. **notion: a `parent` naming a document that does not exist is stored as a dangling served id** — no existence check, so the page serves a parent UUID that 404s on fetch. Confluence/jira refuse the identical shape loudly ("parent … was not imported in this run") and linear NULLs it with a stderr note; notion is the one hierarchy source that says nothing. Fix: resolve notion parents through the same `resolve_parents` bookkeeping.

26. **A fresh (non-append) import unlinks the previous DB before reading a single corpus record.** A bad record then leaves the previous corpus destroyed, a schema-only empty DB, and a stale tokens.yaml beside it. The `--append` path is fully transactional (verified: a failed append leaves the DB byte-identical); the replace path has no such protection. Fix: build into a temp path and rename over on success.

27. **A legacy (pre-#51) DB slips through `connect_ro`/startup with no clear error.** `connect_rw` refuses doc_acl/doc_id DBs, but `connect_ro` and the server lifespan check nothing. Observed: the server boots `degraded`, admin s3 reads serve old rows, gmail/slack listings 500 with `no such column`, github GETs false-404 — and the one message the operator needs ("re-import the corpus") never appears. Running the same two checks in the lifespan would turn that into a clear startup refusal.

28. **`_slack_ts` is the only unbounded probe loop** — every sibling probe exits with SystemExit citing the unbounded-loop incident. The trigger (>10^6 messages in one channel-second) is extreme, but the failure mode is a silent hang.

29. **An explicitly written epoch-0 timestamp on a comment or gmail message is discarded for the synthetic fallback** (`_epoch_field(...) or (prev + 1)` / `or (created + i*3600)`): a stated `1970-01-01T00:00:00Z` parses to a falsy 0 and is replaced by the monotonic default — the exact "time nobody wrote" confusion `_epoch_field`'s docstring exists to prevent. Same `is not None` fix as 15.

### ERB importer

I also ran the full ERB bench import (~512k documents) at this head: it works — 511,961 documents, zero convert failures, 37/38 join checks with zero dangling rows (the 38th is the pre-existing slack label-grant design), FTS counts matching doc counts on all nine sources, 26/26 router spot-GETs, and served ids identical across two fresh regenerations. Two breaks:

30. **Duplicate-Linear-key resolution now disagrees between erb's convert-time relation resolution and byo/store's parent/identifier resolution.** erb's `build_linear_key_index` picks the first issue by DATASET id, and its docstring claims `resolve_cross_references` and `store.linear_issue_by_identifier` "also apply" that rule — true at base (`ORDER BY doc_id`), false at head: both now order by the served uuid. Verified on a fixture (relation 'blocks ENG-9' targets dup A while parent 'ENG-9' and `issue(id:"ENG-9")` resolve dup B) and at bench scale: 3,130 of 22,729 identifiers flip between the rules, and 26,799 of 55,544 resolvable dependency references now target a different issue than `issue(id:)` returns for the same key. Nothing dangles, so the parity suite cannot see it. Fix: erb picks first-by-`synth.linear_id(dsid)` (a pure function of the dataset id, so erb can mirror head's rule locally) and the stale comment gets corrected.

31. **The one test that exercises the real bench import (`ERB_E2E=1`) was not rekeyed and fails at head** — it GETs `/drive/v3/files/dsid_…` with the raw dataset uuid, which the rekeyed drive lookup 404s; the same document through the file's own `served_id` helper returns 200 with the expected owner, so the import is correct and only the assertion's spelling is stale. This is exactly the coverage the default-run parity suite does not include.

### Schema files

`content_id` (confluence) and `record_id` (hubspot) are vendor-faithful, validated at import, with correct affinities end-to-end. Two comments: (a) the schema JSONs were re-serialized with `ensure_ascii=True`, turning em-dashes and ellipses into `\uXXXX` escapes in files that serve as documentation — suggest `ensure_ascii=False`; (b) the refusal text promises an update path ("would be added a second time rather than updated"), but an id matching an existing row is a hard error — and following the message's advice (add the id) on a genuine re-import yields "already claimed by a previous import", with nothing saying re-imports are unsupported outright. One sentence in the cross-run branch would close the loop; the wording should stop promising updates (or an update path added).

### Documentation / comment drift (batch)

- `schemas/README.md` ("the source of truth") documents only github `number` and jira `key` as stateable ids; head also accepts confluence `content_id` and hubspot `record_id` — both mandatory on `--append` into a non-empty source — and fireflies `transcript_id` (23).
- The `SERVED_ID` → `ID_SEED` rename missed 11 comment references to the now-nonexistent symbol (synth.py, test_synth.py, test_store.py ×4, test_importer_byo.py ×5), several load-bearing (the setattr-vs-setitem reasoning).
- `tests/test_importer_byo.py` (fireflies id test) docstring contradicts its own body — it claims fireflies is absent from the registry, then `monkeypatch.setitem`s `store.ID_SEED["fireflies"]` — and cites a test name that does not exist.
- `tests/test_atlassian.py` explains a pinned contract via `_confluence_doc_id`, a function that no longer exists.
- `tests/test_store.py` justifies SOURCE_TABLE's exemption from the silent-replace guard with "upserts on `ON CONFLICT(doc_id)`" — a column this PR deletes, inside the test policing exactly that bug class.
- `tests/test_search.py` test name still says "shared fts index"; this PR retired the shared index.
- `store.py` (fireflies id comment) cites `_byo_fireflies` in `importer.byo` — it lives in erb.py and deliberately omits transcript_id, pointing away from the actual provided-id path (23).
- `tests/test_mcp.py` "a file's number is always NULL (#51)" — false at this revision (INTEGER NOT NULL, part of the PK); the filter is still needed, the stated reason will mislead.
- `tests/test_github.py` "changeset derived from a hash of the doc_id" — head derives it from the pull's served key.

### Non-blocking observations

- The `_acl_clause` wrong-pairing guard is a bare `assert` (vanishes under `-O`) while the sibling arity guard raises ValueError — worth making consistent.
- Grants are append-only across `--append` runs: a re-import can widen readers but never narrow them. Worth a line in the import docs.
- Cross-run association references are a capability removal vs main (targets in earlier append runs are now refused) — deliberate per the docstring, but worth stating in the PR body.
- Dead leftovers: `slack.py` `_search_match` unused `conn`/`roots`; `google.py` `_gmail_resolve` unused `conn`; `_drive_fill_shared` annotates a dict but receives a set; `hubspot.py` garbled merged comment; `slack.py` `client_msg_id` seeded on ts alone (unique per channel only).
- Pre-existing, carried through (optional): the LIKE search fallback does not escape wildcards even though `_like_escape` exists and is used elsewhere; `drive_folder_has_visible` does not exclude trashed files while its listing siblings do.
- Coverage note: the optional-dependency skips (mcp, llama_index, googleapiclient, slack_sdk, botocore) gate exactly the rekeyed serve surface — `test_mcp.py` was substantially rewritten by this PR yet cannot execute without `mcp`. Worth one all-extras CI run before merge.

### Suggestion (follow-up, not blocking)

Dropping doc_id removed the join key for any external per-document verification — our corpus-side ACL cross-check (which measured the #51 defect class over every caller × document pair) cannot run against the new schema, and slack/gmail/drive have no schema field to state ids, so there is no corpus-side workaround. An optional import-time manifest (`--id-map out.json`: dataset doc_id → source_type + served id tuple) would restore auditability without reintroducing doc_id into the DB. Happy to file this as a separate issue/PR after this merges.

backlot/importer/byo.py
```diff
@@ -1237,101 +1652,165 @@ def insert(
...
+                            + (f"{claimed!r}" if claimed is not None else "a previous import")
+                            + (f" in repo {scope!r}" if scope else "")
                         )
-                    self.jira_prefix_holders[prefix] = container
-                    self.jira_prefixes[container] = prefix
+                    self.tracker_ids[claim] = did
+                    if src == "jira":
+                        self._claim_jira_prefix(provided, container, where)
+                else:
+                    self._require_provided_id(src, where)
```
nuri-yoo, line 1968: github `subtype:"file"` rows cannot be appended at all: the file-row branch above sets `cols["number"] = None` before this identity branch, so on `--append` into a DB that already has github rows this `_require_provided_id` fires for every file row — and providing `number` cannot help because it is discarded first (reproduced both variants at 3af6064). This contradicts the sharding intent in `tests/test_importer_byo.py` ("under --append the file may land in a later shard"): a sharded github corpus with file rows is unloadable past the first shard. Since a file is addressed by `(repo, path)` by design, the append-time identity check for file rows could run on `(repo, path)` and let the deferred pass probe the number, rather than requiring one.

backlot/importer/byo.py
```diff
@@ -842,14 +971,184 @@ def _assign_github_comment_id(self, stored_id: str) -> int:
...
+        held = self.jira_prefixes.get(container)
+        if held is not None and held != prefix:
+            raise SystemExit(
+                f"{where}: key {provided_key!r} would name project {container!r} "
+                f"{prefix!r}, but its keys already name it {held!r}"
+            )
+        self.jira_prefix_holders[prefix] = container
+        self.jira_prefixes[container] = prefix
+
+    def _slack_ts(self, channel: str, seed: str, created_ts):
```
nuri-yoo: Re-importing a slack corpus with `--append` silently duplicates every message (regression vs main): the record's deterministic ts is already in `_slack_ts_taken` — taken by its own previous row — so the salt walk hands it a fresh fraction and the message imports as a second row (observed `1704164645.131500` + `.886191`, same content; main keeps one row). The schema has no field to state a ts, so a corpus cannot opt out. This is the "duplicated in silence" mode the `_require_provided_id` docstring says the guard exists to end. A schema-level `ts` (the symmetry this PR gave confluence `content_id` / hubspot `record_id`) would fix it structurally; short of that, the probe needs to recognize "taken by the same document" the way the DEFERRED_ID claim check does.

backlot/importer/byo.py
```diff
@@ -907,32 +1285,58 @@ def seed_tracker_ids(self) -> None:
...
+            self._slack_ts_taken.setdefault(channel, set()).add(ts)
+        # Every row carries a key, stated or derived, and for a claim the difference is immaterial:
+        # both mean the value is taken. The `dataset id` side of `tracker_ids` is unknowable for a
+        # row from an earlier run, so it is recorded as None; the check that reads it only needs to
+        # know the claim belongs to a DIFFERENT document than the one now claiming it.
+        # All four probed sources, not just github and jira: without confluence's and hubspot's
+        # here, a record stating an id an existing row already held went straight through the
+        # upsert's `DO UPDATE` and REPLACED that row -- silently losing a document at the id it
+        # was reachable by, which is the whole failure #51 exists to end.
+        for src in DEFERRED_ID:
```
nuri-yoo, line 1475: Cross-run claims are seeded only for DEFERRED_ID sources, but fireflies (provided `transcript_id`) and s3 (`(bucket, key)`) can also collide with a previous run's row — and there the `ON CONFLICT ... DO UPDATE` insert silently replaces the existing document and the `INSERT OR IGNORE` grants merge both runs' readers. Reproduced at 3af6064: run 2's transcript steals run 1's id, run 1's content is gone, and `fireflies_acl` holds both readers, so run 1's reader now reads run 2's document; `readers: []` re-statement cannot revoke. The github control case raises SystemExit correctly. This contradicts the stated contract ("a duplicate address is now an import error"). Seeding claims for every source with a providable id, not just DEFERRED_ID, would close it.

backlot/importer/byo.py
```diff
@@ -1011,10 +1415,10 @@ def add(self, rec: dict, where: str = "record") -> None:
         doc_id = _doc_id(rec)
         # Recorded, not deduplicated: `seen` answers "is this document in the corpus" for the
         # cross-reference resolution further down. Two records sharing a (source, doc_id) are both
-        # written, and the row-level `INSERT OR REPLACE` leaves the later one — which is what a
-        # direct import of the same documents produces. One real corpus has four such pairs (three
-        # across sources, one within jira); skipping the repeat instead would keep the earlier
-        # document and diverge.
+        # written, and the row-level upsert (`ON CONFLICT(<served key>) DO UPDATE`, below) leaves the
```
nuri-yoo, line 1565: This comment is now false for keyless probed sources: two records sharing a `doc_id` import as TWO documents (the second gets a fresh probed id) rather than the upsert leaving one row — reproduced on jira/github/confluence at 3af6064; gmail still dedupes via pure hashing. On main all sources deduped through the doc_id upsert. The real-corpus jira duplicate pair this comment cites now produces two issues.

backlot/importer/byo.py
```diff
@@ -1237,101 +1652,165 @@ def insert(
...
             odisp=None,
         ):
+            # A parent is named by the target's dataset id; what gets STORED is the target's
+            # served id (#51). notion resolves right here — its id is a pure hash of the seed,
+            # so it is the same answer whether the parent has been loaded yet or not. A DEFERRED
+            # source cannot: its key is only assigned once the whole corpus has been seen, so the
+            # reference is recorded and written by `resolve_parents`. linear's `parent_key` is an
+            # IDENTIFIER, not an id, and is left exactly as the corpus wrote it.
+            if par is not None and src in ("confluence", "jira"):
+                self._parent_seeds.setdefault(src, {})[did] = par
```
nuri-yoo, line 1816: When two records share a doc_id, `parent` attaches to the wrong row: `_parent_seeds` is keyed by dataset id and `resolve_parents` resolves through `self.keys[(src, did)]`, which is last-writer-wins. Reproduced at 3af6064: the record that declared the parent serves `parent_id = NULL`, and the record that declared none gets the parent. Same path applies to jira subtasks.

backlot/importer/byo.py
```diff
@@ -1237,101 +1652,165 @@ def insert(
...
+            # blanket `INSERT OR REPLACE`: two records that resolve to the same key still leave the
+            # later one (DO UPDATE), which is what a direct import of the same documents produces
+            # (see the `seen.add` comment above) — but a conflict on any OTHER unique index (notion's
+            # data_source_id) raises IntegrityError instead of SQLite's REPLACE algorithm silently
+            # deleting the row already holding that value. `names` always includes every column
+            # `_service_columns` + the additions above set for this src, so DO UPDATE overwrites
+            # every one of them, same as OR REPLACE did.
+            key_cols = store.id_columns(src)
+            update_cols = [n for n in names if n not in key_cols]
             conn.execute(
```
nuri-yoo, line 2008: This upsert replaces the root but nothing deletes the root's existing child rows — children are written `INSERT OR REPLACE` under deterministic sequence ids (`::s{j}` / `::c{j}` / `::a{j}`, gmail `::m{i}`), so a version with fewer children overwrites 1..n and leaves the old tail in place. Reproduced at 3af6064: a fireflies transcript re-imported with 1 sentence serves root content `'A: one'` while `fireflies_sentences` still serves 3 rows including the removed one (breaking the invariant above that content IS the sentence concatenation); a gmail thread and notion comments keep removed rows the same way; and two in-run records sharing a doc_id serve MERGED children, falsifying "what a direct import of the same documents produces". Deleting the parent's child rows on root upsert (or refusing pure-source re-import like probed sources) closes it.

backlot/importer/byo.py
```diff
@@ -1237,101 +1652,165 @@ def insert(
...
                 # MATERIALIZE the identifier the server would otherwise synthesize per request.
-                # Same reasoning as created_ts above: an id that is served has to be resolvable,
-                # and the app's reverse index is built from stored columns — so a serve-time-only
-                # identifier came back "Entity not found" from `issue(id: "ENG-749")` even though
-                # the API had just handed the caller that exact string. Deterministic, so the
-                # served value is unchanged; it is just written down now.
+                # An id that is served has to be resolvable, and every lookup reads a stored
+                # column — so a serve-time-only identifier came back "Entity not found" from
+                # `issue(id: "ENG-749")` even though the API had just handed the caller that exact
+                # string. Deterministic, so the served value is unchanged; it is just written down.
```
nuri-yoo: This materialization has no probe and no taken-set, in a 9,000-value per-team space (`synth.linear_identifier` = `hnum % 9000 + 1`): a collision surfaced after only 97 generated doc_ids at 3af6064, and both rows then serve `identifier='ENG-2686'` while `issue(id:)` resolves only the first — the second is unreachable at the only human-facing id it advertises. The PR's rationale probes hubspot because a 9×10^9 space still collides at scale; linear's 9×10^3 reaches 50% collision odds around 112 keyless issues per team. The same probe shape as the siblings (per-team taken-set, preloaded on append) would close it.

backlot/importer/byo.py
```diff
@@ -1237,101 +1652,165 @@ def insert(
...
+            # served id (#51). notion resolves right here — its id is a pure hash of the seed,
+            # so it is the same answer whether the parent has been loaded yet or not. A DEFERRED
+            # source cannot: its key is only assigned once the whole corpus has been seen, so the
+            # reference is recorded and written by `resolve_parents`. linear's `parent_key` is an
+            # IDENTIFIER, not an id, and is left exactly as the corpus wrote it.
+            if par is not None and src in ("confluence", "jira"):
+                self._parent_seeds.setdefault(src, {})[did] = par
+                par = None
+            elif par is not None and src == "notion":
+                par = synth.notion_id(par)
```
nuri-yoo, line 1826: Two issues meet at this line. (a) Now that `parent_id` is stored as the served id here, `_parent_field` in `routers/notion.py` (line 191) double-hashes it — `synth.notion_id(row["parent_id"])` on a value that already went through `synth.notion_id`. Every page/block inside every database advertises a `parent.database_id` that 404s on GET (reproduced at 3af6064); the router should serve `row["parent_id"]` verbatim. (b) There is no existence check: a `parent` naming a document that was never imported is stored as a dangling served id and the page serves a parent UUID that 404s — confluence/jira refuse the identical shape loudly and linear NULLs it with a stderr note; notion says nothing. Resolving notion parents through the same `resolve_parents` bookkeeping would fix both the timing and the check.

backlot/routers/atlassian.py
```diff
@@ -877,7 +887,7 @@ def _match(r) -> bool:
                 "url": page["_links"]["webui"],
                 "entityType": "content",
                 "lastModified": synth.rfc3339_millis(
-                    r["updated_ts"] or r["created_ts"] or synth.epoch(r["doc_id"])
+                    r["updated_ts"] or r["created_ts"] or synth.epoch(r["key"])
```
nuri-yoo: `r["key"]` is a jira column — `confluence_pages` has no `key`, so a page with no timestamps 500s with IndexError (reproduced at 3af6064). Should be the same spelling `_confluence_page` uses (`synth.epoch(str(r["id"]))`). The `or`-truthiness also routes a legal `0` timestamp into the fallback.

backlot/routers/github.py
```diff
@@ -1881,7 +1879,7 @@ def _hunk_around(file_row, line: int | None) -> str:
 def _gh_comment(owner: str, repo: str, number: int, c, api_base: str = "") -> dict:
     ts = c["created_ts"] or synth.epoch(c["id"])
```
nuri-yoo: `synth.epoch(c["id"])` now receives an INT after the TEXT→INT rekey, and `_digest` calls `.encode()` on it — AttributeError → 500 on `/issues/{n}/comments` when a comment has `created_ts=0` (reproduced at 3af6064). Same pattern in `_gh_review_comment` (line 1775) for `/pulls/{n}/comments`. `synth.epoch(str(c["id"]))` restores it.

backlot/store.py
```diff
@@ -1637,67 +2232,65 @@ def slack_messages_at_created_ts(conn, channel, created_ts, visible_ids=None) ->
...
 
 
-def slack_reply_authors(conn, root_doc_id, visible_ids=None) -> list[str]:
+def slack_latest_reply_ts(conn, channel, thread_ts, visible_ids=None) -> str | None:
+    """The ts of a thread's last reply — Slack's ``latest_reply``.
+
+    An exact MAX over the stored column, which costs one indexed lookup. Synthesizing it as "the
+    root's base second plus the reply count" would report a ts no message need actually have."""
+    sql = (
+        "SELECT MAX(ts) FROM slack_messages WHERE channel = ? AND thread_ts = ? AND thread_seq > 0"
```
nuri-yoo: `MAX(ts)` on the TEXT ts column is a lexicographic max (base computed `MAX(created_ts)` as INT): any thread whose reply seconds differ in digit count — the 1970-boundary era this PR series explicitly supports — serves the wrong reply's ts on both `conversations.history` and `conversations.replies`. Reproduced at 3af6064: `latest_reply: 9.989736` when the real latest reply is `10.617499`. Selecting the reply ordered numerically (`ORDER BY created_ts DESC, thread_seq DESC LIMIT 1`) restores it.

backlot/store.py
```diff
@@ -1577,16 +2172,16 @@ def list_slack_top_level(
         hi = ts_hi if ts_hi is not None else (1 << 62)
         sql += " AND created_ts >= ? AND created_ts <= ?"
         params += [lo, hi]
-    clause, cparams = _acl_clause("slack_messages", visible_ids)
-    sql += clause + " ORDER BY doc_id LIMIT ? OFFSET ?"
+    clause, cparams = _acl_clause("slack", visible_ids=visible_ids)
+    sql += clause + " ORDER BY ts LIMIT ? OFFSET ?"
```
nuri-yoo: This `ORDER BY ts` (and every ordering that goes through `_order_by("slack")` — `list_documents` at 1055, `children` at 2142, the search LIKE fallback at 2095) compares the TEXT ts lexicographically: a second-10 message serves before a second-9 message whenever digit widths differ (reproduced at 3af6064). The order stays total so pagination does not skip, but the chronology is wrong. A fix that types or zero-pads `ts` covers all sites at once; patching individual queries does not.

backlot/routers/google.py
```diff
@@ -435,49 +435,64 @@ def _gmail_query(conn, mailbox, ids, q: str) -> list:
...
+def _gmail_resolve(conn, served_id: str) -> str | None:
+    """Validate a served Gmail id's SHAPE and hand it back — a thread is keyed on the root
+    message's own id (#51), so there is nothing left to translate, only to reject.
+
+    Kept as a named step rather than inlined because the shape check must run BEFORE any lookup:
+    an unparsable id is 400 INVALID_ARGUMENT whether or not it would have resolved. No
+    ``visible_ids``: the ACL read stays in the caller (`store.gmail_thread`), so an id naming a
+    thread the caller cannot see is not-found, never a different answer."""
+    _gmail_check_shape(served_id)
+    return served_id
```
nuri-yoo: Returning the spelling as given made `threads.get` case-sensitive while `messages.get` still folds case (regression vs base, which lower()ed here): the exact `thread_id = ?` lookup misses on an uppercase spelling and the route falls through to the case-folding document lookup — answering **200 with a one-message thread** for a 2-message thread (reproduced at 3af6064). `gmail_by_id`'s own docstring pins case-folding as the contract. `return served_id.lower()` restores it.

backlot/routers/google.py
```diff
@@ -1999,30 +2026,30 @@ def _drive_mime(row) -> str:
 def _drive_file(conn, row, shared: bool | None = None, me: str | None = None) -> dict:
     """The served ``files`` resource for a stored row. ``me`` is the caller's email, which decides
     the per-caller ``ownedByMe`` (None for the admin/service token, which owns nothing)."""
-    created = row["created_ts"] or synth.epoch(row["doc_id"])
+    created = row["created_ts"] or synth.epoch(row["id"])
```
nuri-yoo: A legal `created_ts == 0` (a corpus-declared 1970-01-01T00:00:00Z) is falsy and routes to the hash fallback here, in `_gmail_ts` (654), and in `_drive_facts` (800, which also feeds `q` time filters) — serving `createdTime: 2023-01-15…` / `internalDate: 1696389917000` for an explicitly written epoch-0 (reproduced at 3af6064). The truthiness predates this PR, but the PR reseeded the fallback from doc_id to the served id, silently changing the served value across the rekey. `is not None` tests, as the slack/atlassian rework already does, restore it.

backlot/graphql/linear_resolvers.py
```diff
@@ -585,15 +585,15 @@ def _issue(row, info) -> dict:
     sort orders, bot actors and shared access are declared because `@linear/sdk`'s fragment
     selects them, and resolve empty because a document corpus has nothing behind them."""
     identifier = row["identifier"] or synth.linear_identifier(
```
nuri-yoo: This fallback seeds `synth.linear_identifier` on the stored UUID, while the importer seeds on the dataset id (`byo.py` materialization) — verified they diverge (`PLA-4821` vs `PLA-4442` for the same issue). It is currently unreachable because the loader always materializes `identifier`, i.e. latent code contradicting the invariant. An assert like github's `_issue_number` would keep it from going live silently.

backlot/routers/hubspot.py
```diff
@@ -597,14 +622,22 @@ async def list_associations(
...
     has_more = len(rows) > limit
     rows = rows[:limit]
     out: dict = {
         "results": [
             {
-                "toObjectId": synth.hubspot_record_id(r["to_doc_id"]),
+                # The target's own served id, stored on the association row itself (#51) —
+                # there is no join and no re-hash, because the link names the target by the id
+                # the API reports it under.
+                "toObjectId": r["to_id"],
```
nuri-yoo: Real HubSpot v4 associations send `toObjectId` as a JSON number (`"toObjectId": 5790939450`); this serves the TEXT column verbatim as a string, and the paging cursor at 653 echoes the same string. The official python client models `to_object_id` as int, so typed SDKs reject or silently coerce. `int(r["to_id"])` at the boundary, both places.

backlot/synth.py
```diff
@@ -241,13 +305,14 @@ def _uuid_from(seed: str) -> str:
     return f"{h[0:8]}-{h[8:12]}-{h[12:16]}-{h[16:20]}-{h[20:32]}"
```
nuri-yoo: `_uuid_from` slices the digest raw, so the served notion/linear UUIDs do not pin the RFC-4122 version/variant bits: measured over 2,000 ids, ~7% pass an RFC-4122 regex and ~1.4% pass strict v4 - real Notion and Linear ids are v4, and strict validators (zod `.uuid()`, `IsUUID(4)`) reject ~93% of every served notion/linear id. Forcing the version nibble to `4` and the variant into `[89ab]` here is deterministic and one place; it renumbers every served UUID, and since this PR already renumbers everything, this is the natural moment.

backlot/synth.py
```diff
@@ -117,17 +105,46 @@ def gmail_message_id(key: str) -> str:
...
 
     Threads share this space, as they do in real Gmail — a thread key is the root message's
-    ``doc_id``, so a single-message thread reports the same value for ``id`` and ``threadId``, which
-    is exactly what the real API does. That is also why one reverse index resolves both."""
+    ``seed``, so a single-message thread reports the same value for ``id`` and ``threadId``, which
+    is exactly what the real API does. That is also why the stored ``served_id`` column and a
+    re-hash of the thread key resolve both: a message's own id is a column read
+    (``store.gmail_by_served_id``), while its ``threadId`` re-derives this same function over
+    ``thread_id or seed`` rather than reading the root row (see ``routers.google._gmail_ids``)."""
     return f"{hnum(key, salt='msg', length=16) % GMAIL_ID_MAX:016x}"
```
nuri-yoo: `:016x` zero-pads: a measurable fraction of gmail ids get a leading `0`, real Gmail ids are unpadded int64-hex, and the real API resolves the unpadded spelling - the mock serves the padded one and 404s the unpadded (verified 200/404 at 3af6064). `:x` fixes it.

backlot/synth.py
```diff
@@ -117,17 +105,46 @@ def gmail_message_id(key: str) -> str:
...
 
-def github_number(doc_id: str) -> int:
-    return hnum(doc_id, 0, 8) % 90_000 + 1
+def gdrive_file_id(seed: str) -> str:
+    """A Drive file's served id (#51, task 12): 33 characters, base64url alphabet
+    (``[A-Za-z0-9_-]``), starting with ``1`` — the shape of a modern real Drive file id.
+    Reversible via the stored ``gdrive_files.served_id`` column (see
+    ``store.gdrive_by_served_id``), not hashed at serve time.
+
+    Derived from the RAW DIGEST BYTES, unlike ``drive_folder_id``'s hex slice: hex spans only 16
```
nuri-yoo, line 144: This argument (hex "reads noticeably unlike a real id") applies equally to `drive_folder_id` above, which kept `"0A" + hex[:26]`: 28 chars with a hex body matches neither a My-Drive folder id (file-shaped, 33-char base64url) nor a shared-drive id (`0A` + 17 base64url). Worth extending the same raw-digest-bytes derivation to folders.

backlot/importer/byo.py
```diff
@@ -1078,8 +1482,9 @@ def register(email: str | None, name: str | None = None) -> None:
         # `meta: {"number": 3}` claim issue 3 in a repository just as a top-level `number`
         # would — a spelling no schema describes, that shadows a real issue, and that the
         # uniqueness check below would then refuse an import over. Both ids are ordinary
-        # `meta` content again: carried through, never promoted to the served column.
-        for reserved in ("number", "key"):
+        # `meta` content again: carried through, never promoted to the served column. Read from
+        # DEFERRED_ID so a source that gains a stateable id is covered by declaring it there.
+        for reserved in DEFERRED_ID.values():
```
nuri-yoo: fireflies breaks the rule this comment states: the top-level→extras forward list below lacks `transcript_id`, so the field `fireflies.schema.json` declares never reaches the id assignment — while `meta.transcript_id` seeds extras, is NOT popped here (only DEFERRED_ID values are), and is promoted to the PRIMARY KEY. Reproduced at 3af6064: top-level `transcript_id` → ignored, hashed id served; `meta.transcript_id` → served verbatim. The shipped `sample_corpus.jsonl` shows the consequence: its transcript provides `transcript_id` plus a matching `transcript_url`, and at head it serves under a hashed id while `transcript_url` still points at `/view/<provided id>` — a URL whose id `transcript(id:)` does not resolve. Adding `transcript_id` to the forward list and popping it from meta like the DEFERRED_ID fields closes it. (jira top-level `changelog` falls through the same gap — stored NULL while `meta.changelog` is stored; a systematic schema-vs-loader sweep found exactly these two.)

backlot/store.py
```diff
@@ -435,30 +817,45 @@ def connect_rw(path: Path, *, busy_ms: int = 60_000) -> sqlite3.Connection:
...
+    # A DB built before this branch has one shared `doc_acl` table, keyed corpus-wide by `doc_id`
+    # rather than a table per source. An append onto it would write the new source's grants into
+    # the per-source tables SCHEMA creates below while every pre-existing grant stays behind in
+    # `doc_acl`, which nothing reads -- every document from before the append silently
+    # becomes invisible to every scoped token. There is deliberately no backfill here: `doc_acl`
+    # has no source column, so it cannot say which source a colliding `doc_id`'s grant belonged
+    # to, and copying its rows into the per-source tables blind would silently re-create exactly
+    # the cross-source union this branch was written to remove (see the `ACL_TABLE` comment
+    # above). The only correct move is a fresh re-import.
+    if conn.execute(
```
nuri-yoo, line 854: `connect_rw` refuses a legacy doc_acl/doc_id DB here, but `connect_ro` and the server lifespan check nothing, so a pre-#51 DB boots the server `degraded`: admin s3 reads serve old rows, gmail/slack listings 500 with `no such column`, github GETs false-404 (reproduced at 3af6064) — and the one message the operator needs ("re-import the corpus") never appears. Running these same two checks in the lifespan would turn that into a clear startup refusal.

backlot/importer/byo.py
```diff
@@ -842,14 +971,184 @@ def _assign_github_comment_id(self, stored_id: str) -> int:
...
+        """
+        base = int(created_ts) if created_ts is not None else synth.epoch(seed)
+        taken = self._slack_ts_taken.setdefault(channel, set())
+        # Keyed on the message's OWN seed, where the serve-time version keyed a reply on its
+        # thread root — that shared key is precisely what made two replies in one second
+        # indistinguishable. A reply still sorts after its root because its `created_ts` is the
+        # root's plus its position, which is where thread order actually comes from.
+        candidate = synth.slack_fmt_ts(base, seed)
+        salt = 0
+        while candidate in taken:
```
nuri-yoo: This is the only probe loop left unbounded — every sibling probe exits with SystemExit citing the unbounded-loop incident. The trigger (>10^6 messages in one channel-second) is extreme, but the failure mode is a silent hang (reproduced with an exhausted fraction space at 3af6064). A bounded walk with the same SystemExit would match the siblings.

backlot/importer/byo.py
```diff
@@ -1563,32 +2119,32 @@ def resolve_cross_references(self) -> None:
...
-        # every issue is loaded. `Issue.children` reads `parent_doc_id`, so without this a corpus would
+        # every issue is loaded. `Issue.children` reads `parent_id`, so without this a corpus would
         # serve `parent` but an empty `children`, and the two directions would disagree.
         if counts.get("linear"):
-            key_to_doc: dict[str, str] = {}
-            for did, ident in conn.execute(
-                "SELECT doc_id, identifier FROM linear_issues WHERE identifier IS NOT NULL "
-                "ORDER BY doc_id"
+            key_to_id: dict[str, str] = {}
+            for issue_id, ident in conn.execute(
```
nuri-yoo, line 2396: Ordering duplicate-identifier candidates by the served uuid here (and in `store.linear_issue_by_identifier`) diverged from erb's `build_linear_key_index`, which still picks first-by-DATASET-id and whose docstring claims this code "also appl[ies]" that rule (true at base, `ORDER BY doc_id`). Verified on a fixture at 3af6064: a relation naming `ENG-9` targets dup A while `parent: ENG-9` and `issue(id:"ENG-9")` resolve dup B; on the full ERB bench, 26,799 of 55,544 resolvable dependency references now target a different issue than `issue(id:)` returns for the same key (3,130 of 22,729 identifiers flip). Nothing dangles, so the parity suite cannot see it. Since the served uuid is a pure function of the dataset id, erb can mirror this rule locally (`first by synth.linear_id(dsid)`) — and the stale comment at erb.py:1428-1439 should be corrected either way.

khj809:
Thanks — this was an unusually useful review. I reproduced all 31 numbered items against 3af6064 with executed scripts before touching anything, and every one held. All 31 are fixed, in five commits (d6db7f5, d12a2fb, 0cd8c2c, 4c48134, f13e777), plus the comment/doc batch and the non-blocking list.

Two places I resolved differently from the suggestion, and one thing your coverage note caught:

**2 (slack `--append` duplicates).** I did not add a schema-level `ts`. A corpus would then have to state one per reply as well as per message, and slack would need github's whole deferred-claim pass to arbitrate a provided ts against a probed one — and `--append` would start requiring a ts for every new message, which is a capability loss. Instead the probe recognises a re-imported message by what identifies it in the absence of a stated id: same channel, same second, same author, same text. The check runs only against ts values that were in the DB *before* the run, so two such records in one corpus stay two documents. If you'd still rather have the field, say so and I'll add it — the machinery for a fifth stated id is now in place (`_states_own_id`).

**3/4 (fireflies and s3 cross-run claims).** Seeded as you asked, but through a per-source "does this record state its own id" predicate rather than by widening `DEFERRED_ID` — a synthesized id must keep landing back on its own row (a re-imported gmail/notion/drive/linear record does), so the claim has to key on the record stating the value, not on the source having a column for one. Two consequences worth flagging: `meta.transcript_id` is now ordinary meta content (only the schema's own `transcript_id` claims, matching how github's `number` already worked — this is item 23), and grants now REPLACE a row's own rather than accumulating, which also closes your non-blocking note that a re-import could widen a reader set and never narrow it.

**Your coverage note was right, and it found a regression of mine.** `fastmcp` 2.x imports `mcp.shared.exceptions.McpError`, which `mcp` 2.0 renamed — so an unbounded pair resolves to two packages that cannot import each other and all 11 MCP tests fail inside fastmcp's own module body. Pinned `fastmcp>=3`. With every extra installed: **1,145 passed, 2 skipped**.

Notes on the rest:

- **7** is trimmed rather than delete-and-rewrite, so a child sitting under a still-provisional parent key is left for `_settle` to move. gmail/slack thread members are full documents, so the dropped ones take their ACL rows and FTS entries with them.
- **13** is fixed once, in a registry (`store.ORDER_COLUMNS`), so all four sites that go through `_order_by("slack")` plus `list_slack_top_level` are covered — not per query.
- **11 and 29 are the same bug seen twice**: 11 was unreachable through a corpus until 29 was fixed, because `_epoch_field(...) or default` discarded the stated 0 before it could reach the serializer. Fixed together, and the `or`-truthiness swept across gmail, drive, confluence, jira and the comment/message clocks.
- **16 and 30** are both "two places derive the same id and disagree". 16 asserts now, like github's `_issue_number`. For 30, erb orders by `synth.linear_id(dsid)` — a pure function of the dataset id, so the convert pass applies the server's tie-break without a database — and the stale docstring is corrected.
- **19** turned out to be all-or-nothing: forcing the version nibble and variant bits takes strict-v4 conformance from 27/2000 to 2000/2000, and the id stays a pure function of its seed.
- **26** moves the old DB aside and moves it back on any exception, so replacing a corpus costs a rename rather than a copy of the whole file.
- **The one claim I could not confirm** is the garbled merged comment in `hubspot.py` — I read the file and found nothing malformed. If you still have the line number I'll fix it.

Your follow-up suggestion (`--id-map out.json`) is a good idea and I agree it belongs in its own PR — please do file it. It is the one thing this PR genuinely takes away, and a manifest restores it without putting `doc_id` back in the schema.

nuri-yoo:
I re-ran the full repro archive against a586ba2: all 31 items behave as your comment describes, nothing to re-open. Highlights: the stated-id refusals read exactly right ("this row states ('bucket', 'key') = ..., which a document from an EARLIER import already answers"); a failed fresh import leaves the previous corpus intact; slack re-append keeps one row while two identical records in one corpus stay two documents; the linear collision pair comes out as ENG-2686/ENG-2687; erb's ENG-9 fixture resolves the same issue on all three paths; UUIDs pass strict v4 at 2000/2000; gmail message ids are unpadded while the opaque tokens stay padded; jira comment ids reach the wire as numeric strings; connect_ro refuses a legacy DB with a clear message. On 7, I also confirmed a dropped gmail thread member takes its FTS entry and ACL rows with it.

The hubspot comment survives at a586ba2: `backlot/routers/hubspot.py:174-176`, right above `"id": row["id"]`. Three fragments are merged ("The stored column ..., not a re-hash of / HubSpot's id space is probed on a collision, so the row's own stored value / with the value this very row was actually assigned and is looked up by"). It probably wants to say: the stored column, never a re-hash, because the id space is probed on collision and re-deriving here could disagree with the value this row was assigned and is looked up by.

nuri-yoo:
On the slack `ts` question: I would not add the field now. Your fingerprint fix already covers re-import identity, and our corpus neither states a ts nor cites one anywhere in its prose, so a stated value would have nothing to agree with.

The one future case I can see is prose citing Slack permalinks (`/archives/<C..>/p<ts>`) so an agent can follow the link across sources. But a permalink needs the channel id as well as the ts, and channel ids are synthesized from the channel name, so a stated `ts` alone does not make the URL citable. Since every served id is a pure function of the dataset id and independent of content, that case is better served by a two-pass flow: import once, read the served ids back, inject the permalinks into the citing documents, and import the final corpus. That needs no schema change, only that the id manifest I am going to propose also emits container ids (channels, folders, teams) alongside document ids.

A stated `ts` still makes sense for a hand-authored corpus whose author writes permalinks directly. Since `_states_own_id` makes a fifth stated id cheap, I would defer it until such a corpus actually shows up.

I'll prepare the `--id-map` manifest PR on a branch off this one: import-time emission of dataset id to (source_type, served id), plus container ids per the above.

nuri-yoo (approved):
LGTM

PR #60, opened 2026-08-18
linear: a provided identifier's prefix is a fact about its team

### LINKED ISSUE #52: Linear: a provided identifier resolves, but its prefix teaches the team nothing — one team serves two spellings and two teams can share one

nuri-yoo:
Supersedes #54, which built this on the startup index #58 has since removed. Closing that one rather than force-pushing it: the mechanism here is different enough that the diff would not be a rebase of it.

## What is wrong on `main`

Two things #46 built for jira never reached Linear. A provided identifier resolves, but it teaches the container nothing. Loading a corpus that writes one `ENG-7` into a team whose name derives `PP`:

| | `linear_issues.identifier` | `linear_teams.served_key` | `team(id: "ENG")` |
|---|---|---|---|
| `main` (03f3f59) | `ENG-7`, `PP-711` | `PP` | `None` |
| this branch | `ENG-7`, `ENG-711` | `ENG` | `payments-platform` |

One team serves two prefixes at once, the issue object serves `identifier: "ENG-7"` beside `team { key: "PP" }`, and the team the corpus called ENG is unreachable at its own key. Real Linear derives an identifier FROM its team's key, so those three are one defect. A second team providing ENG-prefixed identifiers also loads without a word, and a mistyped `identifier: "7"` renames the team to `7` for every one of its issues.

#51 closed with #58, so there is no later identity work left to defer this to. #58 added the `linear_teams.served_key` column this needs and writes the name-derived key into it.

## What this does

The 1:1 both-directions refusal jira got in `0eff060` now covers linear teams, on the same claim machinery: `_claim_jira_prefix` is generalized into `_claim_prefix`, which takes the nouns a refusal is worded with per source. A team's settled key lands in `linear_teams.served_key`, and every serving surface reads it through the new `store.linear_team_keys`: the team object, the teams filter, the compiled issue filter, and `team(id:)`. The issue filter reads it down the `and`/`or` recursion too, which was compiling `team.key` against the derivation while the same filter at top level compiled against the corpus's, so one query written two ways disagreed.

`team(id:)` also settles its tie the other way round now. A key a team's own issues spell out beats one another team merely derives from its name, told apart by re-deriving, so the team the corpus called ENG is not left unreachable by a neighbour whose name happens to shorten that way. Two containers deriving one key stays as legal as it has always been.

The column carries the cross-run story. A team's key is settled by its first import and seeded back from `served_key` on every later one, so an `--append` that would rename the team is refused instead of silently leaving the stored rows under a prefix that is no longer the team's key. That is jira's rule for a project, and it is here for the same reason: a stored identifier cannot say whether its prefix was the corpus's or this mock's, since both spellings share the one column.

Within one load, a container is settled at the END. A keyless issue is stamped the moment its record lands, when the team's key is only what the corpus has spelled out so far, so `restamp_linear_identifiers` reassigns the container's materialized rows in dataset-id order once every record has been read. What a keyless row ends up with is then a function of the row set rather than of which side of the provided line it sat on. The free-number probe is bucketed by PREFIX rather than by team, because a stated prefix can be the key another team derives from its name and a re-stamp landing on that team's row would leave one document answering at an id that fetches its neighbour. `_linear_identifiers` is bucketed the same way for the same reason: keyed by the derived key, it left the probe walking a set that did not contain the team's own identifiers.

The same pass settles the other way a stamp can go wrong. A keyless row can take a spelling a later record states outright, and the probe that placed it had no way to know. The stamped row gives it up, because the corpus's own issue is the one whose documents cite that id.

`identifier` carries a pattern now, for the reason jira's `key` carries one. The prefix is a fact about the whole team, so a single mistyped identifier would otherwise rename it for every one of its issues.

## Dropped from #54

Two of its pieces are gone rather than carried over, both because #58 moved the ground under them.

Releasing a superseded id claim (a repeated `doc_id` re-stating a different id) is no longer reachable. For a source whose id is deferred, two records sharing a dataset id take different keys and BOTH rows survive, so the first row does still hold the id its claim names, and a third record providing it is refused by the served-id collision check with an accurate message.

A later shard can no longer take a spelling an earlier run materialized. Telling a materialized identifier from a provided one in the DB is exactly what dropping `doc_id` removed, so the shard that would rename the team is refused instead.

## Tests

`tests/test_linear.py` gets the four serving-side tests from #54 unchanged. The nine importer tests are rewritten against the post-#58 model, reading rows back through `served_id("linear", did)` since there is no `doc_id` column to join on. Two of #54's are dropped with the behaviour they pinned, and `test_byo_a_teams_key_is_settled_by_its_first_import` is new.

Applied to `main` with no source changes, 14 of the 15 fail. The 2 that pass do so vacuously: with no re-stamp pass, nothing moves, so nothing can land on another team's identifier.

```
$ uv run pytest tests
1103 passed, 24 skipped in 16.24s

116 files already formatted

$ uv run ruff check .
All checks passed!

$ cd examples/using-official-sdk/linear && npx tsc --noEmit && npx tsx index.ts
(typechecks; serves ENG-49121 and DES-128743 through @linear/sdk)
```


backlot/importer/byo.py
```diff
@@ -1158,25 +1158,47 @@ def _require_provided_id(self, src: str, where: str) -> None:
...
+
+        Jira and linear keep separate maps: they are separate namespaces, and a refusal names the
+        one it came from."""
+        kind, label, plural = self._PREFIX_NOUNS[src]
+        prefixes, holders = (
+            (self.jira_prefixes, self.jira_prefix_holders)
+            if src == "jira"
+            else (self.linear_prefixes, self.linear_prefix_holders)
+        )
+        prefix = str(provided).rsplit("-", 1)[0]
```
khj809: This parse blocks the import of the real ERB corpus outright.

43 of the 35,308 documents under `data/raw/generated_data/sources/linear` carry a slug-shaped `key` — `ENG-453210-kms-hsm-deployment-lifecycle-telemetry-orbiter`, `DES-221341-microcharge-composer-csv-first-export`, `PM-27145-commit-phasing-and-usage-velocity-credits-architecture`, and 40 more. `_byo_linear` (erb.py:1769) passes the value through verbatim as `identifier`, and `rsplit("-", 1)[0]` here reads everything before the LAST hyphen as the prefix.

I dropped one of those keys, verbatim, into the RT_DOCS fixture and ran `erb.import_structured`: main loads it, this branch aborts.

```
line linear/dsid_lin_3: identifier 'ENG-453210-kms-hsm-deployment-lifecycle-telemetry-orbiter'
would name team 'engineering' 'ENG-453210-kms-hsm-deployment-lifecycle-telemetry',
but its identifiers already name it 'ENG'
```

The jira raw data has 4 keys of the same shape, but `_byo_jira` never passes `key` through, so `_claim_jira_prefix` has never seen one — there is no precedent here, this breakage is new. And the ERB path loads with `validate=False`, so the new `pattern` cannot catch it either; what fires instead is the 1:1 refusal, with a message that blames the team rather than the record.

Reading a linear prefix as `split("-", 1)[0]` (identical behaviour on a well-formed identifier) makes all 43 harmless. Otherwise, normalising the bench key in `_byo_linear` with the `_LINEAR_KEY` regex already sitting at erb.py:1037 would do it.

backlot/schemas/linear.schema.json
```diff
@@ -35,7 +35,8 @@
     },
     "identifier": {
       "type": "string",
-      "description": "Human identifier, e.g. \"ENG-123\". Default: synthesized as <TEAMKEY>-<n> from the doc_id. Not required to be unique — the bench's own keys repeat — and `issue(id:)` resolves a repeated identifier to the first matching issue."
+      "pattern": "^[A-Z][A-Z0-9]{0,4}-[1-9][0-9]*$",
```
khj809: This pattern is narrower than the mock's own derivation. It caps a prefix at 5 characters and requires a leading letter, but `synth.linear_team_key` has no such bound:

```
platform-infra-reliability-and-cost-ops → key='PIRACO' → 'PIRACO-8079'  (fails the pattern)
3d-printing                             → key='3P'     → '3P-8079'      (fails the pattern)
```

For those teams a keyless corpus materialises and serves exactly those strings, yet a corpus that states the identifier its own documents cite is refused — the mock will not accept as input what it hands out.

Worth settling one way or the other: cap `linear_team_key` at 5 characters (which matches real Linear's own limit on a team key), or widen the pattern to everything the derivation can produce.

khj809 (commented):
Reviewed in two separate worktrees (PR HEAD `f0d0963`, main `03f3f59`).

**Repro and fix both check out.** On main the table in the description reproduces exactly — `identifier` splits into `ENG-7`/`PP-711`, `served_key` stays `PP`, `team(id: "ENG")` is `null`, and a `team.key` filter returns nothing whether it is written flat or under `and: […]`. On this branch all four agree on `ENG`. Applied to main, 14 of the new tests fail; on the branch they all pass. Full suite is green (the one failure, `test_mirage_patchers_name_every_host_constant_mirage_ships`, fails on main too and is unrelated), and `ruff format --check` / `ruff check` pass.

The supporting claims hold up as well: `_derived_in` expands over the distinct team names, so two teams sharing one key both match a filter; `url` and `branchName` are derived from `identifier` at serve time; and FTS indexes only title/content, so a re-stamp cannot invalidate the index. The `_team_keys` cache lives on a context dict built per request, so it cannot go stale, and `served_key` is `NOT NULL`, so the `str(None)` path in the seed is unreachable. Thanks to the free-then-reassign order and the dataset-id sort in the re-stamp, loading the same corpus twice produces identical identifiers.

Two inline comments below. The first one blocks the import of the real ERB corpus.

One small thing: the arithmetic in "14 of the 15 fail … The 2 that pass" does not add up. 16 items are collected (4 in `test_linear.py`, 12 in `test_importer_byo.py`, including the 5 `mistyped` parameters), and on main that is 14 failing, 2 passing.

backlot/importer/byo.py
nuri-yoo: Confirmed and fixed in f18481b, with the split moved to the first hyphen as you suggested.

Reproduced your abort verbatim through `erb.import_structured` over the RT_DOCS tree with one slugged document, on f0d0963:

```
ABORTED: line linear/dsid_lin_3: identifier 'ENG-453210-kms-hsm-deployment-lifecycle-telemetry-orbiter'
would name team 'engineering' 'ENG-453210-kms-hsm-deployment-lifecycle-telemetry',
but its identifiers already name it 'ENG'
```

The abort needs a canonical sibling to collide with. Without one it is quieter and worse: the slug alone renamed the team and took its keyless issues with it.

```
# f0d0963, one slug + one keyless issue, validate=False
identifiers : ['ENG-453210-kms-hsm-deployment-lifecycle-telemetry-orbiter',
               'ENG-453210-kms-hsm-deployment-lifecycle-telemetry-8843']
served_key  : [('engineering', 'ENG-453210-kms-hsm-deployment-lifecycle-telemetry')]
```

I took the `split("-", 1)[0]` option rather than normalising in `_byo_linear`. Normalising changes the identifier 43 documents are served at, and those 43 are the same before and after this PR on main, so that belongs to whatever decides what the bench's `key` field means. Worth noting for that decision: `build_linear_key_index` (erb.py:1443) indexes the same slug verbatim, so a `dependencies` entry naming `ENG-453210` already resolves to nothing for those issues. I have not touched it here.

All three sites that read a linear prefix go through one `_linear_prefix` now. Keeping them in sync is not cosmetic: the claim, the cross-run seed off `served_key` and the probe's taken-set have to agree about where a prefix ends, or a provided slug lands in a bucket the probe never consults and a keyless sibling is handed a spelling the corpus already wrote.

Jira keeps its last-hyphen split, on your point that `_byo_jira` never passes the bench `key` through. Verified: both `raw.get("key")` reads in erb.py (1443, 1769) are linear's, and the raw jira data does carry 4 slug-shaped keys that never reach the importer.

Scanned the whole raw corpus under the new rule, since one document loading is not the same as the corpus loading:

```
design              derived=DES   prefixes={'DES': 3437}
engineering         derived=ENG   prefixes={'ENG': 23168}
product-management  derived=PM    prefixes={'PM': 8703}

teams with more than one prefix (refused):     none
prefixes held by more than one team (refused): none
teams whose prefix != derived key (renamed):   none
```

`test_byo_a_slug_shaped_key_claims_only_its_leading_prefix` pins it at `validate=False`, since that is the path the ERB import takes and the pattern cannot stand in for it.

backlot/schemas/linear.schema.json
nuri-yoo: Confirmed, and fixed in f18481b by widening the pattern. Your two examples, run against it:

```
platform-infra-reliability-and-cost-ops    key=PIRACO   ident=PIRACO-928     REJECTED
3d-printing                                key=3P       ident=3P-928         REJECTED
```

I went with widening rather than capping `linear_team_key`, for two reasons.

Capping does not actually cover the second case. `3P` is 2 characters and still fails `^[A-Z]…` on the leading digit, so the character class has to widen either way, at which point the length bound is the only thing the cap buys.

And the cap is not free. `_key` returns one initial per word with no bound, so capping renumbers every issue of any team whose name runs past five words. The bench's own teams are all short (DES, ENG, PM, BO, MC) so it would not show up there, but it is an id-stability change for corpora already imported, and this PR is not the place to decide that quietly.

The pattern is now `^[A-Z0-9]+-[1-9][0-9]*$`, which is exactly what the derivation can produce: `_key` guarantees at least one character, has no maximum, and can start with a digit. What it still refuses is the shape a typo actually makes, which is what the pattern was added for. `TOOLONGPREFIX-7` came out of the parametrized test since it is now legal by design, replaced by `ENG_7` and `ENG-7x`.

`test_byo_a_derived_identifier_is_accepted_as_input` states the invariant as a round trip over both of your team names: derive the identifier, feed it back as input, load with validation on. That is the property that was broken, so it seemed better to pin it than to pin the pattern string.

If capping `linear_team_key` at 5 is worth doing on its own terms, real Linear's limit is a decent argument for it and I am happy to open an issue. It changes served identifiers, so it wants its own PR.

backlot/schemas/linear.schema.json
```diff
@@ -35,7 +35,8 @@
     },
     "identifier": {
       "type": "string",
-      "description": "Human identifier, e.g. \"ENG-123\". Default: synthesized as <TEAMKEY>-<n> from the doc_id. Not required to be unique — the bench's own keys repeat — and `issue(id:)` resolves a repeated identifier to the first matching issue."
+      "pattern": "^[A-Z0-9]+-[1-9][0-9]*$",
```
khj809: `import_structured` is fixed, but the artifact `backlot export` writes is read back with validation ON, so the same 43 documents are refused here instead.

These two lines are `export_byo`'s own output, trimmed to the required fields (the second is the bench's `dsid_lin_3` verbatim):

```jsonl
{"source_type":"linear","doc_id":"dsid_lin_2","team":"engineering","title":"Caching epic","content":"Umbrella.","author_email":"maya.chen@redwoodinference.com","identifier":"ENG-8","created":1767225600}
{"source_type":"linear","doc_id":"dsid_lin_3","team":"engineering","title":"KMS/HSM deployment lifecycle telemetry","content":"x","author_email":"ava.chen@redwoodinference.com","identifier":"ENG-453210-kms-hsm-deployment-lifecycle-telemetry-orbiter","created":1767312000}
```

main:

```
LOADED identifiers=['ENG-453210-kms-hsm-deployment-lifecycle-telemetry-orbiter', 'ENG-8']
       teams=[('engineering', 'ENG')]
```

this branch:

```
REFUSED: line 2: dsid_lin_3 [identifier]:
'ENG-453210-kms-hsm-deployment-lifecycle-telemetry-orbiter'
does not match '^[A-Z0-9]+-[1-9][0-9]*$'
```

So `backlot export` still produces an artifact `backlot import` cannot read, and the equivalence `test_erb_to_byo_round_trip_builds_an_equivalent_database` asserts breaks on this data: the direct path loads, the exported-and-reloaded path aborts. (Reproduced end to end too — the same RT_DOCS tree through `_import_via_byo` passes on main and raises here.)

The two paths need to agree on what a bench key becomes. Normalising in `_byo_linear` — cutting the slug down to `ENG-453210` where `_linear_prefix` already cuts the claim — would make both sides carry a clean identifier, at the cost of changing the served identifier for those 43 documents relative to main. Loosening the pattern to admit a trailing slug is the other option, but that gives back the typo guard the pattern exists for.

khj809 (commented):
Re-checked `f18481b` in the worktree. Both findings are fixed and verified: the slug keys now claim `ENG` and `erb.import_structured` over the RT_DOCS tree loads where it aborted before, and the widened pattern accepts `PIRACO-8079` / `3P-8079` / `9LI-8079` while still refusing `eng-7`, `ENG 7`, `7`, `ENG-0007`, `ENG_7`, `ENG-7x`. I also confirmed the commit message's corpus-wide claim over the whole raw tree — design→DES (3,437), engineering→ENG (23,168), product-management→PM (8,703), every split-first prefix equal to its derived key, so no team is renamed and nothing is refused. Suite and lint green.

One inline comment: the same 43 documents still block the other half of the round trip.

backlot/schemas/linear.schema.json
nuri-yoo: Confirmed and fixed in 78fa873. Reproduced your refusal through `_import_via_byo` first:

```
REFUSED: line 16: dsid_lin_3 [identifier]:
'ENG-453210-kms-hsm-deployment-lifecycle-telemetry-orbiter'
does not match '^[A-Z0-9]+-[1-9][0-9]*$'
```

I measured the normalising option before choosing, and it rules itself out. Cutting the slug to `ENG-453210` lands **25 of the 43 on an identifier another document already holds**:

```
as-is      : 17634 repeated identifiers across 5055 groups
normalized : 17664 repeated identifiers across 5060 groups
delta      : +30 documents pushed onto an identifier another document already holds
```

(The 5,055 baseline is the same number the repo already documents on `linear_issues.parent_key`, so the scan agrees with what is on record.) `linear_issue_by_identifier` answers the first row by served id, so each of those 30 leaves one document unreachable at the only human-facing id it advertises. That is the failure this branch exists to remove, reached by a different route, and it is a worse trade than a slug sitting in the identifier column.

So I took the third option rather than either of the two. The pattern now checks only the half it has standing to check:

```
^[A-Z0-9]+-\S+$
```

The prefix is what is a fact about the whole team — it becomes the team's key, which is the entire reason a mistyped identifier is dangerous. Everything after the first hyphen cannot rename anything, and an identifier is an opaque lookup string here (repeats and all), so validating that half only refused corpora for no gain. `eng-7`, `ENG 7`, `7`, `ENG_7`, `ENG-` and `ENG- 7` are all still refused; `ENG-0007` and `ENG-7x` are not, and the parametrized test says so on purpose. A well-shaped prefix that is simply the wrong one (`ENGG-7` for a team the corpus calls ENG) was never catchable by a pattern — the 1:1 refusal at load is what names that, and the schema description now says so instead of implying more.

`test_byo_a_slug_shaped_key_claims_only_its_leading_prefix` loads the slug corpus both ways and asserts the two agree. The validated half is your round trip in miniature; the unvalidated half is `import_structured`, where the pattern never fires and `_linear_prefix` is the only thing between a slug and a renamed team.

Two things I found while checking whether this shape exists elsewhere, both on main and neither touched here:

**jira's `key` pattern has the identical defect, and it is a harder dead end.** `^[A-Z][A-Z0-9]{1,9}-[1-9][0-9]*$` caps at 10 characters and requires a leading letter, but `jira_project_key` is `_key` plus 6 hex digits, so it runs past 10 as soon as a project name has five words, and inherits a leading digit from a name like `3d-printing`. And unlike linear, `_require_provided_id` *demands* a key when appending to a project that already has issues, so the corpus cannot state the key the mock just served it:

```
$ project = "platform-infra-reliability-and-cost-ops"
project key served : PIRACO73723D
issue key served   : PIRACO73723D-726
append stating it  : REFUSED: line 1: j-2 [key]:
                     'PIRACO73723D-2' does not match '^[A-Z][A-Z0-9]{1,9}-[1-9][0-9]*$'
```

Identical output on main (03f3f59), so it predates this branch. That project's corpus is un-appendable: the append path requires a key and validation refuses the only correct one. Happy to open an issue, or fix it here if you would rather have it in one place — the fix is the same shape as this one.

**The bench's slug keys already break relation resolution, independently.** `build_linear_key_index` (erb.py:1443) indexes the raw `key` verbatim, so a `dependencies` entry naming `ENG-453210` resolves to nothing for those 43 issues. That is the thing normalising would genuinely fix, and it wants to be decided together with what the served identifier should be — separate from this PR either way.

`1107 passed, 24 skipped`; ruff format and check clean. I also have a full-corpus sweep running (all 512k documents through `export_byo`, then every written record through `record_errors`) to see whether any other field refuses what the export writes. Will report what it finds.

nuri-yoo: Sweep is done, and it clears the question I could not answer by reading: is `identifier` the only field where the export writes something the import refuses?

Ran the real `export_byo` over the whole bench tree, then every record it wrote through `record_errors` — the same call `byo.load` makes.

```
exported 511,961 documents -> 581,294 records in 236s

pattern ^[A-Z0-9]+-[1-9][0-9]*$   (as reviewed)
  validated 581294 records
  distinct validation failures: 43
  all 43: [identifier] does not match — nothing else, no other source, no other rule

pattern ^[A-Z0-9]+-\S+$           (78fa873)
  validated 581294 records
  distinct validation failures: 0 | total failing records: 0
```

So the surface you found was the only one of its kind in the artifact, and it is closed. Worth noting the sweep also covers the other two patterns in the repo by construction: `hubspot.record_id` (`^[0-9]+$`) and `jira.key` never fire on this data — jira's because `_byo_jira` drops the bench `key`, which is also why its own defect stays invisible until someone appends.

I also loaded the real linear shard (all 35,308 documents, through `byo.load` with the roster) on both revisions and compared row by row:

```
main    (03f3f59): 35308 issues, (id, identifier) sha256 = 4ce7bbc87ac25a8d
                   teams = [('design','DES'), ('engineering','ENG'), ('product-management','PM')]
this branch       : 35308 issues, (id, identifier) sha256 = 4ce7bbc87ac25a8d
                   teams = [('design','DES'), ('engineering','ENG'), ('product-management','PM')]

IDENTICAL
```

Not one identifier moves on the real corpus, and no team is renamed. Every provided prefix there already equals its team's derived key, so the claim is satisfied without the re-stamp doing anything — which is the outcome I wanted from a change whose whole risk is renumbering ids somebody holds.

CI green on 78fa873.

khj809 (approved):
LGTM

PR #61, opened 2026-08-18
Write down what an import served each record as (--id-map)

### LINKED ISSUE #59: Add `--id-map` option on `backlot import`
After #58 is merged, the DB table does not store dataset-side identifier (doc_id) for each document in favor of the real served id. However, sometimes we still need the doc_id for various debugging purposes, such as ACL cross check between dataset and backlot-imported data.
Hence, as discussed in #58, we decided to add an opt-in `--id-map <path>` option on `backlot import` to emit the file storing a mapping from doc_id to (source_type, served_id).

nuri-yoo:
#58 keyed every table on the id the mock actually serves and dropped `doc_id` from the schema, which is the right shape for the API surface: a client only ever sees the served id, so that is what the tables should hold. The cost is that a dataset id is now a seed that exists nowhere after the import returns. Anything that has to reason about a record from the corpus side — an ACL cross-check between what the dataset declares and what backlot serves, a pass that rewrites prose to cite served ids — has no way to get from `doc_id` to a row.

This adds an opt-in `--id-map <path>` to `backlot import` that writes that join as a JSON manifest, so the schema keeps #58's shape and the mapping lives in a file the import emits rather than in a column.

```
backlot import corpus.jsonl --id-map ids.json
```

The manifest has two halves. `documents` maps each source's dataset id to the row's full served key, taken from the loader's own record of what it wrote — settled through the deferred passes, never re-derived, so it cannot disagree with the row. `containers` maps container names to the ids a client needs to address them (slack channel ids, drive folder ids, linear team id/key, jira project keys, confluence space keys); those are pure functions of the name and are read back from the tables or spelled by the same `synth` call the routers use. Space and project numeric ids are serve-time spellings of the routers and are deliberately left out.

```json
{
 "format": "backlot-id-map/1",
 "containers": {
  "linear": {"engineering": {"id": "17bf8201-99d0-4ab3-9aae-009da890d186", "key": "ENG"}},
  "slack": {"incidents": {"id": "C9CB165E444"}}
 },
 "documents": {
  "confluence": {"cf-handbook": {"id": 33619977}},
  "github": {"gh-gateway-limiter-py": {"repo": "gateway", "number": 52278}}
 }
}
```

Two scope decisions worth flagging. The file covers one run, so a corpus loaded in shards with `--append` produces one manifest per shard rather than a cumulative one — the alternative is reading back and merging a prior file, which silently mixes runs when a path is reused. And `--dry-run --id-map` is refused rather than writing an empty file: the map records ids an import assigned, and a validation pass assigns none.

`tests/test_id_map.py` covers that every record lands in the map under its defaulted-or-stated dataset id and that each entry selects exactly one row in its source's table, that container ids match what the routers spell, that `--append` emits only its own run while containers stay the DB's current state, and that `--dry-run` is refused. Full suite locally: 1091 passed, 24 skipped. `ruff format --check` and `ruff check` clean.

Closes #59


tests/test_id_map.py
```diff
@@ -0,0 +1,177 @@
+"""--id-map: the manifest joining a corpus's own dataset ids to the ids the DB serves."""
```
khj809: This test file can be folded into `test_cli.py` since this is the part of CLI features.

nuri-yoo: Fair point that a flag should not earn its own file by default, but I do not think `test_cli.py` is where these four land, because none of them enter through the CLI:

| test | entry point | what it asserts |
| --- | --- | --- |
| `test_every_document_lands_in_the_map_under_its_served_key` | `byo.load(..., id_map=)` | every manifest entry selects exactly one row, in that source's own id columns |
| `test_containers_carry_the_ids_the_routers_serve` | `byo.load(..., id_map=)` | container ids match `synth` and the `linear_teams`/`jira_projects` rows |
| `test_append_emits_only_the_run_it_belongs_to` | `byo.load(..., reset=False)` | the run boundary: documents are this run's, containers are the DB's current state |
| `test_dry_run_refuses_the_flag` | `byo.run(dry_run=True)` | the refusal, raised by the importer |

They read the DB back and compare against `store.id_columns` and `synth`, which is `test_importer_byo.py`'s job — that file already has the `_write` helper, the `Settings` fixture and the same imports.

What I did miss is the CLI side. `--id-map` went into `_BYO_ONLY`, and three places in `test_cli.py` do not know about it yet: `test_byo_options_reach_the_byo_importer_typed` does not assert it arrives as a `Path`, the option tuple in `test_import_help_shows_every_option_it_accepts` omits it, and the parametrize list in `test_a_byo_option_under_the_bench_type_is_refused` omits it — so the bench-type refusal I added is untested.

So my preference is to split rather than fold: the three CLI-surface assertions above go into `test_cli.py` alongside the options they belong with, the four manifest tests move into `test_importer_byo.py`, and `test_id_map.py` goes away. That satisfies the no-new-file point without putting `byo.load()`-level DB comparisons in the CLI file.

Happy to do it the other way if you would rather keep everything about one flag in one place — say which and I will push it.

khj809: Your suggestion seems better, I couldn't consider the `test_importer_byo.py` place. Could you please do so?

nuri-yoo: Done in 29717a2. `tests/test_id_map.py` is gone; the four manifest tests are now a `# --- the --id-map manifest ---` section at the end of `test_importer_byo.py`, which already had `_write`, the `Settings` fixture and the `store`/`synth` imports they need. Names are prefixed so they read in place (`test_every_document_lands_in_the_id_map_under_its_served_key`, `test_a_dry_run_refuses_the_id_map`, and so on).

The CLI half went into `test_cli.py` next to the options it belongs with: `--id-map` now has to arrive at the importer as a `Path`, appear in `import --help`, and be refused under `--type enterpriserag-bench`.

That last one turned up something while I was there. `test_a_byo_option_under_the_bench_type_is_refused` parametrized over bare flag strings, which only works for boolean options — click stops at the missing argument before the refusal the test is about, so the assertion sees "Option '--id-map' requires an argument." instead. It is parametrized over argv fragments now, which also covers `--roster`, absent until now for the same reason.

Suite: 1093 passed, 24 skipped (1091 before, plus the two new refusal cases). `ruff format --check` and `ruff check` clean.

khj809 (commented):
Three notes on the logic; the manifest's shape and the container ids themselves check out (they match what the routers spell, and replies / thread messages land in the map under their defaulted ids).

backlot/importer/byo.py
```diff
@@ -2963,6 +3011,8 @@ def _scanned():
     # it — see the docstring). On append the prior value already reflects earlier loads.
     prior = 0 if reset else int(store.read_meta(conn, "source_documents") or 0)
     store.write_meta(conn, "source_documents", prior + source_docs)
+    if id_map is not None:
+        _write_id_map(loader, conn, Path(id_map))
```
khj809: This runs inside the salvage-protected region, after the DB commit, `tokens.yaml` and `oauth.generate`. So a bad `--id-map` path (missing parent dir, read-only, permissions) throws, `load_records`' `except BaseException` restores the previous DB, and the data dir is left mismatched:

```
after first import: [('first corpus',)]   tokens users: ['a@x.com']
RAISED: FileNotFoundError .../no/such/ids.json
after rollback, DB: [('first corpus',)]  tokens users: ['b@x.com']
```

The restored DB's owner has no token, and the token holder has no data. A completed import is also thrown away for a typo in an output path, and the CLI prints a raw traceback rather than a message. With `--append` there is no salvage, so the DB keeps the appended rows while the command dies — a re-run then appends twice.

Either validate the destination up front in `run()`, beside the `--dry-run` refusal, or write the manifest after `load_records` returns so an output-file problem cannot reach the load.

backlot/importer/byo.py
```diff
@@ -2728,8 +2728,54 @@ def resolve_parents(self) -> None:
...
+    under, settled through the deferred passes — never re-derived, so it cannot disagree with the
+    row. It covers THIS run only: an ``--append`` emits the appended documents, and a sharded
+    corpus concatenates one manifest per shard. ``containers`` is the current DB state instead
+    (container ids are pure functions of their names, identical across runs): read back from the
+    tables where the import stored them (linear, jira), or spelled by the same ``synth`` call the
+    router serves them with (slack, drive, confluence). Space/project NUMERIC ids are serve-time
+    spellings of the routers and are deliberately absent.
+    """
+    documents: dict[str, dict] = {}
+    for (src, did), written in sorted(loader.keys.items()):
```
khj809, line 2791: `loader.keys` is last-writer-wins per dataset id, but when the corpus STATES its ids two records sharing a `doc_id` leave two rows — six rows loaded, three entries emitted:

```
jira: [('PAY-1','one'), ('PAY-2','two')]
github: [('acme/app',5,'one'), ('acme/app',6,'two')]
s3: [('eng','a.md','one'), ('eng','b.md','two')]
MAP: {"github": {"gdup": {...6}}, "jira": {"dup": {"key": "PAY-2"}}, "s3": {"sdup": {...b.md}}}
```

An ACL audit joining through the file then reads the wrong row, or none.

I don't think the fix belongs in the manifest, though: `self.keys` is also what `resolve_cross_references` resolves through (L2346), so in that state a linear parent/relation or a hubspot association already binds to whichever row was written last, with nothing in the corpus saying which was meant. That is an import-level ambiguity worth failing on, and once it does the per-`doc_id` shape here holds as an invariant.

The guard has to be "one dataset id landed on more than one row", not "a `doc_id` repeats": ERB's pairs (L1566) carry no stated key or number, and those collapse to one row as intended — keyless jira duplicates settle on the same key, confluence reuses its memo, slack reuses the ts. Detectable at insert, comparing against `self.keys` before it is overwritten (provisional keys already differ per row), or afterwards through `_claimed`, which is per row.

Happy for that to be its own PR since it predates this one — but then this file shouldn't drop rows silently in the meantime.

backlot/importer/byo.py
```diff
@@ -2985,6 +3040,11 @@ def run(
     """
     corpus = Path(corpus)
 
+    if dry_run and id_map is not None:
+        # The map records ids an import ASSIGNED; a validation pass assigns none, so an empty or
+        # stale file would be worse than the refusal.
+        raise SystemExit("--id-map records the ids an import assigns; --dry-run assigns none")
```
khj809, line 3103: Every other option conflict in this command (`--bundled` + a path, a BYO flag under the bench type) is a `typer.BadParameter` — exit 2, with usage and a param hint. This one exits 1 with a bare line and no usage. Since c4faef9 the options are declared in one place, so the conflict check reads better next to them in `cli.py`; keep this guard for the library entry point if you want both.

backlot/importer/byo.py
nuri-yoo: Confirmed, and worse than the DB alone — the restored corpus and the tokens that describe it disagree:

```
after first import: DB [('first corpus',)]  tokens users: ['a@x.com']
RAISED: FileNotFoundError(2, 'No such file or directory')
after rollback, DB [('first corpus',)]  tokens users: ['b@x.com']
```

Took both halves of your suggestion in b58597e, because they cover different moments. `_check_id_map_destination` opens the path for append once in `load_records`, before the salvage rename, so a typo costs nothing — and it sits there rather than in `run()` so `load()` and `load_records()` get it too. Then `_load_records` only builds the manifest (it needs the loader and the connection) and returns the text, and `load_records` writes it after the salvage is cleaned up. A destination that passes the check and breaks anyway now costs the manifest and nothing else:

```
append, bad dest -> SystemExit: --id-map .../nope/ids.json: No such file or directory
  rows before: [('first',)]  after: [('first',)] -> UNCHANGED
read-only dir -> SystemExit: --id-map .../ro/ids.json: Permission denied
  db exists: False (load never started)
late write failure -> FileNotFoundError
  import survived: [('kept',)]  tokens on disk: True
```

Three tests: the up-front refusal parametrized over replace and append, asserting `_dump_tables` and `tokens.yaml` are both untouched, and one that fails the write at the last moment through `_id_map_manifest` and checks the rows and tokens survive.

backlot/importer/byo.py
nuri-yoo: Reproduced, six rows and three entries:

```
jira:   [('PAY-1', 'one'), ('PAY-2', 'two')]
github: [('acme/app', 5, 'one'), ('acme/app', 6, 'two')]
s3:     [('eng', 'a.md', 'one'), ('eng', 'b.md', 'two')]
MAP: {"github": {"gdup": {"number": 6, ...}}, "jira": {"dup": {"key": "PAY-2"}}, "s3": {"sdup": {"key": "b.md"}}}
```

Agreed the manifest is the wrong place, and I did it at the import level in b58597e rather than as a follow-up, since it is a few lines and it is what makes the per-`doc_id` shape here an invariant instead of an assumption. The guard is at the `self.keys[(src, did)] = key` assignment: if that dataset id already resolved to a DIFFERENT key, the second row is refused. Comparing keys rather than `doc_id`s is what you asked for and it is what keeps the legitimate pairs working — I checked the shapes you named:

```
loaded, no refusal. total = 9
jira:       {"pair": {"key": "PAYDF384A-901"}, "shared": {...}}     # keyless duplicates -> one row
confluence: {"cpair": {"id": 6715206}, "shared": {...}}             # memo reused
slack:      {"spair": {"channel": "general", "ts": "1704164645..."}} # ts reused
linear:     {"shared": {...}}                                       # one dataset id across three sources
```

ERB's own jira records were the case I was most worried about, and they are safe for a concrete reason: `_byo_jira` passes no `key`, so the pair the comment at L1562 mentions goes through the `repeat` branch and lands on one row. Two tests, one per direction — the stated-id duplicate refused, and keyless duplicates still collapsing to one row and one entry.

The message names both rows rather than just complaining about the id, since the useful question is which of the two the author meant.

backlot/importer/byo.py
nuri-yoo: Right, and it also skipped usage entirely. Moved to `cli.py` in b58597e, next to the `--bundled` conflict:

```
Usage: backlot import [OPTIONS] [CORPUS]
Try 'backlot import -h' for help.
╭─ Error ─────────────────────────────────────────────────────────────────────╮
│ Invalid value for '--id-map': --id-map records the ids an import assigns;    │
│ --dry-run assigns none                                                      │
╰─────────────────────────────────────────────────────────────────────────────╯
exit code: 2
```

Kept the `run()` guard for the library entry point as you suggested, with a comment saying a command line never reaches it. Tested on both sides: exit 2 with usage in `test_cli.py`, the `SystemExit` in `test_importer_byo.py`.

khj809 (approved):
All three addressed in b58597e, and I re-ran each one here.

The destination check before the salvage rename plus the write after it covers both moments — a bad path now costs nothing and leaves the data dir alone:

```
bad dest -> SystemExit --id-map .../no/such/ids.json: No such file or directory
  db [('first',)] -> [('first',)]   tokens ['a@x.com'] -> ['a@x.com']
```

The duplicate guard fires on split keys only, which is the shape that matters: the stated pair is refused naming both rows, while keyless jira/confluence duplicates still settle onto one row through the `repeat` branch and one dataset id across three sources is untouched. Confirmed `_byo_jira` passes no `key`, so ERB stays safe. `--dry-run --id-map` is exit 2 with usage now, and the `run()` guard reads clearly as the library-side one.

Full suite here: 1159 passed, 2 skipped (all extras synced).

Minor, not worth holding this up: `_check_id_map_destination` opens the path with `"a"` (L2758), which CREATES it. So a load that fails after the check leaves a 0-byte file behind:

```
invalid corpus -> SystemExit line 2: source_type must be one of [...]
  ids.json exists: True   size: 0
```

Same "empty file" the `--dry-run` refusal argues against, though a truncated JSON fails loudly downstream rather than silently, so the cost is small. Unlinking it when it did not exist before, or probing the parent directory instead, would close it whenever you next touch this.

backlot/importer/byo.py
nuri-yoo: Good catch — fixed in 9aa4da6. `open("a")` stays, since probing the parent directory would miss an existing file that cannot be written, but one this check created is unlinked again and `load_records` creates it for real:

```
invalid corpus -> SystemExit line 2: source_type must be one of [...]
  ids.json exists: False
```

A file that was already there keeps its contents, since deleting it is not this check's business. One test covers the three directions (nothing left after a refused load, a pre-existing file untouched, a successful load overwriting it). It also turned out `test_an_id_map_that_fails_at_the_last_moment_keeps_the_import` was relying on the check to create the file it unlinks, so that would have started passing for the wrong reason — the directory is simply empty now. 1100 passed, 24 skipped here.

nuri-yoo:
> All three addressed in [b58597e](https://github.com/brekkylab/backlot/commit/b58597e4b783821e289b39fb0a3678c732bd5f61), and I re-ran each one here.
> 
> The destination check before the salvage rename plus the write after it covers both moments — a bad path now costs nothing and leaves the data dir alone:
> 
> ```
> bad dest -> SystemExit --id-map .../no/such/ids.json: No such file or directory
>   db [('first',)] -> [('first',)]   tokens ['a@x.com'] -> ['a@x.com']
> ```
> 
> The duplicate guard fires on split keys only, which is the shape that matters: the stated pair is refused naming both rows, while keyless jira/confluence duplicates still settle onto one row through the `repeat` branch and one dataset id across three sources is untouched. Confirmed `_byo_jira` passes no `key`, so ERB stays safe. `--dry-run --id-map` is exit 2 with usage now, and the `run()` guard reads clearly as the library-side one.
> 
> Full suite here: 1159 passed, 2 skipped (all extras synced).
> 
> Minor, not worth holding this up: `_check_id_map_destination` opens the path with `"a"` (L2758), which CREATES it. So a load that fails after the check leaves a 0-byte file behind:
> 
> ```
> invalid corpus -> SystemExit line 2: source_type must be one of [...]
>   ids.json exists: True   size: 0
> ```
> 
> Same "empty file" the `--dry-run` refusal argues against, though a truncated JSON fails loudly downstream rather than silently, so the cost is small. Unlinking it when it did not exist before, or probing the parent directory instead, would close it whenever you next touch this.

 fixed in 9aa4da6. `open("a")` stays, since probing the parent directory would miss an existing file that cannot be written, but one this check created is unlinked again and `load_records` creates it for real:

```
invalid corpus -> SystemExit line 2: source_type must be one of [...]
  ids.json exists: False
```

A file that was already there keeps its contents, since deleting it is not this check's business. One test covers the three directions (nothing left after a refused load, a pre-existing file untouched, a successful load overwriting it). It also turned out `test_an_id_map_that_fails_at_the_last_moment_keeps_the_import` was relying on the check to create the file it unlinks, so that would have started passing for the wrong reason — the directory is simply empty now. 1100 passed, 24 skipped here.

PR #63, opened 2026-08-18
github: a file may hold several snapshots at one path, and the newest is the one served

### LINKED ISSUE #62: github: a repo holds one snapshot per path, so a corpus carrying a file's history cannot be imported
## Summary

`github_items` can physically hold several `kind='file'` rows for one `(repo, path)` — the
PRIMARY KEY is `(repo, number)` and each row draws its own number. The importer nevertheless
refuses such a corpus, because `_existing_file_number` hands the second row the first row's
number and the corpus-wide identity check then sees two documents resolving to one served id.

The result is that a corpus which models a file's history — one snapshot per pull that touched
it, which is what a document set derived from real repository activity naturally looks like —
is **unloadable**, with an error that points at the wrong field.

Measured on a 10.5k-document corpus: 45 `(repo, path)` groups, 100 file rows, and the import
aborts on the first one.

## Reproduction

Four minimal corpora, current `main` (03f3f59):

| corpus | result |
|---|---|
| two doc_ids, same `(repo, path)` | **refused** |
| two doc_ids, same `(repo, path)`, explicit distinct `number` (11 / 22) | **refused**, identical message |
| one doc_id, same path, twice | accepted (upsert, later wins) |
| two pull_requests + one file, all naming the path | accepted |

```
line 2: this row resolves to ('repo', 'number') = ('acme/gw', -1), which 'd1' in this corpus
already resolves to. Two documents cannot share the one id the API serves them at -- one would
be unreachable at it. Give one of them a different id, or (if they are the same document) the
same one.
```

## Defect 1 — the diagnostic names a remedy that cannot work

This is independent of the design question below and is worth fixing on its own.

`byo.py:1919`, in the file branch, overwrites any corpus-provided number *before* the identity
check runs:

```python
if src == "github" and cols.get("kind") == "file":
    cols["number"] = self._existing_file_number(container, cols.get("path"))
```

So "Give one of them a different id" cannot be followed for a file row — row 2 of the
reproduction sets `number: 22` and is still refused at `-1`. The message also reports a
PROVISIONAL internal number (`-1`, `-240`) and never names `path`, which is the field actually in
conflict. Readers reasonably conclude that synthesized numbers collided.

A file-specific message should name the real conflict and the real remedies:

```
line 633: this file row states (repo, path) = ('douro/enterprise-sales-amer',
'config/pipeline_coverage_targets.yaml'), which 'scn-20260815-00449-r0' already states. A file is
addressed by (repo, path) and its `number` is ignored, so giving it a different number will not
separate them. Two snapshots of one file are not representable yet (see Defect 2 below); keep
one `file` row for the current content and model the edits as pull_request rows naming the path
in `changed_paths`.
```

## Defect 2 — the github model has no time dimension

The refusal is a symptom. The model keeps exactly one state per repo, which the code already
says plainly.

`routers/github.py:764` (`get_tree`):

> **We keep no history**, so any `ref` — a branch name, or a sha from /branches, /commits or
> git/ref — resolves to the repo's CURRENT files. […] Because there is no before/after state, a
> pull's changeset is synthesized out of this same snapshot rather than diffed from it […] which
> is also why a file the changeset reports as `removed` is still present here.

`routers/github.py:736` (`get_git_ref`):

> Any ref resolves to the repo's snapshot commit, as `/branches` and `git/trees` already do —
> the mock keeps no history and the schema carries no branch list, so ref EXISTENCE is not
> knowable here.

Concretely today:

- one commit sha and one tree sha per repo (`_repo_commit_sha`, `_repo_tree_sha` — both `sha1` of
  the repo name alone)
- `/repos/{o}/{r}/contents/{path}` takes no `ref` parameter at all, so real GitHub's `?ref=` is
  accepted and silently ignored
- `_pr_files` builds a pull's changeset from the *current* snapshot and derives hunks from it, so
  a pull's diff is not a diff between two states
- `get_repo_file` is `SELECT * … WHERE repo=? AND kind='file' AND path=?` with no `ORDER BY` and
  no `LIMIT` — with two rows present it returns an arbitrary one

A corpus that states a file's content at several points in time is therefore carrying strictly
more information than the model can represent, and the importer rejects it rather than
representing or degrading it.

## What already works in favour of a fix

- **The primary key is not the obstacle.** Prototyped by stubbing `_existing_file_number` to
  return `None`: both snapshots load and coexist under distinct numbers.

  ```
  ('acme/gw', 18016, 'file', 'src/a.py', 'v1', 1735689600)
  ('acme/gw', 53951, 'file', 'src/a.py', 'v2', 1735689600)
  ```

  (`get_repo_file` then returns `v1`, the arbitrary-row problem above — the second half of the
  work.)
- **Blob addressing is already version-correct.** `_blob_sha(content) = sha1(content)`
  (`routers/github.py:1042`), content-addressed exactly as git is, and `git/blobs/{sha}` scans the
  repo's files for a content match. Distinct snapshots already get distinct, correct blob shas
  and that route needs no change.
- Every listing and search endpoint already filters `kind='file'` out, so the blast radius is
  confined to the file-serving routes.

## The constraint any design has to respect

`(repo, path)` is currently the **only** handle for recognising a re-imported file: `doc_id` was
dropped at #58, and `_existing_file_number` exists precisely so an `--append` of a file lands back
on its own row instead of duplicating (`test_byo_a_github_file_is_appendable_and_keeps_its_number`).

Allowing N rows per `(repo, path)` destroys that handle unless a **version discriminator** is
introduced, so the corpus must be able to state one. This is the crux of the design, not a
detail.

## Proposed direction

Add an optional version discriminator to file rows and make it part of the file's identity.

**Schema.** A nullable column on `github_items`, e.g. `ref TEXT` (a commit sha, a branch name, or
a corpus-chosen snapshot label).

A partial unique index on `(repo, path, ref)` does **not** work, and this is worth stating
explicitly because it is the obvious first move: SQLite treats NULLs in a UNIQUE index as
distinct, so it enforces nothing for exactly the rows that need it — every file row in today's
corpora leaves `ref` unset.

```sql
CREATE TABLE t(repo TEXT, path TEXT, ref TEXT, kind TEXT, n INTEGER);
CREATE UNIQUE INDEX i ON t(repo, path, ref) WHERE kind = 'file';
INSERT INTO t VALUES('r','p',NULL,'file',1);
INSERT INTO t VALUES('r','p',NULL,'file',2);   -- accepted
```

The discriminator therefore has to be total. `created_ts` is the natural fallback, since every
file row already has one (`NOT NULL`, filled from the corpus or from `synth.epoch(doc_id)` and so
stable across re-imports):

```sql
CREATE UNIQUE INDEX idx_github_file_snapshot
    ON github_items(repo, path, COALESCE(ref, created_ts)) WHERE kind = 'file';
```

`ref` unset then means "the snapshot taken at `created_ts`", and a single row per path behaves
exactly as it does today.

**Identity / import.** `_existing_file_number` looks up `(repo, path, COALESCE(ref, created_ts))`.
Note that `WHERE ref = ?` bound to `None` never matches in SQL, so the lookup has to be written
with `IS` / `COALESCE` rather than `=` — the same trap as the index. A discriminator the repo does
not yet hold is a new snapshot; the same one is a re-import and keeps its number. Re-import
recognition is restored, and stating a `ref` is how a corpus says "this is a different point in
time" — the same shape `--append` already uses for issues (`number`) and jira (`key`).

The trade-off to state in the schema comment: with `ref` unset, a file's identity becomes
time-dependent, so re-importing the same content under a new `created` adds a snapshot rather than
updating one. That is consistent with the project's existing posture for stated ids ("there is no
update path", `_require_provided_id`), but it is a semantic choice and should be documented rather
than inferred.

**Serving.**

- `get_repo_file(repo, path)` resolves HEAD as **`ORDER BY created_ts DESC LIMIT 1`,
  unconditionally**. Not "the `ref IS NULL` row if one exists": with several rows carrying no
  `ref` — which is every colliding group in a real corpus — that rule selects nothing in
  particular. The function has no `ORDER BY` and no `LIMIT` at all today, which is a latent bug
  independent of this issue.
- `list_repo_files` / `list_repo_file_paths` must collapse to one row per path
  (`GROUP BY path` taking `MAX(created_ts)`), or a path appears several times in a tree and in a
  directory listing.
- `/contents/{path}` gains `?ref=`, and `git/trees/{ref}` resolves per-ref rather than to the one
  snapshot. Both can fall back to HEAD when the ref is unknown, keeping current behaviour for
  corpora that state nothing.
- `_pr_files` can then derive a real before/after when both snapshots exist, instead of
  synthesizing from one. This is the part that turns the existing "the diff applies with real
  `git`" guarantee from a per-file-snapshot fiction into an actual diff — optional, and the
  largest piece.

**Call sites assuming one row per `(repo, path)`**

| location | change |
|---|---|
| `store.get_repo_file` (2498) | resolve HEAD, add `ORDER BY`/`LIMIT` |
| `store.list_repo_files` (2401) | dedupe to HEAD |
| `store.list_repo_file_paths` (2411) | dedupe to HEAD |
| `store` file COUNT (2494) | count distinct paths |
| `byo._existing_file_number` (1181) | key on `(repo, path, ref)` |
| `byo` identity block (1906–1919) | file rows claim `(repo, path, ref)` |
| `routers/github._contents_response` (823) | accept `?ref=` |
| `routers/github.get_tree` (764) | resolve per-ref |
| `routers/github._RepoFiles.get` (1559) | ref-aware for pull diffs |
| `routers/github.get_blob` (857) | none — content-addressed already |

## Verified against a real corpus

Prototyped by stubbing `_existing_file_number` to return `None` (i.e. no row adopts another's
number) and importing a 10,533-document corpus whose github rows carry 45 multi-snapshot paths.

**Storage and import need no corpus change.** All 893 file rows load; nothing is dropped:

```
file rows loaded: 893
(repo, path) holding more than one row: 45  (rows involved: 100)
```

**Serving does not work without the two rules above**, which is what makes them requirements
rather than polish. With the relaxation alone:

```
config/ondol/dual_run_settlement_checks.yaml  (2 snapshots)
   number=39006   created_ts=1756790024  <- get_repo_file returns this
   number=26411   created_ts=1771570990
   newest by created_ts = 26411             <- HEAD is this one
```

`get_repo_file` returns whichever row the scan reaches first (the older snapshot here), and the
tree listing repeats the path:

```
duplicated paths in listing:
  config/pipeline_coverage_targets.yaml: 3
  config/meadow_objection_kit.yaml: 3
  config/pipeline_coverage.yaml: 2
  config/region_pricing_rules.yaml: 2
  config/security_questionnaire_library.yaml: 2
```

A silent wrong answer is worse than the current hard refusal, so the HEAD rule and the listing
collapse are part of the change, not follow-ups.

**`COALESCE(ref, created_ts)` is a valid key for this corpus with no corpus change**: across all
893 file rows, `(repo, path, created_ts)` has 0 duplicates, and no colliding group shares even a
calendar date.

**Size the design for more than the current numbers.** The same corpus has path-hygiene defects —
17% of file paths redundantly restate the repo, and hyphen/underscore spellings of one file coexist.
Canonicalising those *merges* paths that are currently distinct, so the snapshot counts rise:

| corpus state | colliding paths | rows | max snapshots on one path |
|---|---:|---:|---:|
| as shipped | 45 | 100 | 3 |
| repo prefix stripped | 49 | 109 | 3 |
| + separators normalised | 58 | 130 | 4 |

So a corpus cannot clean up its paths until this lands, and the implementation should not assume
two snapshots per path.

## Alternatives considered

- **Keep the constraint; require corpora to model history as pull_requests + one file row.**
  What the current model expects, and it does import cleanly. But it discards the content of every
  earlier snapshot, and a pull's changeset then describes an edit whose before-state the corpus
  no longer holds. For a corpus built to be grounded in, deleting that content is a real loss.
- **Take the newest row and drop the rest at import.** Silent data loss; the project's own posture
  elsewhere is to refuse rather than quietly discard (see `_require_provided_id`).
- **Key file rows on `(repo, path, blob_sha)`.** No new corpus field needed and it is naturally
  content-addressed, but two snapshots with identical content collapse, and the ordering that
  decides HEAD would still have to come from somewhere.

## Scope

Defect 1 is small and independent — worth landing on its own regardless of what is decided here,
since the current message actively misleads.

Defect 2 is a real design change. Worth stating plainly: if the intended posture is that the mock
serves a repository's *current* state and history lives only in issues and pulls, that is a
coherent position — but then `git/trees/{ref}`, `git/ref/{ref}`, `/commits/{sha}` and `?ref=`
are advertising an axis the store does not have, and the importer should say so in the terms
above rather than failing on a number the user cannot influence.

khj809:
Closes #62.

A file is addressed by `(repo, path)` but stored one row per snapshot, so a corpus that records a
file's edits — one row per pull that touched it — now loads instead of being refused.

`github_items` already permitted it: the PRIMARY KEY is `(repo, number)` and every row draws its own
number. What refused it was `_existing_file_number` keying on the path alone, which handed the second
row the first's number and left the corpus-wide identity check seeing two documents at one served id.

## Identity is the snapshot

`(repo, path, COALESCE(ref, created_ts))`, with `ref` a new optional BYO field naming the change a
snapshot came from.

`COALESCE`, not a plain `(repo, path, ref)`: SQLite treats NULLs in a UNIQUE index as **distinct**, so
a nullable third column enforces nothing for exactly the rows that omit it — every row in a corpus
that does not use refs.

```sql
CREATE UNIQUE INDEX i ON t(repo, path, ref) WHERE kind = 'file';
INSERT INTO t VALUES('r','p',NULL,'file',1);
INSERT INTO t VALUES('r','p',NULL,'file',2);   -- accepted
```

`created_ts` makes the key total without asking a corpus for a field it may have no use for. The
cost, stated in the schema comment: with `ref` unset a file's identity is time-dependent, so
re-importing the same content under a new `created` adds a snapshot rather than replacing one — the
same "no update path" the stated ids already have.

## Every read of a file now says which row it picks

| read | before | now |
|---|---|---|
| `get_repo_file` | no `ORDER BY`, no `LIMIT` — with two rows, whichever the scan reached first | newest `created_ts`, `number` breaking a tie |
| listings, file count | one entry per row | one entry per path |
| `contents` | no `ref` parameter; a real client's `?ref=` silently ignored | `?ref=` selects a named snapshot |
| `git/blobs/{sha}` | scans the listing | scans every snapshot |

HEAD is resolved **among the rows the caller can see**. Resolving corpus-wide and ACL-filtering after
answers 404 for a path whose newest snapshot is restricted, which tells the caller a newer one
exists. Pinned by a test, and verified by mutation — reverting that one argument makes
`get_repo_file` return `None` for a path the caller can read.

A ref no snapshot names still answers HEAD rather than 404: a ref is also a git ref, and the
no-history simplification `get_tree` documents means `?ref=main` from a real client has to resolve to
the current file.

`git/blobs/{sha}` **did** need a change, contrary to what I wrote in the issue. A blob sha is
content-addressed so each snapshot already had its own, but the route enumerated through the
now-collapsed listing, which 404'd every blob but HEAD's. A test caught it.

## Two documents at one snapshot are still refused — in the file's own terms

The generic message told the author to give one of them a different id. That cannot be done for a
file row: `number` is ignored there, so the advice was unfollowable, and the number it printed
(`-1`, `-240`) was an internal provisional naming nothing the author wrote. The file's message names
`path` and points at `ref`.

## Upgrade

Opening a DB from before the column adds it rather than dying on `no such column: ref`. Unlike a
served id, `ref` unset is exactly what every pre-existing row is, so it backfills losslessly and no
re-import is needed.

## Out of scope

- `git/trees/{ref}` still resolves any ref to the current file set (the documented no-history
  simplification). Per-ref trees need a ref → snapshot-set mapping.
- a pull's changeset is still synthesized from one snapshot rather than diffed base→head. This is
  now *possible* where both snapshots exist, and is the larger piece #62 flagged as optional.

## Verification

Tests: **1151 passed**, 2 skipped. The one failure,
`test_mirage_patchers_name_every_host_constant_mirage_ships`, is a pre-existing dependency canary
(mirage grew `CALENDAR_API_BASE` and `FORMS_API_BASE`) and fails on `main` untouched by this branch.

7 new tests across the three layers under change: store (HEAD, listing collapse, ACL-scoped HEAD,
column backfill), importer (two snapshots load, one instant refused naming the path, `ref` separates
snapshots sharing an instant), routes (one tree entry served at HEAD, `?ref=`, every snapshot keeps
its blob). The example corpus gains the post-merge snapshot of `gateway/limiter.py`, so the
schema-coverage guard exercises `ref` on a real edit rather than as filler.

Against a 10.5k-document corpus whose github rows carry **58 multi-snapshot paths, four deep** at the
worst (the state after the redundant repo prefixes and hyphen/underscore spellings in its paths are
normalised, which *merges* paths and so raises the count from 45):

```
[1] every file row survived the import: corpus=893 served=893
[2] paths holding a history: 58  rows in them: 130
[3] two snapshots sharing a served id: 0
[4] paths whose served file is not the newest snapshot: 0
[5] paths listed more than once in a tree/contents listing: 0
[6] superseded snapshots unreachable by blob sha (25 sampled paths): 0
[7] count_repo_files(douro/core-services) = 9 vs distinct paths 9
```

Both existing releases of that corpus, which `main` refuses outright, now import unmodified.


nuri-yoo (approved):
Verified by execution against head 26efb82, alongside a read of the diff.

**Suite**: 1094 passed, 0 failed, 24 skipped on my machine (the skips are extras I do not have installed; the mirage canary you mention does not collect here, so nothing failed at all).

**Both corpora, real imports.** The current release of the 10.5k corpus imports unmodified on this head (main refuses it), and a copy with the path hygiene applied — repo prefixes stripped, hyphen/underscore spellings unified — imports too: 893/893 file rows served, 59 multi-snapshot paths holding 132 rows at depth 4, zero `(repo, number)` duplicates, and spot-checked groups serve the newest snapshot as HEAD. My group count is 59 against your 58 because my spelling unification picks the newest snapshot's spelling as canonical; same class, one group of difference, nothing structural.

**Design points I checked rather than took on faith.** The COALESCE index is the right call and your NULL-distinct repro is accurate — a plain nullable third column enforces nothing for exactly the rows that omit it. The upgrade path is safe for a subtle reason worth having on the record: no pre-existing DB can violate the new unique index at creation, because the old `_existing_file_number` collapsed a second row at the same path onto the first row's number and the REPLACE left only one row standing — the defect that made history unloadable is also what guarantees legacy DBs have at most one row per path. And resolving HEAD among caller-visible rows is the leak-safe choice; I confirmed the same property holds for `?ref=` naming a snapshot the caller cannot see (falls back to visible HEAD, indistinguishable from an unknown ref, so nothing reveals a restricted snapshot exists).

One non-blocking observation: `?ref=` on a directory listing being ignored is documented at the route, which is enough — per-ref trees are correctly out of scope with the ref-to-snapshot-set mapping.

On our side, once this merges: we apply the gated path hygiene to the released corpus (152 prefix strips, 12 spelling groups, the 41 scattered-filename judgments), start emitting `ref` from the generator (the PR that touched the file, which also gives you the reverse link you noted), and re-pin our serving gate off the pre-#58 head it currently sits on.

PR #64, opened 2026-08-18
jira: a project must be allowed to state the key it was served

nuri-yoo:
Follow-up to #60, where this turned up while checking whether the "refuses what it serves" shape existed anywhere else.

Rewritten after review. The first version loosened the schema pattern; that was the wrong half. The pattern was a correct transcription of the real API rule, and the defect is in the derivation.

## The dead end

An `--append` MUST state a jira key: `_require_provided_id` refuses a jira record without one once the project has issues, because a probed key draws a fresh value and lands the row a second time. So the key the mock serves has to be a key the corpus is allowed to write back. For two shapes of project name it was not, which leaves those projects with no legal key at all.

```
project            platform-infra-reliability-and-cost-ops
project key served PIRACO73723D
issue key served   PIRACO73723D-726
append stating it  REFUSED: line 1: j-2 [key]:
                   'PIRACO73723D-2' does not match '^[A-Z][A-Z0-9]{1,9}-[1-9][0-9]*$'
```

`CreateProjectDetails.key` states the whole rule: "Project keys must be unique and start with an uppercase letter followed by one or more uppercase alphanumeric characters. The maximum length is 10 characters." `jira.schema.json` already enforced exactly that, and its description already said so. What did not satisfy it is `synth.jira_project_key`, which is `_key` (one initial per word, no upper bound) plus a 6-digit hash — so it passes ten characters at a five-word project name, and inherits a leading digit from a name whose first word starts with one. Its docstring claimed a "valid Jira shape" the code could not deliver.

## The fix

The readable half gives way, since the digest is what keeps keys distinct. It is trimmed to whatever `JIRA_PROJECT_KEY_MAX` leaves after the 6-hex suffix, and leading digits are dropped so the key starts with a letter. A container whose initials are all digits keeps the `PROJ` fallback rather than starting with the digest.

```
container                                  before               ok  after        ok
payments                                   PAYDF384A            Y   PAYDF384A    Y
customer-support                           CS0C6941             Y   CS0C6941     Y
engineering                                ENG8C85B6            Y   ENG8C85B6    Y
3d-printing                                3P3C01D3             N   P3C01D3      Y
3-5                                        351BA709             N   PROJ1BA709   Y
platform-infra-reliability-and-cost-ops    PIRACO73723D         N   PIRA73723D   Y
a-b-c-d-e-f-g-h-i-j-k-l                    ABCDEFGHIJKL45E55A   N   ABCD45E55A   Y
1234-5678                                  15E5957E             N   PROJE5957E   Y
```

Only an already-invalid key moves, and that is a property rather than an observation: a key satisfying the real rule is at most ten characters and letter-led, so its readable half is at most four and starts with a letter, leaving both the trim and the digit-strip as no-ops. Measured over 40,000 generated container names, 0 currently-valid keys change and 0 derived keys violate the rule; collisions stay at 0 across 60,000 multi-word names, since the digest is still 6 hex of the full container name.

The bench's own projects are `customer-support` and `internal-support`, both unchanged, so no corpus needs re-deriving for this.

## Scope check

The same probe over all four sources that demand a stated id on append — fresh import, read the served id back, append stating a different id of the same shape:

```
                                              f58d4c7        this branch
github      number      13838                 ACCEPTED       ACCEPTED
jira        key         PIRA73723D-424242     REFUSED        ACCEPTED
confluence  id          7066509               ACCEPTED       ACCEPTED
hubspot     record_id   4139862163            ACCEPTED       ACCEPTED
```

jira was the only one.

## Also here

`linear.schema.json` no longer says its pattern is wider than "real Linear's 1-5 character key". Linear states no shape for a team key: not in the GraphQL schema, not in the `Team` reference, not in its docs. That number was asserted in #60 without a source, and the pattern stands on what this mock can hand out, which is the only bound there is to check against.

`routers/atlassian.py`'s `_JIRA_KEY_RE` is deleted. Nothing referenced it, and its comment described the pattern this change reasons about.

## Tests

`test_jira_project_key_is_a_shape_real_jira_can_issue` pins the derivation against `^[A-Z][A-Z0-9]{1,9}$` over eight container shapes, and `test_jira_project_key_leaves_an_already_valid_key_alone` pins the no-op property on the names the bench carries. `test_byo_a_project_can_state_the_key_it_was_served` is the append round trip, and `test_byo_a_mistyped_jira_key_is_refused` keeps `A-1` and `PAY-1-x` out.

```
$ uv run pytest tests
1138 passed, 24 skipped in 17.58s

116 files already formatted

$ uv run ruff check .
All checks passed!
```


khj809:
The ground rule of this project is "always minimize the fidelity gap between the real APIs". I searched the Jira's [openapi docs](https://dac-static.atlassian.com/cloud/jira/platform/swagger-v3.v3.json) and found the description of `key` in `CreateProjectDetails`:

> "Project keys must be unique and start with an uppercase letter followed by one or more uppercase alphanumeric characters. The maximum length is 10 characters."

This is exactly same as `^[A-Z][A-Z0-9]{1,9}$` where this PR tries to change. Server/DC is stricter still — its default `jira.projectkey.pattern` is `([A-Z][A-Z]+)`, which admits no digit at all, under an admin-configurable maximum that ships as 10 — so a key Cloud rejects is rejected everywhere.

So the pattern was not "real Jira's rule, which happens to be inconvenient here". It was the one place in this repo where that rule was written down, and it was doing its job: it caught a derivation that produces keys real Jira can never issue. Widening it blesses them instead — every corpus imported after this serves `PIRACO73723D-726` to a client that validates against the real shape and gets a rejection we chose.

The defect is in `synth.jira_project_key`, and its own docstring says so:

> Deterministic, valid Jira shape (uppercase letter start, uppercase alnum).

Both halves of that are false, in exactly the two ways this PR documents: `_key` appends one initial per word with no upper bound, and it inherits a leading digit from a name like `3d-printing`. The fix belongs there — cap the readable half at `10 - 6 = 4` characters and drop a leading digit, reusing the existing `"PROJ"` fallback:

```python
JIRA_PROJECT_KEY_MAX = 10  # real Jira Cloud's cap on a project key

def jira_project_key(container: str) -> str:
    digest = _digest(container)[:6].upper()
    readable = _key(container, "PROJ").lstrip("0123456789")[: JIRA_PROJECT_KEY_MAX - len(digest)]
    return (readable or "PROJ") + digest
```

```
container                                  current              ok  candidate    ok
payments                                   PAYDF384A            Y   PAYDF384A    Y
customer-support                           CS0C6941             Y   CS0C6941     Y
eng-serving-runtime                        ESR0E3D97            Y   ESR0E3D97    Y
3d-printing                                3P3C01D3             N   P3C01D3      Y   <-- key changes
3-5                                        351BA709             N   PROJ1BA709   Y   <-- key changes
platform-infra-reliability-and-cost-ops    PIRACO73723D         N   PIRA73723D   Y   <-- key changes
a-b-c-d-e-f-g-h-i-j-k-l                    ABCDEFGHIJKL45E55A   N   ABCD45E55A   Y   <-- key changes
1234-5678                                  15E5957E             N   PROJE5957E   Y   <-- key changes
```

The keys that move are precisely the keys that are already invalid. That is provable, not just observed — a key that satisfies the real rule is at most 10 characters and letter-led, so its initials are at most 4 and start with a letter, so both `lstrip` and the truncation are no-ops and the key comes out byte-identical. The digest is still 6 hex of the full container name, so the collision residual documented on `jira_project_key` is unchanged; I checked uniqueness over 4096 generated names and every derived key satisfies `^[A-Z][A-Z0-9]{1,9}$`.

"Capping the derivation would change the key every issue of an affected project is served under" is the reason this PR gives for not doing it, and I don't think that reason should carry any weight at this stage. Those keys are wrong; a corpus imported under them was imported under a bug. Throw that data away and re-import — that is already the answer this project gives for every other change to an imported document, and there is no installed base here whose stability is worth encoding a permanent divergence from real Jira to protect.

The schema then goes back to `^[A-Z][A-Z0-9]{1,9}-[1-9][0-9]*$`, which already carries the exactly-one-hyphen guarantee `rsplit("-", 1)` depends on, and both tests in this PR survive as they are: the round trip passes because the served key is now a legal one to write back, and `A-1` / `PAY-1-x` stay refused.

Two smaller things while the schema is being touched: `schemas/README.md` quotes the old pattern verbatim as the corpus-author contract, so it needs to move with whatever this lands on, and `_JIRA_KEY_RE` in `routers/atlassian.py` has no reference anywhere in `backlot/` or `tests/` — its comment describes "an uppercase-led alnum prefix" and it is probably cheaper to delete the constant than to keep the comment true.

khj809:
> Same question stands for `linear_team_key` and a 5-character cap, raised on #60.

I looked for that one in the official docs too, and could not find any stated rule — no length, no character set, no case. The [official Linear GraphQL schema](https://raw.githubusercontent.com/linear/linear/master/packages/sdk/src/schema.graphql) says only this about a team key:

- `Team.key: String!` — "The team's unique key, used as a prefix in issue identifiers (e.g., 'ENG' in 'ENG-123') and in URLs."
- `TeamCreateInput.key: String` — "The key of the team. If not given, the key will be generated based on the name of the team."
- `TeamUpdateInput.key: String` — "The key of the team."

Grepping the whole schema for any length or character-set language turns up nothing that applies to a key.

So the 5-character cap looks like observed client/UI behaviour rather than a documented constraint, which makes it a different question from jira's, not the same one. Jira states its rule in the API contract, so serving a key outside it is a divergence from the documented API; Linear states no rule, and the behaviour this repo already has — deriving the key from the team name — is exactly what `TeamCreateInput.key` describes. A cap there would be cosmetic realism with a real cost (`linear_team_key` carries no hash suffix on purpose, so a served identifier keeps matching the spelling the corpus cites), so I would leave it alone unless someone can show the API itself rejecting a longer key.

nuri-yoo:
Confirmed. `CreateProjectDetails.key` in `swagger-v3.v3.json` says exactly what you quoted, so `^[A-Z][A-Z0-9]{1,9}$` is real Jira's rule, and the description I replaced already said so.

The id-churn argument does not hold either. `store.py`'s guard refuses every pre-#58 database with "re-import this corpus from scratch", so there is no installed base to protect, and the bench's keys are unaffected regardless: `customer-support` stays `CS0C6941`, `internal-support` stays `IS835720`.

Your candidate reproduces exactly. Over 40,000 generated container names, 0 currently-valid keys change and 0 derived keys violate the real rule; collisions stay at 0 across 60,000 multi-word names. `1138 passed, 24 skipped` rebased onto `8d1c01b`, both tests here unchanged, and an append stating `PIRA73723D-2` is accepted.

Linear is a separate question, as you say. The SDL states no length, character-set or casing rule for a team key, and neither does the Apollo Studio reference for `Team`, `linear.app/docs/teams`, `/api-and-webhooks`, or anything GitHub code search turns up. #60's merged schema description asserts `real Linear's 1-5 character key`, which nothing supports.

khj809:
The derivation fix reads right, and I verified it on `1f39032c`: full suite green with all extras (the one failure, `test_mirage_patchers_name_every_host_constant_mirage_ships`, fails identically on `origin/main`), ruff clean, and both round-trip parametrizations fail against main's code, so the regression guard has teeth.

Three leftovers in `tests/test_importer_byo.py`, all of them the first approach still talking:

- **3585-3592** — the docstring describes code this PR no longer has: "`synth.jira_project_key` is `_key` plus a 6-digit hash, so it passes real Jira's 10-character limit as soon as a project name has five words". It doesn't, as of this change. "stating the served one was refused by validation" is the same thing in the past tense.
- **3614** — "Widening the pattern to admit the served key must not give that up." Nothing widens the pattern in the final state; the pattern is untouched from main.
- **3582** — the ids `key-longer-than-real-jiras-10` and `key-starting-with-a-digit` name properties the served keys no longer have (`platform-infra-reliability-and-cost-ops` now serves `PIRA73723D`, ten characters and letter-led). The name shapes are still the right thing to parametrize over — the ids just want to describe the name rather than the old key, e.g. `six-word-name` and `name-starting-with-a-digit`.

What each test pins is right; it is only the prose that still argues for the version that got dropped. Worth a pass before this lands, since a reader who trusts the docstring learns a rule about `jira_project_key` that stopped being true in the same commit.

nuri-yoo:
All three were the first approach still talking. Addressed in `1cfecfe`, prose only, kept as its own commit so the delta against the revision you verified is one diff with no assertion in it.

`test_byo_a_project_can_state_the_key_it_was_served`: the two facts are now attributed to the name shape and to the derivation as it was, not stated as present tense about `jira_project_key`. Ids are `six-word-name` and `name-starting-with-a-digit`, describing what is parametrized rather than a key property the cap removed.

`test_byo_a_mistyped_jira_key_is_refused`: "capping the derivation is what makes the served key legal to state, and the rule a corpus is held to does not move with it: this pattern is the one main already enforced."

I swept the rest of the branch diff for the same drift and those three were all of it. `1138 passed, 24 skipped`, ruff clean, and both parametrizations still collect and pass under the new ids.

Correcting what I first wrote about `test_mirage_patchers_name_every_host_constant_mirage_ships`: it is collected, not absent. It skips on `pytest.importorskip("mirage.core.google._client")`, because `mirage-ai` sits in the `mirage` extra and I install `dev` only. CI installs `.[dev]` too, so it is skipped there as well.

Chased it since your run reaches it and mine did not. Installing the extra does not get there either: the import dies on `mfusepy.py:136: OSError: Unable to find libfuse` before the assertion, and macFUSE is a kernel extension. Stubbing the FUSE binding out, which is not what the test asserts on, the assertion does fail, and for the reason the test was written to catch:

```
mirage 0.0.5 ships host constants backlot.integrations.mirage does not rebind:
  mirage.core.google._client.CALENDAR_API_BASE = https://www.googleapis.com/calendar/v3
  mirage.core.google._client.FORMS_API_BASE    = https://forms.googleapis.com/v1
```

`mirage-ai[fuse]~=0.0.4` admits 0.0.5, and the shim rebinds eight constants that do not include those two, so they stay pointed at the real Google. Nothing to do with this branch, which does not touch the shim. The part worth deciding separately is that the tripwire is unarmed in CI: whether the shim should rebind Calendar and Forms, or the test should name them as deliberately out of scope, and only then whether CI installs the extra.

khj809:
`1cfecfe` checks out. Parsing `tests/test_importer_byo.py` at both revisions and comparing the ASTs with docstrings stripped, the only executable difference on the whole file is the two parametrize ids — `key-longer-than-real-jiras-10` → `six-word-name`, `key-starting-with-a-digit` → `name-starting-with-a-digit`. No assertion moved, so what I verified on `1f39032c` still holds. This looks ready to me.

On the mirage tripwire: thanks for chasing it down, and your reading of it is right — my run reached the assertion only because I had the `mirage` extra installed, and the two constants it names have nothing to do with this branch.

Please take it out of here and give it its own PR. It is a real finding with its own decision attached — whether `backlot.integrations.mirage` should rebind `CALENDAR_API_BASE` and `FORMS_API_BASE`, or the test should name them as deliberately out of scope, and then whether CI installs the extra at all — and none of that shares a reason to land with a jira key derivation. Worth saying plainly in that PR's description that the tripwire is currently unarmed in CI, since a test that skips everywhere reads as passing: `mirage-ai[fuse]~=0.0.4` admits 0.0.5, the shim rebinds eight constants that do not include those two, and so a Calendar or Forms call under the shim goes to the real Google today.

khj809 (approved):

PR #65, opened 2026-08-18
Drop the fields no vendor response carries, and serve the one that does

khj809:
A BYO field earns its place only if the data comes back out. Most of these did not — accepted,
stored, and served nowhere, so a corpus writing one had it silently dropped.

The criterion is **"does the real API response carry it?"**, not "is the code dead". That cuts both
ways: it removed nine fields, and it made `gmail.bcc` a *serving* fix rather than a deletion.

## Removed

| field | why no response carries it |
|---|---|
| `meta` (×11 schemas) | not a vendor field; proven to drop every value it accepts |
| `slack.title` | a Slack message has no title |
| `jira.severity`, `jira.squad` | neither is a Jira field; both would be `customfield_*`, and no custom field is served |
| `jira.subtype` | `issuetype` is Jira's own word and what the API returns |
| `confluence.reviewers`, `.confidentiality`, `.owner_team` | Confluence exposes labels; none of these |
| `google_drive.collaborators` | Drive exposes owners/permissions; access already rides `readers`/`group` |

### `meta`

Declared as "extras **merged into the record**". It merged nothing. The importer seeded a dict from
it, overlaid the declared names from the top level, then consumed only names it knew — so a declared
key was overwritten and an undeclared one was read and discarded. Imported the example corpus and
searched the whole DB for the twelve `meta` keys it demonstrated, and their values:

```
jira_epic  review_cadence_days  pinned  triaged_by  pagerduty_incident  slack_thread
importance  source_system  synced_from  checksum_sha256  crm_source  crm_deal
                          -> all NOT SERVED, keys and values alike
```

It is also how **2,015 jira keys and 635 github numbers went missing** — written as `meta.key` /
`meta.number`, dropped, and the records served under synthesized ids their own prose never cited.
The generator moved both to the top level (`enterprise-mock-experiments` `f56c5a3`); this is
backlot's half: the place they were written should not have taken them in silence.

Nothing is lost: **every key `_service_columns` consumes is reachable from the top level**, either
as a declared field of its source or — for `subtype` — read from the record directly. (An earlier
revision of this description said "89 of 90 are declared", which overstated by one: `versions` sat
in the convenience list declared by no schema at all. It is dead and this branch removes it.)

An unknown key is now the ordinary `additionalProperties: false` refusal, naming the field.

### `slack.title`

It was the one schema of eleven that left `title` out of `required`; the column was `NOT NULL` so
every producer wrote `''`; and the router turned a non-empty one into a `*bold*` lead line — a
message shape Slack does not have. Whether a converter folds a source document's title into the text
is that converter's decision, not something this schema should offer.

The FTS index keeps a `title` column for every source so its shape and the queries over it stay
uniform; for a titleless source it is fed a constant (`store.TITLELESS`).

## Fixed rather than removed: `gmail.bcc`

Bcc is blind copy — a header the Gmail API **does** return, on the sender's own copy; a recipient's
has it stripped in transit. The column was written and served nowhere, so a dead-code scan flags it,
but deleting it would have thrown away real information. The gap was on the serving side.

`_gmail_message` now takes the caller and emits the header only when the caller is the message's
author. A reader who is not the sender must not learn who was blind-copied, and an admin/service
caller has no email so it is not the sender either. Both call sites (`messages.get`, `threads.get`)
pass it, and the `raw` RFC822 form follows the same rule because it serializes the same headers.

## Also kept: `linear.parent_key`

Serving never reads it, but the importer writes it and a later pass reads it back via SQL to resolve
`parent_id`, deliberately retaining it when the parent sits outside the corpus slice. A column rather
than a dict so `--append` resolves across runs. It is the only column left that its own serving
module does not name, and it is importer-internal by design.

## ERB

The converter still reads `owner_team` / `confidentiality` / `collaborators` / `reviewers` from its
**raw** input to derive ACL — that is where they belong. It just stops copying them into the BYO
record. `_byo_slack` already discarded the title.

## Verification

Imported the 10.5k v3 corpus before (on `main`) and after:

```
every surviving column, byte-identical row for row:
  slack_messages  gmail_messages  jira_issues  github_items  gdrive_files
  confluence_pages  linear_issues  notion_pages  hubspot_objects
  fireflies_transcripts  s3_objects            -> all IDENTICAL

ACL grants unchanged on all eleven sources (slack 15,440 · gmail 6,828 · jira 3,545 · … · s3 589)
```

Dropped columns: `slack_messages.title`, `jira_issues.{severity,squad}`,
`confluence_pages.{reviewers,confidentiality,owner_team}`, `gdrive_files.collaborators`. Nothing
else moved.

Tests **1174 passed**, 2 skipped. The one failure,
`test_mirage_patchers_name_every_host_constant_mirage_ships`, is a pre-existing dependency canary
(mirage grew `CALENDAR_API_BASE` / `FORMS_API_BASE`) that fails on `main` untouched by this branch.

Tests that pinned the old behaviour were removed rather than inverted — the schemas' own
`additionalProperties: false` is what refuses a removed field, and that mechanism is already pinned
by `test_unknown_top_level_key_rejected`. `test_byo_meta_cannot_smuggle_a_tracker_id` became
`test_byo_a_tracker_id_is_read_from_the_field_its_schema_declares`, asserting the corpus can no
longer express the smuggle at all — a stronger guarantee than the value being ignored.

## Fixed after review

Two regressions this branch introduced, both reported with repros and both now covered by tests:

- `search_documents` / `count_search` read `title` directly on their **LIKE fallback**, so a build
  without FTS5 (or a DB whose index was never built) raised `no such column: title` on a slack
  search where it used to answer. The three statements ask `title_expr` for the source's title now.
  The fallback had no test on either revision; it has one.
- `CREATE TABLE IF NOT EXISTS` leaves an older DB's `slack_messages.title` in place, and it was
  `NOT NULL`, so an `--append` onto one died on the constraint. `_DROPPED_COLUMNS` mirrors
  `_ADDED_COLUMNS` and drops it on open — lossless, since nothing reads it. The other removals are
  nullable and degrade to NULL on their own.

Plus prose that outlived its mechanism (the `meta` seeding comment, `_states_own_id`'s note about
`meta.transcript_id`, `_service_columns`' "(+ meta)", the schemas README's people-and-scope bullet,
store's module docstring calling `title` a core column on every doc table).

**Breaking:** a corpus setting any removed field now fails validation naming it. Measured on v3: 0
rows affected. The bundled hello corpus, the example corpus and the fixtures are updated.


backlot/store.py
```diff
@@ -2151,7 +2165,7 @@ def search_documents(
         elif lit and re.search(r"\w[^\w\s]\w", lit):
             order_sql = (
                 "(instr(lower(t.content), lower(?)) > 0 "
-                f"OR instr(lower(t.title), lower(?)) > 0) DESC, {fts}.rank"
+                f"OR instr(lower({title_expr(source_type, 't')}), lower(?)) > 0) DESC, {fts}.rank"
```
nuri-yoo, line 2072: `title_expr` reaches the FTS branch, but the LIKE fallback below this hunk still reads `title` directly: `search_documents`' `WHERE (title LIKE ? OR content LIKE ?)` and its `CASE WHEN title LIKE` tiebreak, and the same WHERE in `count_search`. With no FTS index, a slack search now dies instead of falling back.

Reproduced on this head (one-message corpus, `slack_fts` dropped): `store.search_documents(conn, "latency", "slack")` raises `OperationalError: no such column: title`, and `/slack/api/search.messages` answers 500. The same scenario on `8d1c01b` answers 200 through the LIKE path. The docstring still promises the fallback ("FTS5-ranked; LIKE fallback"), and `_fts5_ok`/`_has_fts` exist precisely for builds without FTS5.

`title_expr(source_type)` in those three statements fixes it. The path has no test on either revision, so a no-FTS regression test alongside would keep it fixed.

backlot/store.py
```diff
@@ -276,7 +276,7 @@ def id_seed_scope(source_type: str) -> str | None:
 -- standalone message carries NULL.
 CREATE TABLE IF NOT EXISTS slack_messages (
     channel TEXT NOT NULL, ts TEXT NOT NULL, author_email TEXT NOT NULL,
-    title TEXT NOT NULL, content TEXT NOT NULL,
+    content TEXT NOT NULL,
```
nuri-yoo, line 280: `CREATE TABLE IF NOT EXISTS` leaves a DB imported before this change with the old `title TEXT NOT NULL` column, and the importer no longer writes it, so `--append` onto such a DB dies with a raw `sqlite3.IntegrityError: NOT NULL constraint failed: slack_messages.title` (reproduced: built on `8d1c01b`, appended one slack doc on this head). The nullable drops degrade fine: appended confluence/jira/gdrive rows simply leave the old columns NULL. Slack is the one hard failure.

The repo already has both shapes for this class: `_ADDED_COLUMNS` for lossless column changes, `refuse_pre_served_db` for a readable refusal. Dropping the legacy column on open (`ALTER TABLE slack_messages DROP COLUMN title` is lossless now that nothing reads it) or refusing with the usual "re-import" sentence would keep the cross-version story consistent.

Related nit: the module docstring above ("Every doc table carries the same three core columns (`author_email, title, content`)") now has slack as a counterexample.

backlot/schemas/README.md
```diff
@@ -117,17 +117,17 @@ backlot import generated.jsonl --dry-run && backlot import generated.jsonl
   unknown top-level key is almost always a typo). Gmail's `content` and its `messages[].content`
   are the one exception to non-empty content: a thread opened or continued by a header-only message
   (auto-ack, bare forward) is real, and dropping it would renumber the rest of the thread.
-- **Permissive** — the free-form `meta` object and the loosely typed per-service extras
-  (`reactions`, `attachments`, `issuelinks`, `reviews`, `changelog`, …), which the loader stores
-  as JSON without a fixed shape.
+- **Permissive** — the loosely typed per-service extras (`reactions`, `attachments`,
+  `issuelinks`, `reviews`, `changelog`, …), which the loader stores as JSON without a fixed shape.
+  Each is a declared field of the source that has it: there is no free-form object, so a key no
```
nuri-yoo, line 122: The "Per-service people and scope" bullet further down this file (untouched by the diff) still documents confluence `confidentiality`/`owner_team`/`reviewers`, drive `collaborators`, and jira `severity`/`squad` as accepted optional fields, so a corpus written from that paragraph now fails validation naming them. It needs the same trim as this bullet; slack `participants`, gmail `mailbox_owner`, and `author_name` are the survivors.

backlot/importer/byo.py
```diff
@@ -1734,16 +1722,9 @@ def register(email: str | None, name: str | None = None) -> None:
             grant_types = [("org", org)]
 
         # structured extras: rec.meta merged with convenience top-level keys
```
nuri-yoo: This header comment still describes the deleted mechanism ("rec.meta merged with convenience top-level keys"), and a few more pieces of prose outlived the removal: the `_states_own_id` comment ("`meta.transcript_id` is ordinary meta content, stripped before extras are seeded" — the strip loop is gone), `_service_columns`' docstring ("Map generic BYO fields (+ meta)"), and the `transcript_id` note inside this tuple whose "which is what the loop above exists to prevent" now points at nothing. All four were accurate on `8d1c01b`; each now teaches a reader a mechanism this branch removed.

backlot/importer/byo.py
```diff
@@ -1734,16 +1722,9 @@ def register(email: str | None, name: str | None = None) -> None:
...
-        # A tracker id is read from the field the schema declares it in, and from nowhere
-        # else. `meta` is documented free-form, so seeding `extras` from it let
-        # `meta: {"number": 3}` claim issue 3 in a repository just as a top-level `number`
-        # would — a spelling no schema describes, that shadows a real issue, and that the
-        # uniqueness check below would then refuse an import over. Both ids are ordinary
-        # `meta` content again: carried through, never promoted to the served column. Read from
-        # DEFERRED_ID so a source that gains a stateable id is covered by declaring it there.
-        for reserved in (*DEFERRED_ID.values(), "transcript_id"):
-            extras.pop(reserved, None)
+        # Every per-source field is read from the top level, and each name below is one the schema
```
nuri-yoo, line 1722: "each name below is one the schema declares" has one counterexample: `versions`, further down this tuple, is declared by no schema. Checked every name against all eleven schemas' `properties`: the other 86 unique names are declared, `versions` alone is not. The entry predates this PR (it came in with the #44 rename), but this comment and the PR body's "89 are declared top-level fields" both now overstate by one. Since no record can carry the key any more, the entry is dead and can go.

nuri-yoo (commented):
Reviewed by running both revisions (`8d1c01b` and `b79384e`) side by side. The description's verification claims reproduce: imported the v3 corpus (10,537 docs, with its roster) on both, and every surviving column is hash-identical row for row across all 47 tables (76,152 rows); removed-field usage in that corpus is 0 rows, and the 2,015 jira / 635 github id-lift counts match exactly. The removals check out at the mechanism level too: on `8d1c01b`, `meta.mime_type`/`meta.state` were consumed only through the seeding path this PR deletes, a top-level `jira.subtype` landed in no column at all, and Drive `permissions.list` already answered from ACL grants rather than the `collaborators` column. The Bcc rule holds at route level as well as in the unit test: the sender's own token gets the header, another reader on the same thread and the admin/service caller do not, and `threads.get` and the `raw` form agree with `messages.get`.

Suite: with every extra installed (mcp, llamaindex plus the hubspot reader, examples, mirage), the Docker-gated mcp-atlassian test replicated without Docker (the package from pip plus a `getaddrinfo` patch standing in for `--add-host`), and `ERB_E2E=1` run on both revisions, base and head fail on exactly one and the same test: the pre-existing mirage 0.0.5 canary the description already names. Nothing this PR touches regresses under the full matrix.

Two things do need a fix, both introduced on this branch; the inline comments carry the repro details.

- The slack LIKE fallback in `search_documents`/`count_search` still reads `title`, so a corpus without an FTS index now 500s on `search.messages` where `8d1c01b` answered through the fallback.
- `--append` onto a DB imported before this change hits the old `title TEXT NOT NULL` column and dies with a raw `IntegrityError`. The nullable drops degrade gracefully; slack is the one hard failure, and neither `_ADDED_COLUMNS` nor `refuse_pre_served_db` covers it.

The rest is prose that outlived the mechanism it described (the schemas README's people-and-scope bullet, a few comments in `byo.py`, the store module docstring) and one dead `versions` entry, all inline.

khj809:
All resolved — thanks, both bugs reproduced exactly as described.

**LIKE fallback** (`bae0ce5`) — `title_expr(source_type)` in all three statements: `search_documents`' WHERE and its `CASE WHEN` tiebreak, and `count_search`'s WHERE. Your repro now answers instead of raising:

```
before: search_documents / count_search -> OperationalError: no such column: title
after:  1 hit / 1
```

Added the regression test you suggested, covering a titleless source and a titled one including the title-first tiebreak.

**`--append` onto an older DB** (`bae0ce5`) — went with the drop rather than the refusal, so an existing DB keeps working: `_DROPPED_COLUMNS` mirrors `_ADDED_COLUMNS` and drops the legacy `slack_messages.title` on open. Your repro goes from `IntegrityError` to `append OK`. Test added.

**`versions`** — verified: 87 convenience keys, 86 declared, `versions` alone is not. Removed, and the PR body's "89 of 90 are declared" is corrected — the accurate claim is that every key `_service_columns` consumes is reachable from the top level, `subtype` via `rec` directly.

**Prose** — all four in `byo.py`, the schemas README people-and-scope bullet (now just slack `participants`, gmail `mailbox_owner`, `author_name`), and store's module docstring, which now names `TITLELESS` instead of calling `title` a core column on every doc table.

Also rebased onto `ceadc84`: `#61` landed after this branch was cut and two of its id-map fixtures give a slack message a `title`, which only CI saw since it tests the merge result. Dropped the field from both; `--id-map` on the v3 corpus still emits slack under `(channel, ts)`. CI green.

nuri-yoo (approved):
LGTM.

Both bugs verified fixed on `54bdddf` — the LIKE fallback answers where it raised, `--append` onto an older DB writes instead of hitting the dropped column, and each new test fails on `b79384e`, so they pin the mechanisms rather than the outcomes. Re-checked what the rebase could have moved: the v3 corpus imports column-identical against `ceadc84`, and the `--id-map` manifest for it is byte-identical to the base's. Suite is 1,190 with every extra installed plus `ERB_E2E=1`, one failure and it is the pre-existing mirage canary.

Two non-blocking notes, neither worth holding this:

- A legacy `slack_fts` still holds the titles it indexed, and `--append` only adds the new ids, so an upgraded-not-reimported DB can match a slack message on a word its served text no longer contains. It predates the fix commit (it follows from the `title` removal itself, not from `_DROPPED_COLUMNS`) and a fresh import is unaffected. Worth a follow-up if keeping an older DB is meant to be airtight; rebuilding the index for whatever source `_reconcile_columns` touched would do it.
- `store.py:19` is 149 characters. E501 is off here for long literals and measured-behaviour tables, and this is prose — the rest of the file tops out at 108. `ruff format` will not wrap inside a docstring.

khj809:
Both non-blocking notes fixed in `277c95d`.

**Stale index after an upgrade** — reproduced your scenario (legacy DB whose one message had `quarterly` only in its title): it matched, and now returns nothing while the content it does serve still matches.

`_reconcile_columns` reports which sources it touched and `connect_rw` rebuilds just those, only where an index already exists — a fresh import builds its own at the end of the load. Split `build_fts_for(conn, src)` out of `build_fts` for it: correcting one table should not cost a rebuild of the whole corpus. Test pins that the dropped-title word stops matching and the content word keeps matching.

**`store.py:19`** — rewrapped; the file tops out at 108 again.

Also re-audited slack `title` end to end, since it kept coming up:

```
schema properties          no 'title'   (and not in `required`)
slack_messages columns     no 'title'
slack_fts                  keeps the column (uniform index shape), 0 rows non-empty
conversations.history      no 'title' key on any message object
search.messages            no 'title' key on any match
hello + example corpora    0 slack rows carrying one
tests (AST scan)           0 slack dict literals carrying one
```

The one survivor is `users.list` → `profile.title`, which is Slack's job-title field on a user profile and unrelated to the message shape.

Suite 1188 + the mirage canary; v3 corpus still imports clean. CI green.

nuri-yoo:
> Both non-blocking notes fixed in `277c95d`.
> 
> **Stale index after an upgrade** — reproduced your scenario (legacy DB whose one message had `quarterly` only in its title): it matched, and now returns nothing while the content it does serve still matches.
> 
> `_reconcile_columns` reports which sources it touched and `connect_rw` rebuilds just those, only where an index already exists — a fresh import builds its own at the end of the load. Split `build_fts_for(conn, src)` out of `build_fts` for it: correcting one table should not cost a rebuild of the whole corpus. Test pins that the dropped-title word stops matching and the content word keeps matching.
> 
> **`store.py:19`** — rewrapped; the file tops out at 108 again.
> 
> Also re-audited slack `title` end to end, since it kept coming up:
> 
> ```
> schema properties          no 'title'   (and not in `required`)
> slack_messages columns     no 'title'
> slack_fts                  keeps the column (uniform index shape), 0 rows non-empty
> conversations.history      no 'title' key on any message object
> search.messages            no 'title' key on any match
> hello + example corpora    0 slack rows carrying one
> tests (AST scan)           0 slack dict literals carrying one
> ```
> 
> The one survivor is `users.list` → `profile.title`, which is Slack's job-title field on a user profile and unrelated to the message shape.
> 
> Suite 1188 + the mirage canary; v3 corpus still imports clean. CI green.

Checked non-blocking fix correctly addressed in `277c95d`. LGTM again

khj809:
@nuri-yoo Sorry for the additional commits after review. There were more remaining legacies than I expected, and some of them (`doc_acl`) should have been removed from #58 but it wasn't. I removed all of them and confirmed all tests pass.

nuri-yoo:
> @nuri-yoo Sorry for the additional commits after review. There were more remaining legacies than I expected, and some of them (`doc_acl`) should have been removed from #58 but it wasn't. I removed all of them and confirmed all tests pass.

No worries. I was tracking the commits through Gmail notifications, so nothing came as a surprise. Approval stands.

PR #66, opened 2026-08-19
mirage: rebind the Calendar and Forms hosts mirage 0.0.5 added

nuri-yoo:
Split out of #64 at review request. The finding surfaced there while explaining why `test_mirage_patchers_name_every_host_constant_mirage_ships` never fired for me, and it carries its own decision, so it lands on its own.

## The tripwire is unarmed, which is why nobody saw this

A test that skips everywhere reads as passing. This one skips in two places, and CI is one of them.

```
1. pytest.importorskip("mirage.core.google._client")
   mirage-ai[fuse] is in the `mirage` extra. CI runs `pip install -e ".[dev]"`, so CI skips it.

2. mfusepy.py:136: OSError: Unable to find libfuse
   With the extra installed but no macFUSE, the import dies before the assertion.
```

I installed macFUSE (5.3.3, kext approved and rebooted) to get past the second gate rather than reason about it. On `eb08e6d` with `--extra dev --extra mirage`:

```
FAILED tests/test_integrations.py::test_mirage_patchers_name_every_host_constant_mirage_ships
AssertionError: mirage ships host constants backlot.integrations.mirage does not rebind,
so they stay pointed at the real vendor:
{'mirage.core.google._client.CALENDAR_API_BASE': 'https://www.googleapis.com/calendar/v3',
 'mirage.core.google._client.FORMS_API_BASE': 'https://forms.googleapis.com/v1'}
```

The test is doing exactly what its docstring promises: "Fails when mirage grows one, which is the moment to decide whether the mock serves it."

## What changes

`point_google_at` rebinds both, for the reason `DRIVE_UPLOAD_BASE` already is: backlot serves neither route, so pointing them at the mock turns a call that would have reached the vendor into a 404 that says so. Verified against the app's served paths — `/calendar/v3` and `/forms/v1` are absent from all 111, next to `/upload/drive/v3`:

```
/calendar/v3       -> NONE (404)
/forms/v1          -> NONE (404)
/upload/drive/v3   -> NONE (404)
/docs/v1           -> ['/docs/v1/documents/{document_id}']
/slides/v1         -> ['/slides/v1/presentations/{presentation_id}']
```

`FORMS_API_BASE` takes `/forms/v1` rather than the `{base}/v1` mirage itself composes, because `/docs/v1` IS served: on the single-host spelling a Forms call would land on the docs route instead of failing.

The floor moves to `~=0.0.5`. `_rebind_mirage_constants` raises on a name it cannot find in any imported `mirage.core.*`, and 0.0.4's `_client` ships seven constants without these two, so on 0.0.4 `point_google_at` would fail outright rather than degrade. Compared both sdists:

```
0.0.4: TOKEN_URL DRIVE_API_BASE DOCS_API_BASE SLIDES_API_BASE SHEETS_API_BASE
       GMAIL_API_BASE DRIVE_UPLOAD_BASE
0.0.5: ... + CALENDAR_API_BASE (line 31) + FORMS_API_BASE (line 32)
```

## A stale reason in the docstring, corrected

`point_google_at` said the constants are patched because "Google exposes no host config". That is not why, and it is worth fixing so the next person does not go looking for a config that is right there.

`GoogleConfig.api_base` exists — in 0.0.4 as well — and its own comment describes it as a "single-host override for every Google API ... used to point backends at a fake server". Every base getter in `_client` consults it first.

What rules it out is the SINGLE. Under one `api_base`, mirage composes three getters identically:

```
docs_base   -> http://127.0.0.1:9999/v1
forms_base  -> http://127.0.0.1:9999/v1
slides_base -> http://127.0.0.1:9999/v1
```

Three APIs on one prefix, while the mock serves docs at `/docs/v1` and slides at `/slides/v1`. Per-constant rebinding is what keeps them apart, and that is what the docstring now says.

## Left out on purpose

Arming the tripwire in CI is the follow-on question, not this PR. It needs `libfuse` on the runner for the import to survive, which is a build-time dependency added for one test, and it is worth deciding on its own terms now that the failure it was hiding is fixed. Flagging it since the same silence will hide the next constant mirage adds.

## Verification

```
$ uv run pytest tests -n0            # --extra dev --extra mirage
1153 passed, 22 skipped in 40.58s

116 files already formatted

$ uv run ruff check .
All checks passed!
```

CI will run with `.[dev]` and so will skip the test this PR fixes; the run above is the one that exercises it.


khj809:
The substance checks out. I pulled the 0.0.5 wheel: `CALENDAR_API_BASE` and `FORMS_API_BASE` are exactly the two constants it added, `calendar_base`/`forms_base` read them as `_client` globals so the existing sweep reaches them, and `docs_base`, `slides_base` and `forms_base` do all compose `{base}/v1` under `api_base`, so the single-host argument holds as written. `GoogleConfig.api_base`'s comment says what you quote. `~=0.0.5` is the only pin in the repo and `uv.lock` is untracked, so nothing else is owed. Leaving the CI arming to its own decision is right.

What's left is documentation consistency, all of it inside the file:

- **`mirage.py:8`** — the module docstring still opens with "Google and GitHub do not expose one: they read the host from module-level constants". That is the claim this PR corrects sixty lines down, so the file now says both things, and the wrong one is what a reader meets first.
- **`mirage.py:88`** — `DRIVE_UPLOAD_BASE`'s comment still calls itself "the only constant here whose vendor value would be reached by a caller who believed the whole client was redirected", while the two constants added below say they are there "for the reason `DRIVE_UPLOAD_BASE` is". Whoever trusts the older line reads the new pair as pointless and drops them.
- **`mirage.py:99`** — the value is right, the reason isn't reachable on this path: with `FORMS_API_BASE = f"{base}/v1"` a Forms call is `{base}/v1/forms/{id}`, and the mock serves docs at `/docs/v1/documents/{document_id}` — a 404 either way, not the docs route. The docs/forms collision is the single-`api_base` failure the docstring above already rules out.

Same stale claim also survives in `examples/using-mirage/README.md:73` and the headers of `gmail.py`, `gdrive.py` and `unified.py`, plus README:83's list of what gets redirected, which is now nine constants. Fine as its own change if you'd rather keep this one to the shim.

Nothing else from me — with those sentences fixed this is good to go.

khj809 (approved):

PR #67, opened 2026-08-19
A corpus must state its authors, containers and clocks

khj809:
A BYO corpus could omit facts the served resource cannot exist without — an author, a container, a document's own clock — and nothing failed. Three layers covered for each other: the schema did not ask, the loader invented a value, and the router defaulted whatever was still NULL. A three-field Slack record was served from `unknown@example.com`, in a channel called `slack`, at a timestamp hashed from its `doc_id`, with **zero users in the roster** — which emptied `users.list` and made the `--token` ACL path every example documents impossible to run.

Now a record states those facts, and every producer in the repo states them.

```
/_mock/users        0 users  ->  2
users.list                []  ->  ava@acme.com, bob@acme.com
a thread's replies  one speaker at a hashed second
                              ->  two speakers, at 18:00:00 / 18:00:40 / 18:06:00 as written
```

## The contract

Every source requires `source_type`, `content`, its container (`channel` / `mailbox` / `folder` / `repo` / `project` / `space` / `teamspace` / `bucket` / `object_type` / `team`), `author_email` and `created`. `title` too, except Slack (a message has none) and HubSpot (a note has no name — and HubSpot never serves one). A child row states its own author and its own second. And the requirement is on the VALUE, not just the key: `minLength: 1` on every container, every clock, `status`, `issuetype` and `state`, because `"created": ""` used to satisfy the type and then die at the insert as `IntegrityError: NOT NULL constraint failed`.

Per source, whatever the vendor always reports: jira's `status` and `issuetype`; linear's `state`; drive's and notion's `updated`; hubspot's `properties`; a fireflies meeting's `duration`; a github issue's or pull's `state`.

Every field made required is missing from **0.00–0.23%** of the 600,581-record artifact this branch started from. What that artifact shows as 82–100% absent stays optional — `mime_type` (the `subtype` carries it), `notion.properties`, `s3.updated`, `confluence.version_number`, and the ids the importer assigns.

Three rules follow the vendor rather than the measurement:

- **A gmail Subject may be empty, and empty means absent** — RFC 5322 §3.6 gives `to` and `subject` the same 0..1 bounds, and real Gmail returns the headers a message has. The router used to invent a `To` from the mailbox name, so this mock could never reproduce a Bcc-only send, and served `Subject: ""` for a message stating none. Both headers are omitted now, as `Cc` already was.
- **A jira issue need not name a reporter.** Atlassian's own words: "the reporter field is not required by default, and users can leave it empty" — and the REST API omits the field outright for an anonymous reporter (JRACLOUD-67358). It was redundant here too, since `_jira_issue` serves `reporter_email or author_email` and the author is required.
- **A linear `identifier` stays derived.** Linear declares it `String!`, but the mock reconstructs Linear's own `TEAM-number` form and always serves one, so nothing diverges — and requiring it would delete the affordance that lets a corpus name a team without inventing Linear's key space.
- **A fireflies sentence states its `start_time`** — a rule this mock imposes, not one Fireflies states: the live SDL declares `Sentence.start_time` a nullable `Float` (#68). Every sentence a real transcript serves carries a reading, and the derivation that filled a gap was a speaking-rate estimate — a fabrication about when somebody spoke, unlike the identifier's faithful reconstruction. The type refuses `null` as well as absence, or the derivation stays reachable through it.

A refusal names the field and the condition that put the rule in force. `_when_clause` read `const` and `enum` predicates only, so the two rules stated as negations reported bare; the either-or child rows reported jsonschema's `is not valid under any of the given schemas`, which names neither alternative — the string #47 existed to remove.

```
'state' is a required property         ->  ... when subtype is not "file" (or absent)
'author_email' is a required property  ->  ... when host_email is absent
{'start_time': 0} is not valid ...     ->  'text' or 'content' is a required property
```

## Most of the missing values were already in the data

The ERB importer was discarding them. Four parsers each read the same person reference differently; one recognizer now reads every form the corpus writes (`Name (Role)`, `Role (Name)`, `(Name, Role)`, leading table noise). One comment-line parser serves jira and linear alike, and peels a stamp's clock only when a label follows it:

```
before   author "14",  body "05 - Aisha Patel (Support): Acknowledged ticket…"
after    author Aisha Patel, role Support, 2026-03-13 14:05, body "Acknowledged ticket…"
```

| | before | after |
|---|---|---|
| jira comments parsed | 25,319 | **35,732** |
| jira comment bodies carrying the stamp as text | 15,572 | **0** |
| jira comment authors read | 655 | **28,829** |
| linear comment authors read | 83,124 | **135,204** |
| gmail child senders missing | 52,211 | **10,981** |
| gmail root senders replaced by the mailbox owner | 1,961 | **0** |

`comments` given as one blank-line-separated blob is split rather than iterated character by character, which silently discarded every comment in 42 jira docs.

Where the bench genuinely names nobody, ERB says so: a desk label becomes the desk, a customer lands on the counterparty's domain, a label naming an activity ("Design review") is not a person at all and its passage keeps the label, and a row nobody signed carries an explicit `unknown@<org_domain>`. The roster is the only thing that grants a bearer token: the 167 employees of the directory authenticate, and nobody else does — a Gmail header stating a real address makes that person a principal, not an account. A name the roster does not hold but that reads as a person is still a principal, owning documents and authoring comments with no way to sign in, because the bench names people outside the directory and attribution is not authentication. A string that does not read as a person at all becomes no principal: a name is capitalised, so a lowercase token says the field caught the prose around it ("comment by Aisha Patel", "Priya Mehta   on"), and a desk word anywhere in it says a desk ("Acme NetOps", "FinNova SRE"). 893 of 14,979 such names passed only for being two capitalised tokens.

Reading the reference itself got four more fixes. Each was found by re-exporting the artifact and diffing it by record id — a sample would have missed every one — and three of them are defects the rule above introduced:

- **`Marta Kovač` beside `Marta Kovac`** — one canonical, two addresses. `_slug` dropped the accented letter where `canonical` folds it, so whichever spelling the corpus stated first named the person, and one of the two answers is short a letter. Folding also lands `Tomáš Novák` on the address the directory already holds for them.
- **`(Launch day, Sean Gallagher)`** — the comma branch read only its head, so a name written second was nobody, though the parenthetical and the dash already read both orders.
- **`: Priya Shah - Eng feedback:`** — a comment whose own label is empty leaves its colon at the head of the name, and a name beside a colon is not a name.
- **`NoteWave (Customer, Maria Chen)`, `Maya Chen (Customer)`, `Customer - Ravi Menon`** — a customer's staff read as this org's, which the directory collides with: it holds a Maya Chen, an Aisha Patel and an Emily Zhao of its own. A side that is nothing BUT the counterparty marker gives the name beside it away; a side that merely contains the word does not — `Customer Success - Aisha Patel` is a desk of ours and "(Priya Shah, Support) — Updated customer" is one of ours saying what she did, and reading either as the counterparty put 13 employees on a customer's domain. 796 rows were being served as employees.
- **`Support Maya Chen`, `by Marcus Li`, `PM Priya Rao`** — a desk word anywhere in a label disqualifies the whole of it, and the bench writes the desk and the person as one run with nothing to split on, so 6,600 labels naming somebody resolved to nobody. The runs INSIDE a label are scanned now, longest first, or a four-token name would lose to its own prefix.
- **`[Support - Marisol Rivera 2026-03-13 09:42 UTC] …`** — a bracket closes a label by itself, and the bench writes three of these: the stamp inside the bracket with the name, the stamp in a bracket of its own ahead of it, and the name after the bracket rather than in it. None was read, so 97 lines became body with no author at all and 26 more gave up `[Support` as the label. A bracket closes the label only when it names somebody itself, or the third shape loses its name to the body.

After them the diff holds no row where a person changed hands wrongly. 12,492 comment attributions move: 8,056 from a synthesized desk address to a person the corpus names or to the counterparty's own domain, 3,553 to an explicit `unknown@` with the label kept in the passage, and 874 from unattributed to somebody the corpus names. Every row that crosses the org boundary crosses it the right way — 23 of the counterparty's staff off this domain, two of our own back onto it — and the roster holds no heading, no phrase and no twin of a real name.

The counterparty rule was checked over the labels themselves rather than by sampling: of the 55,400 distinct labels the bench writes across its 197,300 comments it changes 583, and every one of those states "customer" or "client" outright.

One label the diff leaves standing: "Customer Success (internal note)" is this org's own desk, and the counterparty test reads its name as the counterparty, so it lands on a synthesized outside domain. It did that before this branch too, and closing it means teaching that test which "customer" is a department here — left alone rather than guessed at.

Where the bench states a fact in a shape the importer could not read, it reads it now. Four such
shapes turned up only once the whole artifact was re-exported and read back — a sample had missed
every one:

| the shape | what was lost | rows |
|---|---|---|
| `IT (Jordan Liu) 2026-03-10 15:20 - body` | the stamp follows the label, so the label ran to the clock's colon and the minutes stayed at the head of the body | 192 |
| `Security (2026-03-12): body` | a stamp where a name goes became part of an address (`security.20260312@`) and the date was lost with it | 3,068 |
| `BrightCart (Customer): body` | an external commenter served as an address inside the org's own domain | 2,675 |
| a jira issue whose reporter the bench never names | the only record left with no author | 1 |

**Verified by re-exporting the whole artifact and validating every record: 581,294 records, no
contract gaps.** Those records serve 511,961 documents — a slack thread is one record and many
messages, a github path many snapshots — and 14,132 users where an unattributed corpus served
none — 13,965 of them attributed but tokenless, the 167 in the directory able to sign in. The served result carries none of the old defects: no jira comment body opens with a stamp
fragment (15,572 did), no jira or linear comment lacks an author (24,333 and 94,912 did), and
`state` is `open` or `closed` on every pull request.

The table above counts what the converter WRITES; the DB holds 35,724 of those jira comments,
because one jira pair shares a `doc_id` (8 comments against 5) and a child is keyed on its
parent's, so the later record's five overwrite and `_trim_children` trims the tail — the documented
behaviour for a repeated id, not a loss.

## The loader stops covering for the corpus

`backlot/importer/byo.py` loses 129 lines (54 added, 183 removed). The author, container and clock are read straight off the record. `_thread_seconds` is what remains once both ends of a thread state a clock: a check that each reply follows the message before it, since a Slack ts is identity as well as a time. The checks the schemas now make are gone from here; what stays is what a single-record schema cannot state — github's field pairings, the thread's ordering, the cross-corpus reference passes.

`validate=False` is deleted, so ERB's output passes the same gate as a hand-written corpus. Measured at 90 s over 600,581 records, against an import whose FTS build alone takes longer.

Two deletions worth naming, since a reader will look for what replaced them: `_report_unclocked` reported only paths whose every snapshot left `created` unstated, which the requirement now refuses outright — and #63's other half, which refuses two snapshots sharing one `created`, is untouched. #53's re-grounding of a clockless root on its replies cannot fire at all once a root states its own clock.

The routers serve what the row holds. jira's `status` and `issuetype`, github's `state`, linear's workflow state: the substitutes behind them were unreachable, and an unreachable default hides the next regression.

## Eight vendor divergences closed on the way

- **github `repo`** — `repo` occupies the API's bare `name` slot; the owner is a path segment of its own, supplied from the org this mock serves, and `full_name`, `html_url` and every URL template are composed from the two. An owner-qualified value was accepted, stored, and served as `name` verbatim with `full_name` doubled up — `acme/acme/gateway` — while `/repos/{owner}/{repo}` could no longer find the row. GitHub's own rule for a repository name is now the constraint — the charset and the 100-character limit, the two halves of the sentence GitHub publishes — so the record is refused where it is read rather than at serving time.
- **s3 `bucket`** — a bucket is the first half of a served object's identity and the host label in the virtual-hosted URL S3 hands back, so a name S3 would refuse cannot be served either: a slash splits the address, an uppercase letter or an underscore cannot appear in a host label, a one-character name is below the three S3 requires, and S3 refuses both adjacent periods (`logs..2026`) and a name shaped like an IP address. A dot itself S3 allows — at the cost of virtual-hosted HTTPS — so this does too.
- **github `state`** — the artifact held `state: "merged"` on 8,047 pull requests and the router served it verbatim. GitHub returns only `open` or `closed`; the merge moves to `merged_at`, which this mapper had been dropping entirely. An `enum` makes the old value unrepresentable.
- **fireflies empty sentences** — a `{}` sentence validated and became a stored row that the transcript's own `content` does not contain, breaking the inverse those two are defined by. 20 such sentences shipped in the artifact.
- **a mailbox found under one spelling only** — `users/me/messages` is the one listing whose container comes from the caller rather than from the request path, and it derived that container by slugifying the caller's local part. A corpus that states the mailbox as the owner's ADDRESS — the obvious thing to state now that `mailbox` is required — served an empty listing over a mailbox holding 48 messages, `messagesTotal: 0`, and a `Delivered-To` of `someone@example.com@example.com`. The address now resolves against the mailboxes the corpus stated.
- **a thread listing that was two listings** — `threads.list` counted the threads the caller had WRITTEN in, so a mailbox owner was served none of the mail they received, while `threads.list?q=` filtered their MAILBOX and passed the caller's address where a mailbox name goes. Measured on a 26,996-message mailbox: 1,475 threads without a query and 0 with one, against 5,878 and 881 now — each thread listed once, whichever of its messages matched, and `messagesTotal` stops standing in for `threadsTotal`, since a thread of five messages is one thread. `in:` joins the operators it belongs with: unhandled it fell through to free text and FTS-matched nothing, so `in:anywhere` — which asks for MORE mail than a plain search — emptied the mailbox, and `label:inbox` missed every message whose labels the corpus leaves to the default it is served under.
- **a hubspot object type addressed in one number only** — `/crm/v3/objects/deals` served nothing for a corpus that states `object_type: "deal"`; those deals were reachable only under `/objects/deal`, as a custom object, while the vendor's own path returned the empty page a standard type gets in a portal that holds none. Measured against api.hubapi.com with a key that authenticates but holds no CRM scopes — which is what makes the question answerable, since HubSpot resolves the type BEFORE authorizing the call: every standard object answers to both spellings (403 MISSING_SCOPES either way), and an unrecognized word answers **400** `{"status": "error", "message": "Unable to infer object type from: nonsuch"}` — no `category` — where this mock answered 404. Both spellings now reach the same records — listing, search, batch read, get-one and associations alike, and every spelling of one type at once, so a corpus that states both does not serve half its records under each. An **objectTypeId** is a third spelling: `/crm/v3/objects/0-3` IS deals, and the thirteen published ids for the types this mock enumerates each answer 403 like their names do, while an id no standard object holds (`0-6`, `0-41`, `2-12345`) is 400 `Invalid object or event type id`.

- **a hubspot standard type the mock had never heard of** — `/crm/v3/objects/invoices` answered 400 `Unable to infer object type from: invoices`, the message for a typo, because the standard-type table held the CRM's original thirteen names. Asked of api.hubapi.com on 2026-08-24 with the same scope-less key — type resolution answers before the scope check does — twenty three more names resolve exactly as `deals` does: the commerce set (orders, carts, subscriptions, commerce_payments, payments, invoices, quote_templates, discounts, fees, taxes), plus deal_splits, leads, appointments, courses, listings, services, users, goal_targets, marketing_events, campaigns, communications, postal_mail and partner_clients. `audit_logs`, `objects` and `tickets_pipelines` answer 400 and are absent for that reason. A type the vendor resolves has to answer an empty page rather than an error: a corpus holding no invoices is not the claim that HubSpot has no such type, and the 400 exists to tell a caller its type is a typo. Eighteen of the new singulars resolve and are mapped, while `appointment`, `course`, `listing` and `payment` answer 400 where their plurals answer 403 — the singular is not a spelling every standard type answers to. The id space is wider still, `0-1` through `0-165` resolving apart from eleven gaps, but a scope-less key cannot say which type an unpublished id names, so only the thirteen published pairings are mapped and an unmapped id stays a 400 rather than resolving to a guess.

## Corpora, tests, docs

`backlot/data/hello.jsonl` — the corpus `backlot serve` runs with no arguments — gains the child seconds, modified times and CRM properties it was leaving to the loader. The reference corpus, the 34 inline example corpora and the TypeScript one all state their facts, and a test holds each of them, the README snippets included: a snippet is the first corpus anybody writes.

`tests._helpers.complete` reads the required list off the schemas, so a test states only the fields it is about and a helper cannot drift from the contract it stands in for. `complete(..., _omit={"path"})` is how a test whose subject IS one missing field says so. The tests about the contract, and about refusal, still build their records by hand.

Seven tests are retired rather than rewritten. Each pinned a defaulting path that no longer exists — a clockless root re-grounded on its replies, an undated comment taking the document's clock plus a position, a reply landing on the next free second.

The bench e2e is retired too, for its own reason: it fetched the corpus, imported all 581,294 records and asserted one Drive file's owner — one bit of coverage for ~4.6GB of raw and an 11.7GB database per revision, behind an env var that kept it out of CI. What it guarded, that owner resolution is a property of the whole corpus rather than of a subset, is guarded better by re-exporting the artifact and diffing it by record id: that reports every changed owner instead of the one a single assertion expects.

## Verifying

```bash
uv sync --all-extras && uv run pytest -q
```

Green. Also checked from an installed wheel, run from a directory the checkout cannot shadow: 11 schemas ship, slack's required list is `author_email / channel / content / created / source_type`, and the demo corpus validates against its own contract.


backlot/importer/byo.py
```diff
@@ -1819,11 +1760,7 @@ def register(email: str | None, name: str | None = None) -> None:
                 extras[k] = rec[k]
         subtype = rec.get("subtype")
         parent_id = rec.get("parent")
-        # created_ts must never be NULL (the server sorts/filters by it; a NULL would need a
-        # runtime null-check). Fall back to the same deterministic synth.epoch the server would have
-        # synthesized for a missing ts, so the served time is unchanged — just materialized now.
-        created_written = _epoch_field(rec.get("created"), where, "created")
-        created = synth.epoch(doc_id) if created_written is None else created_written
+        created = _epoch_field(rec["created"], where, "created")
```
nuri-yoo, line 1759: `""` passes the schema (`created` is `type: ["integer","number","string"]` with no `minLength`) and `_epoch_field` answers `None` for it, because `_time_given` classifies the empty string as "the corpus did not write a time". That was correct while `synth.epoch(doc_id)` stood behind it. With the fallback gone, `None` reaches the insert:

```
$ echo '{"source_type":"slack","channel":"general","content":"hello",
         "author_email":"ava@acme.com","created":""}' > a.jsonl
$ backlot import a.jsonl
IntegrityError: NOT NULL constraint failed: slack_messages.created_ts
```

With a reply it dies earlier, in `_thread_seconds`:

```
TypeError: '<=' not supported between instances of 'int' and 'NoneType'   (byo.py:285)
```

The same path exists for `c["created_ts"]` (line 2215) and `rep["created"]` (line 284). This is the rule #53 set -- "a field the author filled in with something this importer cannot read is a typo to report, not a default to take" -- and `""` is now exactly that, since the key is required. `_time_given` cannot simply start rejecting `""` (fields whose absence stores NULL still rely on it), so the narrow fix is in `_epoch_field`: an empty string is unreadable rather than unwritten when the caller is reading a required field. The existing test at `tests/test_importer_byo.py:633` covers unreadable non-empty values only.

A raw `IntegrityError` for a corpus-level mistake is also the class we fixed in #65 (`NOT NULL constraint failed: slack_messages.title` on `--append`).

backlot/schemas/fireflies.schema.json
```diff
@@ -74,7 +74,23 @@
...
+          },
+          {
+            "required": [
+              "content"
+            ]
+          }
+        ],
+        "description": "A transcript utterance. Its text is required: `content` IS the sentence concatenation, so a sentence carrying none becomes a stored row the transcript's own text does not contain, and the inverse the two are defined by no longer holds. `start_time` is required too, because every sentence the API serves carries one.",
+        "required": [
+          "start_time"
```
nuri-yoo, line 91: `start_time` is required here while its declaration at line 60 still reads `"type": ["integer","number","null"]` and still documents a default ("Default: the running clock, so the sequence stays ordered"). So the requirement is shape-deep:

```
{"text":"hi","start_time":null}          -> record_errors: []
```

and the loader fills it through `synth.fireflies_fill_times` at `byo.py:1592`, which is the speaking-rate derivation this description calls "a fabrication about when somebody spoke":

```
sqlite> select seq, start_time, end_time, body from fireflies_sentences;
1|0.0|0.5|hello
2|0.5|1.0|there
```

Dropping `"null"` from the type and rewriting that description closes it. The `minLength: 1` treatment `title` and `content` got in this branch is the same move.

Worth a word on the justification too. The live schema declares `Sentence.start_time` as `Float`, nullable, so "every sentence the API serves carries one" is a property of the responses rather than of the contract. That is a fine basis for a constraint this mock imposes; it reads differently when written as the vendor's rule, which is what #60 had to unwind for `real Linear's 1-5 character key`.

Ref: #68, which introspects `api.fireflies.ai` and records `Sentence.start_time` / `end_time` / `speaker_id` as `Float` / `Float` / `Int`. Not `docs.fireflies.ai`, which does not match the API.

backlot/schemas/fireflies.schema.json
```diff
@@ -74,7 +74,23 @@
             "description": "Seconds from the start of the meeting. Default: the next sentence's start, or a speaking-rate estimate for the last one."
           }
         },
-        "additionalProperties": false
+        "additionalProperties": false,
+        "anyOf": [
```
nuri-yoo, line 77: This `anyOf` and the one at `linear.schema.json:196` report the error the comment removed from `byo.py` in this same branch named as useless -- "whose error ('is not valid under any of the given schemas') names neither field and so tells the author nothing":

```
fireflies, sentence with no text
    t [sentences/0]: {'start_time': 0} is not valid under any of the given schemas
linear, comment with neither content nor body
    t [comments/0]: {'author_email': 'a@b.com', 'created_ts': 1} is not valid under any of
                    the given schemas
```

The record-level constraint you added ten lines below does the right thing with the same requirement, and its message names a field:

```
fireflies, neither content nor sentences
    t: 'content' is a required property
```

`allOf` + `if: {"not": {"required": ["text"]}}` / `then: {"required": ["content"]}` gives the child rows the same treatment, at the cost of naming `content` rather than `text` as the missing one. Whatever the shape, the string above is the one #47 existed to remove.

backlot/schemas/github.schema.json
```diff
@@ -277,6 +289,22 @@
           "path"
         ]
       }
+    },
+    {
+      "if": {
+        "properties": {
+          "subtype": {
+            "not": {
```
nuri-yoo, line 300: `_when_clause` renders `const` and `enum` predicates only (`validation.py:127-131`); a `{"not": {"const": ...}}` falls through to `continue`, so this rule reports with no condition attached:

```
github issue with no state
    t: 'state' is a required property
```

That is the line #47 was written about -- "dozens of identical `<root>: 'path' is a required property` lines, none of which said that `path` is required of source files and of nothing else." The two fireflies conditionals are the same: their `if` carries no `properties` at all, so a record that states neither `host_email` nor `author_email` is told only that `author_email` is required, and never learns that `host_email` would satisfy the schema.

Rewriting this as `if: {properties: {subtype: {const: "file"}}}, else: {required: ["state"]}` is not equivalent -- your own note in `_when_clause` explains why: `properties` alone does not demand the field, so an `else` would lift the rule for a record that states no `subtype`. Teaching `_when_clause` the negated form instead (`when subtype is not "file" (or absent)`) keeps the semantics and restores the clause.

backlot/schemas/jira.schema.json
```diff
@@ -213,8 +218,14 @@
     }
   },
   "required": [
-    "source_type",
+    "author_email",
     "content",
+    "created",
+    "issuetype",
+    "project",
+    "reporter",
```
nuri-yoo: Jira documents the opposite: "The reporter field is not required by default, and users can leave it empty, which leads to the appearance of the '?' Anonymous." JRACLOUD-67358 goes further -- for issues whose reporter is anonymous the REST API omits the field. So `reporter` does not belong in the "whatever its vendor always reports" set that `status` and `issuetype` do.

It is also redundant here. `_jira_issue` resolves `reporter = _jira_actor(row["reporter_email"] or row["author_email"], site)` (`routers/atlassian.py:619`), and `author_email` is required by this branch, so a record that omits `reporter` already serves a reporter. And the property's own description at line 133 still says "Default: the author." -- a default documented for a field that can no longer be omitted.

Dropping it from `required` keeps the vendor, the router and that sentence in agreement.

Ref: https://support.atlassian.com/jira/kb/issues-created-by-anonymous-user-in-jira/
Ref: https://jira.atlassian.com/browse/JRACLOUD-67358

backlot/schemas/s3.schema.json
```diff
@@ -28,7 +29,8 @@
     },
     "bucket": {
       "type": "string",
-      "description": "The bucket this object belongs to (default: the source_type)."
+      "pattern": "^[a-z0-9][a-z0-9.-]{1,61}[a-z0-9]$",
```
nuri-yoo: The three rules the description enumerates are transcribed exactly. Two more from the same AWS list are not, and the pattern accepts both:

```
{"bucket": "a..b"}          -> record_errors: []
{"bucket": "192.168.1.1"}   -> record_errors: []
```

AWS: "Bucket names must not contain two adjacent periods." / "Bucket names must not be formatted as an IP address (for example, 192.168.5.4)." (There are also the reserved `xn--` / `sthree-` / `amzn-s3-demo-` prefixes and the `-s3alias` / `--ol-s3` / `.mrap` / `--x-s3` / `--table-s3` suffixes.) Since the stated reason for constraining this field at all is that "a name S3 would refuse cannot be served either", either the pattern takes the two cheap ones -- `(?!.*\.\.)` and an IP guard -- or the sentence narrows to the charset it actually enforces. The test cases at `tests/test_schema.py:355` are worth extending with the same two.

One aside on the reasoning rather than the rule: AWS's own guidance is that a period makes virtual-host-style HTTPS unusable ("you can't use virtual-host-style addressing over HTTPS, unless you perform your own certificate validation"), so the host-label argument in the description cuts against allowing dots -- which S3 nonetheless does, so the pattern is right and the sentence is what overreaches.

Ref: https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html

backlot/schemas/github.schema.json
```diff
@@ -15,15 +15,17 @@
...
+      "minLength": 1
     },
     "doc_id": {
       "type": "string",
       "description": "Stable id (default: dsid_<sha256(source_type+title+content)>)."
     },
     "repo": {
       "type": "string",
-      "description": "The repo this document belongs to (default: the source_type)."
+      "pattern": "^[A-Za-z0-9._-]+$",
```
nuri-yoo, line 27: "GitHub's own charset for a repository name is now the constraint" is accurate to the word -- GitHub states it as "The repository name must not exceed 100 characters, and can only contain ASCII letters, digits, and the characters `.`, `-`, and `_`." The pattern implements the second half of that sentence; `maxLength: 100` would finish it.

Two further restrictions are real but undocumented by GitHub (a repository named `.` or `..` is reserved, and a name cannot end with `.git`), so if you add them, they want a different framing than the charset -- GitHub's docs issue #44518 is them acknowledging the limits are unpublished.

Ref: https://docs.github.com/en/repositories/creating-and-managing-repositories/creating-a-new-repository

backlot/routers/google.py
```diff
@@ -676,16 +724,19 @@ def _gmail_message(row, fmt: str, caller_email: str | None = None) -> dict:
     headers = [
         {
             "name": "Delivered-To",
-            "value": row["to_addr"] or f"{row['mailbox']}@{get_settings().org_domain}",
+            "value": row["to_addr"] or _mailbox_address(row["mailbox"]),
         },
         {"name": "MIME-Version", "value": "1.0"},
         {"name": "Subject", "value": row["title"]},
```
nuri-yoo: `To` moved into `optional` on the grounds that RFC 5322 allows a message with no destination field. The same table in section 3.6 gives `subject` the same bounds:

```
| to      | 0 | 1 |
| subject | 0 | 1 |
```

`Subject` is still in the mandatory list here, so a corpus that states the absence of a subject -- which `gmail.schema.json` now explicitly invites, "Required but MAY be empty" -- is served a header that real Gmail would not include:

```
[('Delivered-To','ava@acme.com'), ('MIME-Version','1.0'), ('Subject',''),
 ('From','Ava <ava@acme.com>'), ...]
```

Moving `("Subject", "title")` into `optional` beside `To` makes the two headers consistent with each other and with the clause the schema cites. If instead the empty header is deliberate, the schema sentence "the API serves it as one" is the thing to fix, since an absent header and an empty one are different responses.

Ref: https://www.rfc-editor.org/rfc/rfc5322#section-3.6

backlot/routers/hubspot.py
```diff
@@ -228,14 +228,68 @@ def _keep(raw) -> list[str] | None:
...
+
+
+def _aliases(object_type: str) -> list[str]:
+    """Every spelling of this object type — what was asked for, then the standard object's other
+    spelling. Order matters: a corpus that states both keeps the one the caller named."""
+    other = _SINGULAR.get(object_type) or _PLURAL.get(object_type)
+    return [object_type] if other is None else [object_type, other]
+
+
+def _resolve_type(request: Request, object_type: str) -> str | None:
```
nuri-yoo: `get_object`, `batch_read` and `list_associations` accept either spelling for any row (`row["object_type"] not in _aliases(...)`), while `list_objects` and `search_objects` resolve to a single stored spelling. For a corpus that states both, the two halves disagree:

```
corpus: one record with object_type "deal", one with "deals"

list   /objects/deal    -> ['4331617102']  ['Acme renewal']
list   /objects/deals   -> ['2013641391']  ['Beta']
get    /objects/deal/4331617102   -> 200      get /objects/deal/2013641391  -> 200
get    /objects/deals/4331617102  -> 200      get /objects/deals/2013641391 -> 200
search /objects/deal    total=1              search /objects/deals total=1
```

A portal has one `deals` type, so no HubSpot response can show a record under `/objects/deal/{id}` that `/objects/deal` never lists. Either the two list paths query the whole alias set, or the importer canonicalises a standard type's container so both spellings land in one place -- which also makes the description's "Both spellings now reach the same records, listing, search, batch read, get-one and associations alike" true as written.

Two smaller things in the same function:

- `0-3` is a documented address for deals ("you can use object endpoints and replace the `{objectTypeId}` in the request URL with the desired object"), and your own measurement note says a recognized type answers 403. `_TYPE_ID` sends it to the malformed-id message instead. Serving the 403-equivalent for the thirteen ids you already enumerate would match both.
- A custom type is still reachable by bare base name, which HubSpot removed in June 2025: supported forms are `p_{object_name}`, the objectTypeId and the fully qualified name, and a base name now answers `Unable to infer object type from: vendor`. Pre-existing rather than introduced here, but this is the function that decides it.

Ref: https://developers.hubspot.com/docs/guides/api/crm/understanding-the-crm
Ref: https://developers.hubspot.com/changelog/breaking-change-removed-support-for-referencing-custom-object-types-by-base-name

backlot/importer/erb.py
```diff
@@ -1572,33 +1907,56 @@ def _byo_jira(dsid, raw, P):
...
+    for seq, c in enumerate(parsed_comments, start=1):
+        # A name-shaped label that names an ACTIVITY ("Design review") is not a person, so its
+        # passage keeps the label and the comment is attributed by what the label does name.
+        named = c["person"] if c["person"] and not _names_an_activity(c["person"]) else None
+        comments.append(
+            _rec(
+                id=f"{dsid}::c{seq}",
+                content=(c["body"] if named else c["body_with_label"]),
+                author_email=(
+                    P.resolve(named, role="author")
```
nuri-yoo: `_person_like` is the only gate between a comment label and a principal, and it accepts two-token headings that `_names_an_activity` does not list. On the full bench import (`Loaded 511961 documents`), 14,979 of the 15,146 user principals are minted rather than from the directory, and all 14,979 arrived through this branch -- none through the desk/label path -- so the denylist is carrying the whole load. What comes out includes:

```
Root cause | Action items | Next steps | Slack refs | Slack thread | Log excerpt
Not assigned | To be assigned | No CSM yet | No SE assigned | Mitigation applied
Resolution planned | Provided details | internal comment | reporting user | PR author
Zendesk import | Policy engine | for completeness | by email | by ticket
```

697 comment rows (44 jira, 653 linear) are attributed to that set, and because `named` is truthy the passage loses its label:

```
action.items@  'Priya to add unit tests; SRE to add alert; Support to deliver final RCA...'
slack.refs@    '#inc-hightower, #support, #oncall. Confluence: Hightower Postmortem page...'
next.steps@    'finalize PD runbook, validate SCIM mapping job, publish customer notice...'
```

Two further shapes worth naming:

- `comment by Aisha Patel` -> `comment.by.aisha.patel@` (3 rows, plus 9 in linear). The label names a person and the whole label became the author, so a named person's comment attaches to a phantom instead of to them.
- `'Priya Mehta   on'` and `'Assignee   at'` are registered principals. The double space is a peeled stamp: `Priya Mehta on 2026-03-12` loses the date and keeps the preposition, then passes `_person_like` on three tokens. `priya.mehta.on@` is a phantom twin of a real name.

To be clear about the blast radius: none of these get a token (`tokens.yaml` holds exactly the 167 directory people), and 697 of ~201k comment rows is 0.3%. What they do is appear in `/_mock/users` and `users.list` as people, and take text out of the bodies they head.

A denylist cannot close this -- the set of headings a bench writes is open. Inverting the test would: attribute only when the label resolves to a name the roster already holds (the treatment `_byo_linear` gives via `display_email`), and let everything else fall to `label_email`, which keeps `body_with_label`. That is also the rule the description states -- "A name the roster holds becomes that principal; a name it does not becomes a display-only address" -- which the jira path does not currently follow, since `P.resolve` mints. If that is too large for this branch, the scoped version is to require the label to resolve before `named` is set, and leave the rest to a follow-up.

backlot/importer/erb.py
```diff
@@ -1515,41 +1833,58 @@ def _byo_drive(dsid, raw, P):
...
         content=content,
-        author_email=author_email,
+        author_email=(author_email or P.unattributed()),
         author_name=author,
-        subtype=("pull_request" if raw.get("pr_number") else "issue"),
-        state=raw.get("state"),
+        subtype=subtype,
+        # Open unless the bench says otherwise: GitHub reports a state on every issue and
+        # pull, and a file row carries none at all.
+        state=(None if subtype == "file" else _GH_STATE.get(raw_state, "open") or "open"),
```
nuri-yoo: `subtype` is computed four lines above as `"pull_request" if raw.get("pr_number") else "issue"`, so `subtype == "file"` cannot hold here and the `None` branch is unreachable. If ERB is expected to grow file rows later, a comment saying so would keep it; otherwise `state=_GH_STATE.get(raw_state, "open")` says the whole rule. (The trailing `or "open"` is also redundant -- every value in `_GH_STATE` is truthy.)

backlot/importer/erb.py
```diff
@@ -1852,13 +2252,15 @@ def _byo_fireflies(dsid, raw, P):
...
-    # `content` is omitted for the same reason — it is DEFINED as the sentence concatenation, so
-    # emitting it would double the artifact's largest field and could drift.
+    # Parsed here, then timed here: every sentence the API serves carries a `start_time`, so the
+    # corpus states one. `synth.fireflies_fill_times` spreads a run of sentences sharing one clock
+    # reading across its window, which is the same rule the server would apply -- applying it once,
+    # at conversion, is what makes the reading a fact of the corpus rather than of the loader.
+    # `content` is still omitted: it is DEFINED as the sentence concatenation, so emitting it would
+    # double the artifact's largest field and could drift from the sentences it restates.
     sentences = parse_fireflies_transcript(_ff_transcript_text(raw), internal + external)
+    _ff_seconds = _ff_duration(raw.get("duration_minutes"))
```
nuri-yoo: `_ff_duration` returns minutes, so `_ff_seconds` holds minutes and the `* 60` on the next line is what makes seconds. The same call is repeated at line 2284 for `duration`. `ff_minutes`, computed once and used in both places, matches `byo.py:1592`'s spelling of the same value.

nuri-yoo (commented):
Reviewed at `7cdc9da`. `uv sync --all-extras`, suite `1317 passed, 3 skipped` (ERB_E2E opt-in, no Docker, `llama_index.readers.hubspot` unavailable here); ruff check and format clean over the 54 touched files.

The whole bench imports on this head: `Loaded 511961 documents`, 0 skipped, exit 0. Every record the converter produces satisfies the new contract at full scale, and the counts agree with the description where they overlap (35,724 jira comments against your 35,732; 15,146 user principals against 15,047). The serving fixes reproduce: a mailbox stated as an address answers `messagesTotal 4 / threadsTotal 2`, `threads.list` returns the mailbox's threads once each with and without `q`, `in:anywhere` and `label:inbox` match where they matched nothing, a message with no `to` serves no `To` header, and `/crm/v3/objects/deal` reaches records stored as `deals`.

The two deletions I went looking for are both justified: `_report_unclocked` (#63) reports only paths whose snapshots all left `created` unstated, and #63's other half still refuses two snapshots sharing one `created`; #53's re-grounding cannot fire without a clockless root. One sentence each in the description would save the next reader the check.

Four things I would want resolved before merge, details inline: a `"created": ""` passes validation and then dies on `IntegrityError` rather than the message `_epoch_field` exists to print, which is the one hole in #53's rule; two new child-row constraints report as `is not valid under any of the given schemas`, the string the comment this branch deletes named as useless and the one #47 removed, and three more conditionals report with no `when` clause; `jira.reporter` in `required` states something Atlassian documents as optional and this mock already resolves to the author; `fireflies.sentences.start_time` is required while its type still admits `null`, so the derivation this branch calls a fabrication stays reachable.

The rest are small: two vendor patterns that stop one rule short of the sentence they cite, a `Subject` header the RFC clause behind `To` treats the same way, a HubSpot alias set that `list` and `get` read differently, ERB attributing 697 comments (0.3%) to labels rather than people, one unreachable branch, one local named for the wrong unit, one container requirement that stops at the key, and three comments that outlived their mechanism.

Three numbers to reconcile: the description gives 600,581 records for the reference artifact and 581,294 for the re-export, and my import reports 511,961 documents; and `byo.py` measures 41 added / 161 deleted, 120 net rather than 153. Per the #65 convention, `erb.py` at 508 changed lines also wants an `ERB_E2E=1` run on both revisions -- I have the head only, so a before/after over the tables this branch means to change is still open.

backlot/importer/byo.py
```diff
@@ -2362,11 +2277,6 @@ def insert(
...
         # in `thread_seq`, which is what `users.messages.list` / `users.threads.get` page over.
         for i, msg in enumerate(messages or [], start=1):
-            # The key is required, its value may be EMPTY: 2.3% of real thread messages are
-            # headers with no body (an auto-ack, a bare forward), and dropping those would drop
-            # messages from the middle of a thread and renumber the rest.
-            if "content" not in msg:
-                raise SystemExit(f"{where}: each gmail message needs 'content'")
             # No fallback to the ROOT's author, unlike a slack reply: a thread's messages have
             # different senders by definition, so attributing an unattributed one to whoever
             # opened the thread would invent a sender. It falls through to `unknown@<org_domain>`.
```
nuri-yoo: Three comments now describe mechanisms this branch removes:

- this line, "It falls through to `unknown@<org_domain>`" -- the fallback was `cols.update(author_email=email or f"unknown@{org_domain}")`, and `author_email` is required on a gmail message now;
- line 2287, "else the root's clock + an hour per position" -- a message's `created` is required, so `_message_second`'s default is unreachable (through the `""` hole I left on line 1763, and nowhere else);
- `_epoch`'s docstring, "Returns None for a missing/unparseable value, so the router falls back to the deterministic synthesized timestamp" -- that fallback is what this branch deletes.

Same class as the four you fixed in #65: each teaches a reader a mechanism that is gone.

backlot/importer/byo.py
```diff
@@ -1665,7 +1606,7 @@ def add(self, rec: dict, where: str = "record") -> None:
         # would keep the earlier document and diverge.
         seen.add((src, doc_id))
         gcol = store.grouping_col(src)
-        container = str(rec.get(gcol) or src)  # channel / mailbox / folder / repo / project / space
+        container = str(rec[gcol])  # channel / mailbox / folder / repo / project / space
```
nuri-yoo, line 1605: The container is required now but may still be empty, and `str(rec[gcol])` takes it verbatim where `rec.get(gcol) or src` used to coerce it:

```
{"source_type":"slack","channel":"","content":"hi","author_email":"a@b.com","created":1}
    -> record_errors: []      -> a container whose name is ""
```

`author_email` is protected by `format: email` (`'' is not a 'email'`), and `title`, `content` and `key` got `minLength: 1` in this branch, so the containers plus `status` / `issuetype` / `state` are the fields where the requirement stops at the key. `minLength: 1` on them makes the contract say what the loader now assumes.

backlot/schemas/linear.schema.json
```diff
@@ -36,7 +36,7 @@
     "identifier": {
```
nuri-yoo, line 37: Not this branch's sentence, but it is in the description you are editing and it is mine, from #64: "its `Team.key` is described as the prefix an identifier carries and nothing more". The SDL reads

```graphql
"""
The team's unique key. Change this at your own risk.
"""
key: String!
```

The conclusion holds -- Linear states no shape for a team key -- but that parenthetical attributes to Linear a description it does not give. Worth trimming to the part the SDL supports while the field is open.

Ref: https://raw.githubusercontent.com/linear/linear/master/packages/sdk/src/schema.graphql

khj809:
All fifteen addressed. Each reproduction re-run against the branch rather than reasoned about, so the notes below are what the code now answers.

| | verified by |
|---|---|
| `created: ""` reaching the insert | all three paths refuse it as a corpus error — `<root> [created]: '' should be non-empty`, `[replies/0/created]`, `[comments/0/created_ts]`. No `IntegrityError`, no `TypeError` |
| fireflies `start_time` nullable + documented default | type is `["integer","number"]`; the description now says the rule is the mock's, cites the live SDL, and drops the derivation |
| `anyOf` naming neither field | `'text' or 'content' is a required property` / `'content' or 'body' is a required property` |
| `_when_clause` skipping negated predicates | `'state' is a required property when subtype is not "file" (or absent)`; `'author_email' … when host_email is absent` |
| jira `reporter` required | dropped from `required`; "Default: the author." is accurate again beside it |
| s3 adjacent periods / IP-shaped | both refused; `tests/test_schema.py` gains `logs..2026`, `192.168.1.1` and `192.168.1.1.eng` as the discriminator |
| github `maxLength` | 100 |
| gmail `Subject` | end-to-end: a message stating `""` is served with no `Subject` header at all; one stating a subject still carries it |
| hubspot list/search vs get-one | one alias set for every path; `0-3` resolves to deals. **The third point is not implemented** — see below |
| `_person_like` minting headings | every name you listed is gone, and so is the class: the roster now holds no heading, no phrase and no twin of a real name |
| unreachable github `file` branch, `or "open"` | both gone |
| `_ff_seconds` holding minutes | `ff_minutes`, computed once |
| three comments describing removed mechanisms | gone, with `_message_second` |
| containers and enums stopping at the key | `minLength: 1` on all eleven containers plus `status` / `issuetype`; `state` carries an `enum` |
| the Linear `Team.key` sentence | trimmed to what the SDL supports |

**The one I did not do.** A custom object reachable by bare base name. Measured against `api.hubapi.com` on 2026-08-24 with the same scope-less key: `orders` and `order` answer 403 — they are standard objects, not custom ones — while `p_orders` and `p12345_orders` answer **400** `Unable to infer object type from: …`. This portal declares no custom object, so the key cannot show what an existing one answers to, and the one testable implication of the changelog reads the other way: `p_<name>` is not a form the API accepts in general. Refusing the bare name on that basis would break a corpus against a client that still uses it, so the comment now records it as unimplemented rather than decided, with the measurement. Reopen if you read the changelog differently — it is the one thread where I am overriding the premise rather than the code.

What the same key did settle: the standard-type table held thirteen names and the API resolves **thirty-six**. `invoices`, `orders`, `subscriptions`, `leads` and twenty more answered 403 exactly as `deals` does, so `/crm/v3/objects/invoices` was returning the 400 meant for a typo. Eighteen of the new singulars resolve; `appointment`, `course`, `listing` and `payment` do not, and are left out.

**On `_person_like`.** You were right that a denylist cannot close it, and inverting the test would have cost more than it bought — 14,158 of the 14,979 minted principals are ordinary person names carrying ~392k attributions, against 893 that are not people. So the test got two rules instead: a name is capitalised, and a desk word anywhere in a label means a desk. Following that through the artifact turned up five more shapes, three of them defects those rules introduced:

- `_slug` dropped an accent where `canonical` folds it, so `Marta Kovač` and `Marta Kovac` shared a canonical but not an address
- the comma branch read only its head, so `(Launch day, Sean Gallagher)` was nobody
- a label of its own being empty left its colon at the head of the name
- a side that is nothing BUT the counterparty marker gives the name beside it away — `Maya Chen (Customer)`, `Customer - Ravi Menon` — which was serving **796 rows** of a customer's staff as employees of this org, the directory holding a Maya Chen and an Aisha Patel of its own. A side that merely contains the word does not: `Customer Success - Aisha Patel` is a desk of ours
- a bracket closes a label by itself, in three shapes, none of which was read — 97 lines became body with no author at all

That last one is where the two names your comment called phantom twins came from. `Aisha Patel Hypothesis` was never a heading: the line reads `[Platform - Aisha Patel 2026-03-13 11:12 UTC] Hypothesis: …`, and it now attributes to Aisha Patel with the heading back in the passage. A denylist entry for "hypothesis" would have sent her comment to `unknown@` instead.

12,492 comment attributions move in total: 8,056 from a synthesized desk address to a person the corpus names or to the counterparty's own domain, 3,553 to an explicit `unknown@` with the label kept in the passage, 874 from unattributed to somebody the corpus names. Every row that crosses the org boundary crosses it the right way — 23 of the counterparty's staff off this domain, two of our own back onto it. The counterparty rule was checked over the labels themselves rather than by sampling: of the 55,400 distinct labels across 197,300 comments it changes 583, and every one states "customer" or "client" outright.

`test_qst_0001_owner_is_maya_chen` is retired too. It fetched the corpus, imported all 581,294 records and asserted one Drive file's owner — one bit for ~16GB per revision, behind an env var that kept it out of CI. Re-exporting the artifact and diffing it by record id reports every changed owner instead of the one a single assertion expects, and it is what found all five shapes above.

Suite: 1,365 passed, 2 skipped (docker, an optional llama-index reader). Roster settles at 14,132 users — 167 in the directory holding tokens, 13,965 attributed and unable to sign in.

backlot/schemas/gmail.schema.json
```diff
@@ -199,19 +201,25 @@
               "number",
               "string"
             ],
+            "minLength": 1,
             "description": "Sent time (default: the root's + an hour per position)."
           }
         },
         "required": [
-          "content"
+          "author_email",
```
nuri-yoo, line 209: `author_email` and `created` are required of a message now, and both still document a default a corpus can no longer take: line 145 reads "Sender (default: the root's author)" and line 205 "Sent time (default: the root's + an hour per position)". Against `4c649af`:

```
{"source_type":"gmail","mailbox":"ava@acme.com","title":"t","content":"c",
 "author_email":"a@b.com","created":1,"messages":[{"content":"m"}]}

    t [messages/0]: 'author_email' is a required property
    t [messages/0]: 'created' is a required property
```

Both were reachable on `main`, where `messages.items.required` was `["content"]`, so the two sentences were accurate documentation until this hunk. Same class as the two you fixed this round: fireflies `start_time` dropped "Default: the running clock", and jira `reporter` made "Default: the author." true again by leaving `required`.

The rest of that class is still there. 24 sentences promise a default for a field that is now required, and every one of the 24 was true on `main` -- the field was optional there and the description is byte-identical on both sides, so promoting it is what made the sentence false:

| where | the sentence |
|---|---|
| top-level `created`, all ten sources | "Default: synthesized from doc_id" |
| all nine named containers (`channel`, `project`, `space`, `team`, `mailbox`, `folder`, `teamspace`, `object_type`, `bucket`) | "Default: the source_type" |
| `google_drive.updated` / `notion.updated`, both required | "Default: created + 1h" / "Default: created" |
| `slack.replies[].created`, required | "default: one second after the previous message" |
| `gmail.messages[].author_email` / `created` | the two above |

Checked one of each rather than reading them: `'created' is a required property`, `'channel' is a required property`, `'updated' is a required property`. The 24/0 split between "made false here" and "already false" is off the base's own schemas rather than off the diff, since a promotion in one hunk and the sentence it falsifies are rarely in the same one.

Two of these describe something that is still alive on the serving side -- `google.py:710` really does serve a thread's nth message at the root's clock plus an hour, and it is what a row with a NULL time gets. That is a router behaviour for a row nobody wrote a time into, not a default a corpus can ask for, and the schema is the corpus-facing document. So either the sentences go, or they say which of the two they mean.

nuri-yoo (approved):
Re-ran all fifteen against `4c649af` rather than reading the diff for them. Fourteen answer differently now, and the reproductions are the ones from the original comments:

`created: ""` is a corpus error on all three paths (`[created]`, `[replies/0/created]`, `[comments/0/created_ts]`), with no `IntegrityError` and no `TypeError`. fireflies `start_time` refuses `null` and its description now names the rule as the mock's. Both `anyOf` errors name their fields, and both negated conditionals render their clause: `'state' is a required property when subtype is not "file" (or absent)`. jira `reporter` validates when absent. s3 refuses `a..b` and `192.168.1.1` while `192.168.1.1.eng` still passes, which is the right discriminator. github `repo` is `maxLength: 100`. A gmail message stating `title: ""` is served with no `Subject` header. hubspot lists both spellings of one type under either path and resolves `0-3`. The unreachable `file` branch, the redundant `or "open"`, `_ff_seconds` and the three stale comments are gone. Containers, `status`, `issuetype` and `state` all stop `""`.

The `_person_like` set: every one of the twenty-one labels I listed now parses to a role with no person, so the passage keeps its label. The two shapes I called phantoms read as people instead of as twins -- `comment by Aisha Patel` and `Priya Mehta on 2026-03-12` both attribute to the person the line names, and `Assignee at` attributes to nobody. That is a better answer than the inversion I proposed, and the 14,158-of-14,979 measurement is the reason it is: inverting would have moved ~392k attributions off the people who made them. `[Platform - Aisha Patel ...] Hypothesis:` is the case that shows the denylist alternative was worse.

Suite here: `1365 passed, 2 skipped` (docker, `llama_index.readers.hubspot`), exit 0, 1367 collected, `ruff check` and `format --check` clean over the 56 touched files.

On the bare base name: agreed, and the measurement is not what settles it either way. A portal holding no custom object cannot answer what one holding a custom object answers to, so `p_orders` returning 400 there is consistent with the changelog rather than against it. Neither of us can read that with this key, and refusing a name a live corpus uses on the strength of a changelog line is the worse failure. Recording it as unimplemented with the measurement beside it is the right shape. The standard-set widening from thirteen to thirty-six, and the four singulars left out, I am taking on your measurement.

Approving. One new comment below, in the gmail hunk that made a message's `author_email` and `created` required: it is prose rather than behaviour, and the same class as the two description sentences already fixed in this round, so it does not hold the branch. What makes it worth a comment is that all 24 of them read correctly on `main` and this branch is what made them false, so nothing outside it will surface them.

backlot/schemas/gmail.schema.json
khj809: All 24 gone in `ee89136`, counted off the schemas the same way you did — the sweep now reports 0 required fields whose description mentions a default, across all eleven.

One more in the same sentence, which your list did not reach: slack's `created` carried the rule twice. After "Default: synthesized from doc_id" it went on to offer a reply "its own `created` when it has one, else one second after the message before it", and a reply states one now, so the alternative is unreachable too. The ordering rule on the reply's own field stays — a ts is identity as well as clock.

Two corrections to the comment, neither of which changes the fix:

`bucket` is not one of them. s3's description opens "The bucket this object belongs to, named the way S3 names one: …" and never offers a default, so the nine are `channel` twice — fireflies and slack — plus `project`, `space`, `team`, `mailbox`, `folder`, `teamspace`, `object_type`. Your total of 24 is right; the per-name list is off by that one substitution.

`google.py:710` is not alive. `gmail_messages.created_ts` is `INTEGER NOT NULL`, so no row can have the NULL time that branch answers — the same shape as `_message_second`, which this branch deleted for the same reason. That holds for every `created_ts` in the schema: all seventeen tables declare it `NOT NULL`, so the ten `created_ts` fallbacks across atlassian, github and google are unreachable. `updated_ts` is nullable everywhere it exists, so those fallbacks are real, and a NULL `updated` is what the two sentences you flagged were describing.

I went one step past that and then backed it out, which is worth recording since the branch would otherwise look like it ignored it. Four routers read a timestamp as `row["updated_ts"] or …`, where `or` treats 0 as unstated and 0 is 1970-01-01T00:00:00Z — the second `_gmail_ts` carries a comment about, and that `atlassian.py` and `google.py` already read with `is not None`. I wrote a corpus stating `updated: 0`, watched notion serve the record's creation time instead, fixed all ten sites and pinned it with a mutation-tested assertion. Then I checked whether the bench ever states it: **zero rows at epoch 0 anywhere in the corpus**, earliest second 1002345678, and notion and s3 hold no rows at all here. So the reproduction was a corpus I wrote rather than a row that exists, and the change was consistency with an idiom the project already settled — not something this branch's own data exercises. On an approved PR that is scope I should not add on an unobserved premise, so it is reverted: `ee89136` is the ten schema files and nothing else. The seven `or` sites are still there for whoever wants them, with the reasoning above.

khj809: Correcting my own count in the note above: the `or` reads are **11**, not the "ten" and "seven" I wrote — github 663 / 1338 / 1339 / 1768 / 1808, hubspot 180, notion 215 / 250 / 324, s3 290 / 316. Ten was the number of edits, since 1338 and 1339 sat in one hunk. All 11 are on the branch as they were; nothing about the revert changes.

nuri-yoo: Re-ran the sweep on `ee89136` and it reports 0 across all eleven schemas. The slack sentence is the better catch of the two of us: the rule a second time over, on the field whose requirement closes it.

Both corrections right, and both from reading the wrong thing.

`bucket`: the sweep has a row per file and field, so `channel` is in it twice, fireflies and slack. Writing the nine out by name I folded those into one and filled the gap with s3's container instead of going back to the sweep, which would have shown that `bucket` offers no default at all. The nine are `channel` twice plus `project`, `space`, `team`, `mailbox`, `folder`, `teamspace`, `object_type`.

`_gmail_ts`: I read the router where the table was the authority. An `is not None` guard shows the case was weighed, not that it occurs, and `gmail_messages.created_ts` is `INTEGER NOT NULL` -- one line of DDL, and the same reason `_message_second` went.

PR #72, opened 2026-08-20
fireflies: a client generated from the vendor schema binds to what is served

### LINKED ISSUE #68: fireflies: a client generated from the vendor schema cannot bind to the served one
`backlot/graphql/fireflies.graphql` names types the vendor's schema does not name, puts the
per-speaker analytics fields on the wrong type, and omits four `transcripts` arguments. A client
generated from the real Fireflies schema fails against the mock at each point below.

Verified 2026-08-20 against `a92e3fa`, and the Fireflies surface is byte-identical at `75b1bc9`
(current `origin/main`), by introspecting `https://api.fireflies.ai/graphql` with a
workspace token, plus one real `transcript` response. Every row is from that introspection —
**not** from `docs.fireflies.ai`, which does not match the API (see the last section).

## `Sentence`

- `ai_filters: AiFilters` — the live type is `AIFilters`.
- `AIFilters` carries a `text_cleanup: String` field the mock's type does not declare.
- The resolver serves `ai_filters: None` unconditionally (`fireflies_resolvers.py:129`), and the
  SDL states this as intent ("Always null"). The live API returns the object, populated:

  ```json
  "ai_filters": { "text_cleanup": "<the sentence text>", "task": null, "pricing": null,
                  "metric": null, "question": null, "date_and_time": null, "sentiment": null }
  ```

  So `sentence.ai_filters.text_cleanup` null-dereferences against the mock alone. The classifier
  flags being null is faithful — they were null in the real response too — but the enclosing
  object is not optional.

## Type names, which a generated client binds by name

| Mock | Live schema |
|---|---|
| `AiFilters` | `AIFilters` |
| `Attendee` | `MeetingAttendeeModel` |
| `Categories` | `AnalyticsCategories` |
| `Analytics` | `MeetingAnalytics` |
| `JSON` (scalar) | no such scalar — `apps_preview` is `Apps` |

`Analytics` is more than a rename. The live schema **has** an `Analytics` type and it is something
else entirely — `{ team: TeamAnalytics!, users: [UserAnalytics!]! }`, workspace-wide rather than
per-meeting. The mock uses the vendor's name for a different shape, so a client binding `Analytics`
gets the wrong fields rather than a missing type.

## `Speaker` conflates two live types

Live:

- `Transcript.speakers: [Speaker!]`, where `Speaker { id: Float, name: String }`
- `MeetingAnalytics.speakers: [AnalyticsSpeaker!]`, where `AnalyticsSpeaker { speaker_id: Int,
  name, duration, word_count, longest_monologue, monologues_count, filler_words, questions,
  duration_pct, words_per_minute }`

The mock serves one `Speaker` carrying the analytics fields at both positions, and the SDL states
this as intent ("the same values as `analytics.speakers`"). So:

- `transcript { speakers { id } }` is valid against the vendor and an unknown-field error here.
- `words_per_minute` is absent from the mock everywhere.

## `Attendance` shares no field name with the live type

| Mock `Attendance` | Live `MeetingAttendance` |
|---|---|
| `displayName`, `email`, `joinedAt`, `duration` | `name: String!`, `join_time: String!`, `leave_time: String` |

## Scalars

| Field | Mock | Live |
|---|---|---|
| `Transcript.date` | `DateTime` | `Float` |
| `Transcript.channels` | `[String]` | `[Channel!]!` — objects (`id: ID!`, `title`, `members`, …) |
| `Transcript.apps_preview` | `JSON` | `Apps { outputs: [AppOutput!] }` |
| `MeetingInfo.summary_status` | `String` | `SummaryStatus` enum (`processing`/`processed`/`failed`/`skipped`/`default`) |
| `User.num_transcripts` | `Int` | `Float` |

`DateTime` is itself a real scalar, so declaring it is fine — only `Transcript.date` uses it where
the vendor uses `Float`. `summary_status` serialises to the same JSON string either way; what
differs is the generated client's enum binding.

## Missing `transcripts` arguments

The mock accepts twelve, the live field sixteen. Absent, each an unknown-argument error here:
`title`, `date`, `organizer_email`, `participant_email`. Their behaviour, probed against the live
API rather than assumed:

- `title` matches on **substring** — `title: "tes"` returns the meeting titled `test`.
- `date` buckets by **day**, not instant. Filtering on either of two same-day meetings' exact
  `date` returns both; a timestamp from another day returns `[]`. (Both sample meetings fall on one
  day, so "same calendar day" is the narrowest reading the data supports, not a proven window.)
- `organizer_email` and `participant_email` are the singular forms of the existing `organizers:`
  and `participants:` filters.

## List element nullability

Live marks elements non-null where the mock does not: `transcripts: [Transcript!]`,
`users: [User!]`, `Transcript.{participants, fireflies_users, sentences, speakers}`,
`Summary.{keywords, topics_discussed, transcript_chapters}`, `User.integrations`, and the
`organizers:`/`participants:` arguments. Conversely `Transcript.id` and `User.user_id` are
nullable in the live schema and non-null here.

## Fields the vendor declares and the mock does not

The comparison above ran mock-to-live. Run the other way, these are live fields no client can
select here — the same class of break, found while scoping the fix:

| Type | Missing |
|---|---|
| `Transcript` | `privacy: String`, `is_live: Boolean`, `workspace_users: [String!]`, `shared_with: [SharedWith!]` |
| `Summary` | `notes: String`, `short_overview: String`, `extended_sections: [SummarySection!]` |
| `User` | `user_groups: [UserGroup!]`, `is_calendar_in_sync: Boolean`, `recent_transcript: String` |
| `Speaker` | `id: Float` (part of the split above) |
| `AIFilters` | `text_cleanup: String` (above) |
| `AnalyticsSpeaker` | `words_per_minute: Float` (above) |

What the live API returns for them, which is what settles null-versus-empty: `privacy: "link"`,
`is_live: false`, and `workspace_users`, `shared_with`, `user_groups` all `[]` — not null.

Out of scope, and deliberately: the live `Query` root also has `apps`, `active_meetings`,
`user_groups`, `analytics`, `channels`, `channel`, `contacts`, `bites`, `bite`, `askfred_threads`,
`askfred_thread`, `live_action_items`, `rule_executions_by_meeting` and `auditEvents`. Those are
whole API surfaces this mock does not model, not fidelity gaps in what it already serves.

## Scope: all of this is serving-level

No DDL change and no import-schema change is required. Verified against the store schema and the
stored corpus, item by item.

SDL-only, because the served values are already right or already null:

| Item | Why nothing else moves |
|---|---|
| `Attendee` → `MeetingAttendeeModel` | its five field names already match the live type exactly |
| `Categories` → `AnalyticsCategories` | fields identical |
| `Analytics` → `MeetingAnalytics` | fields identical |
| `Attendance` → `MeetingAttendance` field names | the list is served null (there is no column); the non-null element fields are never reached |
| `summary_status` → `SummaryStatus` | the resolver already hardcodes `"processed"`, a member of the enum |
| `num_transcripts` `Int` → `Float` | same JSON number |
| element non-nullability, `id`/`user_id` nullability | declaration only |
| `Summary.notes`, `short_overview`, `extended_sections` | null, as the other LLM renderings already are |
| `User.is_calendar_in_sync`, `recent_transcript` | null, as the other account state already is |
| `Transcript.privacy`, `is_live` | constants: the corpus is finished recordings |
| `Transcript.workspace_users`, `shared_with`, `User.user_groups` | the empty list the vendor itself returns |

Resolver work, each backed by a column that already exists:

| Item | Backing |
|---|---|
| `ai_filters` object + `text_cleanup` | `fireflies_sentences.body` |
| `date` → `Float` | `created_ts`; `_millis()` already returns a number |
| `apps_preview` → `Apps` | needs no data — serve `{ outputs: [] }`, which is what the live API returns |
| `channels` → `[Channel!]!` | `fireflies_transcripts.channel` (NOT NULL) and `fireflies_channels.group_id` back `id`/`title`; the rest are null and `members` is `[]` |
| `Speaker` / `AnalyticsSpeaker` split | `fireflies_sentences` already carries `speaker_name` **and** `speaker_id`, so `Speaker { id, name }` is a lookup, not a new column |
| `words_per_minute` | derived from the analytics JSON's `word_count` and `duration` (seconds), as `duration` is already derived from the last sentence |
| the four `transcripts` arguments | `title`, `created_ts`, `organizer_email`, and the `participants` JSON array — `store._fireflies_where` already has the `COALESCE(organizer_email, author_email)` and `json_each` clauses the singular forms need |

Two things to get right while implementing:

- `Transcript.speakers { id }` needs the name-to-`speaker_id` map from `fireflies_sentences`. Done
  naively that is one query per transcript, so an N+1 across a 50-item page — batch it.
- `words_per_minute` divides by a duration that can be zero; guard to null rather than raising.

## Not verifiable with this token

`Transcript.analytics` is gated — `paid_required`, tier `pro_or_higher`. The `Analytics` and
`Speaker` findings above therefore rest on introspection; no real analytics response was observed.

## Correcting this report's original premise

This issue previously claimed `Sentence.start_time`/`end_time` were `String` and `speaker_id` was
`ID`, citing two `docs.fireflies.ai` pages. The live schema declares `Float`, `Float` and `Int` —
what the mock already serves — and a real response carries bare float seconds:

```json
{ "index": 0, "speaker_id": 0, "speaker_name": "Speaker 1",
  "start_time": 1.83, "end_time": 2.79, "text": "..." }
```

The seconds-vs-string question that blocked the original report does not exist: no conversion at
the resolver boundary, and no change to `synth.fireflies_fill_times`, the REAL columns, `keyword`
with `scope: sentences`, or the `duration` derivation. The two docs pages agreeing was not
corroboration — they are one source, and it does not match the API. Check this schema by
introspection.

khj809:
Closes #68.

The served Fireflies schema now follows the vendor's own, verified by introspecting and querying
`api.fireflies.ai` rather than the docs pages — which disagree with the API, and were the basis for
the reverse claim that `Sentence`'s times and speaker id were strings.

- **Type names** a generated client binds by name: `AIFilters`, `MeetingAttendeeModel`,
  `AnalyticsCategories`, `MeetingAnalytics`. The last matters most — the vendor's `Analytics` is a
  different type (workspace-wide), so the old name pointed clients at the wrong shape.
- **`Speaker` was two types.** `Speaker { id, name }` is identity on `Transcript.speakers`;
  `AnalyticsSpeaker` is the talk-time statistics on `analytics.speakers`. The per-meeting speaker
  number joins them and already lives in `fireflies_sentences`, so it is read from there — keyed on
  the NUMBER, which is what Fireflies assigns: two runs diarization never labelled are two
  speakers, and a name-keyed roster would collapse them and leave `Sentence.speaker_id` naming a
  speaker the roster does not list. `analytics.speakers` is one per LABEL by contrast — the stored
  statistics are aggregated by name, so numbers sharing one are a single entry — and the SDL says
  so where a client reads it. `words_per_minute` is derived from the stored word count and talk
  time.
- **Objects where scalars were served**: `channels` is `[Channel!]!`, `apps_preview` is `Apps`,
  `date` is a bare `Float`, `summary_status` is the `SummaryStatus` enum. `MeetingAttendance` takes
  the vendor's field names.
- **Four `transcripts` arguments** the vendor takes and this did not: `title` (substring), `date`
  (a day, narrowed against any `fromDate`/`toDate`), `organizer_email` and `participant_email`.
- **Fields the vendor declares** and this omitted: `Transcript.privacy`/`is_live`/
  `workspace_users`/`shared_with`, `Summary.notes`/`short_overview`/`extended_sections`,
  `User.user_groups`/`is_calendar_in_sync`/`recent_transcript`, `AIFilters.text_cleanup`.
  Where the live API returns an empty list rather than null, so does this.

`Channel.members` is aggregated across a channel's meetings, so it carries the same ACL clause
`transcripts` does: a meeting the caller was denied contributes nobody. Without it the roster named
that meeting's participants and — by carrying an address no visible meeting mentions — said that
such a meeting exists.

The SDL states what a client cannot otherwise tell: `title` and `keyword` match case-insensitively,
folding ASCII, so a cased non-ASCII letter has to be spelled as the text spells it; `date` buckets
by the UTC day (a value in the day before or after returns nothing, whichever zone the caller is
in); and a channel's roster is scoped to the caller.

No DDL or import-schema change: every value comes from a column that already exists.
`Channel.members` and the speaker numbers are bound lazily, so selecting metadata alone still does
not query — and neither costs a statement per row of a page. A page's rosters load in one
statement that both speaker fields read, and a channel's roster is read once per channel: a
20-transcript page is 2 statements for either, where the per-transcript read was 21, or 41 with
both speaker fields selected. A test counts them.

Out of scope and noted in the issue: the live `Query` root has fourteen further fields (`bites`,
`contacts`, `askfred_*`, `auditEvents`, …) that are whole API surfaces this mock does not model.
`Transcript.analytics` is paid-tier gated, so its findings rest on introspection rather than an
observed response.

Verified: 1204 passed, 3 skipped (all extras); ruff clean; trial-merged with #67 (no conflicts,
1370 passed).

### DIFF
--- a/backlot/graphql/fireflies_resolvers.py
+++ b/backlot/graphql/fireflies_resolvers.py
@@ -108,10 +108,15 @@ def _user(email: str | None, display: str | None = None) -> dict | None:
         "name": display or _display_name(email),
         "num_transcripts": None,
         "recent_meeting": None,
+        "recent_transcript": None,
         # account/billing state this mock does not model — see the SDL header
         "minutes_consumed": None,
         "is_admin": None,
         "integrations": None,
+        "is_calendar_in_sync": None,
+        # The mock's ACL groups are a permission mechanism, not Fireflies user-groups, so they are
+        # not served under this name. Fireflies serves the empty list, not null.
+        "user_groups": [],
     }
 
 
@@ -163,20 +168,68 @@ def _summary(row) -> dict:
         "bullet_gist": None,
         "gist": None,
         "transcript_chapters": None,
+        "notes": None,
+        "short_overview": None,
+        "extended_sections": None,
     }
 
 
-def _speakers(row) -> list[dict]:
-    return (store.jcol(row, "analytics") or {}).get("speakers") or []
+def _words_per_minute(word_count, duration_secs) -> float | None:
+    """Words per minute over this speaker's own talk time. Null rather than a division error when
+    the corpus gives a speaker no measurable time."""
+    if word_count is None or not duration_secs:
+        return None
+    return round(float(word_count) * 60.0 / float(duration_secs), 2)
+
+
+def _analytics_speakers(row, numbers: dict) -> list[dict]:
+    """`analytics.speakers` is Fireflies' AnalyticsSpeaker: talk-time statistics keyed by the same
+    per-meeting speaker number the sentences carry. `numbers` maps name -> that number."""
+    out = []
+    for sp in (store.jcol(row, "analytics") or {}).get("speakers") or []:
+        name = sp.get("name")
+        duration = sp.get("duration")
+        out.append(
+            {
+                "speaker_id": numbers.get(name),
+                "name": name,
+                "duration": duration,
+                "word_count": sp.get("word_count"),
+                "longest_monologue": sp.get("longest_monologue"),
+                "monologues_count": sp.get("monologues_count"),
+                "filler_words": sp.get("filler_words"),
+                "questions": sp.get("questions"),
+                "duration_pct": sp.get("duration_pct"),
+                "words_per_minute": _words_per_minute(sp.get("word_count"), duration),
+            }
+        )
+    return out
 
 
 def _analytics(row) -> dict:
     a = store.jcol(row, "analytics") or {}
     return {
+        "_row": row,
         "sentiments": a.get("sentiments"),
-        "speakers": a.get("speakers") or [],
+        # bound explicitly (see RESOLVERS): the speaker numbers live in fireflies_sentences
         # classifier buckets — see the SDL header
         "categories": a.get("categories"),
+    }
+
+
+def _channel(name: str) -> dict:
+    """One Channel. `id` IS the channel name rather than a minted opaque id, because
+    `transcripts(channel_id:)` selects on the name — a client must be able to feed `channels { id }`
+    straight back in."""
+    return {
+        "_channel": name,
+        "id": name,
+        "title": name,
+        # the corpus states which channel a meeting is in, not the channel's own history
+        "created_at": None,
+        "updated_at": None,
+        "created_by": None,
+        "is_private": None,
     }
 
 
@@ -202,11 +255,18 @@ def _transcript(row, info) -> dict:
             for e in (store.jcol(row, "participants", []) or [])
             if isinstance(e, str) and "@external." not in e
         ],
+        # workspace membership and link-sharing state — see the SDL header. Fireflies serves the
+        # empty list for both, not null.
+        "workspace_users": [],
+        "shared_with": [],
+        # The corpus is finished recordings, so no meeting is ever mid-transcription.
+        "is_live": False,
+        "privacy": "link",
         "meeting_link": row["meeting_link"],
         "calendar_id": row["calendar_id"],
         "cal_id": row["calendar_id"],
         "calendar_type": row["calendar_type"],
-        "channels": [row["channel"]] if row["channel"] else [],
+        "channels": [_channel(row["channel"])] if row["channel"] else [],
         "transcript_url": row["transcript_url"],
         "audio_url": row["audio_url"],
         "video_url": row["video_url"],
@@ -220,8 +280,8 @@ def _transcript(row, info) -> dict:
             "silent_meeting": False,
             "summary_status": "processed",
         },
-        "apps_preview": None,
-        "speakers": _speakers(row),
+        # Fireflies serves the wrapper with an empty `outputs`, not null — see the SDL header.
+        "apps_preview": {"outputs": []},
     }
 
 
@@ -237,7 +297,11 @@ def resolve_transcripts(
     toDate=None,
     host_email=None,
     organizers=None,
+    organizer_email=None,
     participants=None,
+    participant_email=None,
+    title=None,
+    date=None,
     user_id=None,
     mine=None,
     channel_id=None,
@@ -265,14 +329,33 @@ def resolve_transcripts(
     if len({h.lower() for h in hosts}) > 1:
         return []
 
+    # The singular forms are aliases of the plural filters, so they combine the same way:
+    # `organizers` is any-of, `participants` is all-of.
+    organizers = [*(organizers or []), *([organizer_email] if organizer_email else [])]
+    participants = [*(participants or []), *([participant_email] if participant_email else [])]
+
+    from_ts = to_epoch_seconds(fromDate)
+    to_ts = to_epoch_seconds(toDate)
+    if date is not None:
+        # `date` selects a DAY, not an instant: the real API returns every meeting sharing the
+        # calendar day of the value passed. Narrowed against any fromDate/toDate already given.
+        day = to_epoch_seconds(date)
+        if day is None:
+            return []
+        start = day - (day % 86400)
+        from_ts = start if from_ts is None else max(from_ts, start)
+        end = start + 86399
+        to_ts = end if to_ts is None else min(to_ts, end)
+
     rows = store.list_fireflies_transcripts(
         ctx["conn"],
         channel=channel_id,
         host_email=hosts[0] if hosts else None,
         organizers=organizers or None,
         participants=participants or None,
-        from_ts=to_epoch_seconds(fromDate),
-        to_ts=to_epoch_seconds(toDate),
+        title=title,
+        from_ts=from_ts,
+        to_ts=to_ts,
         keyword=keyword,
         scope=scope,
         visible_ids=ctx.get("visible_ids"),
@@ -299,6 +382,39 @@ def resolve_transcript_sentences(transcript, info, **_ignored):
     row = transcript["_row"]
     rows = store.fireflies_sentences(_ctx(info)["conn"], row["id"])
     return [_sentence(r, i) for i, r in enumerate(rows)]
+
+
+def _speaker_numbers(info, transcript_id) -> dict:
+    return store.fireflies_speaker_numbers(_ctx(info)["conn"], transcript_id)
+
+
+def resolve_transcript_speakers(transcript, info, **_ignored):
+    """Fireflies' `Transcript.speakers` is identity only — name and the per-meeting number. Both
+    live in fireflies_sentences, so this queries; selecting metadata alone never does."""
+    numbers = _speaker_numbers(info, transcript["_row"]["id"])
+    return [{"id": float(num), "name": name} for name, num in numbers.items()]
+
+
+def resolve_analytics_speakers(analytics, info, **_ignored):
+    """`analytics.speakers` carries the stored talk-time statistics, joined to the same speaker
+    numbers `Transcript.speakers` reports."""
+    row = analytics["_row"]
+    return _analytics_speakers(row, _speaker_numbers(info, row["id"]))
+
+
+def resolve_channel_members(channel, info, **_ignored):
+    """A channel's roster, from the ACL group the channel is grouped by. Bound rather than built
+    eagerly: `channels { id title }` is the common selection and must not query."""
+    ctx = _ctx(info)
+    return [
+        {
+            "user_id": synth.fireflies_user_id(r["email"]),
+            "email": r["email"],
+            "name": r["display_name"] or _display_name(r["email"]),
+        }
+        for r in store.fireflies_channel_members(ctx["conn"], channel["_channel"])
+        if r["email"]
+    ]
 
 
 def resolve_user(_root, info, id=None):
@@ -345,7 +461,12 @@ RESOLVERS = {
         "user": resolve_user,
         "users": resolve_users,
     },
-    "Transcript": {"sentences": resolve_transcript_sentences},
+    "Transcript": {
+        "sentences": resolve_transcript_sentences,
+        "speakers": resolve_transcript_speakers,
+    },
+    "MeetingAnalytics": {"speakers": resolve_analytics_speakers},
+    "Channel": {"members": resolve_channel_members},
 }
 
 
--- a/backlot/graphql/fireflies_resolvers.py
+++ b/backlot/graphql/fireflies_resolvers.py
@@ -126,7 +126,17 @@ def _sentence(row, index: int) -> dict:
         "raw_text": row["body"],
         "start_time": row["start_time"],
         "end_time": row["end_time"],
-        "ai_filters": None,
+        # Fireflies returns the object even when nothing classified: `text_cleanup` carries the
+        # cleaned text, and the classifier flags are null.
+        "ai_filters": {
+            "text_cleanup": row["body"],
+            "task": None,
+            "pricing": None,
+            "metric": None,
+            "question": None,
+            "date_and_time": None,
+            "sentiment": None,
+        },
     }
 
 
--- a/backlot/graphql/fireflies_resolvers.py
+++ b/backlot/graphql/fireflies_resolvers.py
@@ -26,6 +26,14 @@ from backlot import store, synth
 PAGE_DEFAULT = 25
 PAGE_MAX = 50
 
+# Slots this module keeps on the request context (backlot/routers/fireflies.py builds one dict per
+# request, so nothing here outlives the response or crosses a caller). The two lazily-bound fields
+# that query — the speaker numbers and a channel's roster — read through them, so a page of
+# transcripts costs one statement each rather than one per row.
+_PAGE_IDS = "_ff_page_ids"
+_SPEAKERS = "_ff_speakers"
+_CHANNEL_MEMBERS = "_ff_channel_members"
+
 
 def _ctx(info):
     return info.context
@@ -182,9 +190,14 @@ def _words_per_minute(word_count, duration_secs) -> float | None:
     return round(float(word_count) * 60.0 / float(duration_secs), 2)
 
 
-def _analytics_speakers(row, numbers: dict) -> list[dict]:
+def _analytics_speakers(row, speakers: list) -> list[dict]:
     """`analytics.speakers` is Fireflies' AnalyticsSpeaker: talk-time statistics keyed by the same
-    per-meeting speaker number the sentences carry. `numbers` maps name -> that number."""
+    per-meeting speaker number the sentences carry. `speakers` is the roster those numbers come
+    from, joined to the stored statistics by NAME — the only key the analytics JSON has, since
+    synth.fireflies_speaker_stats aggregates by it. Runs diarization never labelled are therefore
+    ONE statistics entry however many numbers they cover, served under one of them, while the
+    roster keeps every number."""
+    numbers = {s["speaker_name"]: s["speaker_id"] for s in speakers}
     out = []
     for sp in (store.jcol(row, "analytics") or {}).get("speakers") or []:
         name = sp.get("name")
@@ -338,7 +351,9 @@ def resolve_transcripts(
     to_ts = to_epoch_seconds(toDate)
     if date is not None:
         # `date` selects a DAY, not an instant: the real API returns every meeting sharing the
-        # calendar day of the value passed. Narrowed against any fromDate/toDate already given.
+        # calendar day of the value passed, and that day is UTC — a value anywhere in the day
+        # before or after returns nothing, whichever zone the caller is in. Narrowed against any
+        # fromDate/toDate already given.
         day = to_epoch_seconds(date)
         if day is None:
             return []
@@ -362,6 +377,9 @@ def resolve_transcripts(
         limit=clamp_limit(limit),
         offset=clamp_skip(skip),
     )
+    # The page's ids, for the batched speaker load. Accumulated rather than assigned: one document
+    # may select `transcripts` twice under aliases, and a second page must not drop the first's.
+    ctx.setdefault(_PAGE_IDS, []).extend(r["id"] for r in rows)
     return [_transcript(r, info) for r in rows]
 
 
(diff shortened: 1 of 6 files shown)


backlot/store.py
```diff
@@ -1849,6 +1857,42 @@ def fireflies_transcript_by_id(conn, transcript_id, visible_ids=None) -> sqlite3
...
+
+def fireflies_channel_members(conn, channel) -> list[sqlite3.Row]:
+    """A channel's roster: everyone who took part in a meeting in it.
+
+    The same choice slack_channel_member_emails documents — membership is the per-channel signal
+    the corpus actually carries. Answering with the channel's ACL group instead would give every
+    channel sharing a group the same members, which the real API cannot produce."""
+    return conn.execute(
+        "SELECT DISTINCT je.value AS email, p.display_name AS display_name "
+        "FROM fireflies_transcripts t, json_each(t.participants) je "
```
nuri-yoo, line 1919: This is the one fireflies query with no ACL clause, and unlike `fireflies_sentences` below it the caller has not been cleared for the rows it reads: a channel's members are aggregated across every transcript in the channel, including ones the caller cannot see.

Corpus: two meetings in `all-hands`, one public, one `visibility: group` on a group the caller is not in. Called as `ava@acme.com`:

```
transcripts         -> [ "Weekly sync" ]      # the board meeting is correctly filtered
channels[0].members -> ava@acme.com, bob@acme.com,
                       hana@acme.com, secret.investor@acme.com
```

The last two addresses appear only in the meeting she was denied, so the roster leaks both the participants and the fact that another meeting exists in that channel. `list_fireflies_transcripts` already threads `visible_ids`, so the fix is the same `_acl_clause` the rest of this family uses, applied to `t`.

The docstring cites `slack_channel_member_emails`, which has the same shape -- but a slack channel's membership is org-visible by construction there, while `fireflies_transcripts` rows are ACL'd per document and this mock's own `transcripts` honours that. If org-wide fireflies membership is the intent rather than an inheritance, it is worth saying so in the docstring, because the two functions are not in the same position.

nuri-yoo (commented):
Reviewed at `1cf1e0e`. Suite `1201 passed, 3 skipped` here (the extra skip is `llama_index.readers.hubspot`, absent from my env); ruff clean on the three touched files; trial-merge with #67 confirmed, `store.py` the only shared file and in disjoint regions.

What I could reach reproduces: the substring match escapes its wildcards, `date:` buckets by day and narrows against a `toDate` inside it, both singular email filters select, `channels { id }` feeds back into `channel_id:`, `summary_status` serialises as an enum member, and metadata-only selection is still 2 queries. A corpus stating `speaker_id: null` is safe -- the loader assigns ordinals, so the `IS NOT NULL` filter drops nothing. The introspection-derived half (type names, the `Speaker` split, the field lists) I took as stated in #68: no workspace token here, and the docs are the wrong source. I checked it for internal consistency instead -- the SDL parses, every lazily-bound field has a resolver, and no non-null field can reach a null through the builders.

Three inline, all before merge. `fireflies_channel_members` takes no `visible_ids`, so `channels { members }` returns participants of transcripts the caller was denied. The speaker map is one query per transcript and two when both speaker fields are selected, which is the N+1 #68 asked to batch. And the roster keys on speaker name where Fireflies keys on the number, so two unlabelled speakers collapse -- the new test's own assertion fails on a transcript with two.

Two small ones inline as well: the title match folds case for ASCII only, and the `date` bucket is a UTC day, which the SDL sentence does not say.

backlot/graphql/fireflies_resolvers.py
```diff
@@ -291,6 +384,39 @@ def resolve_transcript_sentences(transcript, info, **_ignored):
     return [_sentence(r, i) for i, r in enumerate(rows)]
 
 
+def _speaker_numbers(info, transcript_id) -> dict:
+    return store.fireflies_speaker_numbers(_ctx(info)["conn"], transcript_id)
+
+
+def resolve_transcript_speakers(transcript, info, **_ignored):
```
nuri-yoo, line 427: Lazily bound but not batched, so this is the N+1 #68 asked to avoid ("one query per transcript, so an N+1 across a 50-item page -- batch it"). Traced with `set_trace_callback` over a 20-transcript page:

```
{ transcripts(limit:20) { id title } }                                 2 queries
{ transcripts(limit:20) { id speakers { id name } } }                  22   (20x the map)
{ transcripts(limit:20) { id speakers { id }
                          analytics { speakers { speaker_id } } } }    42   (40x the map)
```

The 42 is the part worth fixing first: this resolver and `resolve_analytics_speakers` (line 398) each call `_speaker_numbers` for the same transcript, so the identical statement runs twice per row before any batching question arises. Memoising on `_ctx(info)` closes that half; one `WHERE transcript_id IN (...)` over the page's ids closes the other.

The precedent is #53, where the same shape on `conversations.replies` went from 30 root reads to 0 and the count was pinned by a test. A query-count assertion here would keep it honest, since nothing in the suite currently fails when this doubles.

backlot/graphql/fireflies_resolvers.py
```diff
@@ -291,6 +384,39 @@ def resolve_transcript_sentences(transcript, info, **_ignored):
...
+
+
+def resolve_analytics_speakers(analytics, info, **_ignored):
+    """`analytics.speakers` carries the stored talk-time statistics, joined to the same speaker
+    numbers `Transcript.speakers` reports."""
+    row = analytics["_row"]
+    return _analytics_speakers(row, _speaker_numbers(info, row["id"]))
+
+
+def resolve_channel_members(channel, info, **_ignored):
```
nuri-yoo, line 443: Same trace, and this one repeats work rather than merely doing it per row: the 20 transcripts in the fixture are all in `all-hands`, so the identical statement runs 20 times.

```
{ transcripts(limit:20) { id channels { id members { email } } } }     22 queries
    20x  SELECT DISTINCT je.value AS email, p.display_name ... WHERE t.channel = ?
```

Keying a per-request cache on the channel name makes it one, and that name is already on the object -- `_channel()` stashes it as `_channel` for exactly this lookup.

backlot/store.py
```diff
@@ -1849,6 +1857,42 @@ def fireflies_transcript_by_id(conn, transcript_id, visible_ids=None) -> sqlite3
...
+    Fireflies numbers speakers WITHIN a meeting and reports that number on both `Sentence` and
+    `AnalyticsSpeaker`, so the roster has to come from the rows that already state it rather than
+    from a position in the analytics JSON, which carries no number of its own. DISTINCT over
+    idx_fireflies_sentences_doc, so it seeks the transcript rather than scanning the table.
+
+    A NULL name is a key like any other: diarization that produced no label still numbered the
+    speaker, and synth.fireflies_speaker_stats aggregates that run too, so dropping it here would
+    leave `analytics.speakers` with an entry no number could be found for."""
+    return {
+        r["speaker_name"]: r["speaker_id"]
```
nuri-yoo: The dict keys on `speaker_name` where Fireflies numbers speakers, so two speakers whose label diarization did not produce collapse into one entry, and `ORDER BY speaker_id` decides which survives:

```
sentences   speaker_id 0 (name null), 1 (name null), 2 ("Ava Chen")
speakers    id 1.0 (name null), 2.0 ("Ava Chen")        # 0 is gone
```

That is the invariant `test_fireflies_an_unnamed_speaker_still_carries_its_number:173` states -- `{float(s["speaker_id"]) for s in sentences} == {sp["id"] for sp in speakers}` -- and it holds in the fixture only because `ff-discovery` has one unlabelled run. With two, the test fails.

Returning the rows rather than a name-keyed dict fixes both readers: the roster is one entry per `speaker_id`, and `_analytics_speakers` builds its name-to-number lookup from the same rows (an unlabelled speaker still has no name to join on, which is already true today).

tests/test_fireflies.py
```diff
@@ -86,6 +90,32 @@ def test_fireflies_serves_the_documented_metadata_surface(client, admin_h):
...
+    # `channels.id` IS the channel name, so it can be fed straight back to `channel_id:`
+    ch = t["channels"][0]
+    assert ch["id"] == ch["title"]
+    back = ff_gql(
+        client, '{ transcripts(channel_id: "%s", limit: 50) { id } }' % ch["id"], admin_h
+    ).json()["data"]["transcripts"]
+    assert t["id"] in {x["id"] for x in back}
+    # the roster is who took part in the channel's meetings, not the channel's ACL group, so it
+    # covers this meeting's own participants and anyone else the channel's other meetings named
+    assert {m["email"] for m in ch["members"]} >= set(t["participants"])
```
nuri-yoo: `>=` cannot fail on the leak the members query has: a superset assertion passes just as well when the roster carries participants from transcripts this caller was denied. Asserting equality against the participants of the transcripts the caller can actually see would pin the property the ACL clause is for, and would fail today.

backlot/store.py
```diff
@@ -1783,6 +1784,11 @@ def _fireflies_where(
             "WHERE lower(json_each.value) = ?)"
         )
         params.append(email.lower())
+    if title:
+        # `transcripts(title:)` matches a SUBSTRING, which the live API demonstrates: "tes" returns
+        # the meeting titled "test". Distinct from `keyword`, which `scope` can point at sentences.
+        sql += " AND title LIKE ? ESCAPE '\\'"
```
nuri-yoo, line 1799: `LIKE` folds case for ASCII only, so `title:"LATENCY"` matches `Latency review` (measured) while a non-ASCII title matches case-sensitively -- two behaviours under one argument, neither of them the substring rule #68 verified. `lower(title) LIKE lower(?)` makes it one, and the SDL sentence can then say which one it is. The gdrive path at line 1631 carries the same expression, so whichever way this goes it is worth being deliberate rather than inheriting.

backlot/graphql/fireflies_resolvers.py
```diff
@@ -255,14 +329,33 @@ def resolve_transcripts(
...
+
+    from_ts = to_epoch_seconds(fromDate)
+    to_ts = to_epoch_seconds(toDate)
+    if date is not None:
+        # `date` selects a DAY, not an instant: the real API returns every meeting sharing the
+        # calendar day of the value passed. Narrowed against any fromDate/toDate already given.
+        day = to_epoch_seconds(date)
+        if day is None:
+            return []
+        start = day - (day % 86400)
```
nuri-yoo, line 363: The bucket is a UTC day (`day - day % 86400`), which #68 is careful to call the narrowest reading its two same-day samples support. Since the SDL sentence says only "Meetings recorded on this DAY", naming UTC there would stop a client in another zone from reading its own midnight into it -- and if the real API turns out to bucket in the workspace's timezone, that sentence is where the difference gets recorded.

backlot/store.py
khj809: Reproduced as written: as `ava@acme.com` a public `all-hands` meeting served `members` = ava, bob, hana, secret.investor, the last two appearing only in the meeting she was denied.

`fireflies_channel_members` now takes `visible_ids` and applies `_acl_clause("fireflies", "t", …)`, and `resolve_channel_members` threads `ctx["visible_ids"]`. Same corpus now: ava gets ava+bob, hana gets all four.

The docstring says which position this is in rather than citing the slack one as equivalent: a slack channel’s membership is org-visible by construction, a fireflies transcript is granted per document, so the aggregate needs the clause — and an address no visible meeting mentions also states that a hidden meeting exists. `tests/test_acl.py::test_fireflies_channel_members_name_only_readable_meetings` pins both callers on a two-meetings-one-channel corpus (SAMPLE’s hidden transcript is alone in its channel, so it cannot express this); it fails on the pre-fix resolver.

backlot/graphql/fireflies_resolvers.py
khj809: Traced the same way (my counts run one lower than yours — no ACL statement on an admin token): 1 / 21 / 41 / 21 for the four selections.

Both halves closed. `store.fireflies_speakers(conn, ids)` is one `WHERE transcript_id IN (…)` over the page, and `resolve_transcripts` registers the page’s ids on the request context (accumulated, not assigned, so two aliased `transcripts` fields do not drop each other’s). `_speakers` memoises on that context, so `speakers` and `analytics.speakers` read one load. Now 1 / 2 / 2 / 2.

`tests/test_fireflies.py::test_fireflies_a_page_costs_the_same_queries_as_one_transcript` counts statements over a 5-transcript page and asserts 1/2/2/2; removing either the batch or the memo takes it to 6.

backlot/graphql/fireflies_resolvers.py
khj809: Confirmed — 20 identical statements over the fixture, all-hands for all 20.

`resolve_channel_members` now caches on `channel["_channel"]` for the request. Keyed on the channel and not the transcript because the roster is the channel’s property and `visible_ids` is fixed for the request, so two transcripts in one channel have the same answer. The 20-transcript page is 2 statements, asserted in the same query-count test.

backlot/store.py
khj809: Reproduced with a corpus stating the numbers itself (the BYO loader assigns one ordinal per name, so two unlabelled speakers need the record to distinguish them): sentences 0/None, 1/None, 2/Ava Chen served a roster of 1.0 and 2.0, and `{float(s.speaker_id)} == {sp.id}` failed on the missing 0.

`fireflies_speakers` returns rows grouped `BY transcript_id, speaker_id`, so the roster is one entry per number — 0, 1, 2 above. Where one number carries several labels the first the transcript used is served (one `min(seq)` aggregate, so SQLite takes the bare `speaker_name` from that row). `_analytics_speakers` builds its name→number lookup from the same rows; the analytics JSON aggregates by name, so unlabelled runs stay ONE statistics entry served under one of their numbers, while the roster keeps every number — the docstring says so. Pinned in `tests/test_store.py::test_fireflies_speakers_are_keyed_on_the_number_not_the_name` (fails under `GROUP BY speaker_name`).

tests/test_fireflies.py
khj809: Now equality: the `channel_id:` round-trip query already there selects `participants` too, and the assertion is `members == the union of the participants of that channel’s visible transcripts`. The scoped side — where the superset was the leak — is `tests/test_acl.py::test_fireflies_channel_members_name_only_readable_meetings`.

backlot/store.py
khj809: The measurement is right and the asymmetry reproduces (`title:"LATENCY"` finds `Latency review`; `"örebro"` misses `Örebro sync` while `"ÖREBRO"` finds it) — but `lower(title) LIKE lower(?)` does not make it one behaviour. SQLite’s `lower()` is ASCII-only as well, so it pre-folds exactly what LIKE folds:

```
%örebro%   plain=[]              lower=[]              casefold=[Örebro sync]
%ÖREBRO%   plain=[Örebro sync]   lower=[Örebro sync]   casefold=[Örebro sync]
```

One behaviour is reachable, via a `str.casefold` function registered on the connection (the third column). Three things argued against it, so I kept LIKE and stated the rule instead:

- The vendor is verified case-insensitive for ASCII only in the sense that ASCII is all I can observe: `title:"TEST"` and `keyword:"TEST"` both return the meeting titled `test` on `api.fireflies.ai`. Nothing in that workspace has a cased non-ASCII title, and settling it would mean renaming a transcript in a real workspace.
- A Python callback runs per row: at bench scale (10,173 transcripts, 112 MB of `content`) the `keyword` scan goes 73 ms → 102 ms, and it couples every store read to a store-opened connection, since a raw `sqlite3.connect` would then fail with "no such function".
- Reachability: 1 of the 10,173 bench titles carries a cased non-ASCII letter at all (`Communiqué`), and it matches under either spelling.

So the SDL now says it for both arguments — `title` and `keyword` match case-insensitively, folding ASCII, so a cased non-ASCII letter has to be spelled as the text spells it — and the store comment records why the expression is bare LIKE rather than inherited from `list_gdrive_files_by_name`. Drive’s `name contains` is left alone: same expression, but its own vendor check.

backlot/graphql/fireflies_resolvers.py
khj809: The live API settles it as UTC, so the sentence can say so outright rather than hedging. Probing the two transcripts in that workspace (07:15Z and 07:34Z on 2026-08-20):

```
date=2026-08-19T23:30Z  []          date=2026-08-20T00:30Z  both
date=2026-08-21T01:00Z  []          date=2026-08-20T23:30Z  both
```

A workspace zone ahead of UTC would have matched the first, one behind would have matched the third. The SDL now reads "Meetings recorded on this day, UTC, as epoch millis anywhere within it", and the resolver comment states the same with what rules the alternatives out.

backlot/graphql/fireflies_resolvers.py
```diff
@@ -153,23 +176,76 @@ def _summary(row) -> dict:
...
+
+
+def _analytics_speakers(row, speakers: list) -> list[dict]:
+    """`analytics.speakers` is Fireflies' AnalyticsSpeaker: talk-time statistics keyed by the same
+    per-meeting speaker number the sentences carry. `speakers` is the roster those numbers come
+    from, joined to the stored statistics by NAME — the only key the analytics JSON has, since
+    synth.fireflies_speaker_stats aggregates by it. Runs diarization never labelled are therefore
+    ONE statistics entry however many numbers they cover, served under one of them, while the
+    roster keeps every number."""
+    numbers = {s["speaker_name"]: s["speaker_id"] for s in speakers}
```
nuri-yoo, line 203: The roster is keyed on the number now, and this map turns it back into names, so two numbers sharing a label collapse and the statistics are served under the higher one. A corpus stating its own numbers:

```
sentences   0 "Ava Chen", 1 "Ava Chen"
speakers    0.0 "Ava Chen", 1.0 "Ava Chen"      # both, which is the fix
analytics   speaker_id 1, duration_pct 100.0    # one entry, and 0 is not named
```

Not introduced here -- pre-fix the same corpus served `speaker_id 1` too, and the roster is what changed. And it cannot be resolved from the analytics JSON, which has one entry per name. What the docstring already says covers the unlabelled case ("ONE statistics entry however many numbers they cover, served under one of them"); the same sentence holds for a repeated label, and saying so would stop a reader from taking `analytics.speakers` as one entry per number now that the roster is.

nuri-yoo (approved):
Re-ran the three defects and the two minors against `9124bae`, with the reproductions rebuilt rather than the diff read for them.

The channel roster is scoped: on the two-meetings-one-channel corpus `ava@acme.com` gets `ava, bob` and `hana@acme.com` gets all four, where before ava was served `hana` and `secret.investor` from the meeting she was denied. The page is batched: over a 20-transcript page my counts are 2 / 3 / 3 / 3 for metadata, `speakers`, `speakers` + `analytics.speakers`, and `channels { members }`, against 2 / 22 / 42 / 22 before, and the one extra against your 1 / 2 / 2 / 2 is the ACL statement a scoped token adds. The roster keeps every number: sentences 0/None, 1/None, 2/Ava Chen now serve speakers 0.0, 1.0, 2.0, and `{float(s.speaker_id)} == {sp.id}` holds.

The three new tests fail on `1cf1e0e`, so each pins its property rather than describing it: `AssertionError: assert 6 == 2` on the query count, and both the ACL and the store test outright. `test_fireflies_serves_the_documented_metadata_surface` passes there, so the equality is a tightening rather than the pin -- my comment claimed it would fail today and that was wrong: it runs on an admin token, and an unscoped caller sees the same roster either way. `tests/test_acl.py` is where the property lives.

On `title`: your correction is right and my suggestion was not. `lower(title) LIKE lower(?)` folds nothing extra on SQLite 3.50.4 -- `%örebro%` returns nothing under both, while a `casefold` callback returns `Örebro sync`, exactly your third column. Stating the rule in the SDL for both arguments is the right resolution, and I am taking the 73 ms to 102 ms and the one-title-in-10,173 on your measurement. Same for the UTC probe: I have no workspace token, so the two-transcript boundary sweep is yours, and the SDL now says UTC outright.

Suite here: `1204 passed, 3 skipped` (ERB e2e opt-in, docker, `llama_index.readers.hubspot`), 1207 collected, exit 0, ruff clean over the five touched files.

Approving. One note below on `analytics.speakers`, which this commit does not change and which is a sentence rather than a behaviour.

backlot/graphql/fireflies_resolvers.py
khj809: Reproduced: sentences 0/"Ava Chen", 1/"Ava Chen" serve speakers 0.0 and 1.0 with `analytics.speakers` a single entry at `speaker_id 1`, `duration_pct 100.0`. And it is the roster that changed — the name-keyed map served `speaker_id 1` there before too.

Both sentences now say per LABEL rather than only naming the unlabelled case, in the builder and in the SDL where a client reads it:

> Per-speaker talk time and word counts, computed from the sentences. One entry per LABEL, where `Transcript.speakers` is one per number: the stored statistics are aggregated by name, so two numbers sharing a label are one entry here, under one of the two.

Left as a sentence rather than a behaviour, and the reason is now recorded: nothing in the stored analytics can separate them, since `synth.fireflies_speaker_stats` aggregates by name. Whether the vendor keys its own `analytics.speakers` per number is still unobservable here — `{ analytics { speakers { speaker_id } } }` comes back `paid_required`, `tier: pro_or_higher`, which is the gate #68 noted.

`f63db71`. Suite 1204 passed, 3 skipped; ruff clean.

PR #73, opened 2026-08-20
slack: a token that is present but unusable is not a missing one

### LINKED ISSUE #70: Slack: a malformed bearer token should answer invalid_auth, not not_authed
Slack answers a *malformed or unknown* bearer token with `invalid_auth`, and a *missing* one with `not_authed`. Backlot answers `not_authed` to both, so a connector that branches on the two — retry vs. re-authenticate is the usual split — takes the wrong branch against the mock.

Measured 2026-08-20 against the live API. Both calls need no account, so this reproduces anywhere:

```console
$ curl -s -H "Authorization: Bearer xoxb-not-a-real-token" https://slack.com/api/conversations.list
{"ok":false,"error":"invalid_auth"}

$ curl -s -H "Authorization: Bearer xoxb-not-a-real-token" localhost:8000/slack/api/conversations.list
{"ok":false,"error":"not_authed"}
```

The two agree where a credential is absent entirely, which is the part that is already right:

```console
$ curl -s https://slack.com/api/conversations.list
{"ok":false,"error":"not_authed"}

$ curl -s localhost:8000/slack/api/conversations.list
{"ok":false,"error":"not_authed"}
```

`api.test` also agrees byte-for-byte (`{"ok":true,"args":{}}`), so the envelope machinery is fine — this is specifically about which of the two auth errors gets chosen.

Worth deciding as part of the fix: what counts as "malformed or unknown" for a mock whose tokens are derived from the corpus. A token that is well-formed but belongs to no principal is the interesting case, and it is the one a connector's test suite is most likely to construct.

### LINKED ISSUE #71: Slack: conversations.list omits pending_shared and is_pending_ext_shared
Slack's published example response for `conversations.list` carries 24 fields on a channel object. Backlot serves 22 of them, with no extra fields and matching nesting. The two it omits are `pending_shared` (an array) and `is_pending_ext_shared` (a boolean), both part of the shared-channel family that Backlot already answers in part — `is_shared`, `is_ext_shared`, `is_org_shared` are all present.

Measured 2026-08-20 against the field set at <https://docs.slack.dev/reference/methods/conversations.list>:

```
documented by Slack : 24
served by Backlot   : 22
missing in Backlot  : ['is_pending_ext_shared', 'pending_shared']
extra in Backlot    : none
top-level keys      : ['channels', 'ok', 'response_metadata']   # matches
nested topic keys   : ['creator', 'last_set', 'value']           # matches
```

Reproduce with a served corpus:

```console
$ curl -s "localhost:8000/slack/api/conversations.list?token=admin-service-token" \
    | jq '.channels[0] | keys'
```

Both are constants for this mock — nothing in a BYO corpus expresses a pending external share — so the fix is to emit `"pending_shared": []` and `"is_pending_ext_shared": false` beside the three `is_*_shared` fields. That makes the omission a two-line change rather than a modelling question.

The reason it matters even though the values are inert: a connector that projects the channel object, or validates it against a generated model, sees a shape that real Slack never returns. `conversations.history` has no equivalent gap — it serves all five documented top-level fields and the documented message fields.

Worth a test that diffs the served field set against a committed transcription of Slack's documented example, so a future divergence fails CI instead of waiting to be noticed. `assets/slack-doc.json` on the README branch already holds such a transcription for `conversations.history` and could be the pattern.

khj809:
Closes #70. Closes #71.

**#70 — the two auth errors.** Slack answers `invalid_auth` when a credential was presented and
does not resolve, and `not_authed` only when none was presented. This answered `not_authed` to
both, so a connector branching on them — re-authenticate vs. give up — took the wrong branch.

Measured against `slack.com`, which answers both without an account. It also settles the open
question in the issue: a merely *unknown* token answers the same as a malformed one, so
"malformed" is not a category to recognise — the only question is whether a non-empty credential
arrived.

| Request | Answer |
|---|---|
| no header, no `token` | `not_authed` |
| `Authorization: Bearer` (empty value) | `not_authed` |
| a non-Bearer scheme (`Basic …`) | `not_authed` |
| `token=` (empty value) | `not_authed` |
| any non-empty token that fails to resolve | `invalid_auth` |

`_caller_or_error` decides this once for all eleven methods, so none is left on the old single
error. `_caller` is gone with it — every call site needs the error, not just the caller.

**#71 — the shared-channel family.** `conversations.list` and `.info` build the same channel
object and served three of the five shared-channel fields; `pending_shared` and
`is_pending_ext_shared` are now served beside them. Both are constants — no BYO corpus expresses a
pending external share — but a client validating the object against a generated model saw a shape
real Slack never returns. Confirmed against the live API, which sends both.

A test pins the documented field set as a **floor** for both methods, so a documented field that
goes missing fails in CI. Not an equality check: the live API sends 29 keys where the
`conversations.list` method page documents 24, and `.list` and `.info` do not agree with each other
— both gaps are filed as #74, along with the missing `blocks` on messages. (The issue suggested
reusing `assets/slack-doc.json` from the README branch as the pattern; that file does not exist on
any branch, so the transcription is a constant in the test.)

Verified: 1209 passed, 2 skipped (all extras); ruff clean. Both fixes were mutation-checked — with
the router reverted, the new tests fail and name exactly the two missing fields and the three
`invalid_auth` cases. Trial-merged with #67: no conflicts, 1298 passed.

### DIFF
--- a/backlot/routers/slack.py
+++ b/backlot/routers/slack.py
@@ -2,7 +2,10 @@
 
 Base URL for a client: ``http://<host>/slack/api/`` (methods live under ``/api/``).
 Slack always returns HTTP 200 with an ``{"ok": bool}`` envelope, so auth failures are
-signalled as ``{"ok": false, "error": "not_authed"}`` rather than a 401 status.
+signalled in the body rather than as a 401 status. There are TWO of them and they are not
+interchangeable: a MISSING credential is ``not_authed``, a credential that is present but does not
+resolve is ``invalid_auth``. Connectors branch on exactly that split — re-authenticate versus give
+up — so ``_caller_or_error`` decides it once for every method here.
 """
 
 from __future__ import annotations
@@ -126,8 +129,27 @@ def _slack_ts(value: str | None) -> float | None | bool:
         return False
 
 
-def _caller(request: Request) -> Caller | None:
-    return auth.acl(request).resolve(auth.slack_token(request))
+def _caller_or_error(request: Request) -> tuple[Caller | None, dict | None]:
+    """Resolve the caller, or say which of Slack's two auth errors applies.
+
+    Measured against the live API: a token that is merely unknown answers the same as a malformed
+    one, so "malformed" is not a category this has to recognise — what matters is only whether a
+    non-empty credential was presented at all.
+
+        no header, no `token` param      -> not_authed
+        `Authorization: Bearer ` (empty) -> not_authed
+        `token=` (empty)                 -> not_authed
+        a non-Bearer scheme (e.g. Basic) -> not_authed
+        any non-empty token that fails   -> invalid_auth
+
+    ``auth.slack_token`` already returns None for every case in the first group: it parses only
+    the bearer/token schemes, and an empty query or form value is falsy.
+    """
+    token = auth.slack_token(request)
+    caller = auth.acl(request).resolve(token)
+    if caller is not None:
+        return caller, None
+    return None, _err("invalid_auth" if token else "not_authed")
 
 
 def _full_channel(request: Request, conn, name: str) -> dict:
@@ -150,6 +172,11 @@ def _full_channel(request: Request, conn, name: str) -> dict:
         "is_shared": False,
         "is_ext_shared": False,
         "is_org_shared": False,
+        # No BYO corpus expresses a pending external share, so these are constants — but the
+        # shared-channel family is five fields in every response Slack documents, and a client
+        # validating the object against a generated model sees a shape real Slack never returns.
+        "is_pending_ext_shared": False,
+        "pending_shared": [],
         "unlinked": 0,
         "created": created,
         "updated": created * 1000,
@@ -227,9 +254,9 @@ async def api_test(request: Request):
 
 @router.api_route("/auth.test", methods=["GET", "POST"])
 async def auth_test(request: Request):
-    caller = _caller(request)
-    if caller is None:
-        return _err("not_authed")
+    caller, err = _caller_or_error(request)
+    if err is not None:
+        return err
     who = "service-account" if caller.is_admin else caller.email
     return {
         "ok": True,
@@ -249,9 +276,9 @@ async def auth_test(request: Request):
 )
 async def conversations_list(request: Request):
     conn = auth.conn(request)
-    caller = _caller(request)
-    if caller is None:
-        return _err("not_authed")
+    caller, err = _caller_or_error(request)
+    if err is not None:
+        return err
     types = _slack_types(request)
     if types is None:
         return _err("invalid_types")
@@ -288,9 +315,9 @@ async def conversations_list(request: Request):
 )
 async def conversations_info(request: Request):
     conn = auth.conn(request)
-    caller = _caller(request)
-    if caller is None:
-        return _err("not_authed")
+    caller, err = _caller_or_error(request)
+    if err is not None:
+        return err
     name = _channel_name(conn, _param(request, "channel") or "")
     if name is None:
         return _err("channel_not_found")
@@ -305,9 +332,9 @@ async def conversations_info(request: Request):
 )
 async def conversations_history(request: Request):
     conn = auth.conn(request)
-    caller = _caller(request)
-    if caller is None:
-        return _err("not_authed")
+    caller, err = _caller_or_error(request)
+    if err is not None:
+        return err
     channel_id = _param(request, "channel")
     if not channel_id:
         return _err("channel_not_found")
@@ -392,9 +419,9 @@ async def conversations_history(request: Request):
 )
 async def conversations_replies(request: Request):
     conn = auth.conn(request)
-    caller = _caller(request)
-    if caller is None:
-        return _err("not_authed")
+    caller, err = _caller_or_error(request)
+    if err is not None:
+        return err
     ts = _param(request, "ts")
     name = _channel_name(conn, _param(request, "channel") or "")
     if name is None or not ts:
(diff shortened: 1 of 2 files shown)


backlot/routers/slack.py
```diff
@@ -126,8 +129,27 @@ def _slack_ts(value: str | None) -> float | None | bool:
...
+    """Resolve the caller, or say which of Slack's two auth errors applies.
+
+    Measured against the live API: a token that is merely unknown answers the same as a malformed
+    one, so "malformed" is not a category this has to recognise — what matters is only whether a
+    non-empty credential was presented at all.
+
+        no header, no `token` param      -> not_authed
+        `Authorization: Bearer ` (empty) -> not_authed
+        `token=` (empty)                 -> not_authed
+        a non-Bearer scheme (e.g. Basic) -> not_authed
```
nuri-yoo, line 142: Minor, and not a reason to hold this. Two schemes reach `invalid_auth` here that real Slack answers `not_authed` for, so the table is narrower than it reads. Measured against slack.com alongside the five rows above:

```
Authorization: bearer xoxb-bogus   ->  not_authed     (mock: invalid_auth)
Authorization: token  xoxb-bogus   ->  not_authed     (mock: invalid_auth)
```

`auth.bearer_token` matches `parts[0].lower() in ("bearer", "token")`, so it accepts the scheme case-insensitively and also accepts GitHub's legacy `token` form. Both are defensible there -- RFC 7235 makes the scheme case-insensitive, and the GitHub router needs `token <t>` -- so the place to narrow is this predicate rather than the shared parser: what Slack treats as "a credential was presented" is an exact `Bearer ` prefix, the `token` query param, or the `token` form field.

If that is more precision than the split is worth, the cheaper fix is the line above: "a non-Bearer scheme (e.g. Basic)" reads as though every scheme that is not Bearer lands in the first group, and two of them do not.

tests/test_slack.py
```diff
@@ -107,6 +155,74 @@ def test_slack_conversations_list_defaults_to_public_channels(client, admin_h):
...
+        "pending_shared",
+        "is_pending_ext_shared",
+        "is_member",
+        "is_private",
+        "is_mpim",
+        "updated",
+        "topic",
+        "purpose",
+        "previous_names",
+        "num_members",
```
nuri-yoo, line 192: `num_members` in the floor is right for `conversations.list`, whose example documents it, but this set is also asserted against `.info` at line 215, and Slack gates it there: `include_num_members` is an optional argument, "Set to `true` to include the member count for the specified conversation", default `false`. At this head the argument is ignored altogether --

```
info (no flag)                   num_members present, 1
info include_num_members=false   num_members present, 1
info include_num_members=true    num_members present, 1
```

-- which predates this branch (`test_slack.py:347` already pins it). Nothing to do here: #75 accepts the argument and asserts `"num_members" not in info` for the plain call, which is exactly this. Noting it only so the two floors are read together -- this one has to keep `num_members` for `.list` while #75 subtracts it for `.info`.

Ref: https://docs.slack.dev/reference/methods/conversations.info

tests/test_slack.py
```diff
@@ -107,6 +155,74 @@ def test_slack_conversations_list_defaults_to_public_channels(client, admin_h):
...
+        "/slack/api/conversations.list", headers=admin_h, params={"limit": 1}
+    ).json()["channels"][0]
+    assert _DOCUMENTED_CHANNEL_KEYS <= set(listed)
+
+    # .info builds its channel from the same helper, so it must not drop what .list answers. The
+    # two are NOT the same object in real Slack, which is #74 — this only pins the documented floor.
+    info = client.post(
+        "/slack/api/conversations.info", headers=admin_h, data={"channel": listed["id"]}
+    ).json()["channel"]
+    assert _DOCUMENTED_CHANNEL_KEYS <= set(info)
```
nuri-yoo, line 215: The five live-only keys are served in #75, so the floor catches up there. The one thing that does not carry over is where the comment above says they are documented -- "on the conversation object reference rather than the method page" -- which holds for `conversations.list` but not for `.info`: its own method page's example response carries `is_frozen`, `context_team_id`, `parent_conversation`, `shared_team_ids`, `pending_connected_team_ids` and `properties`.

That matters for how the gap reads rather than for what gets served: #75 serves five of those six and skips `is_frozen` on measured grounds, and its description repeats the same attribution. Worth correcting in whichever of the two branches you touch next, and worth giving `.info` its own transcription eventually, since the two method pages differ.

Ref: https://docs.slack.dev/reference/methods/conversations.info

nuri-yoo (approved):
LGTM.

Reviewed at `c7a8c78`. Suite `1208 passed, 3 skipped` here (the extra skip is `llama_index.readers.hubspot`, absent from my env); ruff clean; trial-merge clean against both #67 and #72, which share no file with this one.

The auth table reproduces. I ran all five rows against slack.com and the mock answers the same in every one, including the two that could have gone either way: `Authorization: Bearer` with an empty value is `not_authed` on both, and `token=%20` is `invalid_auth` on both, so a whitespace-only value counts as presented on each side. `_caller` is gone with no leftovers, eleven of the twelve methods route through `_caller_or_error`, and the twelfth is `api.test`, which real Slack answers `{"ok":true}` for with no credential at all -- so leaving it out is right rather than an omission.

The mutation check reproduces as described: reverting the two fields fails the field-set test naming `is_pending_ext_shared`, and reverting the split fails the three `invalid_auth` parameters plus the uniformity test. The floor constant is a faithful transcription of the `conversations.list` example, `num_members` included.

Three minor things inline, none blocking, and only the first belongs to this branch: two more schemes still diverge on the axis this fixes. The other two are already answered by #75 on top of this -- it accepts `include_num_members` and subtracts `num_members` from the `.info` assertion, and it serves the five conversation-object fields -- so those notes only flag the one thing #75 does not carry over, which is where `.info`'s extra fields are documented.

PR #75, opened 2026-08-20
slack: conversations.* answer the shape the live API does

### LINKED ISSUE #74: slack: conversations.* answer a shape the live API does not, starting with blocks
`conversations.list`, `conversations.info` and `conversations.history` each answer a shape the real
API does not. Measured 2026-08-20 against `slack.com` with a workspace user token
(`channels:read`, `channels:history`) over 8 public channels, so these are observations rather than
readings of the reference — which matters here, because the method pages' example responses turn
out to be abridged in both directions.

## 1. A message carries no `blocks`

The biggest one. Every human-typed message the live API returns carries `blocks`; this serves none.
A client that renders rich text, or validates the message against a generated model, sees a shape
real Slack never returns.

`text` keeps the original markup in the live response, so **nothing about `content` has to change**
— `blocks` is derived beside it. The four shapes, measured:

```jsonc
// "test"
{"type":"rich_text","block_id":"gB9fq","elements":[
  {"type":"rich_text_section","elements":[{"type":"text","text":"test"}]}]}

// "```code fence```"  — text field still carries the backticks
{"type":"rich_text","block_id":"qv5j2","elements":[
  {"type":"rich_text_preformatted","elements":[{"type":"text","text":"code fence"}],"border":0}]}

// "• bullet 1\n• bullet 2"  — text field still carries the bullets
{"type":"rich_text","block_id":"Qy+wn","elements":[
  {"type":"rich_text_list","style":"bullet","indent":0,"border":0,"elements":[
    {"type":"rich_text_section","elements":[{"type":"text","text":"bullet 1"}]},
    {"type":"rich_text_section","elements":[{"type":"text","text":"bullet 2"}]}]}]}

// "`inline` *bold*"  — one section, one text element per run, whitespace its own element
{"type":"rich_text","block_id":"SYhEm","elements":[
  {"type":"rich_text_section","elements":[
    {"type":"text","text":"inline","style":{"code":true}},
    {"type":"text","text":" "},
    {"type":"text","text":"bold","style":{"bold":true}}]}]}
```

`block_id` is 5 characters of base64 alphabet (`+` observed), so it can be seeded off
`(channel, ts)` the way `client_msg_id` already is.

How far to parse is answered by what the corpus actually holds. Measured over the 410 messages the
ERB slack corpus parses into:

| markup | share |
|---|---|
| code fence ` ``` ` | 11.0% |
| emoji `:name:` | 9.5% |
| inline code | 9.0% |
| multi-line | 13.9% |
| bare URL | 4.9% |
| `@name` | 5.1% |
| bullet list | 2.4% |
| bold / italic | 0.5% / 0.2% |
| **no markup at all** | **62.7%** |
| `<@U…>` / `<#C…>` / `<http…\|label>` | **0%** |

So: sections, preformatted, lists and inline styles are all reachable and worth handling. Slack's
own `<@U…>`-style encodings are **not** — the corpus is plain prose, so a mention/link/channel
element parser would be code that never fires. `erb.parse_slack_transcript` is already fence-aware
and joins lines verbatim, so the markup survives import intact.

## 2. `client_msg_id` is served on messages that cannot have one

`client_msg_id` is minted by the client that posted the message. The live API omits it on messages
Slack itself generates — all three `channel_join` messages observed carry only
`type`/`user`/`text`/`ts`/`subtype`, with no `client_msg_id`, `team` or `blocks`. This serves
`client_msg_id` unconditionally, including on rows the corpus marked with a `subtype`.

Scope note: the ERB importer never sets a slack `subtype` (`_byo_slack` writes none), so this is
reachable only from a BYO corpus that states one — `slack.schema.json` does define the field, and
`examples/bring-your-own-corpus/sample_corpus.jsonl` carries a `bot_message`. Low blast radius,
but wrong where it fires.

Deliberately narrow: only `client_msg_id` is proposed for exclusion. `team` is left alone. It is
absent from `channel_join` too, but a `bot_message` is a real posted message rather than a Slack
system notice, and there is no bot message in the measured workspace to check it against — so
excluding `team` on every `subtype` would be generalising past the evidence. `bot_id`, `app_id` and
`username` are out of scope: the mock does not model bot identity and none was observed.

## 3. `conversations.list` omits five fields the live API always sends

All 8 channels returned an identical 29-key object. Five of those keys are not served here:

| field | live value | proposed |
|---|---|---|
| `context_team_id` | the workspace id | the existing `T0000MOCK` constant |
| `shared_team_ids` | `[<workspace id>]` | `["T0000MOCK"]` |
| `pending_connected_team_ids` | `[]` | `[]` |
| `parent_conversation` | `null` | `null` |
| `properties` | an object | `{}` |

All five are documented on the [conversation object
reference](https://docs.slack.dev/reference/objects/conversation-object/) — the `conversations.list`
method page's example simply omits them, which is what #71 was measured against.

`properties` is `{}` rather than populated on purpose. The reference calls its sub-fields
contextual (tabs, canvas, posting restrictions, huddles, …) and the observed values are workspace
UI state — channel canvases, tab layouts — that no BYO corpus expresses. The key is always present
so it must be served, but filling it would invent configuration. For the same reason `use_case`
(observed under `properties` on the two default channels only) stays unserved.

## 4. `conversations.list` and `conversations.info` do not return the same object

`_full_channel` builds one object for both. The live API does not:

| field | `.list` | `.info` |
|---|---|---|
| `num_members` | yes | no |
| `last_read` | no | yes (caller is a member) |
| `is_read_only`, `is_thread_only` | no | on one channel |
| `properties.is_dormant` | yes | no |

The reference agrees: `last_read`/`unread_count` are member-only, `num_members` "may not appear"
for some conversation types. Splitting the two is a structural change to the helper, not a field
addition.

## 5. `conversations.history` omits two top-level fields

Live returns `channel_actions_ts` (null) and `channel_actions_count` (0) on every call; neither is
served, and neither is on the method page. `response_metadata` appeared only when there was a
cursor. `is_limited` also appeared, but the measured workspace is on a free plan and that field
looks like a plan artefact, so it is **not** proposed here — it needs a paid workspace to confirm.

## Deliberately not served

Conditional fields whose condition cannot hold in a corpus this mock serves, all documented on the
conversation object reference: `is_frozen` and `frozen_reason` (Slack Connect / frozen channels —
absent from both `.list` and `.info` on all 8 measured channels, despite appearing in the
`conversations.info` doc example), `enterprise_id`, `connected_team_ids`, `conversation_host_id`,
`internal_team_ids`, `is_ext_ws_shared`, `is_org_default`, `is_org_mandatory`, and the DM-only
`is_open` / `user`. The mock hardcodes `is_shared`/`is_ext_shared`/`is_org_shared` to false and has
no DMs, so serving any of these would be inventing a configuration it does not emulate.

## Scope

Serving-level throughout — no DDL and no import-schema change. `blocks` derives from the existing
`content`; the rest are constants or a restructuring of `_full_channel`. `slack_messages` and
`slack_channels` are untouched.

## Follow-up on #73

#73 (which closes #70 and #71) added a test asserting the served channel object equals Slack's
documented 24-key set exactly, and that `.list` and `.info` agree. Both assertions are now known to
be wrong about the live API: it returns 29 keys, and the two methods differ (items 3 and 4). The
fix in #73 — adding `pending_shared` and `is_pending_ext_shared` — is confirmed correct by the same
measurement; only the assertions need loosening to a subset check.

khj809:
Closes #74.

> **Stacked on #73** — based on `worktree-issue-70-71-slack-auth-and-channel-fields`, which touches
> the same two functions. Review the commits above `worktree-issue-70-71-slack-auth-and-channel-fields`;
> merge #73 first and this rebases to nothing.

Measured against `slack.com` with a workspace token, then diffed key-for-key. After this, the
served shapes match the live ones exactly:

| | live | mock | missing | extra |
|---|---|---|---|---|
| `conversations.list` channel | 29 | 29 | — | — |
| `conversations.info` channel | 29 | 29 | — | — |
| `conversations.history` message | 7 | 7 | — | — |
| `conversations.history` thread root | 15 | 15 | — | — |

**A message carries `blocks`.** Every human-typed message the live API returns has one; this served
none. Derived rather than stored and needing no column — a real response keeps the original markup
in `text` too, so the block is a second rendering of the same string. Four shapes, each transcribed
from a live response and pinned against it: fenced runs become `rich_text_preformatted` (with the
`border` the payload carries), consecutive bullet lines become one `rich_text_list`, inline
`code`/`*bold*`/`_italic_`/`~strike~` become one text element per run carrying `style`, and the rest
stays a `rich_text_section`. Slack's `<@U…>`/`<#C…>`/`<http…|label>` encodings are deliberately not
parsed — measured across the ERB slack corpus they appear in 0% of 410 messages, so that branch
would never run.

**`.list` and `.info` are not the same object.** Building one for both made this mock the only
place a client could rely on it. `num_members` is a `.list` field that `.info` answers only for
`include_num_members=true` — the vendor's own parameter, now accepted — and `last_read` is
`.info`-only and member-only.

**Five fields sent on every live response.** `context_team_id`, `shared_team_ids`,
`pending_connected_team_ids`, `parent_conversation`, `properties`. Where they are documented differs
by method, which is why #71's transcription missed them: the
[`conversations.list` page](https://docs.slack.dev/reference/methods/conversations.list)'s example
response carries none of them and only the
[conversation object reference](https://docs.slack.dev/reference/objects/conversation-object/)
describes them, while the
[`conversations.info` page](https://docs.slack.dev/reference/methods/conversations.info)'s own
example carries all six — the five above plus `is_frozen`, which is still skipped here because the
live response did not send it. The two method pages differ enough to deserve separate
transcriptions. `properties` is served
**empty**: its sub-fields are channel configuration (tabs, a canvas, posting restrictions) that no
corpus states, and furnishing them would invent settings this workspace does not have.

**`client_msg_id` and `blocks` are dropped from a message with a `subtype`.** Both are made by the
posting client, and a live `channel_join` answers only `type`/`user`/`text`/`ts`/`subtype`. `team`
is kept: it is absent from a `channel_join` too, but a `bot_message` is a real posted message and
there was none to measure, so dropping it everywhere would generalise past the evidence.

**`conversations.history`'s envelope** gains `channel_actions_ts` and `channel_actions_count`, and
**omits `response_metadata` on a last page** — the live API drops the key entirely rather than
sending an empty cursor, so a client that subscripts it to decide whether to keep paging worked
here and raised against real Slack. The crawler in `tests/_helpers.py` was such a client and now
terminates on the key's absence.

`conversations.list` is left alone, and the evidence for the two is not symmetric. `.history`
omitting the key is a direct observation of a last page. For `.list` the captured pages carry
`{"next_cursor": ""}`, and Slack's own `.list` example shows the key with a *non-empty* cursor —
neither is an observation of a `.list` last page, which needs a workspace token this environment
does not have. So `.list` keeps the key: matching `.history` would be an inference from one method
to the other, and the whole point of splitting them is that they do not agree. Anyone who can run a
`.list` to exhaustion against a real workspace should settle it.

**`last_read` is sent whatever the caller can read.** It is the *caller's* field, so its value
varies by who asks — that is what it means. Its presence must not: gating the key on visibility gave
`.info` two shapes (29 keys or 28), which defeats the point of pinning the object exactly, and made
the key an ACL oracle — on a private channel `conversations.list` does not show the caller, a
present `last_read` said "you can read this" and an absent one said "you cannot". A caller with
nothing to have read now gets Slack's zero ts, `0000000000.000000`. A service token gets it too: it
is not a person and has read nothing, the same reasoning this branch already applies to
`subscribed`. The zero ts is reasoned from Slack's convention for an empty ts-shaped field rather
than measured — the documented example shows only a real one — but the shape being invariant does
not depend on which sentinel it is.

That `.info` resolves a channel the caller cannot see at all is older than this branch and left
alone: real Slack answers `channel_not_found` there, and `_channel_name` reads the container list
without an ACL clause.

**`client_msg_id` is a v4 UUID**, not the 16-hex token this minted. Every live value is canonical
8-4-4-4-12 with a v4's version nibble and variant bits, so anything validating the field failed
while anything merely reading it passed. `synth._uuid_from` already forces those bits for the same
reason on Notion/Linear ids and is reused.

**`subscribed` depends on who is asking.** It was hardcoded `false` on every thread root; it is the
caller's own state — Slack subscribes you to a thread you started or replied in — so a constant is
wrong in one direction for a participant and the other for everyone else. Now derived from the
caller against the root's author and the reply authors; a service token subscribes to nothing.
`is_locked` is added beside it, present on every live thread root.

Measuring this needed a thread, which confirmed the shape too: in one live `conversations.history`
response a root with replies answered **15** keys where a plain message answered **7** and a system
message **5**, and the root was a strict superset — the thread fields are added, never a
substitution. That is what this mock already did, so only the two fields above changed.

### Deliberately not served

`is_frozen`/`frozen_reason`, `use_case`, `enterprise_id`, `connected_team_ids`,
`conversation_host_id`, `internal_team_ids`, `is_ext_ws_shared`, `is_org_default`,
`is_org_mandatory`, and the DM-only `is_open`/`user` — conditional fields whose condition cannot
hold here. `is_frozen` in particular was absent from both live methods on all 8 measured channels
despite appearing in the `conversations.info` doc example.

`last_read` is not added to messages either, though it sits beside `subscribed` in the live
payload: it is per-caller too, and the observed value equalled `latest_reply` only because that
caller had just read the thread. `unread_count` appears in Slack's documented example but not in
the live response, so it is not served.

`is_limited` is also not served: observed only on a free-plan workspace, so it looks like a plan
artefact rather than part of the shape.
**Only one spelling of `Bearer` is a credential to Slack.** Slack recognises exactly three ways
to present a token — an `Authorization: Bearer <t>` header, a `token` query param, or a `token` form
field — and is stricter about each than the shared `auth.bearer_token` this was routed through. That
parser matches the scheme case-insensitively and also accepts GitHub's legacy `token <t>`; both are
right where they are, since GitHub really does take `token <t>` and RFC 7235 really does make the
scheme case-insensitive. Slack implements neither, so it gets its own parser. Measured against
slack.com, which needs no account to answer:

| presented (`invalid_auth`) | absent (`not_authed`) |
| --- | --- |
| `Bearer <t>` | `bearer <t>`, `BEARER <t>`, `BeArEr <t>` |
| `Bearer  <t>` (repeated spaces) | `token <t>`, `Token <t>` |
| `' Bearer <t>'`, `'Bearer <t> '` | `Bearer<TAB><t>`, `Bearer<t>` |
| `?token=<t>`, form `token=<t>` | `?TOKEN=<t>`, form `TOKEN=<t>` |

The error label is the visible half. The half that matters is that **six spellings authenticated**
here, so a client sending `Authorization: token <xoxb>` passed every test and reached nothing in
production — `test_slack_refuses_a_valid_token_under_an_unrecognised_scheme` pins that direction
with a *valid* token, and asserts `Bearer` still works so this is a narrowing and not a refusal.

Verified: 1241 passed, 3 skipped (all extras); ruff clean. Mutation-checked, each restored from a
copy rather than from HEAD — routing Slack back through the shared parser fails 8 of the new auth
cases, restoring the conditional `last_read` fails both shape tests, and reverting the two
`conversations.*` files fails the pagination and UUID tests. The three skips are environmental:
Docker absent for the MCP test, `ERB_E2E` unset, and `llama_index.readers.hubspot` not installed.

### DIFF
--- a/backlot/routers/slack.py
+++ b/backlot/routers/slack.py
@@ -168,7 +168,7 @@ def _channel_core(request: Request, conn, name: str) -> dict:
 
     What each adds is deliberately not here, because the real API does not agree between the two:
     `num_members` is a `.list` field that `.info` returns only for `include_num_members=true`, and
-    `last_read` is `.info`-only (and member-only). Building one object for both made this mock the
+    `last_read` is `.info`-only (and the caller's). Building one object for both made this mock the
     only place a client could rely on them matching.
     """
     is_private = not store.container_has_public(conn, "slack", name)
@@ -218,18 +218,38 @@ def _listed_channel(request: Request, conn, name: str) -> dict:
     return {**_channel_core(request, conn, name), "num_members": _member_count(request, conn, name)}
 
 
+# Slack's zero timestamp, its "nothing here yet" value for a ts-shaped field. Reasoned from the
+# convention rather than measured: reading the live `last_read` of a channel nobody has opened needs
+# a real token, which this environment has none of.
+ZERO_TS = "0000000000.000000"
+
+
+def _last_read(conn, name: str, caller: Caller, visible_ids) -> str:
+    """The caller's `last_read` for a channel — always a ts, never absent.
+
+    A service token gets the zero ts rather than the channel's newest message: it is not a person
+    and has read nothing, the same reasoning `_subscribed` applies to a thread it cannot have
+    followed. Otherwise this mock models no unread state, so a channel the caller can read reads as
+    caught up."""
+    if not (caller.email or ""):
+        return ZERO_TS
+    return store.slack_latest_ts(conn, name, visible_ids) or ZERO_TS
+
+
 def _info_channel(
-    request: Request, conn, name: str, *, include_num_members: bool, visible_ids
+    request: Request, conn, name: str, *, include_num_members: bool, visible_ids, caller: Caller
 ) -> dict:
     """conversations.info's channel. `num_members` is opt-in here (the vendor's own parameter), and
-    `last_read` is member-only — this mock's caller is always a member of what it can see, and
-    models no unread state, so the whole channel reads as caught up."""
+    `last_read` is the caller's.
+
+    Being the caller's, its VALUE varies by who asks — that is what the field means. Its PRESENCE
+    must not: a key that appears only when the caller can read the channel gives `.info` two shapes,
+    and on a private channel `conversations.list` does not show this caller, the missing key is a
+    straight answer to "can you read this?". A caller with nothing to have read gets `ZERO_TS`."""
     ch = _channel_core(request, conn, name)
     if include_num_members:
         ch["num_members"] = _member_count(request, conn, name)
-    latest = store.slack_latest_ts(conn, name, visible_ids)
-    if latest is not None:
-        ch["last_read"] = latest
+    ch["last_read"] = _last_read(conn, name, caller, visible_ids)
     return ch
 
 
@@ -373,6 +393,7 @@ async def conversations_info(request: Request):
         name,
         include_num_members=want_members,
         visible_ids=auth.visible_ids(request, caller),
+        caller=caller,
     )
     return {"ok": True, "channel": ch}
 
--- a/backlot/routers/slack.py
+++ b/backlot/routers/slack.py
@@ -450,7 +450,7 @@ async def conversations_history(request: Request):
             )
         )
     cursor = next_cursor(offset, len(rows), total)
-    return {
+    body = {
         "ok": True,
         "messages": messages,
         "has_more": bool(cursor),
@@ -459,8 +459,13 @@ async def conversations_history(request: Request):
         # real API sends both keys on every call rather than omitting them.
         "channel_actions_ts": None,
         "channel_actions_count": 0,
-        "response_metadata": {"next_cursor": cursor},
     }
+    # Omitted entirely on a last page, which is what the live API does here — NOT served with an
+    # empty cursor. conversations.list is the other way round (it carries `{"next_cursor": ""}`
+    # even with nothing more), so the two are not a shared convention to factor out.
+    if cursor:
+        body["response_metadata"] = {"next_cursor": cursor}
+    return body
 
 
 @router.api_route(
@@ -896,7 +901,7 @@ def _message(
         # Seeded on (channel, ts), not ts alone: a ts is unique within its channel (see
         # store.ID_COLUMNS), so the same second in two channels produced one client_msg_id — a
         # value real Slack makes globally unique.
-        m["client_msg_id"] = synth.gmail_id(seed, salt="cmid")
+        m["client_msg_id"] = synth.slack_client_msg_id(seed)
         blocks = synth.slack_blocks(text, seed)
         if blocks:
             m["blocks"] = blocks
--- a/backlot/routers/slack.py
+++ b/backlot/routers/slack.py
@@ -139,16 +139,22 @@ def _caller_or_error(request: Request) -> tuple[Caller | None, dict | None]:
 
     Measured against the live API: a token that is merely unknown answers the same as a malformed
     one, so "malformed" is not a category this has to recognise — what matters is only whether a
-    non-empty credential was presented at all.
+    credential was presented at all, and Slack recognises exactly three ways to present one:
+    an `Authorization: Bearer <t>` header, a `token` query param, or a `token` form field.
 
         no header, no `token` param      -> not_authed
         `Authorization: Bearer ` (empty) -> not_authed
         `token=` (empty)                 -> not_authed
-        a non-Bearer scheme (e.g. Basic) -> not_authed
+        any other header scheme          -> not_authed
         any non-empty token that fails   -> invalid_auth
 
-    ``auth.slack_token`` already returns None for every case in the first group: it parses only
-    the bearer/token schemes, and an empty query or form value is falsy.
+    "any other header scheme" is the whole of the rest, not just the obvious `Basic`: the scheme is
+    case-SENSITIVE live, so `bearer` and `BEARER` land here too, as does GitHub's legacy `token <t>`
+    form. `auth.slack_bearer_token` is what draws that line, and is deliberately not the shared
+    `auth.bearer_token` — see its docstring for why the two cannot be one parser.
+
+    ``auth.slack_token`` returns None for every case in the first group: an unrecognised scheme
+    parses to nothing, and an empty query or form value is falsy.
     """
     token = auth.slack_token(request)
     caller = auth.acl(request).resolve(token)
--- a/backlot/routers/slack.py
+++ b/backlot/routers/slack.py
@@ -69,8 +69,13 @@ class SlackSearch(_SlackOk):
     messages: dict = {}
 
 
-_P_LIST = [qp("limit", "integer"), qp("cursor")]
+_P_LIST = [qp("limit", "integer"), qp("cursor"), qp("types")]
 _P_CHANNEL = [qp("channel", required=True)]
+_P_INFO = [qp("channel", required=True), qp("include_num_members", "boolean")]
+
+# The one workspace this mock emulates. Every conversation is in it, shared with nobody, so the
+# sharing ids below are all this team or empty.
+TEAM_ID = "T0000MOCK"
 _P_HISTORY = [
     qp("channel", required=True),
     qp("limit", "integer"),
@@ -152,10 +157,15 @@ def _caller_or_error(request: Request) -> tuple[Caller | None, dict | None]:
     return None, _err("invalid_auth" if token else "not_authed")
 
 
-def _full_channel(request: Request, conn, name: str) -> dict:
-    """A full conversation object (shared by conversations.list and .info)."""
+def _channel_core(request: Request, conn, name: str) -> dict:
+    """The conversation object as BOTH conversations.list and .info answer it.
+
+    What each adds is deliberately not here, because the real API does not agree between the two:
+    `num_members` is a `.list` field that `.info` returns only for `include_num_members=true`, and
+    `last_read` is `.info`-only (and member-only). Building one object for both made this mock the
+    only place a client could rely on them matching.
+    """
     is_private = not store.container_has_public(conn, "slack", name)
-    num = _member_count(request, conn, name)
     created = _channel_created(request, conn, name)
     return {
         "id": synth.slack_channel_id(name),
@@ -177,6 +187,12 @@ def _full_channel(request: Request, conn, name: str) -> dict:
         # validating the object against a generated model sees a shape real Slack never returns.
         "is_pending_ext_shared": False,
         "pending_shared": [],
+        # Single-workspace, nothing shared or nested: the ids are this team and the rest empty.
+        # Absent here, a client generated from the vendor schema has no field to bind them to.
+        "context_team_id": TEAM_ID,
+        "shared_team_ids": [TEAM_ID],
+        "pending_connected_team_ids": [],
+        "parent_conversation": None,
         "unlinked": 0,
         "created": created,
         "updated": created * 1000,
@@ -184,8 +200,31 @@ def _full_channel(request: Request, conn, name: str) -> dict:
         "topic": {"value": f"#{name}", "creator": "USERVICE0", "last_set": created},
         "purpose": {"value": f"Channel for {name}", "creator": "USERVICE0", "last_set": created},
         "previous_names": [],
-        "num_members": num,
+        # Contextual channel configuration — tabs, a channel canvas, posting restrictions. The key
+        # is always present in a real response, but no corpus states any of it, so it is served
+        # empty rather than furnished with settings this workspace does not have.
+        "properties": {},
     }
+
+
+def _listed_channel(request: Request, conn, name: str) -> dict:
+    """conversations.list's channel: the core plus its member count."""
+    return {**_channel_core(request, conn, name), "num_members": _member_count(request, conn, name)}
+
+
+def _info_channel(
+    request: Request, conn, name: str, *, include_num_members: bool, visible_ids
+) -> dict:
+    """conversations.info's channel. `num_members` is opt-in here (the vendor's own parameter), and
+    `last_read` is member-only — this mock's caller is always a member of what it can see, and
+    models no unread state, so the whole channel reads as caught up."""
+    ch = _channel_core(request, conn, name)
+    if include_num_members:
+        ch["num_members"] = _member_count(request, conn, name)
+    latest = store.slack_latest_ts(conn, name, visible_ids)
+    if latest is not None:
+        ch["last_read"] = latest
+    return ch
 
 
 def _channel_names(conn) -> list[str]:
@@ -200,7 +239,7 @@ def _user_obj(conn, email: str) -> dict:
     is_bot = not u and email.split("@")[0].endswith("bot")  # display-only "*bot" speakers
     return {
         "id": synth.slack_user_id(email),
-        "team_id": "T0000MOCK",
+        "team_id": TEAM_ID,
         "name": email.split("@")[0].replace(".", ""),
         "real_name": display,
         "deleted": False,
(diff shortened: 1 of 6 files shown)


nuri-yoo (commented):
Reviewed at `4c3a906`, on top of #73. Suite `1223 passed, 3 skipped` here (the extra skip is `llama_index.readers.hubspot`, absent from my env); ruff clean over the tree.

The key-for-key claims reproduce on the repo's own fixture: `.list` 29 keys, `.info` 29, `.info` with `include_num_members` 30, list-only `num_members` and info-only `last_read`, a plain message 7 and a thread root 15 with the root a strict superset. `conversations.history` carries `channel_actions_ts`/`channel_actions_count` and omits `response_metadata` entirely on a last page while `.list` still sends `{"next_cursor": ""}`. `client_msg_id` comes out canonical v4 (`42b17dda-cfb7-4771-beac-63ae4a7bde20`, version nibble and variant bits included). A `channel_join` answers 6 keys with no `client_msg_id` and no `blocks`, `team` kept as the description explains. `subscribed` is the caller's: the root's author and a reply author both get `true`, a service token `false`. And `blocks` renders all four shapes in one message in the right order -- section, one `rich_text_list` for the bullet run, styled runs as separate text elements, `rich_text_preformatted` for the fence.

Two things I checked rather than assumed. `last_read` is ACL-scoped (`slack_latest_ts` takes `visible_ids`), which is the right call and not what a sibling channel-roster query in #72 does. And `rich_text_list` omitting `offset` is correct rather than a transcription gap: Block Kit documents it as optional and it only sets the starting number of an *ordered* list.

One thing in the code, inline: `.info`'s object loses `last_read` for a channel whose messages the caller cannot read, so the shape this branch is making exact varies by ACL, and the key's presence answers a question about that channel.

Two in the description. It contradicts itself about live `conversations.list`: "it carries `{"next_cursor": ""}` with nothing more, measured" against "this always sends `response_metadata` where live omitted it on a page with no cursor". Which one holds decides whether `.list` should omit the key like `.history` now does, so it is worth settling in the text. And the review instruction says to read the top three commits where there are four (`f28e289`, `a7ea326`, `c10c1d4`, `4c3a906`).

Also inline: the attribution for the five conversation-object fields is `.list`-only, which is the one note from my #73 review that this branch does not already answer.

backlot/routers/slack.py
```diff
@@ -177,17 +187,46 @@ def _full_channel(request: Request, conn, name: str) -> dict:
...
+) -> dict:
+    """conversations.info's channel. `num_members` is opt-in here (the vendor's own parameter), and
+    `last_read` is member-only — this mock's caller is always a member of what it can see, and
+    models no unread state, so the whole channel reads as caught up."""
+    ch = _channel_core(request, conn, name)
+    if include_num_members:
+        ch["num_members"] = _member_count(request, conn, name)
+    latest = store.slack_latest_ts(conn, name, visible_ids)
+    if latest is not None:
+        ch["last_read"] = latest
```
nuri-yoo: `last_read` is conditional, so `.info` answers two different shapes. Measured as `bob@acme.com`, who is not in the people group:

```
caller sees: ['eng-announcements', 'incidents']
  info eng-announcements    ok keys=29  last_read=True   is_private=False
  info incidents            ok keys=29  last_read=True   is_private=False
  info people-confidential  ok keys=28  last_read=False  is_private=True
```

Two things follow. The object is no longer one shape -- which is what this branch is for, since a client validating against a generated model is the case the description opens with -- and the key's presence is a straight answer to "can I read this channel", on a channel `conversations.list` does not show this caller. That the channel resolves at all is older than this branch (`_channel_name` reads the container list without an ACL clause, where real Slack answers `channel_not_found` to a non-member of a private channel), but the new field is what makes the difference readable off the object.

Serving it unconditionally keeps the shape invariant -- Slack's own value for a channel with nothing read is a zero ts rather than an absent key -- and gating on membership instead would match the vendor, at the cost of deciding what membership means here.

While that line is open: the docstring justifies the field with "this mock's caller is always a member of what it can see", and an admin/service token gets it too. This branch decides the opposite for `subscribed` a few lines down, on the grounds that a service token is not a person and follows nothing. `last_read` is member-only in the same sense, so the two could read the same way.

tests/test_slack.py
```diff
@@ -207,12 +207,26 @@ def test_slack_channel_object_carries_slacks_documented_field_set(client, admin_
...
     info = client.post(
         "/slack/api/conversations.info", headers=admin_h, data={"channel": listed["id"]}
     ).json()["channel"]
-    assert _DOCUMENTED_CHANNEL_KEYS <= set(info)
+    assert _DOCUMENTED_CHANNEL_KEYS - {"num_members"} <= set(info)
+    assert "num_members" not in info
+    # and it carries what only .info does
+    assert info["last_read"]
+
+    # Documented on the conversation object reference rather than the method page, and sent on
```
nuri-yoo: This attribution holds for `conversations.list` but not for `.info`: its own method page's example response carries `is_frozen`, `context_team_id`, `parent_conversation`, `shared_team_ids`, `pending_connected_team_ids` and `properties`. So the five are method-page documented for one of the two methods this comment covers, and "#71's transcription missed them" because it was transcribed from `.list` alone rather than because the fields live off the method pages.

Nothing served changes -- this branch serves five of those six and skips `is_frozen` on measured grounds, which the measurement supports either way. It is the sentence that would send the next reader to the wrong page, here and in the description.

Ref: https://docs.slack.dev/reference/methods/conversations.info

khj809:
Both reviews are addressed. Summary of where each note landed:

| Note | Resolution |
| --- | --- |
| #73: auth table narrower than it reads (`bearer`/`token` schemes) | Fixed, `5d111fa` — and it was worse than minor, see below |
| #73: `num_members` gated on `.info` | No change needed; this branch already accepts `include_num_members` and asserts `"num_members" not in info` for a plain call |
| #73 / #75: the five fields are `.list`-attributed, not off the method pages | Fixed, `5d111fa` + `9242cde` (the same sentence was in a second place, above `_DOCUMENTED_CHANNEL_KEYS`) |
| #75: `last_read` gives `.info` two shapes | Fixed, `9242cde` |
| #75: description contradicts itself on live `.list` | Settled in the text |
| #75: "top three commits" where there are more | Fixed — names the base branch instead, so it does not go stale per push |

**The auth one changed severity.** The review measured the error label with a bogus token; measuring
with a *valid* one shows the mock **authenticated** six spellings that live Slack answers
`not_authed` for:

```
Bearer <valid>  AUTHENTICATED     token <valid>   AUTHENTICATED
bearer <valid>  AUTHENTICATED     Token <valid>   AUTHENTICATED
BEARER <valid>  AUTHENTICATED     BeArEr <valid>  AUTHENTICATED
```

So it was not only a wrong error string — a client written against `Authorization: token <xoxb>`
passed every test here and reached nothing in production. Slack got its own parser;
`auth.bearer_token` stays permissive for GitHub. The wider matrix also found the scheme is
case-sensitive and that a tab does not separate scheme from token while repeated spaces do — all
pinned as live-answer rows.

**`last_read`** now always ships. Its value is the caller's and varies, which is the field's
meaning; its presence was varying too, which gave `.info` 29 keys or 28 and made the key an ACL
oracle on a private channel `.list` does not show the caller. A service token gets the zero ts as
well, on the reasoning this branch already applies to `subscribed`.

Two things deliberately left open, both for want of a workspace token:

- The zero ts `0000000000.000000` is reasoned from Slack's convention, not measured. The invariant
  does not depend on which sentinel it is.
- Whether live `.list` omits `response_metadata` on a last page is still unsettled — the captured
  pages carry `{"next_cursor": ""}` and Slack's own `.list` example shows a *non-empty* cursor, so
  neither is a last-page observation. `.list` keeps the key rather than being changed on an
  inference from `.history`.

Also unchanged, as your review concluded: `last_read`'s ACL scoping, and `rich_text_list` omitting
`offset`. `.info` resolving a channel the caller cannot see predates this branch and is noted in the
description rather than fixed here.

Verified: 1241 passed, 3 skipped (all extras); ruff clean. Mutation-checked, each restored from a
copy rather than from HEAD.

nuri-yoo (approved):
Both of mine are answered, re-run against `9242cde`: `.info` gives one 29-key shape across three callers and three channels, the service token gets the zero ts, and the twelve auth spellings I measured against slack.com all match, with `token <t>` still authenticating on github and drive. Suite `1241 passed, 3 skipped`, ruff clean.

Approving.

PR #77, opened 2026-08-21
A GraphQL endpoint describes itself by introspection, so serve that as MCP tools

### LINKED ISSUE #76: linear and fireflies have no MCP example — a GraphQL-only surface the OpenAPI bridge cannot cover
`examples/using-mcp-with-agents` covers 9 of 11 sources. `linear` and `fireflies` have no MCP path, and the OpenAPI bridge cannot give them one: both are **GraphQL-only**, served at `POST /{source}/graphql` with `include_in_schema=False`, so they are absent from `/openapi.json` and `_openapi_bridge.py` has nothing to slice. `GET /_mock/openapi/linear` and `/fireflies` are 404 by construction, not by omission.

The examples already reach MCP three ways, so the gap is only in the third column:

| how | sources |
|---|---|
| backlot's own OpenAPI→MCP bridge | `gdrive` `github` `gmail` `hubspot` `slack` |
| a vendor MCP server pointed at the mock | `atlassian` (docker) · `notion` (`npx @notionhq/notion-mcp-server`) · `s3` (`uvx awslabs.aws-api-mcp-server`) |
| — | **`linear` `fireflies`** |

Both mock endpoints work; only the example is missing:

```
POST /linear/graphql     -> 200
POST /fireflies/graphql  -> 200
```

## Why it matters

Measured against the `backlot10k-r4` question set (21,249 questions), gold that lives **only** in linear or fireflies:

| | questions |
|---|---|
| gold is linear/fireflies only — unreachable over MCP | **1,945 (9.2%)** — 1,212 answerable, 733 denied |
| gold mixes them with a covered source — partially reachable | 549 (2.6%) |
| touched either way | 2,494 (11.7%) |

So an MCP-based agent evaluation structurally cannot answer 9.2% of that set, and the shortfall is invisible unless the subset is reported separately. The two sources are 997 and 497 of the corpus's 10,264 records, so this is not a rounding error.

## The constraint to design around

Linear and Fireflies both ship an **official remote MCP server**. Remote means vendor-hosted, so neither can be pointed at this mock — the official route is closed here, and the limitation should be stated in the example rather than worked around silently.

That leaves community servers, and the blocking issue is that they hardcode the vendor endpoint. Verified in source (2026-08-21):

- `johntoups/mcp-fireflies` — `src/mcp_fireflies/client.py`: `GRAPHQL_ENDPOINT = "https://api.fireflies.ai/graphql"`, a module constant. Only `FIREFLIES_API_KEY` is configurable.
- `tacticlaunch/mcp-linear` — hardcodes `https://api.linear.app/...`; documents only `LINEAR_API_TOKEN` / `LINEAR_OAUTH_ACCESS_TOKEN`.

Neither documents a base-URL override. And because these run as **subprocesses** (`npx` / `uvx`), the in-process redirect backlot already uses for the LlamaIndex Linear reader — `backlot/integrations/llamaindex.py` rewrites `https://api.linear.app` to the mock's base before the request leaves — cannot reach them. `s3` avoids this only because the AWS CLI happens to honour a first-class `AWS_ENDPOINT_URL`.

## Candidates, measured

Stars and last push as of 2026-08-21:

| server | stars | last push | runtime |
|---|---|---|---|
| `jerhadf/linear-mcp-server` | 347 | 2025-05-01 | JavaScript |
| `tacticlaunch/mcp-linear` | 146 | **2026-07-28** | TypeScript |
| `props-labs/fireflies-mcp` | 5 | 2025-03-04 | JavaScript |
| `cassler/fireflies-mcp-server` | 1 | 2025-06-28 | JavaScript |
| `johntoups/mcp-fireflies` | 0 | **2026-05-27** | Python |

For Linear the choice is real: the most-starred server has been untouched for fifteen months, while `tacticlaunch/mcp-linear` is both reasonably adopted and current. For Fireflies there is no meaningfully-adopted server at all — the ceiling is 5 stars — so "pick the widely-used one" has no answer, and whatever is chosen is a judgement call to record rather than a consensus to follow.

## Options

1. **Vendor and patch the endpoint.** The pattern already exists in this tree — `experiments/mcp-vs-mirage/vendor/` holds `google_workspace_mcp` and `slack-mcp-server`. One-line endpoint substitution, pinned to a known commit. Keeps the example driving a real community server, at the cost of carrying a fork.
2. **Contribute an endpoint override upstream** and depend on the released version. Cleanest long-term, slowest, and `tacticlaunch` would have to accept it.
3. **Wrap with a local HTTP proxy / host rebind** so an unmodified server resolves `api.linear.app` to the mock. No fork, but it is per-platform plumbing and TLS makes it awkward.
4. **A GraphQL→MCP bridge in backlot**, alongside `_openapi_bridge.py`. Covers both sources at once, no third party and no hardcoded host, and the schemas are already in `backlot/graphql/`. But it tests *our* tool surface rather than the tooling an agent would really meet — which is the point of using vendor servers for `atlassian`/`notion`/`s3`.

Suggested: **1 for Linear** (`tacticlaunch/mcp-linear`, pinned) and **1 or 4 for Fireflies**, since no community server there has enough adoption to be worth carrying a fork for. Either way the example should say plainly that the official servers are remote-only and why the community one was chosen.

## Acceptance

- `examples/using-mcp-with-agents/linear.py` and `fireflies.py` exist, follow the shape of the other launchers (`build_params` + `run_agent`, `--url` / `--token` / `--agent`), and answer their question against the mock.
- Each file's docstring states that the official MCP server is remote-only, which community server was chosen, and why.
- `tests/test_mcp.py` covers both, as it does the others.
- The README table lists them.

khj809:
Closes #76.

Linear and Fireflies are GraphQL-only, so the OpenAPI→MCP bridge has nothing to slice for them and they were the two sources an agent could not reach over MCP at all. A GraphQL endpoint does describe itself, over standard introspection, so that is what this reads.

`backlot/graphql/mcp_tools.py` turns an introspection result into one tool per root `Query` field — 15 for Linear, 4 for Fireflies — each with a JSON Schema for its arguments and a complete document to send. `examples/using-mcp-with-agents/_graphql_bridge.py` is the transport, posting the document with the caller's own `Bearer` token so the per-token ACL decides every result (one spelling covers both: Linear accepts a bare key or a Bearer token, as the real API does). `linear.py` and `fireflies.py` follow the shape of the sibling launchers.

## The selection set

The one thing OpenAPI never has to decide: a REST operation returns a fixed body, a GraphQL field returns whatever was asked for. The rules are stated in `mcp_tools` rather than tuned — leaf fields always; object fields followed while a depth budget lasts; no re-entering a type already on the path; no field whose argument is required. Input objects expand under that same guard, so `IssueFilter` arrives with its real comparator keys instead of as an opaque blob.

Two rules are about what a tool result costs, and both are measured:

- **A connection is transparent wherever it appears**, keeping `pageInfo` so a page an agent cannot advance never reads as a complete answer. Following nested ones is also what makes an issue's comments reachable: `CommentFilter` has no key for the issue, so the root `comments` tool cannot stand in for `Issue.comments`.
- **A bare list of objects is not selected inside a repeated result** — rows × their own list is a product with no page and no bound. With `Transcript.sentences` inlined, `transcripts` returned 1.1 MB (~276k tokens) for one default page of a 40-meeting corpus and ~55k even at `limit: 5`; without it, 36 KB and 1.8k. `transcript(id:)` still returns every utterance — the same split `examples/using-official-sdk/fireflies.py` writes by hand.

Depth is per-source, because one value does not fit both schemas. Fireflies takes the default 2: `Analytics` has no leaf fields of its own, so anything shallower drops the sentiment split whole. Linear runs at 1, because `Team`/`Project`/`Cycle` each carry dozens of configuration leaves and a second level takes an issue from 507 selected fields to 1,313. `test_linear.py` pins a ceiling that holds at 507 and fails at 1,313.

## Trade, stated

This exercises our own tool surface rather than the community tooling an agent meets in production — which is why `atlassian`/`notion`/`s3` use vendor servers. Both vendors' official servers are remote-hosted with no base-URL override, and the community ones hard-wire `api.linear.app` / `api.fireflies.ai` as subprocesses out of reach of an in-process rewrite, so there is nothing local to redirect. Each launcher's docstring says so.

## Verification

- Both examples run against the mock and answer their questions end to end — the Linear agent quotes the comment that names the root cause, and the Fireflies agent searches with `transcripts` then reads utterances with `transcript(id:)`.
- `tests/test_mcp.py` covers both sources: ACL enforced in both directions, and a nested `IssueFilter` round-trip asserting every returned row matched.
- Derivation rules unit-tested in `tests/test_graphql.py` against a throwaway schema; per-vendor documents checked against the served introspection in `test_linear.py` / `test_fireflies.py`.
- Full suite green with all extras; `ruff check` clean; `mcp_tools` confirmed present in a built wheel.

A review pass also fixed five latent derivation bugs, each with a regression test: an interface- or union-typed root field emitting a document with no selection set; a NON_NULL argument with a default declared `T!` and so unomittable; a page type with no leaf fields interpolating a literal `None`; the page type name assumed to be `PageInfo`; and an error envelope crashing `derive_tools` with a `TypeError` instead of reporting the endpoint's message (the bridge likewise turned a mistyped `--token` into "Connection closed" rather than the mock's 401).

### DIFF
--- a/backlot/graphql/mcp_tools.py
+++ b/backlot/graphql/mcp_tools.py
@@ -1,0 +1,321 @@
+"""Derive MCP tools from a GraphQL introspection result — one tool per root ``Query`` field.
+
+The REST sources reach MCP through ``/openapi.json``: an operation has a fixed request shape and
+a fixed response body, so ``FastMCP.from_openapi()`` is the whole bridge. A GraphQL endpoint
+describes nothing that way — ``POST /linear/graphql`` is one operation taking an arbitrary
+document — which is why Linear and Fireflies had no MCP path at all. This module supplies the
+missing description: it reads the schema the endpoint already serves over standard introspection
+and produces, for each root field, a tool name, a JSON Schema for its arguments, and a complete
+GraphQL document to send.
+
+Everything here is pure — a mapping in, :class:`Tool` objects out. No HTTP, no MCP: the transport
+is ``examples/using-mcp-with-agents/_graphql_bridge.py``, which posts ``tool.document`` with the
+caller's own credential so the mock's per-token ACL decides what comes back.
+
+**One tool per root field, not one ``graphql(query:)`` passthrough.** A passthrough is three lines
+and turns the exercise into "can the model write GraphQL against a schema it has not seen"; none
+of the sibling examples work that way, and a typed toolset is what an agent actually meets.
+
+**The selection set is generated, which is the one thing OpenAPI never has to decide.** A REST
+operation returns a fixed body; a GraphQL field returns whatever was asked for, so something must
+choose. The rules, in full:
+
+* Leaf fields (scalars and enums) are always selected.
+* A Relay connection — an object carrying both ``nodes`` and ``pageInfo`` — is **transparent**
+  wherever it appears: it expands to ``nodes { … } pageInfo { … }``. Unwrapping one at the root is
+  free, so ``issues`` reaches as deep into an ``Issue`` as ``issue`` does; below the root it costs
+  a level like any other object field. ``pageInfo`` always rides along, and for a nested
+  connection it is load-bearing: there is no cursor to pass into ``Issue.comments``, so
+  ``hasNextPage`` is the only thing keeping a truncated answer from reading as a complete one.
+  Following them is also what makes an issue's comments reachable at all: ``CommentFilter`` carries
+  ``id``/``body``/``createdAt`` and no key for the issue, so the root ``comments`` tool cannot
+  stand in for ``Issue.comments``.
+* An object field is followed while a **depth budget** remains, and each level spends one. The
+  default is 2, which is what Fireflies needs to reach ``transcript.analytics.sentiments`` —
+  ``Analytics`` has no leaf fields of its own, so a shallower selection drops the node whole. One
+  depth does not suit both schemas, so the launchers choose: Linear runs at 1, because ``Team`` /
+  ``Project`` / ``Cycle`` each carry dozens of configuration leaves and a second level takes an
+  issue from 516 selected fields to 1,446 for data no agent asks about.
+* A type already on the current path is not re-entered, so ``Issue.parent`` (an ``Issue``) stops
+  there rather than recursing.
+* A field with a required argument is skipped — the generator has no value to supply.
+* An object field that ends up selecting nothing is dropped, because an empty selection set is
+  not a legal document.
+
+Argument schemas follow the same shape of rule: scalars and enums map directly, and input objects
+expand recursively under the same path guard — so ``IssueFilter`` arrives with its real comparator
+keys (``{"title": {"containsIgnoreCase": "…"}}``) rather than as an opaque object the model has to
+guess at, and its self-referential ``and``/``or`` drop out by the guard rather than a special case.
+
+Custom scalars become strings, which is how GraphQL serializes them absent a stated alternative;
+where a vendor accepts more than one spelling (Fireflies' ``DateTime`` takes ISO 8601 or epoch
+millis) the argument's own description carries it, and descriptions are copied through.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any, Mapping
+
+from graphql import get_introspection_query
+
+#: The standard introspection document, from graphql-core itself rather than hand-written — the
+#: bridge sends this and feeds the result straight to :func:`derive_tools`.
+INTROSPECTION_QUERY = get_introspection_query(descriptions=True)
+
+#: How many object levels below a root field's own type the generated selection set reaches.
+DEFAULT_DEPTH = 2
+
+# GraphQL's built-in scalars, plus the two names both vendors use for a loosely-typed object.
+# Anything else is a custom scalar and serializes as a string.
+_SCALARS: dict[str, dict[str, Any]] = {
+    "Int": {"type": "integer"},
+    "Float": {"type": "number"},
+    "String": {"type": "string"},
+    "Boolean": {"type": "boolean"},
+    "ID": {"type": "string"},
+    "JSON": {"type": "object"},
+    "JSONObject": {"type": "object"},
+}
+
+
+@dataclass(frozen=True)
+class Tool:
+    """One MCP tool: a root ``Query`` field, its arguments as JSON Schema, and the document to post.
+
+    ``document`` declares every argument as a GraphQL variable, so it is fixed per tool and the
+    caller's arguments travel as ``variables``. A variable left unsupplied means the argument was
+    not provided at all — which is what the resolvers see — so one document serves every call.
+    """
+
+    name: str
+    description: str
+    input_schema: dict[str, Any]
+    document: str
+
+
+def derive_tools(introspection: Mapping[str, Any], *, depth: int = DEFAULT_DEPTH) -> list[Tool]:
+    """Turn an introspection response into one :class:`Tool` per root ``Query`` field.
+
+    ``introspection`` is the endpoint's reply to :data:`INTROSPECTION_QUERY`, with or without its
+    ``data`` envelope. Tools come back in schema order, which is the order the SDL declares.
+    """
+    body = introspection.get("data", introspection)
+    schema = body["__schema"]
+    types = {t["name"]: t for t in schema["types"]}
+    root = types[schema["queryType"]["name"]]
+
+    tools = []
+    for field in root["fields"]:
+        if field["name"].startswith("__"):
+            continue  # introspection meta-fields are not part of the served surface
+        tool = _tool(field, types, depth)
+        if tool is not None:
+            tools.append(tool)
+    return tools
+
+
+# --- one field -------------------------------------------------------------------------
+
+
+def _tool(field: Mapping[str, Any], types: dict[str, Any], depth: int) -> Tool | None:
+    named = _named(field["type"])
+    selection = _root_selection(named, types, depth)
+    # A field whose type yields no selectable field cannot be asked for at all: the document
+    # would carry an empty selection set, which does not parse. Nothing in either vendor's
+    # schema hits this; it is here so a schema that does loses one tool rather than every tool.
+    if selection is None and named["kind"] == "OBJECT":
+        return None
+
+    args = field["args"]
+    properties, required = {}, []
+    for arg in args:
+        schema = _json_schema(arg["type"], types, frozenset())
+        if schema is None:
+            # Nothing describable in it, so the tool cannot offer the argument — and leaving a
+            # REQUIRED one out means a document the schema rejects, which is worse than one tool
+            # fewer. Neither vendor's schema reaches this; a wider one might.
+            if _is_required(arg):
+                return None
+            continue
+        if arg.get("description"):
+            schema = {**schema, "description": arg["description"]}
+        properties[arg["name"]] = schema
+        if _is_required(arg):
+            required.append(arg["name"])
+
+    input_schema: dict[str, Any] = {
+        "type": "object",
+        "properties": properties,
+        "additionalProperties": False,
+    }
+    if required:
+        input_schema["required"] = required
+
+    return Tool(
+        name=field["name"],
+        description=_description(field, depth),
+        input_schema=input_schema,
+        document=_document(field["name"], [a for a in args if a["name"] in properties], selection),
+    )
+
+
+def _description(field: Mapping[str, Any], depth: int) -> str:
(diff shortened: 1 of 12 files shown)


backlot/graphql/mcp_tools.py
```diff
@@ -0,0 +1,473 @@
...
+    budget: int,
+    path: frozenset[str],
+    indent: int = 1,
+    *,
+    many: bool,
+) -> str | None:
+    parts = []
+    for field in types[type_name]["fields"]:
+        named = _named(field["type"])
+        if named["kind"] in ("SCALAR", "ENUM"):
```
nuri-yoo, line 360: The leaf shortcut runs before the required-argument check on line 362, so the rule the module docstring states without qualification -- "A field with a required argument is skipped" -- holds for object fields only. A leaf field carrying one is selected without it, which is a document the schema rejects:

```
schema     type T { x(need: String!): String }   type Query { t: T }
document   query t {   t {     x   } }
validate   Field 'x' argument 'need' of type 'String!' is required, but it was not provided.
```

Latent: I walked every SCALAR- and ENUM-typed field in both served introspections and neither has a required argument, so no document this repo generates is affected. Worth the two-line move anyway, because the failure mode is the one `_tool` deliberately avoids a few dozen lines up -- there the comment is that a schema which cannot be described "loses one tool rather than serving one that fails on every call", and this path serves one that fails on every call.

backlot/graphql/mcp_tools.py
```diff
@@ -0,0 +1,473 @@
...
+def _connection(type_: Mapping[str, Any]) -> tuple[str, str] | None:
+    """A Relay connection's ``(node type, page-info type)``, or ``None`` if it is not one.
+
+    Carrying both ``nodes`` and ``pageInfo`` is the whole test, and both type names are read off
+    those fields: a schema is free to call its page type something other than ``PageInfo``.
+    """
+    fields = {f["name"]: f for f in (type_.get("fields") or ())}
+    if not {"nodes", "pageInfo"} <= set(fields):
+        return None
+    return _named(fields["nodes"]["type"])["name"], _named(fields["pageInfo"]["type"])["name"]
```
nuri-yoo: `_connection` decides on the two field NAMES and then hands both type names to `_selection`, which does `types[name]["fields"]` -- `None` for a scalar. So a type carrying `nodes` and `pageInfo` of the wrong kind takes `derive_tools` down rather than costing it a tool:

```
nodes: [String!]!  + pageInfo: PageInfo!   ->  TypeError: 'NoneType' object is not iterable
nodes: [Thing!]!   + pageInfo: String      ->  TypeError: 'NoneType' object is not iterable
nodes: [Node!]! (interface) + pageInfo     ->  fine, one tool
```

Same place as the two shapes the review pass already fixed (the page type with no leaf fields, and the page type not named `PageInfo`), and neither vendor schema reaches it. Checking that both named types are OBJECT or INTERFACE here, or defaulting `fields` to `()` in `_selection`, keeps the module's own promise that a schema it cannot describe loses a tool rather than the run.

examples/using-mcp-with-agents/_graphql_bridge.py
```diff
@@ -0,0 +1,140 @@
...
+        endpoint,
+        data=json.dumps({"query": mcp_tools.INTROSPECTION_QUERY}).encode(),
+        headers={"Content-Type": "application/json", "Authorization": f"Bearer {token}"},
+    )
+    try:
+        with urllib.request.urlopen(request, timeout=30) as r:
+            return json.load(r)
+    except urllib.error.HTTPError as exc:
+        body = exc.read().decode(errors="replace").strip()
+        sys.exit(f"introspecting {endpoint} failed with HTTP {exc.code}: {body[:500]}")
```
nuri-yoo, line 72: This works, with one qualification worth putting in the description rather than here. Driving this file as a stdio subprocess with a mistyped token, the endpoint's own body does reach the terminal:

```
introspecting http://127.0.0.1:57781/fireflies/graphql failed with HTTP 401:
{"errors":[{"message":"Please provide a valid API key","extensions":{"code":"unauthorized"}}]}
```

and the client still raises `McpError: Connection closed`, because a stdio server that exits before the handshake cannot surface as anything else. So "rather than 'Connection closed'" reads as though the exception changed; what changed is that the reason is now printed instead of lost, which is the part that matters to whoever is holding the terminal.

While this line is open: nothing tests this file. `tests/test_mcp.py` re-implements the transport, deliberately and with the same reasoning as the REST bridge test, which leaves `_introspect`'s two error paths, the certifi block and the CLI covered by description only. I exercised the happy path myself for both sources (4 tools and 3 rows for fireflies, 15 tools and 5 nodes for linear, neither flagged `is_error`), so this is a note about what CI would catch rather than about what works.

nuri-yoo (approved):
LGTM.

Reviewed at `916cd80`. Suite `1249 passed, 3 skipped` here (ERB_E2E opt-in, `llama_index.readers.hubspot` absent from my env, mcp-atlassian needs Docker), so both new GraphQL bridge tests did run; `ruff check` clean over `backlot tests examples`.

I checked the claim the whole module rests on rather than reading for it: took each endpoint's own introspection, built a client schema from it, and ran `graphql.validate` over every derived document. 15 tools for Linear at depth 1 and 4 for Fireflies at depth 2, matching the description, and **0 invalid documents** on either side. Calling every tool with no arguments returns data except the by-id ones, which fail on exactly their required argument -- so the `required` list and the variable declarations agree. `transcripts` omits `sentences` while `transcript(id:)` selects it (54 against 98 selected lines on my corpus), `with_page_default({})` fills `limit: 10` and leaves `{"limit": 50}` alone, and Linear's `issue` goes from 510 selected lines at depth 1 to 1421 at depth 2, with the test's ceiling sitting between them.

I also drove the real `_graphql_bridge.py` as a stdio subprocess against `backlot.mock_server`, since the tests re-implement its transport rather than importing it: `fireflies` served 4 tools and `transcripts` came back with 3 rows, `linear` served 15 and `issues` came back with 5 nodes, neither flagged `is_error`. That is the one part of this branch nothing else exercises, and it works.

Two latent derivation gaps inline, both the same family as the five the review pass swept and neither reachable from either vendor's schema -- which is why this is an approval rather than a hold. A third note is about the bridge's bad-credential path, where the improvement is real but narrower than the description says.

backlot/graphql/mcp_tools.py
khj809: Fixed in `23664b3` — the required-argument check now precedes the leaf shortcut, so the rule holds for every field kind rather than object fields only. I reproduced your schema first (`validate` gave exactly your error), then added `checksum(algorithm: String!): String` to `MCP_SDL`, which catches it in four places: the new parametrize case, both exact-set cases of `test_the_depth_budget_bounds_the_selection`, and `test_a_derived_document_validates_against_the_schema`.

backlot/graphql/mcp_tools.py
khj809: Fixed in `23664b3` — `_connection` now reads the two kinds as well as the two names, and takes `OBJECT`/`INTERFACE` (so your interface case keeps behaving as it did). Both of your shapes reproduced as the same `TypeError`; they now fall back to the plain object path, which means the tool keeps its place rather than being dropped — `set(tools)` is unchanged in all three cases. I folded them into `test_an_unusable_connection_falls_back_to_the_plain_object_path` as a 3-case parametrize and widened its name. Its old `"pageInfo" not in …` assertion is gone: an empty `pageInfo {}` fails `validate`, which the test already asserts.

examples/using-mcp-with-agents/_graphql_bridge.py
khj809: Agreed, and the docstring now says the narrower thing: the client raises `Connection closed` either way, and what the exit buys is the endpoint`s own reason on the terminal instead of a urllib traceback. Confirmed the same output you saw (`exit=1`, HTTP 401 body on stderr).

On the coverage note — leaving it as is, for the reason you name: `tests/test_mcp.py` re-implements the transport deliberately, same as the REST bridge. I drove the real subprocess for both sources again after this change (fireflies 4 tools / 6 rows, linear 15 tools / 10 rows, neither flagged `is_error`).

khj809:
All three addressed in `23664b3`, details in the threads. Both latent gaps reproduced first, and to confirm they really were latent I diffed every derived tool across the fix: fireflies 4 and linear 15, `document` / `input_schema` / `description` identical on all 19. Suite `1252 passed, 3 skipped` (the +3 are the new parametrize cases), and reverting `mcp_tools.py` alone fails exactly 7 of them. CI green.

PR #79, opened 2026-08-24
github: a file's body is searchable, and a path answers once

### LINKED ISSUE #78: github: a file's body is indexed and unreachable — there is no /search/code
A repository file's body is indexed and ACL-scoped, and no endpoint exposes it. Real GitHub has `GET /search/code`; this mock does not, so the only way to a file's contents is to already know its path.

```
GET /github/search/code?q=INC-0E0C46          -> 404   (no such route)
GET /github/search/issues?q=INC-0E0C46        -> total_count 0
GET /github/repos/{owner}/{repo}/contents/config/reporting/payment_method_mix_export.yaml?ref=pull/340
                                              -> 200, body contains INC-0E0C46
```

## The index is already there

`github_fts` holds every `github_items` row, files included, and it matches:

```
github_items         1,677   (kind='file': 880)
github_fts           1,677
github_fts MATCH '"INC-0E0C46"'  ->  1 hit
    repo=core-payments  path=config/reporting/payment_method_mix_export.yaml  kind=file
```

Nothing needs indexing. `routers/github.py::search_issues` filters the rows back out, correctly for the endpoint it serves — real GitHub's issue search does not return files:

```python
matched = [r for r in cand if r["kind"] != "file" and _issue_qual_match(r, quals)]
```

So the missing piece is a route that inverts that filter and shapes the result like code search. `store.search_documents(conn, free, "github", ids, ...)` already applies the caller's ACL, so scoping comes for free.

## Why it matters

Reaching a file today is a three-hop traversal with no search shortcut:

```
search_issues -> the PR  ->  /pulls/{n}/files (learn the path)  ->  /contents/{path}?ref=pull/{n}
```

Every tool in that chain is exposed over MCP, so it is possible — but only if a client guesses that the answer is in a changed file. Measured on the `backlot10k-r4` question set, 30 of its 374 span-graded questions plant their answer in a `github/file` body. One of them (`q-20260824-00009`, asking for an incident reference that lives only in the file) was failed twice by an agent that found the right PR both times: once it read the PR's comments and gave up, and once — told explicitly to follow the trail — it answered with a *different* reference code from the PR body instead. A single `search/code` call for the code would have ended it.

More generally, "find where this identifier appears" is the natural first move against a code host, and here it silently returns nothing rather than failing.

## Shape to match

`GET /search/code?q=…`, authenticated, returning `total_count`, `incomplete_results`, and `items[]` of `{name, path, sha, url, git_url, html_url, repository{…}, score}`. Qualifiers worth honouring: `repo:`, `path:`, `filename:`, `extension:`, `in:file` / `in:path`. `text_matches` under `Accept: application/vnd.github.text-match+json` would fit the media-type handling this router already does (#49/#50), and is the part that makes a hit useful rather than just locating it.

## One design question

Real code search indexes the **default branch only**. This mock stores several snapshots per path, told apart by `ref` (#63), so a naive implementation returns the same path once per snapshot. Restricting results to the head snapshot per `(repo, path)` — which `_file_head_clause` already defines — matches real behaviour and keeps a result set from being dominated by history. That does mean a code search cannot find a string that exists only in a superseded snapshot; worth deciding deliberately rather than by accident, and worth stating in the docstring either way.

## Acceptance

- `GET /github/search/code` serves the response shape above, ACL-scoped, with the qualifiers listed.
- Head-snapshot behaviour is decided, documented in the handler's docstring, and covered by a test over a path with more than one snapshot.
- A test asserts a string present only in a file body is findable through this endpoint and still absent from `/search/issues`.
- The route appears in `/openapi.json` so `_openapi_bridge.py` picks it up as an MCP tool without further work.

khj809:
Closes #78.

`github_fts` indexes every file, and `/search/issues` filters those rows out — correctly, since real GitHub's issue search does not return files. A file's body was therefore indexed and unreachable: the only route to it was to already know its path.

## `GET /github/search/code`

Real's envelope and item shape — `total_count` / `incomplete_results` / `items[]` of `{name, path, sha, url, git_url, html_url, repository, score}` — ACL-scoped through the same reads every other file route uses. Qualifiers: `repo:` `path:` `filename:` `extension:` `in:file`/`in:path`. `text_matches` under `Accept: application/vnd.github.text-match+json`.

The route is in `/openapi.json`, so `_openapi_bridge.py` exposes it as an MCP tool with no further work.

## One result per `(repo, path)`

The head snapshot's. Real code search indexes the default branch only, and this mock stores a row per snapshot, so without that restriction a path answers once per revision it was ever recorded at and one repo's history crowds out the corpus. The cost is stated in the handler's docstring and pinned by a test: a string surviving only in a **superseded** snapshot is not findable here. It stays reachable at `contents/{path}?ref=` and `git/blobs/{sha}`.

## Shape read off api.github.com, not the docs

| | real | here |
|---|---|---|
| `score` | flat `1.0` on every hit | same |
| blank `q` | `422 Validation Failed` | same envelope |
| qualifier-only `q` | served | same |
| `in:path` hit's `text_matches` | `property: content`, `matches: []` | same |
| item `url` / `git_url` | `/repositories/{id}/…` | `/repos/{owner}/{repo}/…` |

The links take the form this mock serves rather than real's spelling — a link the caller can follow beats one that 404s, which is the rule `_repo_obj` already states for its url templates. The snapshot pin is kept as the head row's own `ref`, so following `url` returns the bytes that were searched.

## Both searches page with a `Link`

Real sends an RFC5988 `Link` on `/search/issues` and `/search/code` alike, and none at all when the results fit on one page. Both endpoints reported `total_count` and then left the caller to compose the next URL; now they carry the header every listing on this surface already had.

`github_link_header` percent-encodes its param values, which a search needs and a listing never did: a `q` carries spaces and colons, and a raw one leaves a URI a client cannot follow back to the query it was paging.

## Also

`store.search_repo_files` is the read behind code search: head-snapshot file rows, corpus-wide or per repo, narrowed by FTS over title+content and/or a path substring. `_fts_relevance_order` is extracted from `search_documents` so both FTS readers share the literal-substring tier — what puts an identifier like `INC-0E0C46` first rather than under coincidental token hits.

## Not included

Real caps a search at 1000 accessible results — `rel="last"` stops there and a page past it is a `422`. This mock has never enforced that ceiling on `/search/issues`, and the `Link` rels here point only at pages it does serve, so the two stay consistent. Worth a change that covers both endpoints and both halves of the behaviour.


nuri-yoo (approved):
Reviewed at `194f76a`, and merged with #82 (`89c7fa5`), against a 19,693-document corpus rather than the fixtures — 1,677 github rows over 45 multi-snapshot paths.

**ACL, re-measured rather than taken.** Every reference code occurring in exactly one github file row (138 of them), narrowed to the files that are not org-readable, searched as a quoted literal with three non-readers' own tokens each: **180 lookups, 0 exposures.** Reader positive control 57/60, the three misses being the superseded-snapshot case the docstring states.

Blank `q` → 422 in real's envelope, `Link` on both searches with `q` percent-encoded, `path:` on whole segments, `repo:` under another owner → `total_count: 0`. 1492 passed with `--all-extras`.

Approving. One minor thing below, a one-liner that can ride here or follow.

**`backlot/store.py:2726` — the no-FTS fallback escapes one `LIKE` and not the other**

```python
narrow += " AND lower(t.path) LIKE ? ESCAPE '\\'"                       # 2710
text, tp = " AND (t.title LIKE ? OR t.content LIKE ?)", [f"%{query}%"]*2   # 2726
```

The path fragment gets `_like_escape` and a comment saying why (`mod_7` must not answer `mod-7`); `query` sixteen lines below gets neither, so on a build without FTS5 a search for `100%` answers with every file. `search_documents`' own fallback has the same gap, so this is inherited rather than introduced — but the module ships `_like_escape` for exactly this, and the new function already uses it once.

khj809:
Fixed in 6ccb1e9. Both LIKEs in `search_repo_files` now go through `_like_escape` — including the path fragment, which was a hand-rolled second copy of it rather than a call, contrary to what the review assumed. Covered by a no-FTS assertion in the store test (`%` returned all four head rows before, none after).

Left `search_documents`' fallback alone: same gap, but closing it there changes every source's search rather than github's file read.

PR #82, opened 2026-08-25
slack: honour `types`, scope every conversations.* method by ACL, and keep speakers inside their channel

### LINKED ISSUE #81: slack: conversations.* ignore the types filter and answer for channels the caller is not in
Three gaps in `conversations.*`, all about **which conversations a caller may see** rather than the
shape of the fields inside one. Measured against `main` @ `369ce75`.

## Repro

Three rows — two public channels and one restricted to `ava` — and a two-person roster:

```jsonl
{"source_type":"slack","doc_id":"m-open","channel":"general","content":"morning all","author_email":"ava@acme.test","author_name":"Ava","participants":["ava@acme.test","bo@acme.test"],"created":"2026-01-05T09:00:00+00:00","visibility":"public","group":"eng"}
{"source_type":"slack","doc_id":"m-secret","channel":"board-comp","content":"the comp band lands at 240k","author_email":"ava@acme.test","author_name":"Ava","participants":["ava@acme.test"],"created":"2026-01-06T09:00:00+00:00","readers":["user:ava@acme.test"],"group":"exec"}
{"source_type":"slack","doc_id":"m-open2","channel":"random","content":"anyone tried the new build","author_email":"ava@acme.test","author_name":"Ava","participants":["ava@acme.test"],"created":"2026-01-07T09:00:00+00:00","visibility":"public","group":"eng"}
```

```yaml
org: acme
org_domain: acme.test
departments:
  Engineering:
    - {name: Ava Stone, email: ava@acme.test}
    - {name: Bo Ryu, email: bo@acme.test}
```

`board-comp` is private and `bo` is not in it. `conversations.list` gets that part right: `ava` sees
all three channels, `bo` sees `general` and `random`.

## 1. `types` is validated and then discarded

`_slack_types` parses the argument and rejects an unknown value with `invalid_types` — correct, and
its docstring states Slack's documented default. But `conversations_list` uses the parsed set only as
a boolean gate:

```python
names = _channel_names(conn) if types & _CHANNEL_TYPES else []
```

so `public_channel` and `private_channel` never separate:

| caller | `types=` | returned |
|---|---|---|
| ava | `public_channel` | `board-comp`, `general`, `random` |
| ava | `private_channel` | `board-comp`, `general`, `random` |
| ava | `public_channel,private_channel` | `board-comp`, `general`, `random` |
| ava | *(omitted)* | `board-comp`, `general`, `random` |
| ava | `im` / `mpim` | *(none)* |
| ava | `bogus` | `invalid_types` |
| bo | `public_channel` | `general`, `random` |
| bo | `private_channel` | `general`, `random` |
| bo | *(omitted)* | `general`, `random` |

Three divergences fall out of that one line:

- `types=public_channel` returns a **private** channel.
- `types=private_channel` returns **public** ones.
- Omitting `types` returns private channels. [`conversations.list`][list] documents the default as
  `public_channel`, and `_slack_types`' own docstring says the same — so the default disagrees with
  the code that reads it.

`types=im` / `mpim` correctly select nothing, which the comment above `_SLACK_TYPES` explains and
which I am not questioning. ACL scoping is sound throughout: `bo` never sees `board-comp` in any
listing. This is a filter that is parsed and then not applied, not a visibility problem.

## 2. The by-id methods answer for a channel the caller cannot see

`conversations_info`, `conversations_members` and `conversations_history` each resolve the channel by
id and answer without asking whether this caller may see it. `bo`, who cannot see `board-comp` in any
listing, addressing it by id:

```
conversations.info     ok=true   name='board-comp'  is_private=true  is_member=true
                                 topic='#board-comp'  purpose='Channel for board-comp'
                                 creator='USERVICE0'  created=1767690000
  …&include_num_members=true     num_members=1        (same as the member's answer)
conversations.members  ok=true   members=['U6677006F42']    # = synth.slack_user_id('ava@acme.test')
conversations.history  ok=true   messages=[]
```

For `ava` (a member) `conversations.history` returns the one message, so history *does* filter content
by ACL. What leaks is the container: that the channel exists, its name, topic and purpose, its creation
time, its member count, and the ids of its members.

[`conversations.history`][history] documents `not_in_channel` for exactly this case ("The token used
does not have access to the proper channel"), and `conversations.info` documents `channel_not_found`.
`ok:true` with an empty `messages` array is a third answer — "this channel exists and you may read it;
it happens to be empty" — which is the opposite of what the mock knows to be true.

The membership list is the sharper half: a channel's members are derived from who has spoken in it, so
`conversations.members` reports a projection of the very messages the ACL is withholding.

`_info_channel` already treats "can this caller read this channel?" as a secret, and says so:

> Being the caller's, its VALUE varies by who asks — that is what the field means. Its PRESENCE must
> not: a key that appears only when the caller can read the channel gives `.info` two shapes, and on a
> private channel `conversations.list` does not show this caller, the missing key is a straight answer
> to "can you read this?"

That care goes into `last_read`'s presence while the response as a whole answers the same question.
The visibility set is right there in the function — `conversations_info` resolves `auth.visible_ids`
and passes it down — but it is spent on `last_read` rather than on whether to answer at all.

## 3. `is_member` is a constant

`_channel_core` hardcodes it:

```python
"is_member": True,
```

The [conversation object][obj] defines it as "whether the user, bot user or Slack app associated with
the token making the API call is itself a member of the conversation". The constant is observable in
`conversations.list` on its own, without §2: `bo` has never posted in `random`, and

```
bo  →  conversations.list  →  random  is_private=false  is_member=true
```

Real Slack lists public channels a user is not in, with `is_member: false`; here every channel a
caller can see claims membership.

## Why it matters beyond shape

A corpus can model a restricted room and grade a system on not revealing that a record exists. Today a
channel id is enough for any authenticated principal to confirm the room, read its name and purpose,
and enumerate who is in it — without ever seeing a message.

## Suggested shape

- `conversations.list`: intersect the resolved names with the requested `types` (`is_private` is
  already computed in `_channel_core` via `store.container_has_public`), and default to
  `public_channel` when the argument is omitted.
- `conversations.info` / `.members` / `.history`: apply the visibility test `conversations.list`
  already uses, and answer `channel_not_found` (info/members) or `not_in_channel` (history) when it
  fails.
- `_channel_core`: derive `is_member` from membership rather than returning a constant.

## Scope

Not in scope here: DM/`mpim` modelling, or the message- and channel-object field work already landed.

[list]: https://docs.slack.dev/reference/methods/conversations.list
[history]: https://docs.slack.dev/reference/methods/conversations.history
[obj]: https://docs.slack.dev/reference/objects/conversation-object

### LINKED ISSUE #83: byo: a speaker in a private channel is not required to be a member of it
A Slack channel's members are the people who have spoken in it (`store.slack_channel_member_emails`),
and what a caller may READ is the union of the ACL grants on that channel's documents. Nothing ties
the two together, so a corpus can state a private channel whose `conversations.members` lists people
who cannot see the channel at all — a pair of facts that cannot both be true of one room.

Speaking in a channel means you were in it. The corpus has no way to say "left the channel", so
within this model a speaker is a member, and a member of a private channel can read it.

## Repro

Two records, neither of them unusual, both accepted by `backlot import` without a warning:

```jsonl
{"source_type":"slack","doc_id":"p1","channel":"board-comp","content":"comp band lands at 240k","created":"2026-01-05T09:00:00Z","author_email":"ava@x.com","visibility":"private","replies":[{"content":"noted","author_email":"bo@x.com","created":"2026-01-05T09:01:00Z"},{"content":"same","author_email":"cara@x.com","created":"2026-01-05T09:02:00Z"}]}
{"source_type":"slack","doc_id":"p2","channel":"exec-only","content":"board deck draft","created":"2026-01-06T09:00:00Z","author_email":"dee@x.com","readers":["user:eve@x.com"]}
```

| speaker | channel | in `conversations.members` | can see the channel |
|---|---|---|---|
| `ava@x.com` | `board-comp` | yes | yes |
| `bo@x.com` | `board-comp` | yes | **no** |
| `cara@x.com` | `board-comp` | yes | **no** |
| `dee@x.com` | `exec-only` | yes | **no** |

`visibility: "private"` is the sharper of the two: it is the documented spelling of a private thread,
grants read to the ROOT AUTHOR alone (`byo.py`, "grant tuples … shared by the whole thread"), and
every replier in that thread lands outside the grant. The corpus author wrote nothing wrong.

## Where the gate would go

`byo.py`'s per-record mapping already carries cross-field checks the schema cannot state, with the
precedent right there: `_github_pairing_errors` raises `SystemExit` for a pull-only field on an
issue. A slack speaker/reader check belongs beside it — the record has the root author, the reply
authors and the resolved grant set all in hand at that point.

What it should DO is the open question, because the two obvious answers each cost something:

- **Reject at import.** Honest, and it makes the invalid corpus unwritable. But `"readers": []`
  means admin-only, deliberately, and a document with an author would then always fail — so the
  check has to exempt the empty-readers case, i.e. state the rule as "a non-empty grant set must
  cover every speaker".
- **Grant every speaker.** No corpus changes needed, but it silently widens `readers` beyond what
  the corpus stated, which is exactly what the empty-readers comment argues against.

A warning rather than an error is the third option, and the weakest: an importer that prints and
proceeds still ships the contradictory DB.

## Scope of the damage today

- The **shipped ERB corpus is unaffected**: `erb.py`'s `grants_for` adds `("org", org)` to every
  Slack document ("channel privacy isn't recoverable from first-names → org-visible"), so it has no
  private Slack channel at all. Measured on the served DB: 5,648,107 org grants over 5,648,107
  messages, 36 channels, 0 private.
- The gap is entirely on the **BYO path**, which is the only way to model a restricted room — so it
  bites precisely the corpora written to grade a system on not revealing that a record exists.

## Related

Beside it, and probably the same decision: `is_member` is true iff the caller has spoken (#81), so
a caller GRANTED read on a private channel who never posted is listed with `is_member: false`,
which real Slack never returns for a private channel. If a corpus gate guarantees speakers ⊆
grantees, that residual case is the only place membership and readability can still disagree.

### LINKED ISSUE #84: slack: a missing required argument is invalid_arguments, not channel_not_found
Slack separates "you did not pass the argument" from "what you passed names nothing". The mock has
only the second answer, so a client that omits a required argument is told the thing it never named
does not exist.

Measured against slack.com with a workspace user token (the same session that settled the
`channel_not_found` question on #81):

| call | live | mock |
|---|---|---|
| `conversations.info` with no `channel` | `invalid_arguments` | `channel_not_found` |
| `conversations.members` with no `channel` | `invalid_arguments` | `channel_not_found` |
| `conversations.history` with no `channel` | `invalid_arguments` | `channel_not_found` |
| `conversations.replies` with no `ts` | `invalid_arguments` | `thread_not_found` |
| `conversations.*` with `channel=` (empty) | `channel_not_found` | `channel_not_found` ✓ |
| `conversations.*` with `channel=not-an-id` | `channel_not_found` | `channel_not_found` ✓ |
| `conversations.replies` with `ts=` (empty) on a readable channel | `thread_not_found` | `thread_not_found` ✓ |

So the rule is a two-stage one, and only the first stage is missing: **an ABSENT required argument
is `invalid_arguments`; one that is present but empty or unparseable is the resolution error.** The
mock collapses both into the resolution error.

`invalid_arguments` is in the documented error table for all four of those methods and appears
nowhere in this repo.

## Why it is worth having

The two answers send a client down different branches. `channel_not_found` is about the workspace —
retry with another id, ask the user to pick a channel again, or treat the channel as deleted.
`invalid_arguments` is about the caller — a bug in the request it just built. A connector under test
against the mock sees only the first, so the branch it takes for its own malformed request is never
the branch it will take in production.

## Shape

`_P_CHANNEL` / `_P_INFO` / `_P_HISTORY` / `_P_REPLIES` already declare which parameters are
`required`, so the check has a declaration to read rather than a new list to keep in step:
`_param(request, name) is None` (absent) against `== ""` (present and empty), which
`routers/slack.py`'s `_param` already distinguishes and every caller currently flattens with
`or ""`.

## Not measurable here

`users.info` (no `user`) and `search.messages` (no `query`) both answer `missing_scope` for this
token, which is checked before the arguments are — the workspace token has neither `users:read` nor
`search:read`. Their `invalid_arguments` rows are documented but unmeasured, so either add the
scopes before deciding those two, or leave them alone.

khj809:
Closes #81. Closes #83. Closes #84.

Which Slack conversations a caller may see, who is in them, and what the API says when it will not answer.

## `conversations.*` (#81)

**`types` selects.** It was parsed — and rejected an unknown value — but then read only as a boolean gate, so `public_channel` and `private_channel` never separated. It now filters on `is_private`, and an omitted argument means `public_channel`, Slack's documented default and what `_slack_types` already claimed to return.

**The by-id methods are scoped.** `conversations.info`, `.members`, `.history` and `.replies` resolved a channel by id and answered without asking whether this caller may see it, so a channel id was enough for any authenticated principal to confirm a private room, read its name, topic and purpose, and enumerate who is in it. `conversations.list`'s visibility test is now one predicate all of them share, and a channel it hides answers `channel_not_found` — the same answer an id that names nothing gets, so the two cannot be told apart. Not the `not_in_channel` that `conversations.history` also documents: that is the case where the token can see the channel and has not joined it, which is a public channel here and never refused.

**`is_member` is the caller's, and membership is per kind of channel.** It was the constant `True`, so every channel a caller could see claimed them as a member. Slack shows the two kinds of channel differently, so they answer from different facts: a public channel is shown to everybody, so membership is who has posted and a caller who never did is `is_member: false`; a private channel is listed to its members only, so there "may read it" and "is in it" are one fact and its grantees ARE its membership, groups expanded to their people. `is_member`, `num_members` and `conversations.members` all read whichever set applies, so a client that stats a channel and then walks it cannot get two answers. A service token is a member of nothing.

**`types` says what it does.** Its default decides whether private channels come back at all, and the OpenAPI slice is the whole of what a generated client — or an agent handed a bridged tool — knows about it. `conversations.list` publishes a description for it; `users.list` no longer advertises the argument, which it never honoured.

## Membership at import (#83)

A channel's members are the people who have spoken in it and what a caller may read is the ACL, with nothing tying the two together — so a corpus could state a private channel whose `conversations.members` listed people that `conversations.info` answered `channel_not_found`. Posting in a channel is being in it, so that pair cannot both be true.

`visibility: private` on a slack record now grants the thread's speakers rather than its root author alone — the documented spelling of a private thread put every replier outside the room they had replied in. A corpus that states its readers explicitly and leaves a speaker out is refused at load, against the DB and before the commit: covering a speaker can take a group grant, and a group's membership is only settled once the whole corpus has been read.

## Arguments (#84)

Slack separates "you did not pass the argument" from "what you passed names nothing"; the mock had only the second, so a client that omitted `channel` was told the thing it never named does not exist. The two send it down different branches — one is the workspace's shape, the other is a bug in the request the client just built. An absent required argument is now `invalid_arguments`, and one that is present but empty or unparseable stays the resolution error. The search family draws the same line with its own name for the blank half: `no_query`, which is what the vendor answers for a query that arrived with nothing in it.

`users.info` is the exception and is left one: live it answers `user_not_found` to every case, an absent `user` included. Its rows are in the test table so the rule is not tidied onto it later.

## Also

`BACKLOT_ENFORCE_ACL` is gone. It was declared on `Settings` and documented as turning ACL enforcement off, but no code path read it.

## Verification

The error answers are **measured**, not reasoned: against slack.com with a workspace user token, on a private channel that token is not in, beside a well-formed id naming no channel and a channel the token is in.

| | private channel the token is not in | id that names nothing | a channel it is in | argument absent |
|---|---|---|---|---|
| `conversations.info` | `channel_not_found` | `channel_not_found` | ok | `invalid_arguments` |
| `conversations.members` | `channel_not_found` | `channel_not_found` | ok | `invalid_arguments` |
| `conversations.history` | `channel_not_found` | `channel_not_found` | ok | `invalid_arguments` |
| `conversations.replies` | `channel_not_found` | `channel_not_found` | `thread_not_found` | `invalid_arguments` |

`channel=` and `ts=` sent empty are the resolution errors rather than `invalid_arguments`, and auth is settled before either — a bad token is `invalid_auth` with or without the arguments. The search methods answer `invalid_arguments` for an absent `query` and `no_query` for a blank one, and `users.info` answers `user_not_found` throughout.

The mock answers all 34 measured cases identically, checked by replaying the table against it. The hidden channel is also absent from `conversations.list`, and live `conversations.list` with `types` omitted returns public channels only.

On a corpus shaped like the reviewed one — a private channel granted to a four-person group, three of whom have posted — the lurker is served `is_member: true`, `num_members: 4`, four members and the full history, while a principal outside the group still cannot see the channel at all; the public channel they have never posted in stays `is_member: false`. Making the counts private-aware costs nothing measurable on the served corpus (6.66s for the whole warm-up pass, against the 12.2s already documented for it).

On #81's own repro corpus every cell of both its tables answers correctly, `bo` never reaches `board-comp` by any route, and `ava` still reads it in full. The membership gate caught a live instance in this repo's own tests — `ava` posted in `#eng-private` while never being in the `engineering` group it is restricted to — and finds none in the served ERB corpus (9ms; its channels are all public, and the org-grant early-out is what keeps that cheap). `crawl_slack` passes `types=public_channel,private_channel`, which is what a crawler must now do against the mock and against the real API alike.


nuri-yoo (commented):
Reviewed at `df6f407`, and merged with #79, against a 19,693-document corpus with 133 slack channels — 27 of them private — rather than the repro corpus.

Both halves of #81 close, measured there:

| | |
|---|---|
| `conversations.list` types | omitted 106 · `public_channel` 106 · `private_channel` 27 · both 133 |
| a private channel by id, to a non-member | `channel_not_found` from `.info`, `.members`, `.history` and `.replies` — the same answer an id naming nothing gets, so none of the four is an oracle |
| the argument absent instead | `invalid_arguments`, which is about the request rather than the workspace |
| a member | reads all four, `is_member: true`, `last_read` set |

The membership gate finds **0 violations** on that corpus (954 principals, 288 groups, 27 private channels), so a full load and `--append` both go through. Suite 1494 passed / 2 skipped, 1521 merged with #79, `--all-extras`.

Two things below. The first is the one I would want resolved before merge.

---

**`backlot/routers/slack.py` — a private channel is listed to somebody `is_member` calls a non-member**

`_is_member` is "has this person posted here", and on a private channel that is not the same set as "may this person read it". A caller granted read who never posted gets:

```
conversations.list    →  {"name": "ccro-staff", "is_private": true, "is_member": false, "num_members": 3}
conversations.history →  ok: true, the messages in full
conversations.members →  three ids, none of them theirs
```

Real Slack has no such state. A private channel appears in your `conversations.list` only because you are in it, and a non-member cannot read a word of its history — `is_private: true` + `is_member: false` + a full history is not a shape a client can be written against.

The docstring's reasoning is right for the case it names and does not reach this one:

> Seeing a channel is not being in it: real Slack lists the public channels a user is not in, and answers `is_member: false` for them.

True of public channels. A private channel is listed to members only, so on that side "can see it" and "is in it" are one fact, and the mock is serving them as two.

Not a corner: **22 of the 27 private channels** in the corpus I served have at least one reader who never posted (`ccro-staff`: 4 principals may read, 3 have spoken). `num_members` is the same set and under-reports for the same reason.

This is the mirror of the import gate the PR adds. #83 refuses a corpus whose *speaker* cannot read the channel; the direction left open is a *reader* of a private channel who is not reported as being in it. Either derive membership from the grant when the channel is private — the grant is the membership, in a model where posting implies reading — or refuse the pairing at import the way the other direction is refused.

---

**`backlot/routers/slack.py:72` — `types` now decides what a client sees and still carries no description**

```python
_P_LIST = [qp("limit", "integer"), qp("cursor"), qp("types")]
```

so `/_mock/openapi/slack` publishes

```json
{"name": "types", "in": "query", "required": false, "schema": {"type": "string"}}
```

Before this PR the argument was inert and an undescribed parameter cost nothing. Now omitting it hides every private channel, and the OpenAPI slice is the whole of what a generated client knows — `examples/using-mcp-with-agents/_openapi_bridge.py` turns exactly this into the tool an agent is handed, with no vendor docs behind it. Nothing in that tool says the default is `public_channel`, that the value is comma-separated, or that `im`/`mpim` are accepted and select nothing here.

Both SDK examples call it bare and now enumerate public channels only:

- `examples/using-official-sdk/slack.py:60` — `client.conversations_list()["channels"]`
- `examples/using-llamaindex-readers/slack.py:55` — `reader._client.conversations_list(limit=200)`

Against the ERB corpus that is invisible (all public). Against a corpus with private channels it silently halves the crawl — which is Slack-correct, and precisely why the parameter now needs to say so. `crawl_slack` was updated; the description is the part that reaches everyone else.

A sentence in `qp("types", ...)` covers it: default `public_channel`, comma-separated, one of `public_channel`/`private_channel`/`im`/`mpim`, unknown → `invalid_types`.

khj809:
Both fixed in `fd60289`.

**Membership** — you're right, and the docstring you quoted was the tell: I took the public-channel reasoning and applied it to a channel Slack only ever shows to its members. Membership now comes from a different fact per kind of channel — a private channel's grantees (groups expanded to their people), a public channel's speakers — and `is_member`, `num_members` and `conversations.members` all read whichever set applies, so stat-then-walk cannot disagree.

On your shape (a private channel granted to a four-person group, three of whom have posted): the lurker gets `is_member: true`, `num_members: 4`, four members and the full history; a principal outside the group still cannot see the channel at all; and the public channel they have never posted in stays `is_member: false`. The `num_members` under-report goes with it. Cost is nothing measurable — 6.66s for the warm-up counts pass on the served corpus, against the 12.2s already documented for it.

**`types`** — described now, and `users.list` no longer advertises the argument at all, since it never honoured it. The two SDK examples are left calling it bare, on your reading that this is Slack-correct and the description is the part that reaches everyone else.

Suite 1496 passed / 1 skipped, `--all-extras`.

nuri-yoo (approved):
Both closed at `fd60289`. Re-checked by running, on the same 19,693-document corpus — 133 channels, 27 of them private.

**Membership.** The exact shape I raised (`ccro-staff`, granted to four people, three of whom have posted), read as the lurker:

| | `df6f407` | `fd60289` |
|---|---|---|
| `conversations.list` | `is_member: false`, `num_members: 3` | `is_member: true`, `num_members: 4` |
| `conversations.members` | three ids, none of them theirs | four, theirs among them |
| `.history` | full | full |

The two cases that had to keep their old answer keep it: a principal outside the grant still gets `channel_not_found` from all four methods and no listing entry, and a public channel the same caller has never posted in is still `is_member: false`.

Stat-then-walk, checked exhaustively rather than on the one channel — every channel listed, its members paged, `num_members` against the length: **0 disagreements across all 133**, and no channel left at `num_members: 0`. On the private side every speaker is now in the members list too, asserted over all 27.

**`types`.** Described, and `users.list` no longer advertises an argument it never honoured — the better half of that fix. Threading `description` through `qp()` puts it where the next such parameter will find it.

**No regressions:** the import gate still finds 0 violations on this corpus, the error table is unchanged (`channel_not_found` from the four by-id methods, `invalid_arguments` for an absent argument), and the `types` counts still read 106 / 27 / 133. Suite 1495, and 1522 merged with #79.

**On the cost.** I could not reproduce 6.66s against 12.2s — my slack surface is 7,015 messages, where the whole counts pass is 1 ms before and 3 ms after. What I could measure is the shape of it, which is the part that generalises: the added work is per private channel × grantees and not per message. With every one of the 133 channels made private and granted to three ~50-person groups, the pass goes 1 ms → 21 ms and the expansion is all of it. Bounded, and paid once at warm-up — but it is the mostly-private corpus, rather than the message-heavy one, that the number to watch belongs to.

Approving.

PR #86, opened 2026-08-25
Templates for issues and pull requests, and a code of conduct

khj809:
The repo has a CONTRIBUTING.md and a CI workflow but nothing that shapes what arrives in an issue or a pull request. This adds those three files, written around what issues here have actually looked like — #23, #49, #68 and #81 are all the same report: a request, what Backlot served, what the vendor served, and the side-by-side measurement that settled it.

## Issue templates

`.github/ISSUE_TEMPLATE/` holds one template per shape an issue here takes:

- **Fidelity gap** — the common one, and built for the group of divergences one measurement pass finds rather than for a single line item. Issue-level prompts for the source and commit, the measurement and its date, and the corpus it was served over — for a mock served over a corpus you supply, those JSONL rows are the reproduction. Then a block per divergence: request, both answers, and whether a client can tell, since which ones are silent is what sets the order they get fixed in. A summary table goes above them once there is more than one, the way #23 opens.
- **Bug** — a crash, a wrong answer, or a failing import: what you ran, what happened verbatim, and which optional dependencies were present, pointing at CONTRIBUTING.md for what each one gates.
- **Feature or new surface** — what's missing, the vendor endpoint it stands in for, what it unblocks, and which layer it lands in: the ingest contract in `backlot/schemas/`, a route's `response_model`, or `backlot/cli.py`.

Each opens its issue under the label its kind now carries — `fidelity`, `bug`, `feature` — and the bug template's one-liner is the `bug` label's own wording, so the line between it and `fidelity` reads the same in both places. `docs` and `infra` issues have no template: they are usually a sentence, and the blank path stays open for them.

`config.yml` keeps blank issues enabled — the longest reports in this repo outgrow any form — and links the contributing guide.

## Pull request template

Asks for the finished state rather than the rounds of work behind it, the measurement that says the new behaviour is right, and verification as **visible** lines: the install used, the surfaces that ran rather than skipped, and the lint result. A skip count means nothing without the install that produced it, and an HTML comment would leave the rendered body carrying the count alone. Three checkboxes: a test that fails without the change, ACL scoping proved for both an admin and a scoped token, and comments that state what was measured rather than the history of the fix.

## What a dev-only run leaves closed

CONTRIBUTING.md named two optional surfaces, and there are more than that — so it now maps every extra and every external tool to what sits out without it, and the templates point at that one list. Three of the gates are module level, which is the part worth knowing: without `.[examples]` all of `tests/test_sdk.py` **and** all of `tests/test_s3.py` sit out, including the SigV4, 404 and 416 endpoint tests that never touch botocore; without `.[mcp]`, all of `tests/test_mcp.py`. `pytest -rs` is in the guide now because it is what makes the difference visible, and the zero-skip install is given as the two commands CI runs: the HubSpot llama-index reader is in no extra, since it pins `hubspot-api-client<9` against the `>=12` that `examples` needs. The guide also no longer says CI runs `pytest -q` — the workflow runs the suite, ruff, and the Linear example — and the templates don't ask for `pytest -n auto`, which `addopts` already supplies.

## Code of Conduct

A short adaptation of the Contributor Covenant 2.1 — pledge, expectations, enforcement — kept to a page. Its expectations are the ones this project runs on: review the work and not the person, and accept correction gracefully, since most of the history here is a measurement that disagreed with someone's assumption. Reports go to oceanjoon@brekkylab.com.

## Verification

Docs only — no source file changes. On `3a7e2d8`, which has `main` merged in, both figures with `git`, Docker, `npx` and `uvx` on PATH:

- `pytest` on a `.[dev]` install: **1466 passed, 23 skipped** — the 23 being the gates the new table names.
- `pytest` on `.[dev,examples,mcp,llamaindex,mirage]` plus `--no-deps "llama-index-readers-hubspot<0.6"`: **1528 passed, nothing skipped**.
- `ruff check .` and `ruff format --check .`: clean, 120 files.


.github/pull_request_template.md
```diff
@@ -0,0 +1,21 @@
...
+<!-- The measurement behind it — the live call, the introspection, the generated client — and
+     the date. Where a fix exists because someone diffed the two side by side, the comment beside
+     it should say so. -->
+
+## Verification
+
+<!-- Paste the counts, don't summarize them. `pytest -n auto` on a `.[dev]` install, plus the
+     extras covering what you touched: `.[examples]` for tests/test_sdk.py, `.[mcp]` + Docker for
+     tests/test_mcp.py. A plain run skips those silently. Also
+     `ruff check . && ruff format --check .`, which gates CI. -->
```
nuri-yoo: Two things here.

Five test surfaces self-skip on a `.[dev]` install, not two. Fresh venv on this head, `pip install -e ".[dev]"`, `pytest -rs`:

```
1462 passed, 23 skipped
```

| file | skipped | needs |
| --- | --- | --- |
| `tests/test_llamaindex.py` | 12 | `.[llamaindex]` |
| `tests/test_integrations.py` | 8 | `.[mirage]` (2), `.[llamaindex]` (3), `.[examples]` (3) |
| `tests/test_mcp.py` | 1, the whole module | `.[mcp]` |
| `tests/test_sdk.py` | 1, the whole module | `.[examples]` |
| `tests/test_s3.py:425` | 1 | `.[examples]` |

`.[examples]` also covers more than `tests/test_sdk.py`: `test_s3.py`'s botocore gate and three `googleapiclient` gates in `test_integrations.py`. The two mirage gates at `test_integrations.py:208,241` are the ones that sat out the `.[dev]` run in #64 while it read as green.

Second, position. All of this is an HTML comment, so once the section is filled in the body carries `N passed, M skipped` and nothing that says which install produced it — and M means nothing without that. #80 already writes the line this template is asking for:

> `pytest -n auto` is green — 1363 passed, 23 skipped under CI's `.[dev]` install; 1425 passed with every extra installed, Docker running and `llama-index-readers-hubspot` present, so nothing is skipped at all.

and #38 names the surfaces instead: "`test_sdk` / `test_mcp` / `test_llamaindex` / `test_editor_apis`: 104 passed — run, not skipped."

A visible line under `## Verification` — the install used, and which optional surfaces ran rather than skipped — keeps that instead of losing it at render.

While here: `.[mcp]` + Docker undersells `tests/test_mcp.py`. `mcp` gates the module at `:31`; Docker gates exactly one of its seven tests (`:134`); `:159` needs `npx` and `:210` needs `uvx`, and neither is mentioned here or in CONTRIBUTING.

.github/ISSUE_TEMPLATE/bug.md
```diff
@@ -0,0 +1,15 @@
...
+---
+
+**What you ran** — the command, the request, or the corpus record.
+
+**What happened** — output or traceback, verbatim.
+
+**What you expected instead**
+
+**Where you ran it** — Backlot version or commit, Python version, OS, and the extras you have
+installed if the failure is in a surface that needs them (`examples`, `mcp`, `dev`).
```
nuri-yoo: Same list as the PR template: `llamaindex` and `mirage` are missing, and `dev` is the baseline rather than an extra that gates a self-skip. A reporter whose failure is in `tests/test_llamaindex.py`, or in the mirage half of `tests/test_integrations.py`, gets no prompt to say so.

.github/ISSUE_TEMPLATE/fidelity-gap.md
```diff
@@ -0,0 +1,24 @@
...
+
+**The real API serves**
+
+**How you know** — the measurement, and the date and commit you made it at. A live call, an
+introspected schema, a generated client that fails to bind, or a vendor SDK that reads the
+response differently. Vendor docs are worth citing, but they have contradicted the live schema
+before now; say so if docs are all you have.
+
+**Can a client tell?** A `200` with the wrong body is worse than an error: nothing downstream
+notices until it runs against the real service.
```
nuri-yoo: The four issues this is written from are each several divergences, not one: #23 is seven, #49 is D1-D8 in three groups, #68 is six typed sections, #81 is three. One `Backlot serves` / `The real API serves` pair means a #23-shaped report either repeats the block by hand or leaves the template behind.

`Can a client tell?` has the same problem a level down. #23 answers it per divergence, in the table it opens with:

| # | Divergence | Silent? |
|---|---|---|
| 1 | `q: sharedWithMe` clause ignored | yes |

Asked once for the whole issue, it collapses seven answers into one — and which five of the seven were silent is the part that set the priority.

There is also no repro slot. #81 opens with three JSONL rows and a roster YAML; #23 with the shared setup its seven divergences came out of. For a mock served over a corpus you supply, those rows are the reproduction, and `Source and request` only asks for the call.

Splitting it fits all four: issue-level prompts for source, how you know (date + commit) and repro, then a divergence block stated as repeatable, with the silent/visible call inside it and a summary table once there is more than one.

nuri-yoo (approved):
Docs-only and the shape is right. Four notes inline; the first two are worth taking before merge.

Measured on `d95e42a`, fresh venv, `.[dev]` only: `1462 passed, 23 skipped` — five surfaces self-skip, not the two the template names.

CONTRIBUTING.md
nuri-yoo, line 1: Two lines outside the diff that this PR's new files now disagree with.

Lines 61-63 carry the same two-of-five extras list as the templates, and the same "Docker + `.[mcp]`" for `tests/test_mcp.py`. Now that the templates copy it, it is worth fixing once here and pointing at it from them.

Line 65 says CI runs `pytest -q`. `.github/workflows/ci.yml` runs `pytest -n auto`, plus a separate `lint` job and a Node job that typechecks and runs the Linear example. The new PR template says `-n auto`, so the two documents name different commands nine lines apart. `-n auto` is itself redundant — `addopts` is already `-q -n auto --dist loadfile` — so plain `pytest` in both places would settle it.

khj809:
All four addressed in `7952b35`. Re-measured here rather than taken on trust, and both counts match yours: `.[dev]` alone is **1462 passed, 23 skipped**; with `[dev,examples,mcp,llamaindex,mirage]` plus `llama-index-readers-hubspot`, and git, Docker, `npx` and `uvx` on PATH, **1524 passed, nothing skipped**.

**The extras list, and where it lives.** Fixed once in CONTRIBUTING.md as a table — every extra and every external tool against what sits out without it — and the templates now point there instead of carrying their own copy. Two things your list surfaced that the table states outright: three gates are module level, so `tests/test_sdk.py`, `tests/test_s3.py` and `tests/test_mcp.py` each sit out whole, and `test_s3.py` is the surprising one, since its SigV4, 404 and 416 endpoint tests never touch botocore. And `.[mcp]` alone does not cover `tests/test_mcp.py`: Docker, `npx` and `uvx` each gate one test — Atlassian, Notion, AWS. `pytest -rs` is in the guide now, since it is what makes any of this visible locally.

**Position.** `## Verification` is three visible lines now — install used, optional surfaces that ran rather than skipped, lint — so the rendered body cannot end up carrying `N passed, M skipped` with nothing that explains M. This PR's own body is filled in that way.

**The fidelity template.** Split as you described. Issue-level: source and commit, how you know with its date, and a reproduction slot for the JSONL rows and roster — for a mock served over a corpus you supply, those rows *are* the repro, and #81 opens with them. Then one block per divergence with `Can a client tell?` inside it, and a summary table above them once there is more than one, in #23's shape.

**CONTRIBUTING lines 61-66.** Both fixed. The `pytest -q` line no longer names a command at all — CI runs the suite, ruff, and the Linear example — and nothing asks for `-n auto`, which `addopts` already supplies.

On CI itself skipping while green: that is #87, which installs all four extras and puts `uvx` and `npx` on the runner, so the numbers above become CI's. This PR only fixes the documents.

The mirage gates at `test_integrations.py:208,241` you traced back to #64 are in the table under `.[mirage]`.

CONTRIBUTING.md
nuri-yoo: The `.[llamaindex]` row is right as written — without it, all twelve of `tests/test_llamaindex.py` sit out. The gap is the other direction: installing it does not close the file. `:119` needs `llama_index.readers.hubspot`, which no extra carries, because pyproject leaves it out deliberately (it pins `hubspot-api-client<9` against the `>=12` that `[examples]` needs) and documents `pip install --no-deps llama-index-readers-hubspot` as the way past.

Measured on `7952b35`: `uv pip install -e ".[dev,examples,mcp,llamaindex,mirage]"` resolves 205 packages, eight of them `llama-index-readers-*`, and not that one. With it installed and `git`, `npx` and `uvx` present:

```
1522 passed, 2 skipped
SKIPPED [1] tests/test_mcp.py:134: Docker not available
SKIPPED [1] tests/test_llamaindex.py:119: could not import 'llama_index.readers.hubspot'
```

Docker is just absent here and the table covers that one. The other is your 1524 minus the reader, so the arithmetic agrees — and #87's CI install is two commands for exactly this reason: `.[dev,examples,mcp,llamaindex,mirage]`, then `--no-deps llama-index-readers-hubspot`. The guide's block has only the first, so someone following it locally carries a skip CI will not.

nuri-yoo:
Went over `7952b35`. Every gate in `tests/` maps to a row now; the three module-level ones are `test_sdk.py:15`, `test_s3.py:425` and `test_mcp.py:31`, and the `git` row is one I missed — `-rs` never shows it on a machine that has git.

You are right about `test_s3.py`, and my row understated it: I listed `:425` as one skip alongside per-test rows, when it sits below thirteen endpoint tests and takes the whole file with it.

One thing left, inline on CONTRIBUTING: the install block is one command short of the zero-skip run.

khj809:
Right on both counts, and fixed in `3a7e2d8`. `main` is merged in, so this branch now has #87.

The install block is the two commands CI runs, and I took the pin from the workflow rather than from the pyproject comment — CI installs `--no-deps "llama-index-readers-hubspot<0.6"`, while `pyproject.toml`'s note beside the `llamaindex` extra says the unpinned `pip install --no-deps llama-index-readers-hubspot`. The guide follows the workflow, since that is the command with a green run behind it. Worth reconciling the pyproject comment separately if `<0.6` is load-bearing. The reader also has its own table row now, so the file it gates is named whether you are missing the extra or just the reader.

Re-measured on `3a7e2d8`, with git, Docker, `npx` and `uvx` all present:

- `.[dev]` alone: **1466 passed, 23 skipped** — the same 23 gates, four more tests behind them since #87.
- Both commands above: **1528 passed, nothing skipped**.
- `ruff check .` and `ruff format --check .`: clean, 120 files.

One thing I cannot account for: the first full run right after the merge came back `1 failed, 1527 passed`, and I lost the failing test's name to a truncated log. Four consecutive full runs since are green at 1528, and `tests/test_testing.py` on its own is green over six runs, so I have nothing to point at — noting it rather than leaving it out of the numbers.

nuri-yoo:
Pinned it: `tests/test_mcp.py::test_mcp_s3_lists_objects`.

Fresh worktree at `3a7e2d8`, fresh venv, both install commands. It failed on the first full run and I kept the report:

```
1 failed, 1526 passed, 1 skipped in 389.28s

tests.test_mcp :: test_mcp_s3_lists_objects   time=49.585s
McpError('Connection closed')
  → await sess.initialize()   (tests/test_mcp.py:123)
```

Same test immediately after, four times: `2.30s`, `5.85s`, `3.75s`, `2.80s`, green every time. Nothing changed but what `uvx` had already fetched for `awslabs.aws-api-mcp-server`. Cold, `uvx` is still resolving and building while the client waits on `initialize`, and the stdio stream closes under it. Worth being exact about the condition: my `~/.cache/uv` was 4.7 GB at the time, so this is that package being new to the cache, not an empty one — `rm -rf ~/.cache/uv/environments-v2` will not recreate it (that directory is 0 B here). `uv cache clean awslabs.aws-api-mcp-server` before the suite is the lever, and the module docstring already predicts the behaviour: "the first run downloads the package", for this test and the `npx` Notion one both.

**This is not only a first-run problem.** `_s3_params` asks for `awslabs.aws-api-mcp-server@latest`, and `@latest` makes `uvx` go to the index and decline to reuse a cached version. So the window reopens on its own the day awslabs publishes a release — on every machine at once, and on CI too, cache restore or not.

Two ways to close it, and they are not equivalent:

- **Pin the server.** `awslabs.aws-api-mcp-server==<version>` instead of `@latest`. The only one that also removes the publish-day recurrence, and it makes the test say which server it proved the mock against.
- **Warm it before the suite.** `uv tool install` the same pin in the CI step above `pytest`, and the fetch stops competing with the client's `initialize`. Locally it is a one-liner in the guide.

On CI: `setup-uv` runs `enable-cache: auto` and restores `~/.cache/uv` on every leg (`cache restored from GitHub Actions cache with key: setup-uv-2-…` in each job), which is why four legs stay green. The key carries no server version and GitHub drops a cache after seven days unused, so a cold leg is a matter of when.

On the pin you asked about: `<0.6` is not load-bearing today. PyPI's latest `llama-index-readers-hubspot` is 0.5.0 (`0.1.1 … 0.4.1, 0.5.0`), so the bound excludes nothing that exists — a precaution against the next minor, which is a reasonable thing for the guide to inherit from the workflow. Matching the pyproject note to it costs one word.

On losing the name: `--junit-xml` is how I got it. It went out with the `Every test ran` guard in `1bde5c5`, but the report is not the step that was turned down — the objection was to failing a green job over a skip, and a report fails nothing. It writes a file that names every test, its time and its traceback. Adding `--junit-xml=junit.xml` back to the pytest line, with no step reading it, is the whole cost, and the next one-off comes with a name attached instead of a truncated scrollback.

PR #87, opened 2026-08-26
CI runs the whole suite, on every Python the package claims

khj809:
The test job installed `[dev]` alone, so every test behind an optional dependency skipped while the job stayed green — the official vendor SDKs, the LlamaIndex readers, the three MCP servers, the mirage shims. It installs all four extras now, plus `llama-index-readers-hubspot` (`--no-deps` past the stale pin its own extra documents, bounded below 0.6), and `uvx` and `npx` join Docker and git on the runner so the MCP servers the suite drives have something to run under. 1528 tests run per leg, none of them skipped. pip and npm are cached, the install being the larger half of the job.

The job runs on 3.11, 3.12, 3.13 and 3.14 — what `requires-python` already claimed — with `fail-fast` off, so one version's break doesn't hide another's. The classifiers name all four.

`mock_server` and `serve_or_connect` take a bind address, loopback by default as before, and `base_url` follows it: a wildcard bind is dialled on loopback, a narrower one at the address it bound, which is the only place it answers. `_free_port` probes in that address's own family, so an IPv6 bind gets an IPv6 port.

The `live_server` fixture widens its bind where something outside loopback dials it — Linux with `docker` on PATH, the predicate the Atlassian MCP test itself opens with. Docker Desktop forwards `host-gateway` to the host's own loopback, so that container reaches a loopback-only mock on macOS; on Linux `host-gateway` is the bridge address, where it never could. That is every Linux contributor running `examples/using-mcp-with-agents/atlassian.py`, not only CI, so the example widens its bind the same way.

`tests/test_testing.py` settles the bind by request rather than by argument: whether Backlot answers on this machine's own routable address, asserted in both directions, plus a single-interface bind and `::1` dialled at their own addresses.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,9 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           node-version: "22"
-      - uses: astral-sh/setup-uv@v10
+      # Pinned to an exact version because that is all this action publishes — there is no
+      # floating `v10` tag to follow, unlike the actions/* above.
+      - uses: astral-sh/setup-uv@v10.0.1
 
       - name: Install
         # Every extra, so nothing self-skips: [examples] is the official vendor SDKs, [mcp] the MCP
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,17 +17,21 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
+          cache: pip
+          cache-dependency-path: pyproject.toml
       - uses: actions/setup-node@v4
         with:
           node-version: "22"
+          cache: npm
+          cache-dependency-path: examples/using-official-sdk/linear/package-lock.json
       - uses: astral-sh/setup-uv@v10.0.1
 
       - name: Install
         run: pip install -e ".[dev,examples,mcp,llamaindex,mirage]"
       - name: Install the HubSpot reader past its stale pin
-        run: pip install --no-deps llama-index-readers-hubspot
+        run: pip install --no-deps "llama-index-readers-hubspot<0.6"
       - name: Unit tests
-        run: pytest -n auto -rs
+        run: pytest -rs
 
       - name: Install the Linear example's dependencies
         working-directory: examples/using-official-sdk/linear
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,22 +8,66 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
+    strategy:
+      # Every version reports; one failing leg must not cancel the others, or a break on 3.14
+      # hides which of 3.11/3.12/3.13 also broke.
+      fail-fast: false
+      matrix:
+        python-version: ["3.11", "3.12", "3.13", "3.14"]
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
         with:
-          python-version: "3.13"
-      - name: Install
-        run: pip install -e ".[dev]"
-      - name: Unit tests
-        # -n auto: pytest-xdist comes from the dev extra; --dist loadfile is in pyproject's addopts.
-        run: pytest -n auto
+          python-version: ${{ matrix.python-version }}
 
-      # Node is for the Linear example at the end of this job: `@linear/sdk` is the only client
-      # Linear publishes and there is no official Python one, so pytest cannot drive it.
+      # Node and uv are here for the SUITE, not only for the Linear example at the end of this job:
+      # tests/test_mcp.py runs the Notion MCP server under `npx` and the AWS one under `uvx`, and
+      # skips whichever is missing. Docker (the Atlassian server) and git (tests/test_github.py)
+      # are already on the runner.
       - uses: actions/setup-node@v4
         with:
           node-version: "22"
+      - uses: astral-sh/setup-uv@v10
+
+      - name: Install
+        # Every extra, so nothing self-skips: [examples] is the official vendor SDKs, [mcp] the MCP
+        # servers, [llamaindex] the readers, [mirage] the virtual-filesystem shims. Whole test
+        # files sit behind a module-level importorskip on these.
+        run: pip install -e ".[dev,examples,mcp,llamaindex,mirage]"
+      - name: Install the HubSpot reader past its stale pin
+        # Deliberately not in [llamaindex]: it pins hubspot-api-client<9 against the >=12 that
+        # [examples] needs, which no resolver can reconcile. The pin is over-restrictive rather
+        # than a real incompatibility (see pyproject.toml), so --no-deps installs the reader
+        # without letting it near the resolver — the same line the extra's comment prescribes.
+        run: pip install --no-deps llama-index-readers-hubspot
+      - name: Unit tests
+        # -n auto: pytest-xdist comes from the dev extra; --dist loadfile is in pyproject's addopts.
+        run: pytest -n auto -rs --junit-xml=junit.xml
+      - name: Every test ran
+        # This job installs everything the suite can ask for, so a skip is a signal, not noise: an
+        # extra that quietly stopped resolving, or a `docker`/`npx`/`uvx` that went missing from the
+        # runner. Either way whole files stop running while the job still reports green, which is
+        # exactly how the SDK, reader and MCP tests came to be skipped here for so long.
+        run: |
+          python - <<'PY'
+          import xml.etree.ElementTree as ET
+
+          skipped = [
+              f"{case.get('classname')}::{case.get('name')} — {s.get('message', '')}"
+              for case in ET.parse("junit.xml").getroot().iter("testcase")
+              for s in case.findall("skipped")
+          ]
+          if skipped:
+              raise SystemExit(
+                  "these tests were skipped, but this job installs every dependency they "
+                  "need:\n  " + "\n  ".join(skipped)
+              )
+          print("no tests were skipped")
+          PY
+
+      # The Linear example is the one service pytest cannot drive: `@linear/sdk` is the only client
+      # Linear publishes and there is no official Python one. It starts the mock through $PYTHON,
+      # so it runs on each matrix leg rather than once.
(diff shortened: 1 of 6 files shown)


backlot/testing.py
```diff
@@ -194,7 +202,7 @@ def mock_server(records: list[dict] | None = None):
             check=True,
             stdout=subprocess.DEVNULL,
         )
-        port = _free_port()
+        port = _free_port(host)
         base = f"http://127.0.0.1:{port}"
```
nuri-yoo: `host` reaches the bind here, but `base` on the next line is a literal `127.0.0.1` and `_healthy` dials it, so the parameter only works for a wildcard. The docstring above says "``host`` is the bind address", which invites the LAN address:

```
>>> backlot.mock_server(host="192.168.0.14")
RuntimeError: Backlot did not become ready on http://127.0.0.1:62235
```

The server is up on that address the whole time — the health probe is knocking on loopback, and the message says the opposite of what happened, ten seconds later. `"::"` is a second one: `_free_port` opens an AF_INET socket, so it dies at `bind` with `gaierror: nodename nor servname provided`.

`mock_server` is public API and this is its new documented parameter, so either derive `base` from `host` when it is not a wildcard, or say in the docstring that `0.0.0.0` and `127.0.0.1` are what it takes and refuse the rest with a message that names the reason.

tests/conftest.py
```diff
@@ -967,6 +968,13 @@ def live_server(sample_settings):
...
     ``settings`` (``admin_token``, ``tokens_path``) even though they're different directories.
+
+    Bound past loopback on Linux only. ``test_mcp.py`` runs the Atlassian MCP server in Docker and
+    reaches back with ``--add-host=…:host-gateway``, which on Linux is the bridge address — a
+    loopback-only server never answers there. Docker Desktop forwards it to the host's loopback
+    instead, so on macOS the wider bind would buy nothing and cost every contributor the firewall
+    prompt that opening a port to the network raises.
     """
-    with backlot.mock_server(SAMPLE) as m:
+    host = "0.0.0.0" if sys.platform == "linux" else "127.0.0.1"
```
nuri-yoo: The wide bind exists for one test — `test_mcp.py:134`, the Atlassian container — and that test is `skipif(not _docker_available())`. On a Linux machine without Docker this fixture opens a module-scoped port on every interface for the length of the run and nothing dials it. `shutil.which("docker")` is the predicate `_docker_available()` already uses; ANDing it in leaves CI exactly as it is and keeps the narrow bind everywhere the wide one buys nothing.

No argument with the macOS half — Docker Desktop forwarding host-gateway to loopback is the reason, and the run proves the Linux side.

.github/workflows/ci.yml
```diff
@@ -8,22 +8,27 @@ on:
...
         with:
           node-version: "22"
+      - uses: astral-sh/setup-uv@v10.0.1
+
+      - name: Install
+        run: pip install -e ".[dev,examples,mcp,llamaindex,mirage]"
+      - name: Install the HubSpot reader past its stale pin
+        run: pip install --no-deps llama-index-readers-hubspot
+      - name: Unit tests
+        run: pytest -n auto -rs
```
nuri-yoo: `-rs` reports skips; nothing here fails on one. What this PR fixes is a snapshot rather than a property: a yanked reader, a runner image that drops Docker, a new gate arriving with a new test, and the job is quietly back to green-over-skips — the state the description opens with. A step that fails when the summary carries a skip would hold it.

Separately, `-n auto` is already in `addopts` (`-q -n auto --dist loadfile`). The comment that used to say so went out with this diff, so the flag now reads as load-bearing when it is not.

.github/workflows/ci.yml
```diff
@@ -8,22 +8,27 @@ on:
...
+          python-version: ${{ matrix.python-version }}
       - uses: actions/setup-node@v4
         with:
           node-version: "22"
+      - uses: astral-sh/setup-uv@v10.0.1
+
+      - name: Install
+        run: pip install -e ".[dev,examples,mcp,llamaindex,mirage]"
+      - name: Install the HubSpot reader past its stale pin
+        run: pip install --no-deps llama-index-readers-hubspot
```
nuri-yoo: No dependency cache, and the install is the bulk of the job now: on the 3.13 leg it ran 04:12:47 → 04:14:17, roughly 90 seconds of a three-minute job, and that is 205 packages fetched four times per push and again per pull request. `cache: pip` with `cache-dependency-path: pyproject.toml` on `setup-python` takes most of it back, and `setup-node` has `cache: npm` for the Linear steps below.

`llama-index-readers-hubspot` is unpinned, and `--no-deps` means pip will not tell you when it grows a requirement — it will import-fail on all four legs at once. An upper bound costs nothing.

tests/test_testing.py
```diff
@@ -113,6 +128,33 @@ def test_serve_or_connect_does_not_fetch_the_token_over_plain_http_to_a_non_loop
...
+
+        addr = _routable_address()
+        if addr is None:
+            return
+        port = int(m.base_url.rsplit(":", 1)[1])
+        if answers_off_loopback:
+            assert _get(f"http://{addr}:{port}/health")["status"] == "ok"
+        else:
+            with pytest.raises(OSError):
+                socket.create_connection((addr, port), timeout=5).close()
```
nuri-yoo: Three small things in this block.

`if addr is None: return` passes the test while checking half of what its name claims. On a network-isolated runner nothing says so, and `-rs` — which this PR just turned on — would not show it either. `pytest.skip("no routable address")` puts it where the rest of the suite's absences live.

The `else` branch asserts a property of the machine rather than of the server: that nothing at all answers on that port from outside. `_free_port("127.0.0.1")` only reserves it on loopback, as the comment you added to it says, and on Linux the other xdist workers are running `live_server` bound to `0.0.0.0` at the same time — a port collision turns this into a failure that has nothing to do with the bind. Asserting that `http://{addr}:{port}/health` does not answer as backlot narrows it to the thing under test. Low odds and four green legs, so hardening rather than a bug.

Last: this runs the `0.0.0.0` leg on every platform, including the macOS where `live_server` and `atlassian.py` both stay narrow to avoid the firewall prompt. It passed here with the firewall off, so it is a cost rather than a failure, and testing a parameter that exists on macOS on macOS is a fair answer — just noting that the two rationales point different ways.

nuri-yoo (approved):
Checked the run itself: all four legs are `1526 passed`, and the whole 4,888-line log has no `SKIPPED` line. The Atlassian Docker test actually executing on ubuntu is what proves the host-gateway reasoning. Locally on macOS with `.[dev]`: `1464 passed, 23 skipped`, no regression, and `tests/test_testing.py` is 11/11 on 3.14.

Five notes inline. The first two are the ones I would weigh before merge; the rest are follow-up sized.

One coordination note: the `main` ruleset already requires `test (3.11)` … `test (3.14)`, and main's workflow still emits a single `test`. #86 and #57 produce the old context — #57 reads BLOCKED right now — so this has to land before either of them can satisfy the rule.

backlot/testing.py
khj809: Fixed — `base` is derived from the bind now: `0.0.0.0` and `::` are dialled on loopback, anything narrower at the address it bound, and `_free_port` opens an AF_INET6 socket for a colon-bearing host. Both of your reproductions come up:

```
>>> backlot.mock_server(host="192.168.0.8")   # http://192.168.0.8:56870
>>> backlot.mock_server(host="::")            # http://[::1]:56876
```

`tests/test_testing.py` covers a single-interface bind and `::1` as two parametrized cases, each skipping when the machine has no such address to bind.

tests/conftest.py
khj809: Gated on `sys.platform == "linux" and shutil.which("docker")`. CI is unchanged, and a Linux box without Docker keeps the narrow bind — the test that dials the wide one doesn't run there either.

.github/workflows/ci.yml
khj809: `-n auto` is gone from the pytest line — `addopts` carries it, as you say.

The failing-on-skip step was in the first two pushes and came out at the repo owner's request. I've put your argument for it to them and they've chosen to keep it out for now, so it's a decision rather than an oversight. `-rs` stays, so a skip still names itself in the log.

.github/workflows/ci.yml
khj809: `cache: pip` keyed on `pyproject.toml` and `cache: npm` on the Linear lockfile are both in, and the reader is pinned `<0.6`.

It buys less than "most of it back", though — measured on the same commit, cold run against warm:

| leg | no cache | warm cache |
| --- | --- | --- |
| 3.11 | 105s | 78s |
| 3.13 | 90s | 81s |

pip's cache spares the downloads; the wheel installs and the sdist builds are what the rest of that minute-and-a-half is, and they run either way. Worth having at 10-25%, just not the bulk of it.

tests/test_testing.py
khj809: All three addressed.

The silent `return` is `pytest.skip("this machine has no address but loopback to dial")`, and it happens before the server starts rather than halfway through the assertions.

The off-loopback check now asks whether *Backlot* answered — `/health` coming back `status: ok` — through one `_health()` helper both branches share, so another program holding that port makes the narrow case pass for the right reason instead of failing for an unrelated one.

The `0.0.0.0` case still runs on macOS. It is the platform where nothing else exercises the parameter — `live_server` and `atlassian.py` both stay narrow there — so the prompt is the cost of testing it where it is otherwise untested.

nuri-yoo:
Confirmed on `2b42463`:

- `mock_server` dials where it bound. `192.168.0.14`, `::`, `::1`, `localhost` and both wildcards all come up with a matching `base_url` and a healthy `/health`.
- macOS: `1466 passed, 23 skipped`, `tests/test_testing.py` 13/13, ruff clean across 120 files. CI: four legs at `1528 passed`, no `SKIPPED` in 4,972 lines of log.
- The warm figures match — attempt 2 of the run on this commit is 78s on 3.11 and 81s on 3.13.

Could you update the description to match this push? It still describes the previous one.

khj809:
Description rewritten for `2b42463` — the bind derivation, the `docker`-gated fixture, the caches and the pin are in it, and the count is 1528.

Issue #92, opened 2026-08-27
github: no /branches or /tags listing, so a client cannot enumerate a repo's refs


khj809:
**What's missing** — the two repo ref *listings*, which Backlot serves no route for:

| Endpoint | Real GitHub | Backlot today |
|---|---|---|
| [`GET /repos/{owner}/{repo}/branches`](https://docs.github.com/en/rest/branches/branches#list-branches) | the repo's branches | 404 — only `/branches/{branch}`, the single-branch route, exists |
| [`GET /repos/{owner}/{repo}/tags`](https://docs.github.com/en/rest/repos/repos#list-repository-tags) | the repo's tags | 404 |

Both were found by pointing fsspec's `GithubFileSystem` at Backlot for `examples/using-fsspec/github.py`; they are the last two of its six endpoints with nothing to answer them.

**What it unblocks** — any client that enumerates a repo's refs before reading it. Concretely, `fsspec.implementations.github.GithubFileSystem` exposes `.branches`, `.tags` and `.refs` as properties over exactly these two routes; `backlot.integrations.fsspec.github_filesystem_at` points them at the mock, so today they raise `HTTPError: 404` instead of reaching api.github.com. A client that lists refs to pick one to read — rather than being handed a sha — cannot get started.

**Where it lands** — `backlot/routers/github.py`, beside the existing `@router.get("/repos/{owner}/{repo}/branches/{branch}")`. Most of the material is already there:

- `/branches` — one entry, for the repo's single branch. `repo_get` already reports `"default_branch": "main"`, and `_repo_commit_sha(repo)` / `_repo_tree_sha(repo)` already produce the commit the existing `/branches/{branch}` hands back, so the list is that same object in a one-element array.
- `/tags` — `[]`. A corpus states no tags, and an empty array is what real GitHub answers for a repo that has none, so this is a shape a client meets in production rather than a mock-only degenerate case (the same reasoning `commit_statuses` is written on).

**Reproduction**

```python
import backlot, requests

with backlot.mock_server() as m:                       # the bundled corpus, repo `pipeline`
    org = requests.get(f"{m.base_url}/_mock/users").json()["org"]
    h = {"Authorization": f"Bearer {m.token}"}
    for path in ("branches", "tags"):
        r = requests.get(f"{m.base_url}/github/repos/{org}/pipeline/{path}", headers=h)
        print(path, r.status_code)                      # branches 404 / tags 404
    print("single branch:",
[...]

PR #93, opened 2026-08-27
Add fsspec examples

khj809:
Stacked on #57.

## What changed

`examples/using-fsspec/` — three scripts reading the mock through [fsspec](https://filesystem-spec.readthedocs.io/), the filesystem interface pandas, pyarrow, dask and LlamaIndex all resolve URLs through. Each ends by loading a table into pandas, which is the point: `read_csv("s3://…")` in code nobody wrote for Backlot reads your corpus.

| Source | Implementation | How it is pointed at the mock |
|---|---|---|
| S3 | `s3fs` | `client_kwargs={"endpoint_url": f"{base_url}/s3"}` — no shim |
| Drive | `gdrive-fsspec` | `integrations.fsspec.drive_filesystem_at()` |
| GitHub | `fsspec.implementations.github` | `integrations.fsspec.github_filesystem_at()` |

`backlot/integrations/fsspec.py` — Drive and GitHub both hardcode their vendor host with no endpoint argument, and both build some of those URLs inside method bodies where there is no constant to rebind, so each is a subclass rather than a patcher. Drive's also works around three `gdrive_fsspec` defects that reproduce against real Google Drive: `ls` answering a directory out of the cached parent listing, `_strip_protocol` leaving a leading `/` on, and no `files.export` fallback for Google-native files.

`git/trees/{ref}` now resolves a **subtree** sha to that subtree, with paths relative to it. It previously answered the repo root for every ref, which is not a loud failure: a client descending by the sha it read from the parent listing gets the root's entries under the child's name, so `ls("src")` yields `src/src` and `src/config` and a walk recurses until it runs out of stack.

New `[fsspec]` extra, wired into CI and the CONTRIBUTING extras table.

The tree fix was measured against fsspec's `GithubFileSystem`, which is written for the live API: it reads a subtree's sha out of a listing and asks for it, which is how `GET /repos/{owner}/{repo}/git/trees/{tree_sha}` is documented to work. Against Backlot that walk raised `RecursionError`; it now terminates and matches the corpus.

The three Drive workarounds are client defects, not divergences — each was confirmed to reproduce with no Backlot involved, and each override says which one it fixes. Backlot's `files.export` and its `'<id>' in parents` handling were correct throughout.

Not fixed here: Backlot serves no `/branches` or `/tags` **listing**, so `fs.branches` and `fs.tags` are pointed at the mock and 404 rather than reaching GitHub — #92.

## Verification

- **Tests** — `pytest` on a `.[dev,examples,mcp,llamaindex,mirage,fsspec]` install: **1547 passed, 1 skipped** (12 new)
- **Optional surfaces that ran rather than skipped** — all of them; the single skip is the HubSpot reader, which no extra carries (`pip install --no-deps llama-index-readers-hubspot`, as CI does)
- **Lint** — `ruff check . && ruff format --check .`: clean

Each new shim was mutation-checked: disabling it fails its own test and no other. Removing the `branches`/`tags` overrides makes the "never addresses the real GitHub" test print the two `api.github.com` URLs it caught. The package was also rebuilt as a wheel and imported from outside the checkout, since an editable install hides a packaging defect.

- [x] A new endpoint serving corpus content is ACL-scoped, proved for both an admin and a scoped token — no new endpoint; the tree change stays inside the existing route's `visible_ids` scoping

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -72,6 +72,7 @@ file with them:
 | `llama-index-readers-hubspot`, which no extra carries | the HubSpot reader test in `tests/test_llamaindex.py` |
 | `.[mcp]` | all of `tests/test_mcp.py` |
 | `.[mirage]` | the shim tests in `tests/test_integrations.py` |
+| `.[fsspec]` — `fsspec` and `gdrive-fsspec` | the Google Drive and GitHub filesystem tests in `tests/test_integrations.py` |
 | Docker, `npx`, `uvx` | one `tests/test_mcp.py` test each — the Atlassian, Notion and AWS MCP servers |
 | `git` | the tests in `tests/test_github.py` that build a real repo |
 
@@ -82,7 +83,7 @@ rather than a real incompatibility, so it goes in past its own dependencies, at 
 installs:
 
 ```bash
-uv pip install -e ".[dev,examples,mcp,llamaindex,mirage]"
+uv pip install -e ".[dev,examples,mcp,llamaindex,mirage,fsspec]"
 uv pip install --no-deps "llama-index-readers-hubspot<0.6"
 ```
 
--- a/backlot/integrations/fsspec.py
+++ b/backlot/integrations/fsspec.py
@@ -1,0 +1,251 @@
+"""Read a Backlot server through fsspec, the filesystem interface pandas, pyarrow, dask and
+LlamaIndex all consume.
+
+S3 needs nothing from here: ``s3fs`` takes the endpoint as a constructor argument, so
+``S3FileSystem(client_kwargs={"endpoint_url": f"{base_url}/s3"})`` is the whole redirect (see
+``examples/using-fsspec/s3.py``). Its one wrinkle is an upstream ``walk`` bug, worked around by
+``integrations.llamaindex.patch_s3fs_walk`` — the same shim, whichever entry point reached it.
+
+Drive and GitHub are the two that need a module, both because their implementation hardcodes the
+vendor host and takes no endpoint argument. Neither can be redirected by rebinding a constant the
+way the rest of ``backlot.integrations`` does — the hosts they name are built inside method bodies
+— so each is a subclass that replaces those methods:
+
+- ``drive_filesystem_at`` — ``gdrive_fsspec`` (the ``gdrive://`` implementation). Supplies the
+  endpoint, and works around three of its defects that have nothing to do with Backlot: all three
+  reproduce against real Google Drive, and each override names the one it fixes.
+- ``github_filesystem_at`` — fsspec's own ``GithubFileSystem``. Supplies the endpoint for all six
+  of the api.github.com URLs it builds, closes the one host it takes from the SERVER's response
+  rather than building, and swaps HTTP Basic for the bearer scheme Backlot answers.
+"""
+
+from __future__ import annotations
+
+import io
+
+__all__ = ["drive_filesystem_at", "github_filesystem_at"]
+
+# What a Google-native file exports as. A Workspace file has no binary content, so `alt=media` —
+# the only download gdrive_fsspec knows — answers 403 fileNotDownloadable for these three.
+_EXPORT_FORMATS = {
+    "application/vnd.google-apps.document": "text/plain",
+    "application/vnd.google-apps.spreadsheet": "text/csv",
+    "application/vnd.google-apps.presentation": "text/plain",
+}
+
+_DRIVE_FS_CLASS = None
+
+
+def _drive_fs_class():
+    """The subclass, built on first use so importing this module needs no optional dependency."""
+    global _DRIVE_FS_CLASS
+    if _DRIVE_FS_CLASS is not None:
+        return _DRIVE_FS_CLASS
+
+    from gdrive_fsspec import GoogleDriveFileSystem
+    from google.api_core.client_options import ClientOptions
+    from google.oauth2.credentials import Credentials
+    from googleapiclient.discovery import build
+
+    class _BacklotDriveFileSystem(GoogleDriveFileSystem):
+        """``gdrive_fsspec`` pointed at a Backlot server instead of Google."""
+
+        def __init__(self, base_url: str, token: str, **kwargs):
+            self._base_url = base_url.rstrip("/")
+            self._token = token
+            self._export_cache: dict[str, bytes] = {}
+            # "anon" is the one connect() method that runs no OAuth flow; the override below
+            # replaces the service it would have built anyway.
+            super().__init__(token="anon", **kwargs)
+
+        def connect(self, method=None):
+            """The redirect. Drive's discovery document already carries the `/drive/v3` service
+            path in its rootUrl, so the replacement endpoint has to carry it too — the same rule
+            `integrations.llamaindex.point_drive_at` documents for the LlamaIndex reader."""
+            srv = build(
+                "drive",
+                "v3",
+                credentials=Credentials(token=self._token),
+                client_options=ClientOptions(api_endpoint=f"{self._base_url}/drive/v3"),
+            )
+            self.srv = srv
+            self.files = srv.files()
+
+        @classmethod
+        def _strip_protocol(cls, path):
+            """gdrive_fsspec's paths are relative and it leaves a leading "/" in place, so
+            `ls("/folder")` raises FileNotFoundError. Anything that roots paths at "/" — a URL
+            whose path is host-less, fsspec.fuse, plain habit — trips over it."""
+            return super()._strip_protocol(path).lstrip("/")
+
+        def ls(self, path, detail=False, trashed=False):
+            """gdrive_fsspec's `ls` consults `_ls_from_cache` first, which will answer a directory
+            listing out of the cached PARENT listing — with that directory's own entry rather than
+            its children. Dropping the parent entry forces the slow path, which re-fetches and
+            re-caches the parent on its way to the children."""
+            path = self._strip_protocol(path)
+            if path and path not in self.dircache:
+                self.dircache.pop(self._parent(path), None)
+            return super().ls(path, detail=detail, trashed=trashed)
+
+        def _export(self, path) -> bytes | None:
+            """The bytes a read of `path` returns, when Drive will only give them up as an export.
+            None for an ordinary binary file, which `alt=media` downloads normally."""
+            info = super().info(path)
+            mime = _EXPORT_FORMATS.get(info.get("mimeType"))
+            if mime is None:
+                return None
+            if info["id"] not in self._export_cache:
+                self._export_cache[info["id"]] = self.files.export(
+                    fileId=info["id"], mimeType=mime
+                ).execute()
+            return self._export_cache[info["id"]]
+
+        def info(self, path, trashed=False):
+            """`size` on a native file is the length of the stored content, but a read returns the
+            EXPORT of it, which is longer. Anything that trusts `size` to size a read — fsspec.fuse,
+            a range request — truncates the file unless the two agree.
+
+            Only here, not in `ls(detail=True)`: reconciling a whole listing would mean exporting
+            every native file in the directory just to measure it. A bulk listing keeps Drive's own
+            number, so `ls` and `info` can disagree on a Doc, Sheet or Slide deck — `info` is the
+            one that matches what a read returns.
+            """
+            info = super().info(path, trashed=trashed)
+            body = self._export(path) if info.get("type") == "file" else None
+            return info if body is None else {**info, "size": len(body)}
+
+        def cat_file(self, path, start=None, end=None, **kwargs):
+            body = self._export(path)
+            if body is None:
+                return super().cat_file(path, start=start, end=end, **kwargs)
+            return body[start:end]
+
+        def _open(self, path, mode="rb", **kwargs):
+            body = self._export(path) if mode == "rb" else None
+            if body is None:
+                return super()._open(path, mode=mode, **kwargs)
+            return io.BytesIO(body)
+
+    _DRIVE_FS_CLASS = _BacklotDriveFileSystem
+    return _DRIVE_FS_CLASS
+
+
+def drive_filesystem_at(base_url: str, token: str, **kwargs):
+    """An fsspec filesystem reading Backlot's Drive API at `base_url`, authenticated by `token`.
+
+    Paths are relative to My Drive: `"folder"`, `"folder/file"`, `""` for the root.
+    """
+    return _drive_fs_class()(base_url=base_url, token=token, **kwargs)
+
+
+_GITHUB_FS_CLASSES: dict[str, type] = {}
+
+
+def _github_fs_class(api: str) -> type:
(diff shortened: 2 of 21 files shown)


backlot/integrations/fsspec.py
```diff
@@ -0,0 +1,215 @@
...
+        return _GITHUB_FS_CLASSES[api]
+
+    import requests
+    from fsspec.implementations.github import GithubFileSystem
+
+    class _BacklotGithubFileSystem(GithubFileSystem):
+        """``GithubFileSystem`` pointed at a Backlot server instead of GitHub."""
+
+        url = api + "/repos/{org}/{repo}/git/trees/{sha}"
+        content_url = api + "/repos/{org}/{repo}/contents/{path}?ref={sha}"
```
nuri-yoo, line 171: `_open` does not always stop at `content_url`. When the contents response's `content` is empty, or when the decoded bytes start with `version https://git-lfs.github.com/`, it falls through to `self.http_fs.open(content_json["download_url"])` — and Backlot's own `_file_obj` sets `download_url` to `https://raw.githubusercontent.com/{owner}/{repo}/{ref}/{path}`. So a corpus that carries a git-LFS pointer as a file's content makes this filesystem read from the real GitHub.

Reproduced on `7587799` with a two-document corpus whose `assets/model.bin` content is a three-line LFS pointer, `fs.http_fs` swapped for a recorder:

```
fell through to: ['https://raw.githubusercontent.com/acme/platform/main/assets/model.bin']
```

The empty-content branch is unreachable — the corpus schema rejects `''` (`'' should be non-empty`) — so the pointer is the live path. Six is the right count for the URLs the client builds; this seventh comes back from the server, which is why a subclass that fixes the client cannot see it. Serving the bytes the contents response already carried, or rewriting the host, in an `_open` override would close it.

tests/test_integrations.py
```diff
@@ -270,3 +272,133 @@ def test_patch_linear_at_only_rewrites_linear_urls():
...
+
+    monkeypatch.setattr(requests, "get", spy)
+    github_fs.invalidate_cache()
+    github_fs.ls("")
+    for prop in ("branches", "tags"):
+        with contextlib.suppress(Exception):
+            getattr(github_fs, prop)
+
+    assert seen, "nothing was requested, so this asserts nothing"
+    assert not [u for u in seen if "api.github.com" in u], seen
```
nuri-yoo: The guard is narrower than what this docstring and the example README claim for it. It greps `api.github.com`, but the host a read can actually escape to is `raw.githubusercontent.com` (see the comment on `integrations/fsspec.py`), so it would pass through the one case that matters. `assert not [u for u in seen if "github.com" in u]` covers both hosts and stays true for everything the examples do — `html_url` is never requested, only returned.

Separately, the docstring names `repos` alongside `branches` and `tags`, but the loop below touches only the two properties, so the `repos` override is not exercised by anything. Worth knowing before adding it: upstream `repos` sends no auth and the override keeps that, so against Backlot it 401s — measured on `7587799`, `401 Client Error: Unauthorized for url: http://127.0.0.1:49307/github/orgs/acme/repos`. That is still fine for a where-was-it-addressed assertion, which is all this test claims to check.

backlot/integrations/fsspec.py
```diff
@@ -0,0 +1,215 @@
...
+
+        def __init__(self, org, repo, token=None, sha=None, **kwargs):
+            # Replaces, rather than extends, the default-branch lookup in the parent's __init__:
+            # that one builds its URL as a local variable, so passing `sha` is the only way to stop
+            # it reaching api.github.com.
+            self._token = token
+            if sha is None:
+                r = requests.get(f"{api}/repos/{org}/{repo}", timeout=self.timeout, **self.kw)
```
nuri-yoo, line 182: `self.timeout` is still the class default here: the parent applies a `timeout=` kwarg inside its own `__init__`, which runs on the line below. Measured on `7587799` — `github_filesystem_at(..., timeout=(3, 3))` issues this default-branch lookup with `(60, 60)` and the tree listing immediately after it with `(3, 3)`:

```
requests during construction: [(['acme', 'platform'], (60, 60)), (['trees', 'main'], (3, 3))]
```

Popping `timeout` out of `kwargs` and setting `self.timeout` before the request would make the two agree.

CONTRIBUTING.md
```diff
@@ -72,6 +72,7 @@ file with them:
 | `llama-index-readers-hubspot`, which no extra carries | the HubSpot reader test in `tests/test_llamaindex.py` |
 | `.[mcp]` | all of `tests/test_mcp.py` |
 | `.[mirage]` | the shim tests in `tests/test_integrations.py` |
+| `.[fsspec]` — `s3fs` and `gdrive-fsspec` | the Google Drive filesystem tests in `tests/test_integrations.py` |
```
nuri-yoo: The row is narrower than the extra in one direction and wider in the other. `github_mock` calls `importorskip("fsspec")`, and nothing in the base or `dev` dependency sets pulls fsspec in, so the four GitHub filesystem tests sit out alongside the six Google Drive ones. On a `.[dev]`-only install of `7587799`, `-rs` reports both groups:

```
SKIPPED [1] tests/test_integrations.py:362: could not import 'fsspec': No module named 'fsspec'
```

The other direction: `s3fs` gates no test at all — the S3 leg is example-only, since s3fs needs no shim — so naming it here points at a gate that does not exist. This page's own rule is that `-rs` should confirm nothing you meant to run skipped, which is exactly what the omission costs.

```
| `.[fsspec]` — `fsspec` and `gdrive-fsspec` | the Google Drive and GitHub filesystem tests in `tests/test_integrations.py` |
```

backlot/routers/github.py
```diff
@@ -1057,9 +1057,16 @@ async def get_tree(
...
-    git/ref — resolves to the repo's CURRENT files. FILES themselves do have history: a corpus may
-    state several snapshots of one path, and `/contents/{path}?ref=` reaches them (see
+    `ref` selects WHICH tree, exactly as on real GitHub: a SUBTREE's own sha — the one a client
+    reads out of a parent listing's `tree` entry — answers that directory's entries, with paths
+    relative to it. That is how a client walks a repo one level at a time (fsspec's
+    `GithubFileSystem` does), and answering the root instead does not fail loudly: it reports the
+    root's entries under the child's name, so listing `src` yields `src/src` and `src/config` and a
+    recursive walk descends until it runs out of stack.
+
+    **A tree keeps no history**, so every other `ref` — a branch name, or a sha from /branches,
```
nuri-yoo, line 1076: Now that some shas resolve to a tree, the ones that do not land in this sentence, and root is a quiet answer for them. Two cases, both measured on `7587799`:

A 40-hex sha naming nothing in this repo returns the full root, with `sha` set to the root's rather than to what was asked for — `GET git/trees/000...0` answered `200` with `['README.md', 'open', 'secret', 'src']`. Real GitHub 404s. A blanket 404 would break the commit-sha and git/ref cases this paragraph deliberately keeps, so it is not a one-liner.

And a subtree sha resolves per caller, because `entries` is already `visible_ids`-scoped: asking for `secret`'s sha with a token that can see nothing under `secret` returns that caller's root instead of the subtree. No content crosses the ACL — the response held only what that token could already read — and fsspec never asks for a sha it did not just read out of its own listing, so nothing here is reachable through the examples.

The paragraph above now says `ref` selects which tree exactly as on real GitHub, and these two are where it stops doing that. A clause naming them would keep this honest for someone reading the docstring to decide what to trust.

nuri-yoo (approved):
Ran the suite and the three example scripts against `35238ba` on macOS / Python 3.13 with a `.[dev,fsspec]` install: `1512 passed, 19 skipped`. The 19 are the `llamaindex`, `mcp`, `mirage` and `examples` extras I did not install — CI installs all of them and is green on four legs. `ruff check .` and `ruff format --check .` clean. `s3.py`, `gdrive.py` and `github.py` each run end to end and print a DataFrame.

The tree fix is the part I pushed on hardest and it holds. On a five-level corpus (`src/ingest/deep/a/b/c/leaf.py`) `fs.walk("")` descends to the bottom, one request per directory, and terminates. Reversing the subtree off the entries already built is the right call: `_dir_sha` really is a pure function of (repo, path), so the sha a listing hands out is exactly the one recomputed here, and doing it this way keeps the resolution inside the existing `visible_ids` scoping without a second code path.

Five comments, none blocking. One is a reachable path where this filesystem still leaves the mock — `raw.githubusercontent.com`, reached through `download_url` rather than through any URL the subclass builds, which is why counting the client's six did not catch it. The rest are a guard whose assertion is narrower than its stated claim, a `timeout` kwarg that does not reach the first request, a CONTRIBUTING row that understates what the extra gates, and a docstring sentence the subtree change left incomplete.

The delta from `7587799` to `35238ba` is prose only (Drive -> Google Drive plus a reflow); everything above was measured on `7587799` and re-anchored against `35238ba`.

backlot/integrations/fsspec.py
khj809: Reproduced and fixed in 34ac33f. `_open` is replaced rather than extended: it reads the contents response and returns those bytes, with no path to `download_url` at all. Backlot always inlines `content` base64 (`_file_obj`, no size cap) and has no LFS, so there is nothing to fall through to — a corpus that states a pointer as a body means those are the bytes, which is what the new test asserts.

Measured before the fix, with `http_fs` swapped for a recorder:

```
#1 fell through to: ['https://raw.githubusercontent.com/acme/platform/main/assets/model.bin']
```

The docstrings that counted six now say why six was the wrong number to count.

tests/test_integrations.py
khj809: Widened to `github.com` in 34ac33f, and `repos` is now called in the loop so the docstring names only what it exercises.

One correction on the reasoning, because it changes what the guard is worth: widening the pattern alone would NOT have caught the LFS case. That read leaves through `http_fs` — an `HTTPFileSystem`, i.e. aiohttp — not through `requests`, so the spy records nothing while the request goes out. Verified: with the LFS corpus in place and the `github.com` pattern, this test still passed while the read reached raw.githubusercontent.com. So the LFS body gets its own test with an `http_fs` recorder, and the widened pattern stands on its own merit rather than as cover for that case.

The 401 on `repos` is what I see too; left as-is, since where-it-was-addressed is all this asserts.

backlot/integrations/fsspec.py
khj809: Fixed in 34ac33f — `timeout` is popped into `self.timeout` before the lookup and still passed down, so the parent's own handling is unchanged. Same measurement now reads `[(3, 3), (3, 3)]`, and there is a test that fails with `(60, 60)` on the first entry without it.

CONTRIBUTING.md
khj809: Taken verbatim in 34ac33f — the row now names `fsspec` and `gdrive-fsspec`, and both the Google Drive and GitHub groups.

On s3fs: it does gate one test, `tests/test_llamaindex.py::test_s3` (`importorskip("s3fs")`), but that test also needs `llama_index.readers.s3` and `[llamaindex]` already ships s3fs — so the gate is not this extra's, and naming it here pointed at the wrong row either way. It stays in the extra's dependency list, since `s3.py` needs it.

backlot/routers/github.py
khj809: Both reproduced on the `codebase` fixture and named in the docstring in 34ac33f:

```
admin root: ['README.md', 'config', 'src']
  admin            sha=75f0e289 tree=['secret.yaml']
  bob-no-config    sha=9943feb2 tree=['README.md', 'src']
  unknown-sha      status=200 sha=9943feb2
```

The paragraph now says where `ref` stops selecting a tree: an unknown 40-hex sha answers the root with the root's sha, because the fallback that keeps commit and git/ref shas working cannot tell it from one that means nothing; and a subtree sha resolves per caller, because the entries it is matched against are already `visible_ids`-scoped. It also records that no content crosses the ACL — the answer holds only what that token could already read — so the cost is a wrong tree, not a leak.

No test for either: both are limitations rather than behaviour to hold still, and 404-ing the unknown sha is the change this docstring deliberately does not make.

backlot/integrations/fsspec.py
nuri-yoo: Confirmed on `34ac33f`. The repro that printed `raw.githubusercontent.com/acme/platform/main/assets/model.bin` now prints `fell through to: []`, and the pointer comes back as the file's bytes. Replacing `_open` rather than widening the guard is the stronger of the two fixes: there is no fallthrough left to miss.

tests/test_integrations.py
nuri-yoo: Verified on `34ac33f`: `1514 passed, 19 skipped` locally on `.[dev,fsspec]`, and both mutations are single-test kills — delegating `_open` back to upstream fails only the LFS test, dropping the early `self.timeout` fails only the timeout one.

You're right about the pattern. I only saw that URL by swapping `fs.http_fs` for a recorder, which is exactly why a `requests` spy can't, and I proposed it as a fix anyway. Its own test is the right shape; the widened pattern stands on the client's URLs alone.

The 401 on `repos` reads the same to me. Where it was addressed is the whole assertion.

backlot/integrations/fsspec.py
nuri-yoo: Confirmed — the same measurement now reads `[(3, 3), (3, 3)]` on `34ac33f`, and removing the early assignment fails only the new test.

CONTRIBUTING.md
nuri-yoo: Checked the s3fs point and you are right: `test_llamaindex.py:85` does `importorskip("s3fs")` and the `llamaindex` extra ships `s3fs>=2026.6`, so the gate exists but belongs to that row rather than to this one. My comment named it under the wrong extra.

backlot/routers/github.py
nuri-yoo: The paragraph covers both cases I measured, including that the answer holds only what the token could already read. Agreed on leaving them untested: a test there would pin behaviour the docstring is deliberately declining to promise.

nuri-yoo:
The conflict is from the README revision that just landed on `readme-landing-page` (87e30f3), and it should be cheap to resolve: the new README already carries this PR's fsspec row (under **Examples**, phrased "Read it with `pandas`, `pyarrow` or `dask`, through fsspec"), and the other two README lines in this diff predate the "Serve your own enterprise playground" copy change. So on rebase, taking the base branch's README wholesale — dropping this PR's README.md diff entirely — loses nothing. Everything else here is untouched by the revision.

---

Correction to the above: the fsspec row is now *out* of the base README — it pointed at a directory that lands with this PR, and `tests/test_docs.py` fails on unresolved relative links. So on rebase, the README.md diff for this PR becomes exactly one line, inserted in the **Examples** table after the LlamaIndex row:

```
| 🐍 Read it with `pandas`, `pyarrow` or `dask`, through [fsspec](https://filesystem-spec.readthedocs.io/) | [`examples/using-fsspec/`](examples/using-fsspec/) |
```

Everything else in the earlier comment still holds: drop the rest of this PR's README diff.

PR #94, opened 2026-08-28
Add agent skill and plugins

### LINKED ISSUE #90: An agent skill for driving Backlot end to end, shipped from the repo as its own plugin marketplace
**Depends on #57**, which creates the `docs/` surface this builds on: `docs/supported-sources.md` (URL prefix, every endpoint, and the fidelity notes, per service), `docs/auth.md` (the auth scheme per service), `docs/corpus.md` (what a record must state), `docs/configuration.md`. This issue adds no second copy of any of it.

**What's missing** — Backlot has no agent-facing surface. Everything an agent needs to go from "mock our Slack with these three messages" to a client reading them back is here, but only as prose written for a human to read in order: which fields a record of that source must carry, which prefix and auth scheme its clients want, and which of four client paths is the right one (`examples/using-official-sdk/`, `using-mcp-with-agents/`, `using-llamaindex-readers/`, `using-mirage/`). An agent rediscovers the procedure on every task, and gets wrong the two things a wrong answer hides best: it skips `backlot import --dry-run`, so a record short a field surfaces later as a corpus quietly thinner than intended rather than at import time by name; and it reaches for `backlot.mock_server()` — a context manager that dies with the Python process — for a server that has to outlive the turn.

The [Agent Skills](https://code.claude.com/docs/en/skills) format is built for this: a `SKILL.md` loaded when the task matches, read alongside the repo it describes.

**What it unblocks** — the connector developer's first hour. Today that hour is "read the docs, pick a source, learn its record shape, learn its auth, find the example that matches your client"; with the skill it is one sentence to an agent, and what comes back is a running server plus the first real response from a client, checked against the corpus that produced it. The same loop covers the other two things Backlot is for: ACL-scoped retrieval, since `/_mock/users` hands out one identity per person and the skill can serve the same query under two of them, and reproducing an edge case a live account cannot reach — write the record that causes it, `--dry-run`, serve.

**Where it lands** — no change to the server, the corpus format, or the docs. Two new top-level entries, one generated block, one test file:

```
skills/backlot/SKILL.md              # the loop, and one generated routing table — the only copy
.claude-plugin/marketplace.json      # Claude Code: plugins: [{ name: backlot, source: "./" }]
.claude-plugin/plugin.json           # Claude Code: name, description, version
.codex-plugin/plugin.json            # Codex: the same, plus "skills": "./skills/"
.agents/plugins/marketplace.json     # Codex: plugins: [{ source: { source: "url", url: "./" } }]
scripts/gen_docs.py                  # +1 generated block, from the SOURCES dict it already owns
tests/test_skills.py
README.md                            # a three-line "Agent skills" section
```

## Distribution: the repo is its own marketplace, once per harness

`skills/backlot/` is the only copy of the skill. Each harness gets a small manifest pointing at it, so supporting a second one adds manifests, never content.

**Claude Code.** `.claude-plugin/marketplace.json` names `"source": "./"`, so the repo is both the catalog and the plugin. Installing is two commands, and no `pip install` has to come first:

```bash
claude plugin marketplace add brekkylab/backlot
claude plugin install backlot@brekkylab
```

**Codex.** Codex will not take a git repo as a skill source directly — distribution there goes through a plugin. `.codex-plugin/plugin.json` declares it and points its `"skills"` field at `./skills/`, and `.agents/plugins/marketplace.json` makes the repo a Codex marketplace, its `plugins[].source` set to `{"source": "url", "url": "./"}`. That is the same two files' worth of work as the Claude side, over the same `skills/backlot/SKILL.md`, and it is the layout [obra/superpowers](https://github.com/obra/superpowers) and [hashicorp/agent-skills](https://github.com/hashicorp/agent-skills) both ship.

Codex's *repo-level* auto-discovery is deliberately left out: it scans `.agents/skills` in the working tree rather than `skills/`, so serving it would mean a second location for the same skill. A symlink there is worse than it looks — `tests/test_docs.py` walks the repo with `rglob("*.md")`, which traverses directory symlinks, so every skill file would be visited twice. The plugin route covers everyone who installs Backlot; the only reader it misses is someone editing Backlot itself under Codex, who has the file open already.

Three consequences, in the order they matter:

1. **The skill and the surface it describes are in the same commit.** A PR that changes what is served fixes the skill in the same diff, and the tests below fail if it does not. Drift is handled by git rather than by keeping two release channels in step.
2. **The whole repo lands in the plugin cache**, not just `skills/` — measured on an installed whole-repo plugin, whose cached plugin root holds its `README.md`, `docs/` and `scripts/` alongside `skills/`. So the skill points at `docs/supported-sources.md`, `backlot/schemas/slack.schema.json` and `examples/using-official-sdk/slack.py` by relative path, and those paths resolve for someone who installed the skill and never cloned anything. The clone is cheap: 163 tracked files, 12.5 MiB packed, with `data/` and `dataset/` gitignored.
3. **The remaining harnesses come free.** With `skills/` at the repo root, `npx skills add brekkylab/backlot` installs the same skill into the other agents that read the portable format.

Shipping a vendor's own skills from the vendor's own repo-as-marketplace is the established shape — [hashicorp/agent-skills](https://github.com/hashicorp/agent-skills), [microsoft/power-platform-skills](https://github.com/microsoft/power-platform-skills). Packaging the skill inside the wheel was considered and rejected: it has no precedent, and it inverts the dependency, since a skill whose first step may be `pip install backlot` cannot be readable only after that step has run.

## Two axes, so nothing is written twice

`docs/` is cut **by source**: one section per service, holding its prefix, its endpoints, and the fidelity notes that are the valuable part. The skill is cut **by task**: decide the corpus, validate it, isolate it, serve it with the right lifetime, get credentials, wire a client, check the answer. Neither axis restates the other, and per-source prose keeps exactly one home.

So there is no `skills/backlot/references/<source>.md`, and no per-source skill either. A per-source file would be a fourth copy of what `docs/supported-sources.md`, `docs/auth.md` and `examples/*/` already hold — the copy #57 declined to make when it ruled out `docs/clients.md`. Splitting `docs/` by source instead was considered and rejected on two measurements: a source needs 13-56 lines of `docs/supported-sources.md` and 7-34 of `docs/auth.md`, not their 533-line total, so the routing table handing over an exact heading closes most of the gap; and the auth axis does not factor by source at all — 11 sources come to 9 sections there, because Jira and Confluence share one Basic-auth section and the five Google surfaces share one Bearer story, so per-source files would duplicate that prose twice and five times over.

The loop, meanwhile, is the same for every source, which is why it is one skill and not eleven: the source list is not finished (#89), and eleven copies of one procedure drift eleven ways.

Once a server is up, the skill prefers the **served** surface over any description of it: `GET /_mock/openapi/<source>` for the typed per-source slice, and GraphQL introspection for Linear and Fireflies. That is measured rather than written down, so it cannot go stale.

## The one per-source thing, generated

`SKILL.md` carries a routing table — one row per `source_type`, with its URL prefix, its record schema, its section in `docs/supported-sources.md`, its section in `docs/auth.md`, and the `examples/` scripts that exist for it. `scripts/gen_docs.py` writes it into a `<!-- generated:skill-sources -->` block from the same `SOURCES` dict that produces the table in `docs/supported-sources.md`, so this introduces no second mapping and inherits `--check`.

That leaves `description` as the one hand-written thing that decides whether any of this runs, because the request is "mock our Slack", never "use Backlot". It has to name every source:

> Serve a read-only mock of enterprise SaaS knowledge APIs — Slack, Gmail, Google Drive, Docs, Sheets, Slides, GitHub, Jira, Confluence, Notion, Amazon S3, HubSpot, Linear, Fireflies — over a corpus you supply, then point a real client (vendor SDK, MCP server, LlamaIndex reader, mirage) at it. Use when asked to mock, stub or reproduce one of those services, to write or validate a Backlot BYO-JSONL corpus, or to test a connector, RAG pipeline or ACL-scoped retrieval without live vendor accounts.

## SKILL.md — the loop

1. **Check the install** — `backlot --version`, then `backlot status` for whether a corpus is already loaded and what is in it.
2. **Decide the corpus** — `--bundled` for a demo or a shape question, the user's own documents for a connector, hand-written records for reproducing one behaviour. Say which of the three this task is.
3. **Write the records** — the routing table gives the source's schema. `backlot/schemas/<source>.schema.json` is the contract, `docs/corpus.md` says what a record must state, and `examples/bring-your-own-corpus/sample_corpus.jsonl` is the field reference.
4. **Validate before loading** — `backlot import corpus.jsonl --dry-run`. Not optional: it names the record that is short a field, which is the failure that otherwise shows up much later as a thin corpus.
5. **Load into its own data dir** — `BACKLOT_DATA_DIR=/tmp/<task> backlot import corpus.jsonl`, so the task cannot overwrite a corpus the user already had.
6. **Serve, and pick the right lifetime.** A server the user keeps using across turns is `backlot serve` in the background, confirmed by polling `/health`. `backlot.mock_server()` is for a single script that starts and stops the server inside itself — its context manager dies with the process, so it cannot back a server the next turn expects to find. This is the step the skill exists to get right.
7. **Get credentials** — `GET /_mock/users` for the admin token (crawling) and per-user tokens (ACL), and `docs/auth.md` for the form the source's client wants them in.
8. **Wire a client** — which of the four paths this task calls for, and the `examples/` script for it.
9. **Check the first response against the corpus** that produced it, and show it.

## Tests

`tests/test_docs.py` already covers most of this the moment `skills/` exists: its `_markdown_files()` walks every `*.md` in the repo, so a stale path in `SKILL.md` fails the link test, and `gen_docs.py --check` guards the generated block. `tests/test_skills.py` adds only what is not already covered:

1. Frontmatter parses, the directory name matches `name`, and `description` names every source in `SERVICE_SCHEMAS`. A source missing from that line is a skill that never triggers for it, which no other test can see.
2. Every `#anchor` the skill links to resolves to a heading in its target file. `test_every_relative_link_resolves` drops the fragment (`target.split("#", 1)[0]`), so nothing today catches a renamed section — and routing to the right section is the skill's whole job. Headings in `docs/supported-sources.md` embed URL prefixes, so this is the guard that turns a moved prefix into a red build rather than a silent misroute.
3. All four manifests parse, and each resolves to the one real skill: the two `.claude-plugin/` files' `source`, `.codex-plugin/plugin.json`'s `skills`, and `.agents/plugins/marketplace.json`'s `plugins[].source.url` all land on the directory that actually holds `skills/backlot/SKILL.md`. A harness added later gets a row here rather than a copy of the skill.

## Not in scope

- An `examples/using-agent-skills/` directory. Every existing `examples/` dir is runnable client code; a skill is prose an agent reads, so there is nothing to run, and the skill's whole job is to route the agent to the examples that already exist.
- Splitting `docs/` into per-source pages. If the skill's read pattern turns out to be heavy in practice, `gen_docs.py` already owns the mapping, so the answer is a generated per-source view rather than a prose migration — and it should be decided on evidence from a skill that exists.
- A skill eval (`claude plugin eval`), which is how to measure whether the loop above holds up, and its own issue.
- A listing in either official plugin marketplace — Anthropic's, where 150 of the 286 entries are whole-repo sources pointing at a vendor's own repository, and [openai/plugins](https://github.com/openai/plugins) for Codex. That is where discovery actually happens, but each listing is a pointer at this repo and needs nothing here to change.

khj809:
Closes #90.

`skills/backlot/SKILL.md` gives an agent the loop from a request to a served corpus: check the install, pick which of the four corpora the task wants, write records against the source's schema, `--dry-run`, load into a per-task `BACKLOT_DATA_DIR`, serve with the right lifetime, fetch credentials, wire a client, check the first response against the records that produced it. Two of those steps are the reason it exists — `--dry-run` is what names a record short a field, and `backlot.serve()` is a context manager that dies with its process, so it cannot back a server a later turn expects to find.

Per-source prose stays in `docs/`. The skill is cut by task and `docs/` by source, so neither restates the other and there are no per-source skill files. The one per-source thing the skill carries is a routing table, and `scripts/gen_docs.py` generates it from the `SOURCES` mapping it already owns, so `--check` gates it alongside the docs table. `AUTH_SECTIONS` is the one new mapping, because the auth axis does not factor per source — Jira and Confluence share one section and the five Google surfaces share one, so eleven sources come to nine headings.

The repo becomes its own plugin marketplace, one manifest per harness over the same single copy of the skill:

```bash
claude plugin marketplace add brekkylab/backlot && claude plugin install backlot@brekkylab
codex plugin marketplace add brekkylab/backlot && codex plugin add backlot@brekkylab
```

`npx skills add` is not offered. It copies `skills/backlot/` alone, so all 42 of the skill's relative links land outside what it installed — measured on this branch, 42 unresolved of 42. Both plugin routes clone the repository, so the schemas, `docs/` pages and `examples/` the skill routes to are present.

## Trying it before this merges

Those commands read the default branch, which does not carry the manifest yet. Append the branch as a fragment to install from this PR — verified on both harnesses, each cloning tracked files only (4.8 MB):

```bash
claude plugin marketplace add "https://github.com/brekkylab/backlot.git#worktree-agent-skill-plugin"
claude plugin install backlot@brekkylab

codex plugin marketplace add "https://github.com/brekkylab/backlot.git#worktree-agent-skill-plugin"
codex plugin add backlot@brekkylab
```

`tests/test_skills.py` guards what nothing else can see — prose that is wrong while every path in it still resolves. A source missing from the `description` is a skill that never triggers for it. A `#anchor` is unchecked by `test_docs.py`, which drops the fragment. Backlot's own endpoints appear as bare paths, matched here by their `/_` namespace rather than by today's prefix, because a pattern that only recognises `/_meta/` stops checking at the moment the name changes. The `backlot.*` attributes the skill names are read back off the module for the same reason.

## Verified end to end, on both harnesses

Installed from a marketplace and driven by a prompt in an empty directory holding no copy of this repo, so the skill was the agent's only source of Backlot knowledge. Both announced the skill, wrote a corpus, served it, and returned a real `conversations.history` — Slack's envelope, `rich_text` blocks, and `ts` derived from the corpus clock. All 42 relative links resolve inside the installed plugin cache, on both.

Codex found a defect the Claude run could not: it saw the skill and declined it, reporting that no Backlot project files were present. The `description` never said the skill needs no project — Claude inferred it, Codex did not. With that stated, the same prompt on the same model reversed the decision and ran the loop to completion.

`1582 passed, 1 skipped` with all extras synced; `ruff check` and `ruff format --check` clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,19 @@ with backlot.serve() as s:                             # no arguments: a tiny he
 
 The only change from talking to the live service is the base URL.
 
+### Let your coding agent run Backlot instantly
+
+This repo is its own plugin marketplace, so the [agent skill](skills/backlot/SKILL.md) installs with no clone and no `pip install` first:
+
+```bash
+claude plugin marketplace add brekkylab/backlot && claude plugin install backlot@brekkylab
+codex plugin marketplace add brekkylab/backlot && codex plugin add backlot@brekkylab
+```
+
+and prompt like this:
+
+> Mock our Slack workspace with three messages in an #incidents channel, get a server running, then show me what `conversations.history` actually returns for that channel.
+
 ## What it serves
 
 Every source on one local port, each behind the path prefix its own SDK expects.
--- a/scripts/gen_docs.py
+++ b/scripts/gen_docs.py
@@ -1,8 +1,14 @@
 #!/usr/bin/env python3
-"""Regenerate the machine-maintained block of docs/supported-sources.md.
+"""Regenerate the machine-maintained blocks of docs/supported-sources.md and the agent skill.
 
-    python scripts/gen_docs.py            # rewrite the block
-    python scripts/gen_docs.py --check    # exit 1 if stale, or if SOURCES is out of step
+    python scripts/gen_docs.py            # rewrite both blocks
+    python scripts/gen_docs.py --check    # exit 1 if either is stale, or if SOURCES is out of step
+
+The skill's block is a routing table: one row per source, pointing at the schema its records are
+validated against and at the two docs sections that say what it serves and what its clients
+authenticate with. Generated for the same reason the docs table is — those links carry heading
+anchors, and a skill that sends an agent to the top of a 300-line page instead of to the answer
+fails silently.
 
 Why a mapping lives here instead of being derived: nothing in the app answers "which URL prefix
 belongs to which source_type". ``jira`` and ``confluence`` both sit under ``/atlassian``; one
@@ -30,10 +36,13 @@ import sys
 from pathlib import Path
 
 from backlot.main import app
+from backlot.openapi import SOURCE_PREFIXES
 from backlot.validation import SERVICE_SCHEMAS
 
 REPO = Path(__file__).resolve().parent.parent
 SOURCES_DOC = REPO / "docs" / "supported-sources.md"
+AUTH_DOC = REPO / "docs" / "auth.md"
+SKILL_DOC = REPO / "skills" / "backlot" / "SKILL.md"
 
 # source_type -> (display name, URL prefixes). The only place this mapping exists.
 SOURCES: dict[str, tuple[str, tuple[str, ...]]] = {
@@ -56,6 +65,25 @@ SOURCES: dict[str, tuple[str, tuple[str, ...]]] = {
 # One POST each, include_in_schema=False, so they contribute no /openapi.json paths.
 GRAPHQL_ONLY = frozenset({"linear", "fireflies"})
 
+# source_type -> the docs/auth.md heading that covers it. Not derivable from SOURCES, because the
+# auth axis does not factor per source: Jira and Confluence share one Basic-auth section and the
+# five Google surfaces share one, so eleven sources come to nine headings. validate() proves each
+# value still names a heading that is there, which is what stops a renamed section from rotting the
+# skill's links quietly.
+AUTH_SECTIONS: dict[str, str] = {
+    "confluence": "Jira and Confluence",
+    "fireflies": "Fireflies",
+    "github": "GitHub",
+    "gmail": "Gmail, Google Drive, Docs, Sheets, Slides",
+    "google_drive": "Gmail, Google Drive, Docs, Sheets, Slides",
+    "hubspot": "HubSpot",
+    "jira": "Jira and Confluence",
+    "linear": "Linear",
+    "notion": "Notion",
+    "s3": "Amazon S3",
+    "slack": "Slack",
+}
+
 _START = "<!-- generated:{name} start -->"
 _END = "<!-- generated:{name} end -->"
 
@@ -67,6 +95,50 @@ def _first_sentence(text: str) -> str:
     is…" yields the whole first sentence instead of breaking at ".ai".
     """
     return re.split(r"(?<=\.)\s", text.strip(), maxsplit=1)[0]
+
+
+def _anchor(heading: str) -> str:
+    """GitHub's heading slug: drop inline markup, then non-word characters, spaces to hyphens.
+
+    Markup comes off as PAIRED delimiters, not as characters. Stripping ``_`` character-wise would
+    also eat the underscores inside an identifier, which GitHub keeps because they are word
+    characters: `` `test_docs.py` `` has to slug to ``test_docspy``, not ``testdocspy``.
+    """
+    # Backreferenced, so a delimiter closes only with itself. Without `\1` the lazy body stops at
+    # the first delimiter CHARACTER, which for `` `test_docs.py` `` is the underscore.
+    text = re.sub(r"([`*_]{1,2})(.+?)\1", r"\2", heading).strip().lower()
+    return re.sub(r"\s", "-", re.sub(r"[^\w\s-]", "", text))
+
+
+def _slice_key(prefixes: tuple[str, ...]) -> str | None:
+    """The key `/_meta/openapi/<key>` wants for a source served under `prefixes`, or None.
+
+    That endpoint is keyed on ``openapi.SOURCE_PREFIXES``, which is the MCP bridge's namespace and
+    not ``source_type``: Jira and Confluence share ``atlassian``, ``google_drive`` answers to
+    ``gdrive``, and S3 has no entry at all. Derived by prefix overlap rather than restated, so the
+    two namespaces cannot drift apart in the skill's routing table.
+    """
+    keys = [
+        key
+        for key, bridge in SOURCE_PREFIXES.items()
+        if any(prefix.startswith(b) for prefix in prefixes for b in bridge)
+    ]
+    if len(keys) > 1:
+        raise SystemExit(f"prefixes {prefixes} match more than one MCP slice key: {sorted(keys)}")
+    return keys[0] if keys else None
+
+
+def _sections(doc: Path) -> dict[str, str]:
+    """`### Slack — \\`Bearer\\`` -> {"Slack": "slack--bearer"}, keyed on the half before the dash.
+
+    Both reference pages head a service's section with its name, an em dash, and the detail that
+    varies (its URL prefixes, its auth scheme). The name is the stable half and the only half a
+    mapping here can state; the anchor is read off the heading as it stands.
+    """
+    return {
+        heading.split("—", 1)[0].strip(): _anchor(heading)
+        for heading in re.findall(r"^###\s+(.*?)\s*$", doc.read_text(), re.M)
+    }
 
 
 def validate() -> list[str]:
@@ -101,6 +173,28 @@ def validate() -> list[str]:
                 f"{source_type!r} is GraphQL-only, so it needs exactly one /graphql prefix; "
                 f"got {list(prefixes)}"
             )
+
+    # The skill's routing table links into a heading in each reference page, so a name that heads
+    # no section there would be written as a link to the top of the page — which resolves, and
+    # answers nothing. Caught here rather than left to the reader.
+    detail = _sections(SOURCES_DOC)
+    auth = _sections(AUTH_DOC)
+    for source_type in sorted(SOURCES):
+        name = SOURCES[source_type][0]
+        if name not in detail:
+            problems.append(
+                f"{source_type!r}: {name!r} heads no section in {SOURCES_DOC.name} — the display "
+                "name in SOURCES has to match the heading there"
+            )
+        if source_type not in AUTH_SECTIONS:
+            problems.append(f"{source_type!r} has no AUTH_SECTIONS entry — add one")
+        elif AUTH_SECTIONS[source_type] not in auth:
+            problems.append(
+                f"{source_type!r}: AUTH_SECTIONS names {AUTH_SECTIONS[source_type]!r}, which heads "
+                f"no section in {AUTH_DOC.name}"
+            )
+    for source_type in sorted(set(AUTH_SECTIONS) - set(SOURCES)):
+        problems.append(f"{source_type!r} is in AUTH_SECTIONS but not in SOURCES")
     return problems
 
 
@@ -127,11 +221,37 @@ def render_sources() -> str:
     return "\n".join(rows)
 
 
-def replace_block(text: str, name: str, body: str) -> str:
+def render_skill_sources() -> str:
+    """The agent skill's routing table: schema to validate against, and where each answer lives.
+
+    Paths are relative to skills/backlot/, where SKILL.md sits — two levels below the repo root,
+    unlike the docs table's one.
+    """
+    detail, auth = _sections(SOURCES_DOC), _sections(AUTH_DOC)
+    rows = [
+        "| `source_type` | URL prefix | Record schema | What it serves | Auth | OpenAPI slice |",
+        "|---|---|---|---|---|---|",
+    ]
+    for source_type in sorted(SOURCES):
+        name, prefixes = SOURCES[source_type]
+        prefix_cell = " ".join(f"`{p}`" for p in prefixes)
+        schema = f"[`{source_type}.schema.json`](../../backlot/schemas/{source_type}.schema.json)"
+        serves = f"[{name}](../../docs/supported-sources.md#{detail[name]})"
+        section = AUTH_SECTIONS[source_type]
+        scheme = f"[{section}](../../docs/auth.md#{auth[section]})"
+        key = _slice_key(prefixes)
+        slice_cell = f"`{key}`" if key else "—"
+        rows.append(
+            f"| `{source_type}` | {prefix_cell} | {schema} | {serves} | {scheme} | {slice_cell} |"
+        )
+    return "\n".join(rows)
+
+
+def replace_block(text: str, name: str, body: str, doc: Path) -> str:
     start, end = _START.format(name=name), _END.format(name=name)
     pattern = re.compile(re.escape(start) + r".*?" + re.escape(end), re.DOTALL)
     if not pattern.search(text):
-        raise SystemExit(f"no {name!r} marker pair in {SOURCES_DOC}")
+        raise SystemExit(f"no {name!r} marker pair in {doc}")
     # A lambda, not a replacement string: the body is markdown full of backslashes and pipes that
     # re.sub would otherwise read as group references.
     return pattern.sub(lambda _: f"{start}\n{body}\n{end}", text)
@@ -139,7 +259,7 @@ def replace_block(text: str, name: str, body: str) -> str:
 
 def main() -> int:
     parser = argparse.ArgumentParser(
-        description="Regenerate docs/supported-sources.md's generated block."
+        description="Regenerate the docs' and skill's generated blocks."
     )
     parser.add_argument(
         "--check", action="store_true", help="exit 1 if stale or invalid; write nothing"
@@ -152,18 +272,32 @@ def main() -> int:
             print(f"  - {problem}", file=sys.stderr)
         return 1
 
-    current = SOURCES_DOC.read_text()
-    updated = replace_block(current, "sources", render_sources())
-    if args.check:
-        if current != updated:
-            print(
-                "docs/supported-sources.md is stale — run: python scripts/gen_docs.py",
-                file=sys.stderr,
-            )
-            return 1
-        return 0
-    SOURCES_DOC.write_text(updated)
-    print(f"wrote {SOURCES_DOC.relative_to(REPO)}")
+    blocks = (
+        (SOURCES_DOC, "sources", render_sources),
+        (SKILL_DOC, "skill-sources", render_skill_sources),
+    )
+    stale = []
+    for doc, name, render in blocks:
+        current = doc.read_text()
+        updated = replace_block(current, name, render(), doc)
+        if current == updated:
+            # Reported, not passed over in silence: SKILL.md tells its reader to run this command,
+            # and a command that prints nothing leaves them unable to tell it from a no-op.
+            if not args.check:
+                print(f"already current: {doc.relative_to(REPO)}")
+            continue
+        if args.check:
+            stale.append(doc.relative_to(REPO))
+            continue
+        doc.write_text(updated)
+        print(f"wrote {doc.relative_to(REPO)}")
+
+    if stale:
+        print(
+            f"stale, run `python scripts/gen_docs.py`: {', '.join(str(p) for p in stale)}",
+            file=sys.stderr,
+        )
+        return 1
     return 0
 
 
--- a/skills/backlot/SKILL.md
+++ b/skills/backlot/SKILL.md
@@ -1,0 +1,180 @@
+---
+name: backlot
+description: Emulate enterprise SaaS knowledge APIs — Slack, Gmail, Google Drive (with Docs, Sheets, Slides), GitHub, Jira, Confluence, Notion, Amazon S3, HubSpot, Linear, Fireflies — over a corpus you supply, then point a real client (vendor SDK, MCP server, LlamaIndex reader, fsspec, mirage) at it. Applies in ANY directory, an empty one included — it needs only the `backlot` CLI, never a project, a repository, a checkout or a config file, so the absence of Backlot files is not a reason to skip it. Use whenever asked to mock, stub, fake, emulate or reproduce one of those services — prefer this over hand-writing a stub — and when asked to write or validate a Backlot BYO-JSONL corpus, or to test a connector, RAG pipeline or ACL-scoped retrieval without live vendor accounts.
+---
+
+# Backlot
+
+Backlot answers each service's real API over documents you supply, so a connector built on the
+vendor's own SDK can be exercised end to end with no vendor account. One server process serves
+every source at once, so a request for Slack means starting that server over a corpus that has
+Slack documents in it, not starting a Slack-shaped server.
+
+**It brings its own everything.** The corpus is a file you write here and now, so this loop runs in
+an empty directory as readily as in a project. Nothing has to already exist — not a repository, not
+a checkout of Backlot, not a config file, not a seeded workspace. Writing a stub by hand instead is
+the mistake this exists to prevent: a hand-rolled stub agrees with your assumptions about the
+vendor's response, and Backlot agrees with the vendor.
+
+Work the loop below in order. Steps 4 and 6 are the two that get skipped and shouldn't be.
+
+## 1. Check the install
+
+```bash
+backlot --version            # not on PATH -> pip install backlot
+backlot status               # which corpus is loaded, and what is in it
+```
+
+`backlot status` exits non-zero when no corpus is loaded. Read what it prints before deciding
+anything: a corpus that already holds what the task needs makes steps 2-5 unnecessary.
+
+## 2. Decide which corpus this task wants
+
+| The task | The corpus |
+|---|---|
+| a demo, or "what shape does this API answer" | `backlot import --bundled` — 136 records covering every source |
+| exercising a connector against real content | the user's own documents, as BYO-JSONL |
+| reproducing one behaviour (a paginated edge case, an odd MIME type, an empty thread) | records you write for exactly that |
+| a large public corpus | `backlot import --type enterpriserag-bench` |
+
+Say which of the four this is before writing anything. A request naming one service and one
+situation is almost always the third, and wants three records rather than a dataset.
+
+## 3. Write the records
+
+One JSON object per line, `source_type` naming the service. The routing table below gives each
+source's schema — that schema is the contract, and it is what `--dry-run` validates against.
+
+- [`docs/corpus.md`](../../docs/corpus.md) — what every record must state, and the three corpora
+- [`examples/bring-your-own-corpus/sample_corpus.jsonl`](../../examples/bring-your-own-corpus/sample_corpus.jsonl) — the field reference, kept exhaustive by a test
+- [`backlot/schemas/README.md`](../../backlot/schemas/README.md) — the rules each source imposes
+
+A record must state the facts the served document cannot exist without: who wrote it, which
+container it lives in, and when it happened. Everything you leave out is derived deterministically,
+so ids are stable across calls and pages — but an author, a container and a clock are not derivable
+and the import will tell you so.
+
+## 4. Validate before loading
+
+```bash
+backlot import corpus.jsonl --dry-run
+```
+
+A plain `backlot import` validates too, and refuses the whole file rather than loading part of it,
+so a record short a required field cannot slip in either way. What `--dry-run` adds is the two
+things that decide how many rounds this takes:
+
+- **Every problem in the corpus, not the first.** A plain import stops at the first bad line;
+  `--dry-run` reports all of them, so a corpus with four mistakes takes one pass to fix.
+- **Names for the references that resolved to nothing.** These do not fail the import — it exits 0
+  and reports how many documents it loaded — and each one silently drops content from what gets
+  served. A pull declaring `changed_paths: ["a.py", "b.py"]` where only `a.py` has a `file`
+  document loads happily, and `/pulls/{n}/files` then answers with `a.py` alone. A plain import
+  counts those references; `--dry-run` prints the line and the path of each.
+
+The second is the corpus that really does go quietly thin, and it is why this step is not optional.
+
+## 5. Load into its own data dir
+
+```bash
+BACKLOT_DATA_DIR=/tmp/<task> backlot import corpus.jsonl
+```
+
+`BACKLOT_DATA_DIR` defaults to `./data` in the working directory. Set it per task, so this work
+cannot overwrite a corpus the user already had.
+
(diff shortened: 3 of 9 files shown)


README.md
```diff
@@ -39,6 +39,19 @@ with backlot.serve() as s:                             # no arguments: a tiny he
 
 The only change from talking to the live service is the base URL.
 
+### Let your coding agent run Backlot instantly
+
+This repo is its own plugin marketplace, so the [agent skill](skills/backlot/SKILL.md) installs with no clone and no `pip install` first:
+
+```bash
+claude plugin marketplace add brekkylab/backlot   # Claude Code
+npx skills add brekkylab/backlot                  # Codex, Cursor, Copilot, and the rest
```
nuri-yoo: Blocker. `npx skills add` installs `SKILL.md` and nothing else, so all 42 of the skill's relative links break on the harness family this line addresses.

`npx skills` is `vercel-labs/skills`. It is not a marketplace reader — it discovers `skills/<name>/SKILL.md` and copies that skill directory into `.agents/skills/<name>/`. Every link in `SKILL.md` is `../../`, so all of them escape what gets copied. Measured against this branch, not a local path:

```
$ npx skills add "https://github.com/brekkylab/backlot/tree/worktree-agent-skill-plugin" \
    --skill backlot --agent codex --copy -y
  -> ./.agents/skills/backlot

$ find .agents -type f
  .agents/skills/backlot/SKILL.md

relative links: 42
UNRESOLVED:     42
```

What the agent loses is the whole reference layer the skill routes to: the 11 record schemas step 3 says are the contract, `docs/corpus.md`, `docs/auth.md`, `docs/supported-sources.md`, the `sample_corpus.jsonl` described as "the field reference, kept exhaustive by a test", and all five `examples/using-*/` directories step 8 points at. The skill still triggers and still reads, so this fails quietly: a 166-line document in which every pointer 404s.

The two plugin mechanisms are fine, because both clone the repo. I confirmed the Claude Code one directly: the cache at `~/.claude/plugins/cache/brekkylab/backlot/0.1.0` holds the whole tree at this commit and all 42 links resolve inside it. So the body's "All 41 relative links resolve inside the installed plugin cache, on both" is true of the plugin paths and does not cover this line.

`npx skills` does carry a skill's own subdirectories, so bundling is a real option. Verified by adding two files under `skills/backlot/` and reinstalling:

```
.agents/skills/backlot/SKILL.md
.agents/skills/backlot/helper.py
.agents/skills/backlot/references/corpus.md
```

Three ways out, in rough order of how much they preserve: move the referenced material under `skills/backlot/` (which means duplicating or symlinking what `docs/` and `backlot/schemas/` already hold), rewrite the 42 links as absolute `https://github.com/brekkylab/backlot/blob/main/...` URLs, or drop this line and document only the two plugin mechanisms. The second keeps one copy and costs an agent a network fetch per link; the third is the smallest change and honest about what is verified.

skills/backlot/SKILL.md
```diff
@@ -0,0 +1,166 @@
...
+| LlamaIndex `Document` loading | [`examples/using-llamaindex-readers/`](../../examples/using-llamaindex-readers/) |
+| `pandas`/`pyarrow`/`dask` over a filesystem | [`examples/using-fsspec/`](../../examples/using-fsspec/) |
+| a virtual filesystem to `ls`, `cat` and `grep` | [`examples/using-mirage/`](../../examples/using-mirage/) |
+
+One runnable script per service in each. List the directory rather than guessing the filename:
+they follow the service, but Google Drive is `gdrive` and Jira/Confluence share `atlassian`, and
+not every source has a script in every directory.
+
+With a server already up, prefer the served surface over any description of it:
+`GET /_meta/openapi/<source>` is that source's typed OpenAPI slice, and Linear and Fireflies answer
```
nuri-yoo: `<source>` reads as a `source_type` here, and on that reading this 404s for four of the nine REST sources the routing table below lists. The endpoint is keyed on `openapi.SOURCE_PREFIXES`, which is a different namespace from `source_type`.

Measured against a running server on this commit:

```
github 200   gmail 200   hubspot 200   notion 200   slack 200
confluence 404   google_drive 404   jira 404   s3 404
linear 404   fireflies 404      <- these two the sentence already excuses
atlassian 200 (24 paths)   gdrive 200 (6 paths)
```

`s3` is the one with no way to spell it: it has four paths in the app's `/openapi.json` but no entry in `SOURCE_PREFIXES`, so there is no slice to ask for.

This bites. I gave an agent in an empty directory "stand up a mock Jira with two issues, then fetch the typed OpenAPI slice for that source off the running server", with this skill as its only source of Backlot knowledge. It followed the loop correctly through step 7, then reported:

> `curl localhost:8000/_meta/openapi/jira` -> **404** ... One real failure worth flagging: step 8 was a wrong guess on my part (`jira` isn't a valid OpenAPI-slice name even though it's the corpus `source_type`)

It recovered on the next call because the 404 body enumerates the valid keys, which is good server design covering for the skill. A harness that does not retry on a 404 does not get that.

The fix is four lines up. `SKILL.md:131-133` already states this exact mapping for `examples/`: "Google Drive is `gdrive` and Jira/Confluence share `atlassian`, and not every source has a script in every directory." All three facts hold here, and here they hold exactly, which is not true of the examples tree (see the other comment on that line). Naming the keys, or pointing at the 404 body as the way to learn them, would close it.

tests/test_skills.py
```diff
@@ -0,0 +1,190 @@
...
+
+# Backlot's own endpoints, as opposed to the vendor prefixes gen_docs.py already checks. Matched by
+# the whole underscore namespace rather than by one literal prefix: a pattern that recognises only
+# the prefix in use stops matching the moment that prefix is renamed, which is the moment it has to
+# fire. `<source>` is the skill's spelling of a path parameter FastAPI writes as `{source}`.
+_SEGMENT = r"(?:[\w-]+|<\w+>)"
+_OWN_ENDPOINT_RE = re.compile(rf"/_{_SEGMENT}(?:/{_SEGMENT})*|/health\b")
+
+
+def test_every_backlot_endpoint_the_skill_names_is_mounted():
```
nuri-yoo, line 184: This passes on the `/_meta/openapi/` problem, because it checks that the route is mounted and never that the values the skill implies are served.

Everything it extracts from `SKILL.md` and asserts:

```
/_meta/credentials        -> in served: True
/_meta/openapi/<source>   -> /_meta/openapi/{source}   in served: True
/_meta/users              -> in served: True
/health                   -> in served: True
```

`<source>` normalises to `{source}`, the route exists, done. Four of the nine REST `source_type` values 404 at that route right now and this stays green, which is the failure the docstring says it exists to catch.

A path-parameterised endpoint needs its values checked against whatever the server keys them on. `openapi.SOURCE_PREFIXES` is importable and the 404 body already enumerates it. Written against this commit, that fails and names them:

```
AssertionError: the skill's routing table lists source_types that /_meta/openapi/ does not
serve: ['confluence', 'fireflies', 'google_drive', 'jira', 'linear', 's3']; it accepts only
['atlassian', 'gdrive', 'github', 'gmail', 'hubspot', 'notion', 'slack']
```

Whether the assertion should be "every table row resolves" or "the skill names the slice keys explicitly" depends on which way you fix the skill, so I have not proposed a patch.

skills/backlot/SKILL.md
```diff
@@ -0,0 +1,166 @@
...
+so ids are stable across calls and pages — but an author, a container and a clock are not derivable
+and the import will tell you so.
+
+## 4. Validate before loading
+
+```bash
+backlot import corpus.jsonl --dry-run
+```
+
+Not optional. This is the step that names the record that is short a field. Skip it and the same
```
nuri-yoo: The claim in the second sentence does not happen. A record short a required field is named on the writing path too, and nothing loads:

```
$ backlot import missing_field.jsonl --dry-run
INVALID: 1 problem(s) in missing_field.jsonl
  line 2: <root>: 'author_email' is a required property        exit=1

$ backlot import missing_field.jsonl          # no --dry-run
line 2: <root>: 'author_email' is a required property          exit=1

$ backlot status
documents: 0 rows
```

`run()` validates before it writes, so there is no "loads successfully into a thinner corpus" on the missing-field axis. Which matters because an agent reads this and takes away that a plain import tolerates malformed records silently. It does not.

Two things `--dry-run` does buy, and the second is the corpus that really does go quietly thin.

A whole-corpus report instead of a first-failure abort:

```
--dry-run: 2 problems   (line 1 author_email, line 2 created)
import   : line 1 only, then stops
```

And naming each unresolved path reference rather than counting them:

```
--dry-run: NOTE: 2 path references with no matching `file` document
             line 2: src/gateway.py
             line 2 (review comment): src/nowhere.py
           OK: 2 records valid.                    exit=0
import   :   github: 2 path references ... (`--dry-run` names them)
           Loaded 2 documents                      exit=0
```

That second one produces exactly the symptom this paragraph describes. A record declaring `changed_paths: ["src/limiter.py", "src/gateway.py"]` plus one review comment imports at exit 0 reporting "Loaded 2 documents", and the server then answers:

```
GET /pulls/28314/files    -> ['src/limiter.py']     gateway.py gone
GET /pulls/28314/comments -> count: 0               the comment gone
```

So the symptom is right and the mechanism attached to it is not. Swapping in either of the two real reasons would make the paragraph true without weakening it, and the second is stronger than what is written now, since it is the case where the import reports success.

scripts/gen_docs.py
```diff
@@ -69,6 +96,25 @@ def _first_sentence(text: str) -> str:
     return re.split(r"(?<=\.)\s", text.strip(), maxsplit=1)[0]
 
 
+def _anchor(heading: str) -> str:
```
nuri-yoo, line 100: This deletes literal underscores, which GitHub keeps, and the guard in `tests/test_skills.py` cannot catch that because it is the same rule written twice.

The two regexes here disagree with each other about `_`:

```
first  [`*_]      on "a_b" -> 'ab'
second [^\w\s-]   on "a_b" -> 'a_b'      <- \w, so written to preserve _
```

The second is deliberately underscore-preserving; the first strips `_` along with the emphasis delimiters and gets there first.

`tests/test_skills.py:45` reimplements this as `_github_anchor` rather than importing it. The two are textually different and semantically identical: 0 disagreements over 200,000 random headings. So for every generated row the test compares the generator against a copy of the generator's own rule, and can only catch a heading rename, which `validate()` already catches via `_sections`. The docstring on `_github_anchor` names this failure mode ("an approximation that disagrees with GitHub in the same direction for both sides of the comparison would catch nothing") and then meets it.

Nothing is broken today. The skill anchors only into `docs/auth.md` and `docs/supported-sources.md`, and I checked every heading in both against GitHub's rule: zero divergence. But a heading the rule already gets wrong is in the repo, at `AGENTS.md:32`:

```
## Documentation rules (`tests/test_docs.py` enforces all of these)
  this rule: documentation-rules-teststestdocspy-enforces-all-of-these
  GitHub   : documentation-rules-teststest_docspy-enforces-all-of-these
```

Nothing links to it, so it is latent rather than live. What it costs when it does fire, demonstrated by renaming one auth heading to carry a code identifier:

```
### Amazon S3 — `SigV4` (`s3_access_key_id`)

gen_docs wrote:  docs/auth.md#amazon-s3--sigv4-s3accesskeyid
GitHub's anchor: #amazon-s3--sigv4-s3_access_key_id
gen_docs.py --check      -> exit 0
test_skills + test_docs  -> 26 passed
```

A link to the top of `docs/auth.md` instead of to the section, shipped with both gates green.

Two independent fixes. Strip emphasis as paired delimiters rather than as characters, or simply stop stripping the underscore, since the second regex already handles it:

```python
text = re.sub(r"[`*_]{1,2}(.+?)[`*_]{1,2}", r"\1", heading).strip().lower()
```

And have the test import this function instead of copying it, so the duplicate cannot drift and nobody reads the guard as stronger than it is.

scripts/gen_docs.py
```diff
@@ -152,18 +242,28 @@ def main() -> int:
...
-    print(f"wrote {SOURCES_DOC.relative_to(REPO)}")
+    blocks = (
+        (SOURCES_DOC, "sources", render_sources),
+        (SKILL_DOC, "skill-sources", render_skill_sources),
+    )
+    stale = []
+    for doc, name, render in blocks:
+        current = doc.read_text()
+        updated = replace_block(current, name, render(), doc)
+        if current == updated:
```
nuri-yoo, line 283: Writing only on change means the command prints nothing at all when the tree is current, where before it always confirmed.

```
this commit:   $ python scripts/gen_docs.py   ->  (no output)  exit 0
6a2d16f:       $ python scripts/gen_docs.py   ->  wrote docs/supported-sources.md
```

`SKILL.md` closes with "Regenerate with `python scripts/gen_docs.py`", so the reader most likely to run it is the one who gets no evidence it did anything. An `else: print("already current")`, or printing every file checked rather than only the ones rewritten, restores that.

skills/backlot/SKILL.md
```diff
@@ -0,0 +1,166 @@
...
+| Want | Where |
+|---|---|
+| the vendor's own SDK | [`examples/using-official-sdk/`](../../examples/using-official-sdk/) |
+| MCP tools for an agent | [`examples/using-mcp-with-agents/`](../../examples/using-mcp-with-agents/) |
+| LlamaIndex `Document` loading | [`examples/using-llamaindex-readers/`](../../examples/using-llamaindex-readers/) |
+| `pandas`/`pyarrow`/`dask` over a filesystem | [`examples/using-fsspec/`](../../examples/using-fsspec/) |
+| a virtual filesystem to `ls`, `cat` and `grep` | [`examples/using-mirage/`](../../examples/using-mirage/) |
+
+One runnable script per service in each. List the directory rather than guessing the filename:
+they follow the service, but Google Drive is `gdrive` and Jira/Confluence share `atlassian`, and
```
nuri-yoo: `gdrive` holds everywhere; the `atlassian` half holds in one of the five directories.

```
using-fsspec              atlassian=0  jira=0  confluence=0  gdrive=1
using-llamaindex-readers  atlassian=0  jira=1  confluence=1  gdrive=1
using-mcp-with-agents     atlassian=1  jira=0  confluence=0  gdrive=1
using-mirage              atlassian=0  jira=0  confluence=0  gdrive=1
using-official-sdk        atlassian=0  jira=1  confluence=1  gdrive=1
```

Only `using-mcp-with-agents` shares one script. Two directories keep `jira.py` and `confluence.py` separate, and `using-official-sdk` is the one step 8 sends an agent to for "the vendor's own SDK", so it is the likely destination and it splits them.

Bounded harm, since the instruction is to list the directory rather than guess, and that instruction is right. But the caveat as stated teaches a rule that is wrong four times out of five, and the same caveat is exactly correct for `/_meta/openapi/` four lines down, where it is missing (see the other comment).

Also `One runnable script per service in each`: `examples/using-official-sdk/linear` is a directory, a Node project, not a script.

tests/test_skills.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+# manifest -> (how to read its path out, what that path is a path TO). The two marketplaces name a
+# plugin root; Codex's plugin.json names the skills directory itself. Both are relative to the
+# repository root, not to the manifest's own directory — the shape obra/superpowers and
+# hashicorp/agent-skills both ship.
+_MANIFESTS = {
+    ".claude-plugin/marketplace.json": (_claude_marketplace, "skills/backlot/SKILL.md"),
+    ".codex-plugin/plugin.json": (lambda data: data["skills"], "backlot/SKILL.md"),
+    ".agents/plugins/marketplace.json": (
```
nuri-yoo, line 129: Minor asymmetry: this row reads `plugins[0]` positionally while the Claude row above filters by `name == "backlot"`. Add a second plugin to the `.agents` catalog and this silently checks the wrong one. Reusing a by-name lookup for both would make the two rows say the same thing.

The assertion is not weak in general: it would catch a wrong path like `"./plugin/"`. It is only insensitive to which plugin it read, and only in this row.

nuri-yoo (commented):
The loop works. I installed the plugin from a marketplace and drove it with a prompt in an empty directory holding no copy of this repo, twice, on Claude Code. The skill triggered both times and ran the loop to completion: it wrote a BYO corpus, validated it with `--dry-run`, loaded it into a per-task `BACKLOT_DATA_DIR`, served it, fetched credentials, resolved the channel through `conversations.list` and checked `conversations.history` against the records that produced it. `1576 passed, 1 skipped` locally with every extra synced plus the HubSpot reader past its pin (the skip is the Docker-gated Atlassian MCP test; CI has Docker and reports `1577 passed`, 0 skipped, on all four Pythons). `ruff check` and `ruff format --check` clean.

One blocker and three substantive issues, all in prose or in a test rather than in code. Details are inline.

The blocker is `README.md:48`. `npx skills add` is the command this README hands to everyone who is not on Claude Code, and on that path the skill ships with every pointer dead. Inline comment has the measurement.

Two smaller notes that do not anchor to a line.

The body says "All 41 relative links resolve inside the installed plugin cache". I cannot get 41 by any counting rule: 42 link occurrences, 40 distinct targets, 21 distinct files, 22 occurrences carrying an `#anchor`, 0 external links. All 42 do resolve, both in-tree and in the Claude Code plugin cache, so only the number is off.

A directory-source install copies untracked files verbatim. `claude plugin marketplace add <local path>`, which is what a contributor testing locally runs, brought my install to 498 MB because it pulled in `.venv`, `.pytest_cache` and `backlot.egg-info`. The advertised git-URL path is unaffected: the tracked tree is 4.6 MB, which matches the 4.8 MB in the body. Worth a line in CONTRIBUTING if local testing is expected, not a change here.

khj809:
Reproduced all eight before touching anything; all eight hold. Fixed in `12a96d7`.

**`README.md:48` — `npx skills`.** Line dropped. Confirmed your measurement on this branch: it installs `SKILL.md` alone, 42 unresolved of 42. The absolute-URL option would have taken all 42 links out of `test_docs.py`'s link check, which only sees relative ones, so the README now offers the two plugin routes and says why the third is absent. `codex plugin marketplace add brekkylab/backlot` accepts the shorthand — checked separately, it clones and fails only on main's missing manifest.

**`SKILL.md:136` and `tests/test_skills.py:165` — the slice keys.** Fixed together, and generated rather than written: the routing table gained an **OpenAPI slice** column, derived from `openapi.SOURCE_PREFIXES` by prefix overlap, so the two namespaces cannot drift. `—` where no slice exists. Step 8 now says `<key>` comes from that column and not from `source_type`, and that the 404 body lists what it accepts. The test compares the column against `SOURCE_PREFIXES` as a set — row-wise would mean reimplementing the derivation, which could only agree with it. Worth flagging: my first version of that test passed its own mutation for exactly the reason you gave about `_github_anchor`; the set comparison catches it.

**`SKILL.md:63` — `--dry-run`.** Rewritten to your two reasons. Measured both: plain import reports the first bad line and loads 0 rows, so nothing slips in that way; the unresolved-reference case imports at exit 0 with "Loaded 2 documents", and `/pulls/28314/files` then serves `src/limiter.py` alone with `changed_files: 1`, review comment gone.

**`scripts/gen_docs.py:99` — the anchor rule.** Now strips markup as backreferenced pairs (`([`*_]{1,2})(.+?)\1`), so a delimiter closes only with itself. A first attempt without the backreference still ate the underscore, since the lazy body stops at the first delimiter character. `tests/test_skills.py` imports `_anchor` instead of copying it, and five headings are asserted against GitHub's rule directly — including your `AGENTS.md:32` case and the `s3_access_key_id` one.

**`scripts/gen_docs.py:253` — silence.** Prints `already current: <path>` per file.

**`SKILL.md:132` — the examples caveat.** Corrected: `gdrive` throughout, but only `using-mcp-with-agents` folds Jira and Confluence into one script, and `using-official-sdk/linear` is a Node project rather than a script.

**`tests/test_skills.py:120` — positional lookup.** Both catalogs read by name now, via one `_by_name()`.

**The two notes.** Body says 42, not 41. The directory-source install copying untracked files is left alone as you suggested — worth a CONTRIBUTING line, not this PR.

`1582 passed, 1 skipped` with all extras synced; `ruff check` and `ruff format --check` clean; `gen_docs.py --check` exit 0.

nuri-yoo (approved):
All eight are closed. Reproduced each fix against `12a96d7` with all extras synced, rather than reading the commit.

| What | Verified |
|---|---|
| `README.md:48` | `npx skills` line gone; the two remaining commands both clone the repo, so the 42 links resolve |
| `SKILL.md:136` + `tests/test_skills.py:165` | every key in the new **OpenAPI slice** column returns 200 against a live server (`atlassian` 24 paths, `gdrive` 6, `slack` 12), every `—` corresponds to a `source_type` that 404s, and "six of the eleven" measures 6 of 11 exactly |
| `SKILL.md:63` | both replacement reasons hold: a corpus with four mistakes gives `INVALID: 4 problem(s)` under `--dry-run` against one line from a plain import, and the unresolved-reference case imports at exit 0 then serves `changed_files: 1` with `/files` answering `a.py` alone |
| `scripts/gen_docs.py:99` | the five asserted cases pass, and reverting `_anchor` to the character-wise strip fails two of them, so the guard is no longer checking a copy of itself |
| `scripts/gen_docs.py:253` | `already current: docs/supported-sources.md` / `already current: skills/backlot/SKILL.md` |
| `SKILL.md:132` | checked against all five directories: `gdrive` in 5 of 5, `atlassian` only in `using-mcp-with-agents`, `jira.py` and `confluence.py` separate in the two that have them, `linear/package.json` present |
| `tests/test_skills.py:120` | both catalogs read by name |

`1582 passed, 1 skipped` here too, the skip being the Docker-gated Atlassian test. `ruff check` and `ruff format --check` clean, `gen_docs.py --check` exit 0.

Two things I checked beyond the diff. Renaming the route to `/_meta/spec/{source}`, and `/health` to `/healthz`, each still fails `test_every_backlot_endpoint_the_skill_names_is_mounted`, so collapsing parameter names in `_path_shape` did not loosen what it was there to catch. And an agent installed from a marketplace in an empty directory ran the loop to completion on this head, including going straight to `/_meta/openapi/atlassian` for a Confluence corpus.

One correction I owe you. The regex I put in that review comment was broken, and in the same way as the code it was meant to replace:

```
heading:  Documentation rules (`tests/test_docs.py` enforces all of these)
  mine:   documentation-rules-teststestdocspy-enforces-all-of-these
  yours:  documentation-rules-teststest_docspy-enforces-all-of-these
```

Without the backreference the lazy body stops at the first delimiter character, which is the underscore, so it ate the same underscore and left a stray backtick behind. You caught that and said so in the reply; I should have run the patch I was proposing before proposing it. What you landed is right:

```python
re.sub(r"([`*_]{1,2})(.+?)\1", r"\2", heading)
```

Injecting my version into your new cases fails them, which is the guard doing its job.

khj809:
Thanks — and no need to owe me the regex; the guard exists because the rule is easy to get wrong twice, which is what we both did.

Landed the one item from your first review that suggested a change and had not had one: `CONTRIBUTING.md` now has a **Testing the agent skill** section under "Running tests and lint" — install a branch by git URL with the ref as a fragment, why not the checkout's path (a directory source is copied verbatim, `.venv` and all), and a reminder that uninstalling leaves the cache directory behind. `893da12`.

Everything else in the approval reads as confirmation rather than an open thread, so nothing else changed.

PR #95, opened 2026-08-31
GitHub serves a repo's branch and tag listings

### LINKED ISSUE #92: github: no /branches or /tags listing, so a client cannot enumerate a repo's refs

khj809:
Closes #92.

## What changed

`GET /repos/{owner}/{repo}/branches` answers the one branch the corpus snapshots, and honours `?protected=` — that branch is unprotected, so a request for the protected ones gets `[]`. `GET /repos/{owner}/{repo}/tags` answers `[]`. Both are ACL-scoped like every other repo route, and a repository object now carries `tags_url`, since the rule there is a template iff the resource.

fsspec's `GithubFileSystem` builds `.branches`, `.tags` and `.refs` on exactly these two routes and `backlot.integrations.fsspec` already redirected them, so all three reached a 404 until now; `examples/using-fsspec/github.py` opens on `fs.refs`, which is where a client that was handed a repo rather than a sha starts.

Measured against api.github.com on 2026-09-01:

- A listed branch is the SHORT branch — `GET /repos/octocat/Hello-World/branches` gives each item a `commit` of `{sha, url}` and nothing more, where `/branches/{branch}` nests the whole commit under the same key. Serving the longer object here would emit a field real never sends on this route.
- `/tags` on a repo with no tags is `200 []`, not a 404 — octocat/Hello-World answers exactly that, so this is a shape a client meets in production.
- `?protected=` has three answers: only the protected branches for a true value, only the unprotected ones for `false`, and all of them when the parameter is omitted or empty. The value parses the way `?recursive=` does, every non-empty value but `false`/`0` reading true — on fastapi/fastapi (22 branches, one protected) `true`/`1`/`TRUE`/`yes`/`banana` answer 1, `false`/`0` answer 21, empty and omitted answer 22. The single branch here is unprotected, which collapses the last two answers into the same list, so the router's existing `_truthy` covers every value; a `bool` param would instead 422 the ones real accepts.

Not closed here, and filed as #96: the listing names only `main`, while a pull in the same repo advertises a same-repo `head.ref` that real GitHub would list.

## Verification

- **Tests** — `pytest` on a `uv sync --all-extras` install (`dev`, `examples`, `mcp`, `llamaindex`, `mirage`, `fsspec`), at `a4688e4`: 1584 passed, 1 skipped, 0 failed. The skip is `llama-index-readers-hubspot`, which `pyproject.toml` documents as deliberately absent.
- **Optional surfaces that ran rather than skipped** — `fsspec` (the GitHub-over-fsspec tests in `tests/test_integrations.py`, which cover this change), plus the `examples`, `mcp` (Docker and `uvx` both present), `llamaindex` and `mirage` suites; the one skip above is the only test that sat out.
- **Lint** — `ruff check . && ruff format --check .`: clean, 127 files. `scripts/gen_docs.py --check` exits 0.

- [x] A test fails without this change — `/branches` and `/tags` 404 before it, watched failing; the fsspec test was re-run against the pre-change router to confirm the same.
- [x] A new endpoint serving corpus content is ACL-scoped, proved for both an admin and a scoped token — both routes join the visibility loop in `test_github_user_repos`, which asserts 404 for a token that cannot see `vault` and 200 for the admin.

### DIFF
--- a/backlot/routers/github.py
+++ b/backlot/routers/github.py
@@ -1235,10 +1235,15 @@ async def list_branches(
     single-branch route nests the whole commit under the same key. Serving the longer object from
     both would hand a client a field real GitHub never sends here.
 
-    `?protected=` selects, so it is honoured rather than ignored: nothing here is protected, and a
-    client that asked for the protected ones and got an unprotected branch back would read this
-    branch as push-guarded. Real parses the value the way `?recursive=` is parsed (measured:
-    `true`/`1`/`TRUE`/`yes`/`banana` filter, `false`/`0`/empty do not), which is :func:`_truthy`.
+    `?protected=` selects, so it is honoured rather than ignored: a client that asked for the
+    protected branches and got an unprotected one back would read this branch as push-guarded.
+    Real has three answers — only protected branches for a true value, only unprotected ones for
+    `false`, and all of them when the parameter is omitted — and parses the value the way
+    `?recursive=` is parsed, every non-empty value but `false`/`0` reading true. Measured on
+    fastapi/fastapi (22 branches, one of them protected): `true`/`1`/`TRUE`/`yes`/`banana` answer
+    1, `false`/`0` answer 21, an empty value and an omitted one answer 22. The single branch here
+    is unprotected, which collapses real's last two answers into the same list, so :func:`_truthy`
+    covers every value.
     """
     conn = auth.conn(request)
     caller = _require(request)
--- a/backlot/routers/github.py
+++ b/backlot/routers/github.py
@@ -1224,7 +1224,9 @@ async def get_blob(owner: str, repo: str, sha: str, request: Request):
 
 
 @router.get("/repos/{owner}/{repo}/branches")
-async def list_branches(owner: str, repo: str, request: Request):
+async def list_branches(
+    owner: str, repo: str, request: Request, protected: str | None = Query(None)
+):
     """The repo's branches — one, the branch the corpus snapshots and the `default_branch` a repo
     object reports.
 
@@ -1232,11 +1234,18 @@ async def list_branches(owner: str, repo: str, request: Request):
     api.github.com an item's `commit` carries `sha` and `url` and stops there, where the
     single-branch route nests the whole commit under the same key. Serving the longer object from
     both would hand a client a field real GitHub never sends here.
+
+    `?protected=` selects, so it is honoured rather than ignored: nothing here is protected, and a
+    client that asked for the protected ones and got an unprotected branch back would read this
+    branch as push-guarded. Real parses the value the way `?recursive=` is parsed (measured:
+    `true`/`1`/`TRUE`/`yes`/`banana` filter, `false`/`0`/empty do not), which is :func:`_truthy`.
     """
     conn = auth.conn(request)
     caller = _require(request)
     ids = auth.visible_ids(request, caller)
     _require_repo(conn, repo, ids)
+    if _truthy(protected):
+        return []
     sha = _repo_commit_sha(repo)
     return [
         {
--- a/backlot/routers/github.py
+++ b/backlot/routers/github.py
@@ -1223,6 +1223,47 @@ async def get_blob(owner: str, repo: str, sha: str, request: Request):
     }
 
 
+@router.get("/repos/{owner}/{repo}/branches")
+async def list_branches(owner: str, repo: str, request: Request):
+    """The repo's branches — one, the branch the corpus snapshots and the `default_branch` a repo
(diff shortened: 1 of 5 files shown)


nuri-yoo (approved):
Approved.

Measured at 8e6e1fb: `pytest` gives 1583 passed, 2 skipped (Docker absent here, plus the documented `llama-index-readers-hubspot`); `ruff check` and `ruff format --check` are clean; `scripts/gen_docs.py --check` exits 0. Reverting only `backlot/routers/github.py` to main drops 6 tests, so the "fails without this change" claim holds. The example runs, `.refs` issues exactly two requests with nothing addressed at github.com, and the `commit.url` the listing emits resolves to the same sha `/branches/main` reports.

Two things in the prose, neither of which changes the route.

**`?protected=`.** The docstring and the test comment both say `false`/`0` leave the listing alone. Real filters on them: "Setting to true returns only branches protected by branch protections or rulesets. When set to false, only unprotected branches are returned. Omitting this parameter returns all branches." The sentence reads like it was measured on a repo with no protected branches, where the two rules cannot be told apart, and the test comment says as much ("none, there or here"). On octocat/Hello-World `?protected=false` answers all three branches; on a repo whose `main` is protected it answers the other two and omits `main`, while `?protected=` and an omitted parameter still answer all three. The route needs nothing: with one unprotected branch real's three cases collapse into the two `_truthy` gives, and all four values here answer what api.github.com answers. It is the attribution that is off, in `list_branches` and in `test_github_lists_the_refs_a_client_enumerates_before_it_reads`.

**The Verification block is behind this head.** It reports 1 failed and reads `test_mcp_s3_lists_objects` as upstream. That diagnosis still reproduces on its own (`uvx awslabs.aws-api-mcp-server@latest --help` dies on the same `McpError` import), but #98 landed `--with mcp<2` on main and the merge at 8e6e1fb picked it up, so the test passes at this head and the run is 1583 passed, 2 skipped, 0 failed.

PR #97, opened 2026-08-31
A corpus states a Slack channel name real Slack could hold

### LINKED ISSUE #105: Every schema pattern accepts a value with a trailing newline
**What you ran** — a one-record BYO corpus whose Slack `channel` is a legal name with a newline stuck on the end, then the same record served.

```python
import json, pathlib, backlot
from backlot.validation import record_errors

record = {
    "source_type": "slack",
    "channel": "incidents\n",          # legal name + a trailing newline
    "content": "Ava Ng: 502s from the gateway?",
    "created": "2026-02-01T09:00:00Z",
}
print(record_errors(record) or "accepted")

pathlib.Path("corpus.jsonl").write_text(json.dumps(record) + "\n")
# $ backlot import corpus.jsonl --dry-run

with backlot.serve([record]) as m:
    ...  # GET /slack/api/conversations.list
```

**What happened** — it validates, it imports, and it is served.

```
record_errors            -> accepted
backlot import --dry-run -> OK: 1 records valid.        (exit 0)
conversations.list       -> name: 'incidents\n'  name_normalized: 'incidents\n'
```

**What you expected instead** — the refusal `slack.schema.json` already writes down. Its `channel` pattern is `^(?:[a-z0-9_-]|[^\u0000-\u007f])+$`, and a newline is neither branch, so the value should never have reached the importer — the same refusal `"eng platform"` and `"#incidents"` get.

The cause is not the pattern. `jsonschema` applies `pattern` as `re.search(patrn, instance)`, and Python's `$` matches before a final newline, so an anchored pattern stops being anchored at exactly one character. `"incidents\nX"` is refused, which is what narrows it to the trailing newline rather than to `\n` in general. The same pattern in an ECMA-262 engine — the dialect JSON Schema defines `pattern` against — refuses `"incidents\n"`.

**This is every pattern in `backlot/schemas/`, not Slack's.** All six are anchored `^…$` and all six accept a valid value with a newline appended:

| schema | field |
|---|---|
| `github.schema.json` | `repo` |
| `hubspot.schema.json` | `record_id` |
| `jira.schema.json` | `key` |
| `linear.schema.json` | `identifier` |
| `s3.schema.json` | `bucket` |
| `slack.schema.json` | `channel` |

So the fix belongs in `backlot/validation.py` rather than in six patterns — either matching with `re.fullmatch`, or an end anchor ECMA honours too. Doing it in one pattern would leave the other five inconsistent, and rewriting all six would make them Python-specific: the schemas ship in the package and are the documented contract for a BYO corpus, so they should stay readable by a non-Python validator.

Filed as a bug rather than a fidelity gap because the divergence is with Backlot's own spec: the schema states the rule correctly and the validator enforces a weaker one. What it costs is a fidelity consequence, though — `channel` is served verbatim as a conversation's `name` and `name_normalized`, so a corpus can state a name real Slack could not hold and a connector that round-trips it back into a call fails against Slack while passing here. That is the guarantee #97 added the pattern for.

Surfaced while reviewing #97.

khj809:
Closes #105.

## What changed

`slack.schema.json` holds `channel` to Slack's own charset -- `^(?:[a-z0-9_-]|[^\\u0000-\\u007f])+$`, 80 characters -- so `backlot import --dry-run` names the record instead of loading it: `sl-1 [channel]: '#incidents' does not match ...`.

The value is served verbatim as a conversation's `name` and `name_normalized`, so before this a corpus written by hand (or by a model, which reaches for `#incidents` because that is the spelling humans use) produced a `conversations.list` entry the real API cannot return, with `topic.value` doubling the hash to `##incidents`. A connector round-tripping that name back into a call failed against Slack while passing here.

**The anchor holds at both ends now (#105).** `pattern` is defined against ECMA-262, whose `$` matches only at the end of the input; Python's also matches just before a final newline, so `re.search` left all six anchored patterns in `backlot/schemas/` accepting a valid value with a newline appended. The end anchor is corrected in `backlot/validation.py` rather than in the patterns: rewriting them would make the shipped contract Python-specific, and matching with `fullmatch` would trade this divergence for a second one, since the spec's `pattern` is an unanchored partial match. A `$` that is escaped or inside a character class is left alone, so a literal dollar sign is still a value a corpus can carry.

The ERB conversion reduces the bench's channel to the same charset, because a direct `--type enterpriserag-bench` import and an exported artifact both validate through this schema. One bench thread spells its channel `slack/sports` -- the path from the sources root; its directory and the fifteen threads beside it spell it `sports`.

The sibling container fields were checked in the same pass and needed nothing: GitHub `repo` and S3 `bucket` already carry vendor charset patterns, and Jira `project` / Confluence `space` hold a project or space *name*, whose key is either the already-pinned `key` field or synthesized by `synth.jira_project_key` / `synth.confluence_space_key`, which split on `[^a-z0-9]+` and so cannot emit an invalid key.

`conversations.create` and `conversations.rename` both state the ASCII half: "Channel names may only contain lowercase letters, numbers, hyphens, and underscores, and must be 80 characters or less", refused as `invalid_name_specials` ("unallowed special characters or upper case characters") or `invalid_name_maxlength`.

That sentence is not the whole rule, and enforcing it as written would be the same bug pointing the other way. Measured 2026-08-31 against a live workspace: `conversations.list` answers with `일반`, `랜덤` and `식사`, so Slack holds non-Latin names that no vendor page mentions. Every character above U+007F is therefore accepted rather than refused for a rule Slack does not enforce. Whether Slack folds non-ASCII uppercase is not measured -- the token is read-only, and settling it would mean creating channels in a real workspace -- so it is left accepted, the direction that cannot reject a name the live API serves.

`name_normalized` is settled by the same response: for all eight channels, non-Latin included, it comes back identical to `name`, so `routers/slack.py` serving one string for both is right and now says what was measured.

Re-measured across the bench's 285,605 Slack documents: exactly one record moves (`slack/sports` -> `sports`) and every resulting channel satisfies the pattern.

## Verification

- **Tests** -- `pytest -rs` on a `.[dev,examples,mcp,llamaindex,mirage,fsspec]` install plus `llama-index-readers-hubspot<0.6`: `1617 passed in 24.41s`, 0 skipped
- **Optional surfaces that ran rather than skipped** -- all of them: `tests/test_s3.py`, `tests/test_sdk.py`, `tests/test_mcp.py`, `tests/test_llamaindex.py`, `tests/test_skills.py`, and the reader, Google, fsspec and mirage tests in `tests/test_integrations.py`
- **Lint** -- `ruff check . && ruff format --check .`: `All checks passed!` / `127 files already formatted`
- **Generated docs** -- `scripts/gen_docs.py --check`: exit 0

- [x] A test fails without this change -- reverting the pattern fails 6 parametrised cases of `test_a_slack_channel_is_named_the_way_slack_names_one`; reverting the validator's end anchor fails that test's `incidents\n` case and all six of `test_every_anchored_pattern_refuses_a_trailing_newline`; reverting the ERB reduction fails `test_byo_slack_channel_is_a_name_slack_could_hold` on `'slack/sports' == 'sports'`
- [ ] A new endpoint serving corpus content is ACL-scoped, proved for both an admin and a scoped token -- n/a, no new endpoint

### DIFF
--- a/backlot/importer/erb.py
+++ b/backlot/importer/erb.py
@@ -2261,8 +2261,31 @@ def _byo_gmail(dsid, raw, P):
     return [rec], {"owner": owner_email, "people": people, "group": None, "confidentiality": None}
 
 
+# The characters a Slack channel name may not hold, which is `slack.schema.json`'s `channel`
+# pattern read the other way round: uppercase, whitespace and ASCII punctuation. Everything above
+# U+007F stays -- real Slack serves non-Latin names, so reducing one to hyphens would destroy a
+# name it could hold. The schema is the contract; `test_byo_slack_channel_is_a_name_slack_could_hold`
+# validates this function's output against it, so the two cannot drift apart.
+_SLACK_NAME_ILLEGAL = re.compile(r"[^a-z0-9_\u0080-\U0010ffff-]+")
+
+
+def _slack_channel(raw) -> str:
+    """The bench's channel as a name Slack could hold.
+
+    A value the schema would refuse stops the whole import rather than serving a conversation
+    object the real API cannot return, so the conversion states a legal name instead.
+
+    One thread in the bench spells its channel as the path from the sources root, `slack/sports`,
+    rather than as the channel; the directory it sits in and the fifteen threads beside it both
+    spell it `sports`, so the trailing segment is the name. Everything left over is reduced to the
+    charset instead of failing a million-document import over one record.
+    """
+    name = str(raw.get("channel") or "").rsplit("/", 1)[-1].strip().lower()
+    return _SLACK_NAME_ILLEGAL.sub("-", name).strip("-")[:80] or "general"
+
+
 def _byo_slack(dsid, raw, P):
-    channel = raw.get("channel") or "general"
+    channel = _slack_channel(raw)
     _title, content = _title_content(raw)
     participants = _names(raw.get("participants"))
     turns = parse_slack_transcript(content, participants)
--- a/backlot/importer/erb.py
+++ b/backlot/importer/erb.py
@@ -2281,7 +2281,7 @@ def _slack_channel(raw) -> str:
     charset instead of failing a million-document import over one record.
     """
     name = str(raw.get("channel") or "").rsplit("/", 1)[-1].strip().lower()
-    return _SLACK_NAME_ILLEGAL.sub("-", name).strip("-")[:80] or "general"
+    return _SLACK_NAME_ILLEGAL.sub("-", name)[:80].strip("-") or "general"
 
 
 def _byo_slack(dsid, raw, P):
--- a/backlot/schemas/slack.schema.json
+++ b/backlot/schemas/slack.schema.json
@@ -19,8 +19,9 @@
     },
     "channel": {
       "type": "string",
-      "minLength": 1,
-      "description": "The channel this document belongs to."
+      "pattern": "^(?:[a-z0-9_-]|[^\\u0000-\\u007f])+$",
+      "maxLength": 80,
+      "description": "The channel this document belongs to, named the way Slack names one. The ASCII half of the rule is documented -- `conversations.create` and `conversations.rename` both say \"Channel names may only contain lowercase letters, numbers, hyphens, and underscores, and must be 80 characters or less\", refusing the rest as `invalid_name_specials` (\"unallowed special characters or upper case characters\") or `invalid_name_maxlength`. The other half is not: Slack also holds non-Latin names, which no vendor page states and a live workspace demonstrates -- `conversations.list` there answers with `일반`, `랜덤` and `식사` -- so every character above U+007F is accepted rather than refused for a rule Slack does not enforce. What is refused is what Slack refuses: uppercase, whitespace and ASCII punctuation. The value is served verbatim as a conversation's `name` and `name_normalized`, so a name Slack would refuse is an object the real API cannot return, and a connector that feeds the name back into a call fails against Slack while passing here. The leading `#` is the one to watch: it is display sugar a client puts in front, never part of the name, and stored it also doubles into a `##incidents` topic."
     },
     "group": {
       "type": [
--- a/backlot/routers/slack.py
+++ b/backlot/routers/slack.py
@@ -208,6 +208,12 @@ def _channel_core(request: Request, conn, name: str, caller: Caller) -> dict:
     return {
         "id": synth.slack_channel_id(name),
         "name": name,
+        # The same string as `name`, measured rather than assumed: every channel a live
+        # workspace's conversations.list answers with carries a `name_normalized` identical to
+        # its `name`, and that holds for the non-Latin ones too, which is where a normalization
+        # would show if there were one. Slack documents nothing about what it does -- the
+        # conversation object shows only an ASCII example, and the vendor's OpenAPI carries no
+        # description for the field -- so the response is the only source there is.
         "name_normalized": name,
         "is_channel": True,
         "is_group": False,
--- a/backlot/validation.py
+++ b/backlot/validation.py
@@ -13,7 +13,11 @@ import json
 from functools import lru_cache
 from pathlib import Path
 
+import re
+
 from jsonschema import Draft202012Validator, FormatChecker
+from jsonschema import validators as _js_validators
+from jsonschema.exceptions import ValidationError
 
 # __file__-relative, not cwd-relative: these are resources the package SHIPS (see the
 # [tool.setuptools.package-data] entry in pyproject.toml), unlike backlot.config's data_dir,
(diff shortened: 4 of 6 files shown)


backlot/schemas/slack.schema.json
```diff
@@ -19,8 +19,9 @@
     },
     "channel": {
       "type": "string",
-      "minLength": 1,
-      "description": "The channel this document belongs to."
+      "pattern": "^(?:[a-z0-9_-]|[^\\u0000-\\u007f])+$",
```
nuri-yoo, line 22: Not this PR's doing, and not a blocker — but this is the check it adds, so it is the place to name it: a trailing newline gets through. `record_errors` accepts `channel: "incidents\n"` while refusing `"incidents\nX"`, so it is the newline specifically.

The mechanism is Python, not the pattern. `jsonschema`'s `pattern` keyword is `re.search(patrn, instance)`, and Python's `$` matches before a final newline. The same pattern in an ECMA-262 engine — the dialect JSON Schema's `pattern` is defined against — refuses it:

```
Python   "incidents\n" -> PASS
ECMA     "incidents\n" -> refused
```

It is repo-wide rather than local to Slack: `github`'s `repo` accepts `"acme\n"` the same way, and every pattern in `backlot/schemas/` is anchored `^…$`. So the fix belongs in the validator (a `fullmatch`, or an end anchor ECMA also honours) rather than in this one pattern, and doing it here would leave the other five inconsistent. Filing it is probably the right move; I am happy to.

backlot/importer/erb.py
```diff
@@ -2261,8 +2261,31 @@ def sender(msg: dict) -> str | None:
...
+    A value the schema would refuse stops the whole import rather than serving a conversation
+    object the real API cannot return, so the conversion states a legal name instead.
+
+    One thread in the bench spells its channel as the path from the sources root, `slack/sports`,
+    rather than as the channel; the directory it sits in and the fifteen threads beside it both
+    spell it `sports`, so the trailing segment is the name. Everything left over is reduced to the
+    charset instead of failing a million-document import over one record.
+    """
+    name = str(raw.get("channel") or "").rsplit("/", 1)[-1].strip().lower()
+    return _SLACK_NAME_ILLEGAL.sub("-", name).strip("-")[:80] or "general"
```
nuri-yoo: Nit. `strip("-")` runs before `[:80]`, so a name cut at 80 can end on a separator that the strip already passed over: `"a"*79 + " " + "b"*20` comes out as 80 characters ending `aaaa-`.

Schema-legal, so nothing breaks — the pattern allows a hyphen anywhere — which is why it is a nit rather than a finding. Swapping the order to `[:80].strip("-")` removes it (same input gives 79 characters ending `aaaaa`), and the degenerate case is unchanged: an input that is all separators still reduces to `""` either way and falls through to `"general"`.

nuri-yoo (approved):
Approved. Ran it at `8260744`: `pytest` on a `uv sync --all-extras` install is 1600 passed, 2 skipped (Docker and the documented `llama-index-readers-hubspot`, both environmental), `ruff check`/`ruff format --check` clean, `scripts/gen_docs.py --check` exit 0. Both "fails without this change" claims hold, and the pattern one is understated: reverting it fails 6 of the parametrised cases, not 5.

The direction is the part worth saying out loud. Enforcing the documented sentence as written would refuse a name the live API serves, and accepting everything above U+007F is the side of that line a corpus can afford to be wrong on. Leaving non-ASCII case-folding accepted, and saying it is unmeasured rather than guessing, is the same call made twice.

Three checks on the parts a reader has to take on trust, all of which came back clean:

- **The two rules cannot drift apart.** `_SLACK_NAME_ILLEGAL` and the schema pattern agree on every code point I swept (U+0000 through U+02FF, plus CJK, Hangul, an emoji and U+10FFFF): zero disagreements. And `_slack_channel`'s output satisfies the schema on 4,000 random inputs drawn from a pool of the characters that matter, never exceeding 80.
- **The bench sweep reproduces exactly.** `generated_data/sources/slack`: 285,605 documents, 36 distinct channels, and the only value the pattern refuses is `slack/sports` with one occurrence. That file is `sports/1765000000-fantasy-trade-drama.json`, and its directory holds 16 files whose channels are `{'sports': 15, 'slack/sports': 1}` — the neighbours say what the description says they say.
- **Nothing else in the tree has to move.** Every channel in the example JSONL and in a fresh `backlot import --bundled` satisfies the new pattern.

## `name_normalized`, which this PR is what makes reachable

A question rather than a finding, and I could not settle it from a source. `routers/slack.py:211` serves `name_normalized` as the same string as `name`, and non-Latin channel names only become statable in a corpus with this change. Slack does not document what the normalization does: the conversation-object page shows only an example (`"name_normalized": "general"`, identical to `name` for an ASCII name), and walking `slack_web_openapi_v2.json` finds no `description` on the field at all. So I am not claiming a divergence.

You already hold the answer, though — the same `conversations.list` response that produced `일반`, `랜덤` and `식사` also carries their `name_normalized`. If those come back unchanged, the current line is right and worth a comment saying so; if they do not, it is a separate issue and this PR is what makes it worth filing.

khj809:
All three land, and the count correction is right — 6, not 5. My figure was measured before `eng.platform` joined the cases and I never re-took it.

**`name_normalized`** — settled, and it is what you expected. Same workspace, same call:

```
name='랜덤'  name_normalized='랜덤'  equal=True
name='일반'  name_normalized='일반'  equal=True
name='식사'  name_normalized='식사'  equal=True

channels: 8; name != name_normalized: 0
```

So `_channel_core` serving one string for both is correct, and `routers/slack.py` now carries a comment saying that was measured and that the vendor documents nothing about the normalization — nothing to file.

**Trailing newline** — folded into this PR rather than left for #105, since it is the guarantee the pattern was added for. Your diagnosis was exact, including that it is the newline specifically and that it is all six patterns. Fixed in `backlot/validation.py`, for the reason you gave for keeping it out of the patterns: they ship as the documented contract and have to stay readable by a non-Python validator. `fullmatch` I passed over for the mirror of that reason — the spec's `pattern` is an unanchored partial match, so it would trade this divergence for a second one the moment a schema wants a pattern anchored at neither end.

What it does instead is rewrite the `$` metacharacter to `\Z`, which is the whole of the difference between the two dialects' end anchor. An escaped `\$` or a `[$]` inside a class is left alone, and a class that opens `[]` or `[^]` closes on its second `]` rather than its first — both cases are parametrised, because a literal dollar sign is a value a corpus can carry and rewriting one would refuse a record the schema accepts. The message still reports the pattern the schema *wrote*, not the rewritten one.

Coverage is derived from the schemas rather than listed: `test_pattern_samples_cover_every_patterned_field` fails if a pattern is added without a sample, so the next anchored pattern is checked by code that already exists.

**Nit** — taken, `[:80].strip("-")`. Your input reproduces: 80 characters ending `aaaa-` before, 79 ending `aaaaa` after, and the all-separator input still falls through to `general`.

One thing outside the diff, in case it bites elsewhere: I hit `test_mcp.py`'s AWS case failing on `from mcp.shared.exceptions import McpError`, reproducible as plain `uvx awslabs.aws-api-mcp-server@latest --help` with Backlot nowhere in it. It was a stale checkout on my side — #98 had already pinned it — and it is green after merging main.

nuri-yoo:
Re-ran at `ffc231b`. All three land: `'incidents\n'` is refused, all six patterns refuse a trailing newline, `\Z` stays out of the shipped schemas, and the literal-dollar cases (`[$]`, `\$`, `[]$]`, `[\]$]`, `\\$`) survive the rewrite intact.

Your reason for passing over `fullmatch` is better than the suggestion I made — `pattern` is an unanchored partial match, so a full match breaks the first schema that wants neither end anchored.

1615 passed, 2 skipped here. Approval stands.

PR #99, opened 2026-09-01
slack: auth.test answers the caller's own user id

### LINKED ISSUE #88: slack: auth.test answers one hardcoded user_id for every caller
`auth.test` resolves the caller correctly and then reports a constant id. `user` differs per identity; `user_id` does not.

```
admin  {"user": "service-account",         "user_id": "USERVICE0", ...}
ava    {"user": "ava.chen@acme.com",       "user_id": "USERVICE0", ...}
dana   {"user": "dana.whitfield@acme.com", "user_id": "USERVICE0", ...}
```

Meanwhile `users.list` answers with the real derived ids for the same people:

```
U7D69A26700  avachen         ava.chen@acme.com
UCF68E7C66E  danawhitfield   dana.whitfield@acme.com
```

So the mock knows ava's id and `auth.test` declines to use it. `backlot/synth.py:214` is `slack_user_id(email)`, and `backlot/routers/slack.py:366` hardcodes `"user_id": "USERVICE0"` instead of calling it.

## Why it matters

Real Slack's `auth.test` on a user token returns *that user's* `U…` id, and "call `auth.test`, then match its `user_id` against message authors" is an ordinary thing for a connector to do — it is how you identify your own messages, or attribute an action to the caller. Here that comparison can never match: `conversations.history` reports authors as `U7D69A26700`, and `auth.test` claims the caller is `USERVICE0`, which is not any corpus user. A connector doing this correctly against production looks broken against the mock, and one doing it *wrongly* — ignoring `user_id` — passes.

It is also inconsistent within the same surface. Every other Slack endpoint scopes and identifies by the resolved caller; only `auth.test` flattens them.

## What fixed looks like

`user_id` is `synth.slack_user_id(caller.email)` for a corpus user. The admin/service identity has no corpus email, so a constant is defensible there — real Slack would return the bot/app's own id — but it should not be the answer for everyone.

Worth deciding at the same time: `slack.py:237-239` uses the same `USERVICE0` for a channel's `creator` and its topic/purpose `creator`. That is a different question (the corpus does not state who created a channel) but it is the same placeholder, and if it stays, it should be because a channel creator is genuinely unknown rather than by inheritance from this bug.

## Reproduce

```bash
backlot import --bundled && backlot serve
curl -s localhost:8000/slack/api/auth.test -H "Authorization: Bearer usr-29b84da5703116c2a832"
curl -s localhost:8000/slack/api/users.list -H "Authorization: Bearer usr-29b84da5703116c2a832"
```

`usr-29b84da5703116c2a832` is `ava.chen@acme.com` on the bundled corpus; tokens are derived from the corpus emails, so it is the same value on any machine.

Found while documenting the auth surface in #57, where the user-token/bot-token distinction made the `auth.test` response shape worth stating precisely.

nuri-yoo:
Closes #88.

## What changed

`auth.test` answers `user_id` with the caller's own derived id — `synth.slack_user_id(caller.email)`, the same `U…` that `users.list` reports as that person's `id` and `conversations.history` reports as a message's `user`. The admin/service token keeps `USERVICE0`: that identity has no corpus email, and real Slack answers a bot token with the app's own id rather than a person's. The response keys are unchanged, and `docs/auth.md` describes the shape without naming a value, so nothing moves in the docs.

The channel `creator` fields keep the service constant, deliberately: the issue splits that question off — a corpus does not state who created a channel — so whatever those fields end up carrying should be decided on its own rather than inherited from this fix.

`user` is the handle, not the address. Slack's spec pins it beside the id this PR is about — `slack_web_openapi_v2.json`, `paths./auth.test.get.responses.200`, whose user-token example is `"user": "grace"` with `"user_id": "W12345678"` — so the same divergence was standing in the field next door, and against `users.list`'s `name` for the same person. Both now derive through one `_handle(email)`, so a user object's `name` and `auth.test` cannot drift apart.

Slack's auth.test reference (docs.slack.dev/reference/methods/auth.test, read 2026-09-01) frames the method as telling "you" who you are, and its user-token success example carries that user's own `user_id`, where the bot-token example carries the app's id plus a `bot_id`. The failure this closes is the one #88 measured on 2026-08-26: every caller answered `USERVICE0`, an id `users.list` assigns to nobody, so "match auth.test's `user_id` against message authors" — how a client finds its own messages — could never match here while working in production.

## Verification

- **Tests** — `pytest` on a `uv sync --all-extras` install (`dev`, `examples`, `mcp`, `llamaindex`, `mirage`, `fsspec`): 1584 passed, 2 skipped. The skips are Docker (absent on this machine) and `llama-index-readers-hubspot`, which `pyproject.toml` documents as deliberately absent.
- **Optional surfaces that ran rather than skipped** — `mcp` (uvx present, the S3 MCP test ran), `examples`, `llamaindex`, `mirage`, `fsspec`, and the s3/sdk suites; the two skips above are the only tests that sat out.
- **Lint** — `ruff check . && ruff format --check .`: clean, 127 files.

- [x] A test fails without this change — `test_slack_auth_test_identifies_the_caller` against the unfixed router, on each half: `ava@acme.com authors a message in #eng-announcements, but auth.test's user_id USERVICE0 matches none of ['UC20FA2B1C0']`, and for the handle `assert 'ava@acme.com' == 'ava'`.
- [ ] A new endpoint serving corpus content is ACL-scoped, proved for both an admin and a scoped token — n/a: no new endpoint. `auth.test` still resolves the caller through `_caller_or_error` and reveals only the caller's own identity.

### DIFF
--- a/backlot/routers/slack.py
+++ b/backlot/routers/slack.py
@@ -368,9 +368,8 @@ async def auth_test(request: Request):
     object carries for that person — and `user_id` is that caller's own id, the `U…` that
     `users.list` reports and that a message in `conversations.history` names as its author.
 
-    Both were constants before, and the id is the one that costs: "call auth.test, then match
-    `user_id` against message authors" is how a client finds its own messages, and a workspace
-    constant can never match.
+    A workspace constant can never match: "call auth.test, then match `user_id` against message
+    authors" is how a client finds its own messages, so the id is the field that costs.
 
     The admin/service token has no corpus email, so it keeps the service identity: that caller is
     not a person, and real Slack answers a bot token with the app's own id rather than a user's."""
--- a/backlot/routers/slack.py
+++ b/backlot/routers/slack.py
@@ -899,7 +899,14 @@ async def search_all(request: Request):
 
 
 def _search_match(conn, row) -> dict:
-    """A search.messages `matches[]` entry for a slack row."""
+    """A search.messages `matches[]` entry for a slack row.
+
+    `username` is the author's handle, the same string `users.list` gives that person as `name` —
+    Slack's spec spells it that way in this method's own example (`slack_web_openapi_v2.json`,
+    `paths./search.messages.get.responses.200`, whose matches carry `"username": "roach"`). It goes
+    through `_handle` for that reason: a hit and a user object name one person, and a client that
+    reads the id off `user` and the handle off `username` must not get two spellings of them.
+    """
     ch = row["channel"]
     cid = synth.slack_channel_id(ch)
     text = row["content"]
@@ -913,7 +920,7 @@ def _search_match(conn, row) -> dict:
             "is_private": not store.container_has_public(conn, "slack", ch),
         },
         "user": synth.slack_user_id(row["author_email"]),
-        "username": row["author_email"].split("@")[0],
+        "username": _handle(row["author_email"]),
         "ts": ts,
         "text": text,
         "permalink": f"https://{get_settings().org_name}.slack.com/archives/{cid}/p{ts.replace('.', '')}",
--- a/backlot/routers/slack.py
+++ b/backlot/routers/slack.py
@@ -292,6 +292,12 @@ def _channel_names(conn) -> list[str]:
     return [row["name"] for row in store.list_containers(conn, "slack")]
 
 
+def _handle(email: str) -> str:
+    """A corpus email as Slack's handle — the `name` a user object carries and the `user`
+    `auth.test` reports for the same person."""
+    return email.split("@")[0].replace(".", "")
+
+
 def _user_obj(conn, email: str) -> dict:
     u = store.get_user(conn, email)
     display = u["display_name"] if u else email.split("@")[0]
@@ -301,7 +307,7 @@ def _user_obj(conn, email: str) -> dict:
     return {
         "id": synth.slack_user_id(email),
         "team_id": TEAM_ID,
-        "name": email.split("@")[0].replace(".", ""),
+        "name": _handle(email),
         "real_name": display,
         "deleted": False,
         "is_bot": is_bot,
@@ -354,17 +360,24 @@ async def api_test(request: Request):
 
 @router.api_route("/auth.test", methods=["GET", "POST"])
 async def auth_test(request: Request):
-    """Who the token is: `user_id` is the CALLER's derived id, the same `U…` every other route
-    reports for that person (users.list `id`, a message's `user`). Real Slack answers the token
-    owner's own id here, and "call auth.test, then match `user_id` against message authors" is how
-    a client identifies its own messages — a constant can never match.
+    """Who the token is, in the caller's own terms.
 
-    The admin/service token has no corpus email, so it keeps the service constant: that identity
-    is not a person, and real answers a bot token with the app's own id rather than a user's."""
+    Slack's spec pins both fields at once: `slack_web_openapi_v2.json`,
+    `paths./auth.test.get.responses.200`, whose user-token example is `"user": "grace"` with
+    `"user_id": "W12345678"`. So `user` is the HANDLE, not an address — the same `name` a user
+    object carries for that person — and `user_id` is that caller's own id, the `U…` that
+    `users.list` reports and that a message in `conversations.history` names as its author.
+
+    Both were constants before, and the id is the one that costs: "call auth.test, then match
+    `user_id` against message authors" is how a client finds its own messages, and a workspace
+    constant can never match.
+
+    The admin/service token has no corpus email, so it keeps the service identity: that caller is
+    not a person, and real Slack answers a bot token with the app's own id rather than a user's."""
     caller, err = _caller_or_error(request)
     if err is not None:
         return err
-    who = "service-account" if caller.is_admin else caller.email
+    who = "service-account" if caller.is_admin else _handle(caller.email)
     return {
         "ok": True,
         "url": f"https://{get_settings().org_name}.slack.com/",
--- a/backlot/routers/slack.py
+++ b/backlot/routers/slack.py
@@ -354,6 +354,13 @@ async def api_test(request: Request):
 
 @router.api_route("/auth.test", methods=["GET", "POST"])
 async def auth_test(request: Request):
+    """Who the token is: `user_id` is the CALLER's derived id, the same `U…` every other route
+    reports for that person (users.list `id`, a message's `user`). Real Slack answers the token
+    owner's own id here, and "call auth.test, then match `user_id` against message authors" is how
+    a client identifies its own messages — a constant can never match.
+
+    The admin/service token has no corpus email, so it keeps the service constant: that identity
+    is not a person, and real answers a bot token with the app's own id rather than a user's."""
     caller, err = _caller_or_error(request)
     if err is not None:
         return err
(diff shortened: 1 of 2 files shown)


khj809 (commented):
The `user_id` fix is correct: on the fixture corpus `auth.test` -> `users.list` id -> `conversations.history`'s `user` all agree, and reverting the one line does make the new test fail. Three notes.

backlot/routers/slack.py
```diff
@@ -363,7 +370,7 @@ async def auth_test(request: Request):
         "url": f"https://{get_settings().org_name}.slack.com/",
         "team": get_settings().org_name,
         "user": who,
```
khj809, line 390: The `user` question left open in the description is answered by the vendor spec. `slack_web_openapi_v2.json`, `paths./auth.test.get.responses.200` (the user-token success example) carries `"user": "grace", "user_id": "W12345678"` — `user` is the handle, not an email. So the same class of divergence this PR fixes is left standing in the field next to it, and it also disagrees with `users.list`'s `name` (`hana`) for the same person. The fix is one line with the same derivation as `_user_obj`, and nothing in the repo reads `auth.test`'s `user` (neither examples nor docs name a value), so closing it here is cheap. If it does wait, drop the test's `r["user"] == email` assertion — it pins the divergence as the expected value.

backlot/routers/slack.py
```diff
@@ -354,6 +354,13 @@ async def api_test(request: Request):
 
 @router.api_route("/auth.test", methods=["GET", "POST"])
 async def auth_test(request: Request):
+    """Who the token is: `user_id` is the CALLER's derived id, the same `U…` every other route
+    reports for that person (users.list `id`, a message's `user`). Real Slack answers the token
+    owner's own id here, and "call auth.test, then match `user_id` against message authors" is how
+    a client identifies its own messages — a constant can never match.
+
+    The admin/service token has no corpus email, so it keeps the service constant: that identity
+    is not a person, and real answers a bot token with the app's own id rather than a user's."""
```
khj809: Under AGENTS.md this docstring is a source attribution, but the code carries no source. The `docs.slack.dev` citation in the PR body does not survive into the file, and `_caller_or_error` / `_missing_argument` in this same module say "Measured against the live API:". Please add the spec URL or the measurement date — the `auth.test` success example above proves both `user_id` and `user` at once, so it works as the citation as-is. (Minor: both docstrings read "and real answers a bot token…" — "Slack" is missing from "real Slack".)

tests/test_slack.py
```diff
@@ -159,6 +159,31 @@ def test_slack_auth_error_split_is_uniform_across_methods(client):
...
+    answers a bot token with the app's own id rather than a person's."""
+    members = client.post("/slack/api/users.list", headers=admin_h).json()["members"]
+    by_email = {m["profile"]["email"]: m["id"] for m in members}
+    ids = {}
+    for email in ("hana@acme.com", "bob@acme.com"):
+        r = client.post(
+            "/slack/api/auth.test", headers={"Authorization": f"Bearer {tokens[email]}"}
+        ).json()
+        assert r["ok"] is True and r["user"] == email
+        assert r["user_id"] == by_email[email], email
```
khj809: This assertion is one step weaker than the claim. Both `auth.test` and `users.list` go through `synth.slack_user_id`, so it still passes if the two derivations diverge. The scenario the issue describes is a match against `conversations.history`'s `user`, so picking a caller who actually authors a Slack message (ava in the fixture) and comparing against the history author would catch that regression.

backlot/routers/slack.py
nuri-yoo: Fixed in `9028dcf`. You are right that the spec answers it, and that leaving it was leaving the same divergence in the next field — `paths./auth.test.get.responses.200` is `"user": "grace"` with `"user_id": "W12345678"`, so one example fixes both. `user` is now the handle, through a `_handle(email)` that `_user_obj` also calls for its `name`, so the two cannot drift apart the way they had.

The test's `r["user"] == email` is gone with it — it now asserts the handle and that no address survives.

backlot/routers/slack.py
nuri-yoo: Fixed in `9028dcf`. The citation is in the docstring now, in the form the spec supports rather than the docs URL: `slack_web_openapi_v2.json`, `paths./auth.test.get.responses.200`, quoting the user-token example. Since it pins `user` and `user_id` together, it is the one source both halves of this change rest on.

"real answers a bot token" is "real Slack" in both places.

tests/test_slack.py
nuri-yoo: Fixed in `9028dcf`. The assertion was circular — `auth.test` and `users.list` both derive through `synth.slack_user_id`, so it would have survived exactly the drift the issue is about.

It now takes the two fixture users who author messages (ava in `#eng-announcements`, bob in `#incidents`), and matches `auth.test`'s `user_id` against the `user` on a message in `conversations.history`. Against the pre-fix router it fails as the scenario reads: `ava@acme.com authors a message in #eng-announcements, but auth.test's user_id USERVICE0 matches none of ['UC20FA2B1C0']`. The per-caller check stayed, on three tokens.

khj809 (commented):
All three are addressed, and I checked each half by reverting it: `user` back to the address fails at `tests/test_slack.py:181`, `user_id` back to the constant fails at `:194` on the history match. Suite and lint clean here too. Two small things left, neither blocking.

**`search.messages`'s `username` keeps the dots** — `backlot/routers/slack.py:917` is still `row["author_email"].split("@")[0]`, so `vault.keeper@acme.com` is `vault.keeper` in a search hit and `vaultkeeper` everywhere `_handle` now runs. The spec's search.messages example carries `"username": "roach"`, the same handle concept, so this is the helper's third caller:

```python
        "username": _handle(row["author_email"]),
```

It predates this PR (there since the initial release), so a follow-up is fine — but `_handle` is two functions above it now. Worth noting either way that no tokened fixture user has a dotted address, so the dot-stripping half of `_handle` is not exercised by a test.

backlot/routers/slack.py
```diff
@@ -354,16 +360,30 @@ async def api_test(request: Request):
...
+
+    Slack's spec pins both fields at once: `slack_web_openapi_v2.json`,
+    `paths./auth.test.get.responses.200`, whose user-token example is `"user": "grace"` with
+    `"user_id": "W12345678"`. So `user` is the HANDLE, not an address — the same `name` a user
+    object carries for that person — and `user_id` is that caller's own id, the `U…` that
+    `users.list` reports and that a message in `conversations.history` names as its author.
+
+    Both were constants before, and the id is the one that costs: "call auth.test, then match
+    `user_id` against message authors" is how a client finds its own messages, and a workspace
+    constant can never match.
```
khj809: "Both were constants before" states what the code used to be — the history of the fix rather than what was measured, which is the box this PR checks. It reads as an invariant with one word moved:

```suggestion
    A workspace constant can never match: "call auth.test, then match `user_id` against message
    authors" is how a client finds its own messages, so the id is the field that costs.
```

nuri-yoo:
Folded in as `a3422ed` rather than left for a follow-up: `_handle` existing at all was this PR's answer to two derivations drifting, so leaving the third caller out would keep the drift in the one place the helper was added to close.

It is visible in a served corpus, not only in the code — one author, one id, two spellings:

```
users.list             id=U7D69A26700  name='avachen'
search.messages        user=U7D69A26700  username='ava.chen'   <- before
search.messages        user=U7D69A26700  username='avachen'    <- after
```

The test states a dotted address for the reason you gave: no address in the module's shared fixture has one, so the dot-dropping half of `_handle` had nothing exercising it. Reverting the line fails it on `'ava.chen' == 'avachen'`.

Your spec pointer checks out — `paths./search.messages.get.responses.200` carries `"username": "roach"` on its matches — and the line dates to #44, so it did predate this. 1585 passed, 2 skipped; ruff and `gen_docs --check` clean.

I also re-ran the gates over your applied suggestion, since applying one from the web skips them: clean there too.

khj809 (approved):
Looks good

PR #102, opened 2026-09-01
An optional test gate must be reachable from a documented extra

nuri-yoo:
`pytest.importorskip` makes a test self-skip when its module is absent, so a gate whose distribution no extra carries never fails anywhere: a contributor who installed every extra still skips it believing they ran it, and `-rs` reports it without failing. The suite has already paid for this once — #64's review is the measured case, where a `1138 passed, 24 skipped` run read as green while the reviewer's fuller install failed four of the skipped tests. #86 and #87 answered with CONTRIBUTING's gate table and a CI that installs everything, but the table is prose over `pyproject.toml` and the gates are calls in test files, and nothing held the three to each other.

## The change

Two tests in `tests/test_packaging.py`, beside the package-data test that guards the same drift class one layer down. Both read sources, never a test run — failing CI on runtime skips was considered in #87 and kept out at the repo owner's request, and this is deliberately not that: neither test can fire on a machine condition like a missing Docker or a cold network, only on a diff that adds a gate, renames an extra, or moves a file.

`test_the_contributing_extras_are_the_pyproject_extras` holds the prose to the config, in both directions: every `.[extra]` CONTRIBUTING names must exist in `pyproject.toml`, every extra pyproject defines must be named, and every test file the gate table points at must still be on disk.

`test_every_optional_import_gate_is_reachable_from_an_extra` collects every `importorskip`'d module — with `ast` rather than a regex, because `tests/test_sdk.py` spells four of its gates through a loop variable that a literal scan silently drops, and a collector that misses gates proves nothing — and asserts each maps to a distribution the core dependencies or some extra carries. Names PEP 503 normalization cannot reach (`googleapiclient` → `google-api-python-client`, the two mirage shim modules) sit in a small explicit map, so a new gate arrives either with its extra or with a one-line map entry in the same diff. The one deliberate absence, `llama-index-readers-hubspot`, is asserted in both directions: something must still gate on it, and no extra may quietly start carrying it while CONTRIBUTING's "which no extra carries" row says otherwise.

## Verification

- **Tests** — `pytest` on a `uv sync --all-extras` install: 1583 passed, 2 skipped (Docker absent on this machine, plus the documented `llama-index-readers-hubspot`).
- **Each guarded drift injected, each caught** — a new `importorskip("shiny_new_reader")` with no extra: `these importorskip gates name modules no extra's distribution provides … {'shiny_new_reader': ['test_zz_probe.py']}`; the `mirage` extra renamed in pyproject: the two-set mismatch names both directions; `llama-index-readers-hubspot` added to the `llamaindex` extra: `is now carried by an extra; llama_index.readers.hubspot is no longer a documented absence`; CONTRIBUTING pointed at `tests/test_mcpx.py`: `CONTRIBUTING points at test files that are gone`. All four reverted; the tree diff is the two tests.

### DIFF
--- a/tests/test_packaging.py
+++ b/tests/test_packaging.py
@@ -145,6 +145,13 @@ def _norm(dist: str) -> str:
     return re.sub(r"[-_.]+", "-", dist).lower()
 
 
+def pyproject_optional_dependencies() -> dict[str, list[str]]:
+    """`[project.optional-dependencies]` verbatim — the requirement strings, not the parsed names,
+    which is what the aggregate has to be read from."""
+    with open(REPO_ROOT / "pyproject.toml", "rb") as f:
+        return tomllib.load(f)["project"]["optional-dependencies"]
+
+
 def _extras() -> dict[str, set[str]]:
     with open(REPO_ROOT / "pyproject.toml", "rb") as f:
         pyproject = tomllib.load(f)
@@ -155,23 +162,42 @@ def _extras() -> dict[str, set[str]]:
     }
 
 
+def _loop_values(tree: ast.AST, call: ast.Call, target: str) -> list[str] | None:
+    """The strings ``target`` takes, if ``call`` sits inside a ``for`` that binds it to a literal
+    sequence of them. A tuple and a list are the same loop, so both are read.
+
+    Scoped to the enclosing loop rather than to the file: the answer depends on WHERE the call is,
+    and a name is free to mean something else in the next loop down.
+    """
+    for node in ast.walk(tree):
+        if not (isinstance(node, ast.For) and isinstance(node.target, ast.Name)):
+            continue
+        if node.target.id != target or not isinstance(node.iter, ast.Tuple | ast.List):
+            continue
+        if not all(
+            isinstance(e, ast.Constant) and isinstance(e.value, str) for e in node.iter.elts
+        ):
+            continue
+        if any(child is call for child in ast.walk(node)):
+            return [e.value for e in node.iter.elts]
+    return None
+
+
 def _importorskip_modules() -> dict[str, list[str]]:
     """Every module the suite gates on, module -> the test files that gate on it.
 
     Read with ``ast`` rather than a regex because one gate is spelled through a loop variable
     (``for _mod in (...): pytest.importorskip(_mod)`` in ``tests/test_sdk.py``), which a string
     scan for literals silently drops — and a collector that misses gates proves nothing.
+
+    The loop variable is resolved INSIDE the loop that binds it, not from a file-wide map: one
+    map per file is last-write-wins, so a second ``for`` over the same name would drop the first
+    loop's gates, and a second ``for`` that is ordinary iteration would contribute names that gate
+    nothing. Both leave the collector reporting confidently on the wrong set.
     """
     found: dict[str, list[str]] = {}
     for path in sorted((REPO_ROOT / "tests").glob("*.py")):
         tree = ast.parse(path.read_text())
-        loop_names: dict[str, list[str]] = {}
-        for node in ast.walk(tree):
-            if isinstance(node, ast.For) and isinstance(node.target, ast.Name):
-                if isinstance(node.iter, ast.Tuple) and all(
-                    isinstance(e, ast.Constant) and isinstance(e.value, str) for e in node.iter.elts
-                ):
-                    loop_names[node.target.id] = [e.value for e in node.iter.elts]
         for node in ast.walk(tree):
             if not (isinstance(node, ast.Call) and node.args):
                 continue
@@ -182,8 +208,8 @@ def _importorskip_modules() -> dict[str, list[str]]:
             arg = node.args[0]
             if isinstance(arg, ast.Constant) and isinstance(arg.value, str):
                 mods = [arg.value]
-            elif isinstance(arg, ast.Name) and arg.id in loop_names:
-                mods = loop_names[arg.id]
+            elif isinstance(arg, ast.Name) and (bound := _loop_values(tree, node, arg.id)):
+                mods = bound
             else:  # pragma: no cover — a shape this collector cannot read must fail, not skip
                 raise AssertionError(
                     f"unreadable importorskip argument in {path.name}: {ast.dump(arg)}"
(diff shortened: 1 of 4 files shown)


khj809:
Ran both tests against the head and reproduced each point below with an injected probe. The `ast` collector is the right call — a literal scan really does drop `test_sdk.py`'s loop-variable gates, and asserting `_DOCUMENTED_ABSENT` in both directions is what keeps the exception row honest. Four things, the first structural.

**The extras list has a fourth copy, and it is the one that decides whether CI runs a gate.** `.github/workflows/ci.yml:30` hard-codes `pip install -e ".[dev,examples,mcp,llamaindex,mirage,fsspec]"`, which neither test reads. Probe: add `shiny` to `[project.optional-dependencies]`, a `.[shiny]` row to CONTRIBUTING's table, and a test gating on `importorskip("shiny_new_reader")` — both new tests pass, CI never installs the extra, the test skips forever, and `-rs` reports it without failing. That is #64 again.

Worth making impossible rather than tested. pip has no `--all-extras` (pip 26.2.1: only `--group`, one at a time; `uv pip install -e . --all-extras` is rejected outright, and `-r pyproject.toml --all-extras` installs the deps but not the project). A self-referential aggregate extra does work — verified on pip 26.2.1, where the self-reference resolves to the local editable directory, not PyPI:

```toml
# Every extra at once. pip has no --all-extras, so the aggregate is a self-reference.
all = ["backlot[dev,examples,mcp,llamaindex,mirage,fsspec]"]
```

ci.yml and both CONTRIBUTING install lines then become `pip install -e ".[all]"`, the four copies collapse to one, and what is left is a single-file invariant worth three lines here instead of a test that reads a workflow:

```python
aggregated = set(re.search(r"backlot\[([\w,]+)\]", ...).group(1).split(","))
assert aggregated == set(_extras()) - {"all"}
```

`named == extras` keeps passing, since CONTRIBUTING names `.[all]` in the install command.

**`loop_names` is file-scoped and last-write-wins (line 174).** A second `for _mod in (...)` anywhere in the file overwrites the first. With `for _mod in ("A", "B"): importorskip(_mod)` followed by `for _mod in ("slack_sdk",): importorskip(_mod)`, the test *passes* with A and B dropped — the miss the docstring rules out. If the second loop is ordinary iteration (`for _mod in ("alpha", "beta"): seen.append(_mod)`) it fails with `{'alpha': [...], 'beta': [...]}`: names that are not gates at all, while the real gates stay invisible. Resolving the name inside each `ast.For` subtree fixes both; the `slack_sdk` sanity check only covers `test_sdk.py`.

**Only `ast.Tuple` iterables are read (line 170).** `for _m in ["slack_sdk", "googleapiclient"]:` — identical at runtime — hits the `raise AssertionError` branch, so a benign refactor turns the guard red. Same for a module-level tuple constant. Accepting `ast.List` covers it.

**`\.\[([\w,]+)\]` cannot match a hyphenated extra or `.[dev, mcp]` (line 202).** Over `.[dev]`, `.[all-extras]`, `.[dev, mcp]` the comprehension yields only `['dev']`: a correctly documented hyphenated extra fails as "pyproject defines extras CONTRIBUTING never names", blaming the wrong file, and a bogus `.[bogus-extra]` slips through the other direction. `[\w,\s-]+` plus a strip per part.

nuri-yoo:
All four reproduce, and all four are fixed in `56c618c`. The first one is the review — a guard that reads three copies of a list while a fourth decides the outcome is the shape of bug it was written against.

**The aggregate.** Taken as you wrote it. I checked the self-reference before relying on it, since `backlot` is on PyPI and resolving there instead would have installed a release's dependencies quietly: `pip install --dry-run -e ".[all]" --report` names `file:///…/drift` for the backlot entry with `is_direct: true`, at the local `0.0.2.dev9+gc4c1743` version, and a real (non-dry) install brings up `pytest`, `slack_sdk`, `fastmcp`, `llama_index.readers.slack`, `mirage` and `fsspec` together. `uv sync --all-extras` still resolves it too, which the local loop depends on. `pip 26.2.1` has no `--all-extras`, as you said.

`ci.yml` and CONTRIBUTING's zero-skip line are now `.[all]`, and `test_the_all_extra_aggregates_every_other_one` holds the aggregate to the extras. Your probe is the check: with `shiny` defined, a `.[shiny]` row added and a gate on it, the suite now fails instead of passing.

```
`all` installs ['dev', 'examples', 'llamaindex', 'mcp', 'mirage'] but the extras are
['dev', 'examples', 'fsspec', 'llamaindex', 'mcp', 'mirage'] — an extra missing here is
one CI never installs
```

**The loop variable** is resolved inside the `ast.For` that binds it rather than from a file map. Both halves of your probe come out right now: two gate loops keep `alpha_gate` and `beta_gate` instead of dropping them, and a plain `for _mod in ("alpha", "beta")` beside a gate loop contributes nothing.

**`ast.List`** reads as the same loop as a tuple — `for _m in ["slack_sdk", "googleapiclient"]` collects both instead of raising.

**The extras pattern** is `[\w,\s-]+` with a strip per part: over `` `.[dev]`, `.[all-extras]`, `.[dev, mcp]` `` it now yields all three.

1620 passed, 2 skipped; ruff and `gen_docs --check` clean.

tests/test_packaging.py
```diff
@@ -112,3 +113,198 @@ def test_pypi_readme_has_no_relative_links():
...
+
+
+def test_the_all_extra_aggregates_every_other_one():
+    """`.[all]` is what CI installs, so it is what decides whether a gated test runs at all.
+
+    It is a fourth copy of the extras list — `pyproject.toml` defines them, CONTRIBUTING's table
+    names them, its install line and `.github/workflows/ci.yml` install them — and the copy that
+    silently costs coverage: an extra added everywhere else but left out of the aggregate installs
+    nowhere, so every test behind it skips forever and `-rs` reports that without failing. The
+    other three copies are held to each other by the test below; this holds the one that matters.
```
khj809: This paragraph describes the state before `.[all]`. CONTRIBUTING's install line and `ci.yml` now name the aggregate instead of the list, so neither is a copy of it any more: the places that enumerate the extras are pyproject's definitions, CONTRIBUTING's table, and this aggregate — three, and the test below holds two of them to each other rather than three.

```suggestion
    It is one of three places that enumerate the extras — `pyproject.toml` defines them,
    CONTRIBUTING's table names them, this aggregate installs them — and the one that silently
    costs coverage: an extra added to the other two but left out here installs nowhere, so every
    test behind it skips forever and `-rs` reports that without failing. The table and the
    definitions are held to each other by the test below; this holds the copy CI acts on.
```

tests/test_packaging.py
```diff
@@ -112,3 +113,198 @@ def test_pypi_readme_has_no_relative_links():
...
+    test behind it skips forever and `-rs` reports that without failing. The table and the
+    definitions are held to each other by the test below; this holds the copy CI acts on.
+
+    An aggregate rather than a flag because pip has no `--all-extras` (measured on pip 26.2.1),
+    and a self-reference rather than a duplicated list because pip resolves `backlot[...]` to the
+    directory being installed: the resolution report names a `file://` url for it and marks it
+    direct, so the release on PyPI is never consulted.
+    """
+    extras = _extras()
+    assert "all" in extras, "the aggregate CI installs is gone; ci.yml still names `.[all]`"
```
khj809, line 237: The docstring says "`.[all]` is what CI installs", but there's no any assert that checks it.
Would be enough to add just one line:

```suggestion
    assert "all" in extras, "the aggregate CI installs is gone; ci.yml still names `.[all]`"
    assert '".[all]"' in (REPO_ROOT / ".github/workflows/ci.yml").read_text()
```

khj809 (approved):
LGTM with minor one

PR #104, opened 2026-09-01
github: errors answer GitHub's envelope, and the 401 says which credential failed

### LINKED ISSUE #100: github: errors answer FastAPI's envelope, except the two shaped by hand
**Source and commit** — `github`, at `4577a31` (current main).

Real GitHub answers an error as `{"message", "documentation_url", "status"}`, plus an `errors` member on a few. Backlot's github router speaks that envelope in exactly two places — the `X-GitHub-Api-Version` 400 and the q-less code-search 422, both shaped by hand via `exc.github_body` — and every other error on the surface falls through to FastAPI's `{"detail": …}`. The seam for the conversion already exists and says so: `backlot/errors/github.py` claims every `/github` path in `owns()` and its module docstring calls converting the whole surface "a real fidelity improvement; it is also its own change, touching every 404 the router raises and the tests that assert them. It is deliberately not bundled here." This is that change, filed.

**How you know** — measured against api.github.com on 2026-09-01 with curl (a bad bearer token, no token, an unauthenticated GET of a private repository, and the two search requests below), re-confirming a first measurement from 2026-08-27 made while auditing the README's fidelity claims — the README stopped claiming error-byte equality because of it. Backlot's side measured the same day on the bundled corpus at `4577a31`.

**Reproduction** — the bundled corpus: `backlot import --bundled && backlot serve`, then the curl of each block against `localhost:8000/github` with the admin token from `/_meta/users`.

| # | Divergence | Silent? |
|---|---|---|
| 1 | q-less `/search/issues` answers 200 with everything; real answers 422 Validation Failed | **yes** |
| 2 | 401 and 404 answer `{"detail": …}`, not the envelope | no |
| 3 | a missing Authorization header answers "Bad credentials"; real never conflates missing with wrong | no |
| 4 | `per_page=0` answers FastAPI's 422 array; real answers 200 | no |

## 1. A q-less issues search is not an error at all

**Request** — `GET /search/issues`, no `q`.

**Backlot serves** — `200`, with `total_count` and every issue and pull the caller can see: `q` defaults to `""` and an empty query matches everything.

**The real API serves** — `422 {"message": "Validation Failed", "errors": [{"resource": "Search", "field": "q", "code": "missing"}], "documentation_url": "https://docs.github.com/v3/search", "status": "422"}`.

**Can a client tell?** No — this is the expensive case. A client that forgets `q` gets a plausible, ACL-scoped result set and ships; the same request against real GitHub is a hard 422. The route one door over already answers this correctly: `/search/code` raises `_search_validation_failed("q")`, whose body is the real envelope above, so the two searches currently disagree with each other as well as with real.

## 2. 401 and 404 leave the envelope

**Request** — any repo route with a wrong token; any hidden or absent repo with a valid one.

**Backlot serves** — `401 {"detail": "Bad credentials"}` / `404 {"detail": "Not Found"}`.

**The real API serves** — `401 {"message": "Bad credentials", "documentation_url": "https://docs.github.com/rest", "status": "401"}` / `404 {"message": "Not Found", "documentation_url": "https://docs.github.com/rest/repos/repos#get-a-repository", "status": "404"}`. Note `documentation_url` is route-specific on the 404 and generic on the 401 — a conversion has to decide whether to carry that or accept a smaller divergence, and the values are otherwise already real's ("Bad credentials", "Not Found").

**Can a client tell?** Yes, but in its error path, and the SDK the examples ship makes it concrete: PyGithub classifies a 401 by the body's `message`, so against real it raises `BadCredentialsException` and against Backlot the same wrong token raises the generic `GithubException` (measured with `Auth.Token("not-a-real-token")` on both, 2026-09-01). A client catching `BadCredentialsException` — the documented way — catches nothing here.

## 3. A missing header is not a wrong token

**Request** — any repo route with no Authorization header at all.

**Backlot serves** — `401 {"detail": "Bad credentials"}`, the same answer a wrong token gets.

**The real API serves** — two different things, neither of which is that: an auth-requiring endpoint answers `401 {"message": "Requires authentication", …}`, and an unauthenticated GET of a private repository answers `404 {"message": "Not Found", …}` — existence hidden, which is the model Backlot itself follows for a token that cannot see a repo. Real reserves "Bad credentials" for credentials that are present and wrong.

**Can a client tell?** A client distinguishing "I forgot the token" from "my token is bad" — retry logic, credential rotation — reads the wrong branch here.

## 4. A zero `per_page` is tolerated by real and 422'd here

**Request** — `GET /search/issues?q=test&per_page=0`.

**Backlot serves** — `422 {"detail": [{"type": "greater_than_equal", "loc": ["query", "per_page"], …}]}` — FastAPI's validator, whose array shape real never sends. Every `ge=1` page parameter on the surface shares this path, and `errors/github.py::validation_body` currently states no github route answers through the validator — these do.

**The real API serves** — `200`, the parameter tolerated.

**Can a client tell?** Only when a paginator computes an edge value; then it gets a hard error real would have absorbed.

**Where it lands** — the seam is built: `errors/github.py` already owns every `/github` path and renders `github_body` when an exception carries one, so the conversion is attaching real bodies at the raise sites (or synthesizing them in `http_body`), giving `validation_body` an answer, and raising `_search_validation_failed("q")` from `search_issues` the way `search_code` already does. `tests/test_github.py` asserts the current `{"detail": …}` shape in one place (`test_github.py:617`, whose own comment names the FastAPI fallback), and the four blocks above each want a test against the measured body.

nuri-yoo:
Closes #100.

## What changed

Every error on the github surface now answers `{"message", "documentation_url", "status"}`. Two errors already did — the `X-GitHub-Api-Version` 400 and the code-search 422, both shaped by hand through `exc.github_body` — and `backlot/errors/github.py` already owned every `/github` path against the day the rest followed. This is that day: `http_body` renders any exception raised on the surface, and the hand-shaped bodies still win where they exist.

The wording needed almost nothing. "Not Found" and "Bad credentials" were already real's, because the routers were written against measured responses; only the shape around them was FastAPI's.

`documentation_url` is the part that had to be measured rather than derived. Real names the endpoint that failed, not a root — `/branches` gets the list-branches anchor, `/tags` the list-tags one — so a single URL would have been a divergence on every 404. Requesting each route shape against a repository that does not exist returns that route's own anchor, and `ROUTE_DOCS` is the result for all 31 routes. A test holds the table to the routes the app serves, so a route added without an entry fails rather than quietly answering the root.

## The three answers that change, and why they are real's

**A `q`-less issue search is a 422, not a listing (#100's first block).** This is the one a client cannot see: a forgotten query got a plausible, ACL-scoped result set here and a hard 422 in production. `/search/code` next door already answered correctly, so the two searches disagreed with each other as well as with real. Measured against api.github.com — `GET /search/issues` with no `q` is `Validation Failed`, resource `Search`, field `q`, code `missing`. (`/search/code` differs only on a `q` that is absent rather than empty: real rejects that one at the query-string parser, before the API, with a `text/plain` 400. Backlot declares `q` with a default, so both routes see one case, and the 422 is the one to serve.)

**A request with no credential says so.** Real reserves "Bad credentials" for a credential that arrived and did not resolve; one carrying none gets "Requires authentication". A client telling "I forgot the token" from "my token is wrong" read the same answer for both. An `Authorization` header real cannot parse belongs on the missing side, not the wrong side: real ignores it and serves the request anonymously — `Basic …` and a scheme-less value both answer 200 on a public repository — so the caller arrives with no credential rather than a rejected one. Backlot serves nothing anonymously, which is why that lands as the missing-credential 401 instead of as a public read.

**Pagination values real absorbs are absorbed.** Real refuses none of them: `per_page=0`, `per_page=abc`, `page=0`, `page=-1`, `page=` and a `per_page` over the cap are each a 200 with the defaults applied. Backlot declared a floor and an `int`, so FastAPI's validator answered before `clamp_page` was reached and a paginator computing an edge value got a hard error where production absorbs it. The parameters now take a `BeforeValidator` that reads an unparseable value as absent, which `clamp_page` already treated as the default — and the OpenAPI schema stays an integer, which is what real's own spec declares, so the tolerance sits in the runtime where real has it rather than in the contract where real does not.

**A path parameter real has no route for is real's 404.** `/repos/{owner}/{repo}/issues/notanint` is `Not Found` with that route's own anchor, measured; Backlot answered a 422. Real never matches such a path at all, while Backlot declares `number: int` and fails after matching — a difference in how the two arrive, not in what they answer. Any path failure decides it rather than whichever error is reported first, so a request that fails on a path and a query parameter at once has one answer.

## Two knowing divergences

Real is not uniform on the 401's `documentation_url`: a bad token carries the bare root on every route measured, while `/user/repos` with no header carries that route's anchor. Backlot answers the root for both, which is right everywhere except that one route-and-cause pair.

A 405 stays outside the envelope on purpose. Every route declares GET, so a wrong method is refused with `http.HTTPStatus(405).phrase`, and no measurement attributes that string to real — which answers a wrong method per endpoint rather than uniformly. It keeps FastAPI's `detail`, which says plainly that the mock is answering.

## Verification

- **Tests** — `pytest` on a `uv sync --all-extras` install: 1625 passed, 2 skipped. The skips are Docker (absent on this machine) and `llama-index-readers-hubspot`, which `pyproject.toml` documents as deliberately absent.
- **Optional surfaces that ran rather than skipped** — `examples` (the PyGithub checks below are in `tests/test_sdk.py`), plus `mcp`, `llamaindex`, `mirage`, `fsspec` and the s3/sdk suites; the two skips above are the only tests that sat out.
- **Lint** — `ruff check . && ruff format --check .`: clean. `python scripts/gen_docs.py --check`: exit 0.

- [x] A test fails without this change — the two SDK checks are load-bearing in the sense #40 used: with `owns()` returning False, `GitHub.wrong token -> BadCredentialsException: GithubException: 401 {"detail": "Bad credentials"}` and `GitHub.missing repo -> UnknownObjectException: GithubException: 404 {"detail": "Not Found"}`. Against real those same calls raise `BadCredentialsException` and `UnknownObjectException`; against a `detail` body both were the bare `GithubException`.
- [ ] A new endpoint serving corpus content is ACL-scoped, proved for both an admin and a scoped token — n/a: no new endpoint. A repo the caller cannot see still answers exactly as one that is not there.

### DIFF
--- a/backlot/errors/github.py
+++ b/backlot/errors/github.py
@@ -1,35 +1,184 @@
-"""GitHub's error envelope, for the errors Backlot shapes on purpose.
+"""GitHub's error envelope.
 
-Real GitHub answers an error as ``{"message", "documentation_url", "status"}`` — plus a
-``"errors"`` member on a few, where it is sometimes a string rather than the usual array. That is
-not FastAPI's ``{"detail": …}``, so converting the WHOLE github surface is a real fidelity
-improvement; it is also its own change, touching every 404 the router raises and the tests that
-assert them. It is deliberately not bundled here.
+Real GitHub answers an error as ``{"message", "documentation_url", "status"}``, plus an ``errors``
+member on the kinds that carry one. That is not FastAPI's ``{"detail": …}``, and the difference is
+not cosmetic: PyGithub — the SDK ``examples/using-official-sdk/github.py`` drives — picks its
+exception CLASS off the body's ``message``, so the same wrong token raises
+``BadCredentialsException`` against real and a bare ``GithubException`` against a server answering
+``detail`` (measured against both). A client's error handling could not be written here.
 
-What this module carries is narrower: an envelope for an error the router builds deliberately and
-would otherwise have to emit in a shape real never sends — today, the 400 that refuses an
-unsupported ``X-GitHub-Api-Version``. The exception carries its own body as ``github_body``, the
-same way :mod:`backlot.errors.google` carries its extra fields on the exception, and anything
-without one falls through to FastAPI's default: ``http_body`` returning ``None`` is what
-:mod:`backlot.errors` reads as "this vendor has nothing to say about that one".
+``documentation_url`` is per-ENDPOINT and measured, not derived: requesting each route shape against
+a repository that does not exist returns that route's own docs anchor, which is where
+:data:`ROUTE_DOCS` comes from (every route below, none inferred from another).
 
-So ``owns`` claiming every ``/github`` path costs nothing today and is the right seam for the
-broader conversion when it happens.
+An authentication failure is the exception to that: real answers those with the bare
+``https://docs.github.com/rest``, measured on a bad bearer at
+``/repos/{owner}/{repo}/collaborators`` and at ``/user/repos``. Real is not uniform there and
+Backlot does not follow it all the way — ``/user/repos`` with NO Authorization header carries that
+route's anchor, while a bad token on the same route carries the root. Backlot answers the root for
+both 401s, which is this module's one knowing divergence.
 """
 
 from __future__ import annotations
 
+PREFIX = "/github"
+
+# What real sends when the failure is about the credential rather than the resource.
+DOCS_ROOT = "https://docs.github.com/rest"
+
+_DOCS = "https://docs.github.com/rest/"
+
+# Route template -> the anchor real names in an error on that route. Ordered as the router declares
+# them, and read first-match, because two templates can cover one path and FastAPI resolves those by
+# declaration order too (`issues/comments/{comment_id}` before `issues/{number}/comments`).
+# ``tests/test_github.py`` holds this table to the routes the app actually serves.
+ROUTE_DOCS: dict[str, str] = {
+    "/github/search/issues": "https://docs.github.com/v3/search",
+    "/github/search/code": "https://docs.github.com/v3/search",
+    "/github/orgs/{org}": _DOCS + "orgs/orgs#get-an-organization",
+    "/github/orgs/{org}/repos": _DOCS + "repos/repos#list-organization-repositories",
+    "/github/user/repos": _DOCS + "repos/repos#list-repositories-for-the-authenticated-user",
+    "/github/repos/{owner}/{repo}": _DOCS + "repos/repos#get-a-repository",
+    "/github/repos/{owner}/{repo}/issues": _DOCS + "issues/issues#list-repository-issues",
+    "/github/repos/{owner}/{repo}/issues/comments/{comment_id}": (
+        _DOCS + "issues/comments#get-an-issue-comment"
+    ),
+    "/github/repos/{owner}/{repo}/pulls/comments/{comment_id}": (
+        _DOCS + "pulls/comments#get-a-review-comment-for-a-pull-request"
+    ),
+    "/github/repos/{owner}/{repo}/issues/{number}": _DOCS + "issues/issues#get-an-issue",
+    "/github/repos/{owner}/{repo}/issues/{number}/comments": (
+        _DOCS + "issues/comments#list-issue-comments"
+    ),
+    "/github/repos/{owner}/{repo}/pulls": _DOCS + "pulls/pulls#list-pull-requests",
+    "/github/repos/{owner}/{repo}/pulls/{number}": _DOCS + "pulls/pulls#get-a-pull-request",
+    "/github/repos/{owner}/{repo}/pulls/{number}/reviews": (
+        _DOCS + "pulls/reviews#list-reviews-for-a-pull-request"
+    ),
+    "/github/repos/{owner}/{repo}/pulls/{number}/comments": (
+        _DOCS + "pulls/comments#list-review-comments-on-a-pull-request"
+    ),
+    "/github/repos/{owner}/{repo}/pulls/{number}/commits": (
+        _DOCS + "pulls/pulls#list-commits-on-a-pull-request"
+    ),
+    "/github/repos/{owner}/{repo}/statuses/{sha}": (
+        _DOCS + "commits/statuses#list-commit-statuses-for-a-reference"
+    ),
+    "/github/repos/{owner}/{repo}/pulls/{number}/files": (
+        _DOCS + "pulls/pulls#list-pull-requests-files"
+    ),
+    "/github/repos/{owner}/{repo}/git/ref/{ref}": _DOCS + "git/refs#get-a-reference",
+    "/github/repos/{owner}/{repo}/git/trees/{ref}": _DOCS + "git/trees#get-a-tree",
+    "/github/repos/{owner}/{repo}/contents": _DOCS + "repos/contents#get-repository-content",
+    "/github/repos/{owner}/{repo}/contents/{path}": (
+        _DOCS + "repos/contents#get-repository-content"
+    ),
+    "/github/repos/{owner}/{repo}/git/blobs/{sha}": _DOCS + "git/blobs#get-a-blob",
+    "/github/repos/{owner}/{repo}/branches": _DOCS + "branches/branches#list-branches",
+    "/github/repos/{owner}/{repo}/tags": _DOCS + "repos/repos#list-repository-tags",
+    "/github/repos/{owner}/{repo}/branches/{branch}": _DOCS + "branches/branches#get-a-branch",
+    "/github/repos/{owner}/{repo}/commits/{sha}": _DOCS + "commits/commits#get-a-commit",
+    "/github/repos/{owner}/{repo}/readme": _DOCS + "repos/contents#get-a-repository-readme",
+    "/github/repos/{owner}/{repo}/collaborators": (
+        _DOCS + "collaborators/collaborators#list-repository-collaborators"
+    ),
+    "/github/orgs/{org}/teams": _DOCS + "teams/teams#list-teams",
+    "/github/repos/{owner}/{repo}/teams": _DOCS + "repos/repos#list-repository-teams",
+}
+
+# The two routes whose final parameter is a path converter, so it takes every remaining segment:
+# `git/ref/{ref:path}` resolves a ref like `heads/release/2026-03`, and `contents/{path:path}` a
+# nested file. The OpenAPI spec this table is keyed on writes both as an ordinary `{name}`, which is
+# why they are named here instead of being inferred — letting any trailing placeholder run greedy
+# would hand `/repos/{owner}/{repo}` every deeper path that failed to match something longer.
+_TRAILING_PATH_ROUTES = frozenset(
+    {
+        "/github/repos/{owner}/{repo}/git/ref/{ref}",
+        "/github/repos/{owner}/{repo}/contents/{path}",
+    }
+)
+
 
 def owns(path: str) -> bool:
-    return path.startswith("/github")
+    return path.startswith(PREFIX)
 
 
-def http_body(path: str, exc) -> dict | None:
-    """The github-shaped body the router attached to ``exc``, or ``None`` for FastAPI's default."""
+def _matches(template: str, segments: list[str]) -> bool:
+    """Whether a concrete path's ``segments`` fit ``template``."""
+    parts = template.strip("/").split("/")
+    greedy = template in _TRAILING_PATH_ROUTES
+    if len(parts) > len(segments) or (not greedy and len(parts) != len(segments)):
+        return False
+    for i, part in enumerate(parts):
+        placeholder = part.startswith("{") and part.endswith("}")
+        if greedy and placeholder and i == len(parts) - 1:
+            return True  # swallows segments[i:]
+        if not placeholder and part != segments[i]:
+            return False
+    return len(parts) == len(segments)
+
+
+def docs_url(path: str) -> str:
+    """The anchor real names for the route ``path`` reaches, or the root for anything else.
+
+    First match wins, in the router's own declaration order, because two templates can cover one
+    path and FastAPI resolves those the same way.
+
+    The root is also the honest answer for a path that matches no route: real has no endpoint
+    documentation to point at for a URL it does not serve either.
+    """
+    segments = path.strip("/").split("/")
+    for template, url in ROUTE_DOCS.items():
+        if _matches(template, segments):
+            return url
+    return DOCS_ROOT
+
+
+def http_body(path: str, exc) -> dict:
+    """Render an exception into the envelope.
+
+    An error the router shaped by hand carries its own body as ``github_body`` — the version 400
+    and the search 422, both of which have an ``errors`` member no generic rendering could invent —
+    and that wins. Everything else is the three-member envelope over the exception's own detail,
+    whose wording is already real's ("Not Found", "Bad credentials") because the routers were
+    written against measured responses; only the shape around it was FastAPI's.
+    """
     body = getattr(exc, "github_body", None)
-    return body if isinstance(body, dict) else None
+    if isinstance(body, dict):
+        return body
+    detail = exc.detail
+    message = detail if isinstance(detail, str) else str(detail)
+    return {
+        "message": message,
+        "documentation_url": DOCS_ROOT if exc.status_code == 401 else docs_url(path),
+        "status": str(exc.status_code),
+    }
 
 
-def validation_body(errors) -> dict | None:
-    """Nothing to add: no github route answers a bad *parameter* through FastAPI's validator."""
-    return None
+def validation_body(errors) -> dict:
+    """A 422 from FastAPI's own request validation, in real's Validation Failed envelope.
+
+    Real reaches this case far less often than Backlot does — it refuses no pagination value at all
+    (measured: ``per_page=0``, ``per_page=abc``, ``page=0``, ``page=-1`` are each a 200
+    with the defaults applied), which is why the page parameters here no longer declare a floor and
+    clamp instead. What is left arriving at FastAPI's validator is a parameter whose TYPE will not
+    parse, and real answers those per route rather than uniformly — a non-integer issue number is
+    its 404, not a 422. So this is the closest real shape for a case real does not have, not a
+    measured reproduction of one: the envelope is the search 422's, whose ``errors`` entries name
+    the offending field.
+    """
+    entries = [
+        {
+            "resource": "Request",
+            "field": str(e.get("loc", ["?"])[-1]),
+            "code": "invalid",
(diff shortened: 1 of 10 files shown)


khj809 (commented):
Reviewed the diff, ran the suite in a clean worktree (`uv sync --all-extras`: **1589 passed, 1 skipped**, docker only; `ruff check`/`format --check` clean), and re-measured the claims against api.github.com rather than taking them from the body.

**What held up.** Of the 31 `ROUTE_DOCS` entries, the 25 reachable without a token match real exactly — `tags`, `branches`, `branches/{branch}`, `readme`, `contents[/{path}]`, `git/ref`, `git/trees`, `git/blobs`, `statuses`, `commits`, every `issues`/`pulls` shape, and both `orgs` entries. The `q`-less search 422 matches to the field: `resource: Search`, `field: q`, `code: missing`, `documentation_url: https://docs.github.com/v3/search`. `Authorization: Basic …` and a scheme-less value both answer 200 anonymously on a public repo, so classifying an unparseable header as *missing* rather than *wrong* is right. An unknown path is the bare root, as here. The six entries I could not reach unauthenticated (`collaborators`, `repos/../teams`, `orgs/../teams`, `user/repos`, the two searches) I did not verify.

#100's first block — the silent one — is closed, and that is the highest-value part of this change. No regression path in the 401 split: `_require`'s pre-check uses the same parser `resolve_bearer` does, so nothing that resolved before now fails.

Five points below. Only the 405 one has a clean self-contained suggestion; the others need an edit outside this diff, so they carry the sketch instead.

backlot/routers/github.py
```diff
@@ -307,13 +323,21 @@ async def search_issues(
     request: Request,
     response: Response,
     q: str = Query("", description="Issues/PRs search query"),
-    page: int | None = Query(None, ge=1),
-    per_page: int | None = Query(None, ge=1),
+    page: int | None = Query(None),
+    per_page: int | None = Query(None),
```
khj809: **`per_page=abc` / `page=abc` are still 422, so #100's fourth divergence is half-closed.**

The PR body lists `per_page=abc` under "Pagination values real absorbs are absorbed", but the annotation is still `int | None`, so FastAPI's validator answers before `clamp_page` is reached. Re-measured against api.github.com today:

```
real     /repos/python/cpython/issues?per_page=abc               -> 200, 30 items
real     /repos/python/cpython/issues?page=abc                   -> 200
real     /search/issues?q=test+repo:python/cpython&per_page=abc  -> 200
backlot  /github/.../issues?per_page=abc                         -> 422 Validation Failed
```

Closing it needs a tolerant parse, not a type change alone — `clamp_page` already absorbs `None`:

```python
# backlot/pagination.py
def _page_int(v: str | None) -> int | None:
    try:
        return int(v)
    except (TypeError, ValueError):
        return None  # real applies the default rather than refusing
```

with every `page`/`per_page` param declared `str | None` and coerced inside `clamp_page`.

If that is its own change, the fix here is to stop claiming it: drop `per_page=abc` from the absorbed list in the PR body, and name it in `validation_body`'s docstring — which today covers only the per-route-404 case, not the one you measured as a 200.

backlot/errors/github.py
```diff
@@ -1,35 +1,184 @@
...
-    """Nothing to add: no github route answers a bad *parameter* through FastAPI's validator."""
-    return None
+    Real reaches this case far less often than Backlot does — it refuses no pagination value at all
+    (measured: ``per_page=0``, ``per_page=abc``, ``page=0``, ``page=-1`` are each a 200
+    with the defaults applied), which is why the page parameters here no longer declare a floor and
+    clamp instead. What is left arriving at FastAPI's validator is a parameter whose TYPE will not
+    parse, and real answers those per route rather than uniformly — a non-integer issue number is
+    its 404, not a 422. So this is the closest real shape for a case real does not have, not a
+    measured reproduction of one: the envelope is the search 422's, whose ``errors`` entries name
+    the offending field.
```
khj809: **A non-integer path parameter now wears the envelope on a status real does not send.**

Measured: `/repos/python/cpython/issues/notanint` is `404 {"message": "Not Found", "documentation_url": "https://docs.github.com/rest/issues/issues#get-an-issue", "status": "404"}`. Backlot answers `422 Validation Failed` with `documentation_url: .../rest`.

Deferring the per-route conversion is a fair call. What changes here is that the 422 now *looks* measured: `{"detail": [...]}` announced itself as the mock's FastAPI default, and this envelope does not. Either convert the case (a validation error whose `loc[0] == "path"` -> 404 with `docs_url(path)`, already available in this module), or return `None` for it so it keeps `detail` until it is converted.

backlot/errors/github.py
```diff
@@ -1,35 +1,184 @@
...
-    return body if isinstance(body, dict) else None
+    if isinstance(body, dict):
+        return body
+    detail = exc.detail
+    message = detail if isinstance(detail, str) else str(detail)
+    return {
+        "message": message,
+        "documentation_url": DOCS_ROOT if exc.status_code == 401 else docs_url(path),
+        "status": str(exc.status_code),
+    }
```
khj809: **A 405 gets GitHub's envelope with Starlette's wording.**

`POST /github/repos/{owner}/{repo}` answers `{"message": "Method Not Allowed", "documentation_url": ".../repos/repos#get-a-repository", "status": "405"}`. `Method Not Allowed` is `http.HTTPStatus(405).phrase`, not a measured GitHub string, and the router declares GET only, so every wrong method lands here.

I could not measure real's authenticated 405 (no token). Unauthenticated `POST /repos/python/cpython` is real's `401 Requires authentication`, so it is not a 405 there.

Worth measuring. If that is out of scope, leaving it in `detail` is more honest than dressing an unmeasured phrase as real's:

```suggestion
def http_body(path: str, exc) -> dict | None:
    """Render an exception into the envelope.

    An error the router shaped by hand carries its own body as ``github_body`` — the version 400
    and the search 422, both of which have an ``errors`` member no generic rendering could invent —
    and that wins. Everything else is the three-member envelope over the exception's own detail,
    whose wording is already real's ("Not Found", "Bad credentials") because the routers were
    written against measured responses; only the shape around it was FastAPI's.

    A 405 is the exception: the router declares GET only, so a wrong method is refused by Starlette
    with ``http.HTTPStatus(405).phrase`` — a string no measurement attributes to real, which routes
    a wrong method per endpoint (an unauthenticated ``POST /repos/{owner}/{repo}`` is its 401). It
    keeps FastAPI's ``detail`` rather than being dressed as real's until real's answer is measured.
    """
    body = getattr(exc, "github_body", None)
    if isinstance(body, dict):
        return body
    if exc.status_code == 405:
        return None
    detail = exc.detail
    message = detail if isinstance(detail, str) else str(detail)
    return {
        "message": message,
        "documentation_url": DOCS_ROOT if exc.status_code == 401 else docs_url(path),
        "status": str(exc.status_code),
    }
```

backlot/errors/github.py
```diff
@@ -1,35 +1,184 @@
...
+            "message": e.get("msg", "invalid request"),
+        }
+        for e in errors
+    ]
+    return {
+        "message": "Validation Failed",
+        "documentation_url": DOCS_ROOT,
+        "errors": entries,
+        "status": "422",
+    }
```
khj809: **One search route answers two different `documentation_url`s for its own 422.**

```
/github/search/issues                      -> https://docs.github.com/v3/search   (hand-shaped)
/github/search/issues?q=test&per_page=abc  -> https://docs.github.com/rest        (here)
```

`errors.http_body` receives the path; `errors.validation_body` does not pass it through. Three one-liners:

```python
# backlot/errors/__init__.py
return envelope.validation_body(path, errors)
# backlot/errors/github.py
def validation_body(path: str, errors) -> dict:
    ...
    "documentation_url": docs_url(path),
# backlot/errors/atlassian.py, backlot/errors/google.py — same parameter, unused
```

backlot/errors/github.py
```diff
@@ -1,35 +1,184 @@
...
+# nested file. The OpenAPI spec this table is keyed on writes both as an ordinary `{name}`, which is
+# why they are named here instead of being inferred — letting any trailing placeholder run greedy
+# would hand `/repos/{owner}/{repo}` every deeper path that failed to match something longer.
+_TRAILING_PATH_ROUTES = frozenset(
+    {
+        "/github/repos/{owner}/{repo}/git/ref/{ref}",
+        "/github/repos/{owner}/{repo}/contents/{path}",
+    }
+)
+
```
khj809: **`_TRAILING_PATH_ROUTES` has no guard, so the failure `ROUTE_DOCS`'s test prevents can still happen one door over.**

`test_github_documentation_url_names_the_route_that_failed` catches a missing `ROUTE_DOCS` entry. But a route added with a `{name:path}` converter *and* a correct entry still resolves wrong: `_matches` requires an exact segment count for anything not in this set, so the route's deeper paths fall through to `DOCS_ROOT` — silently, which is what the test exists to prevent.

Starlette keeps the truth on the route, so it can be derived rather than restated (verified on this branch — it returns exactly these two):

```python
_TRAILING_PATH_ROUTES = frozenset(
    re.sub(r"\{(\w+):path\}", r"{\1}", r.path)
    for r in github_router.routes
    if any(isinstance(c, PathConvertor) for c in r.param_convertors.values())
)
```

`backlot.routers.github` does not import `backlot.errors`, so that import direction is clean. Alternatively key `ROUTE_DOCS` on `scope["endpoint"]`, which Starlette 1.6 does set (it does not set `scope["route"]`) — then `_matches`, this set, and the declaration-order caveat all go away.

nuri-yoo:
All five reproduce and all five are fixed in `50a287d`.

**Pagination is closed rather than dropped from the claim.** You are right that the body claimed more than the diff did — the `int` annotation answered before `clamp_page` was reached. It absorbs now, and without giving up the schema: `Annotated[int | None, BeforeValidator(...), Query()]` keeps the parameter an integer in `/openapi.json`, which is what real's own spec declares, so the tolerance sits in the runtime where real has it. `per_page=abc`, `page=abc`, `page=`, `per_page=0`, `page=-1` and a per_page over the cap are each a 200 with the defaults applied, matching what you measured.

**The path parameter is real's 404**, not a softened 422: `Not Found` with that route's anchor, which is byte-for-byte what you measured on `/issues/notanint`. That needed the status out of the envelope, so `validation_body` returns `(status, body)` and `main`'s handler honours it — Atlassian returns `(422, …)` unchanged, Google still returns `None`.

**405 taken as suggested**, with your reasoning in the docstring. It keeps `detail`, which announces the mock rather than dressing Starlette's phrase as measured.

**`validation_body` takes the path**, so the search route no longer answers two `documentation_url`s. The case is now hard to reach at all — the pagination parameters absorb, and every other query parameter on this surface is a `str` — so the test calls the function directly rather than pretending a route still reaches it.

**`_TRAILING_PATH_ROUTES` is gone entirely.** Rather than derive the set, `docs_url` asks Starlette which route the path reaches and looks the template up. That deletes the hand-rolled matcher, the converter set and the declaration-order caveat in one go: the three rules it restated — `git/ref` swallowing the rest of the path, `contents/{path}` likewise, `issues/comments/{id}` before `issues/{number}/comments` — are the router's own, and a restatement only has to be wrong once. (`scope["endpoint"]` would work too; keying the table on path templates keeps it readable and keeps the test that compares its keys to `/openapi.json`.)

Reverting each fix fails the test that covers it: `assert 422 == 200` on the pagination values, `assert 422 == 404` on the path parameter, the envelope against `{"detail": "Method Not Allowed"}`, and `.../rest` against `.../v3/search`.

1625 passed, 2 skipped; ruff and `gen_docs --check` clean. The body's absorbed list keeps `per_page=abc`, which the diff now earns.

backlot/errors/github.py
```diff
@@ -1,35 +1,199 @@
...
-def validation_body(errors) -> dict | None:
-    """Nothing to add: no github route answers a bad *parameter* through FastAPI's validator."""
-    return None
+    A QUERY parameter is the residue. Real refuses no pagination value at all, which is why those
+    now absorb (see :func:`backlot.pagination._absorb_page`) and never reach here; what is left is
+    a shape real has no measured answer for, so it keeps the Validation Failed envelope real uses
+    for a parameter it does refuse, with this route's anchor rather than the bare root.
+    """
+    where = errors[0].get("loc", ()) if errors else ()
+    if where and where[0] == "path":
```
khj809: `errors[0]` alone makes this an ordering dependency rather than a rule: a request failing on both a path and a query parameter answers 404 or 422 depending on which error FastAPI happens to report first.

Unreachable today — no `str`-typed exception is left among the query parameters on this surface, and the docstring says so — but `any(e.get("loc", ())[:1] == ("path",) for e in errors)` states the rule instead of relying on the order, at the same length.

Not blocking.

khj809 (approved):
Approving. Verified all five points in a worktree at `50a287d`, independently of the PR body.

The pagination values now absorb (`per_page=abc`, `page=abc`, `page=`, `0`, `-1`, `1.5`, over-cap — each a 200 identical to the default listing, on the repo routes and both searches) while `/openapi.json` keeps `integer`, which is the better half of that fix and reads as real's spec. The path-parameter 404 is byte-identical to what I measured on `/repos/python/cpython/issues/notanint`, including the route anchor. The 405 keeps `detail`. The search 422 names `v3/search`. And `docs_url` delegating to Starlette is further than I suggested — `contents/x/y/z.py`, `git/ref/heads/release/2026-03`, and `issues/comments/5` vs `issues/5/comments` all resolve correctly with the matcher gone.

Reverting each of the four code fixes from a copy breaks exactly one test each, so they are load-bearing as claimed. No regression in the 401 split, the 404 envelope, the `detail` fallback off the github surface, or the Atlassian envelope. 1626 passed, 1 skipped here; ruff and `gen_docs --check` clean.

One non-blocking note inline. Separately, the description now trails the diff in a few places — the pagination paragraph still credits `clamp_page` and the removed floor rather than `_absorb_page` and the kept integer schema, the path-parameter 404 is not mentioned at all, and the heading says one knowing divergence over two. Worth a pass before merge, but nothing that changes the code.

backlot/errors/github.py
nuri-yoo: Taken in `251a4eb` — `any(tuple(e.get("loc", ()))[:1] == ("path",) for e in errors)`, which states the rule rather than reading the first row of a list.

It stays unreachable, so the test calls `validation_body` directly with a path error and a query error in both orders: both answer 404, and a query error alone is still the 422. Against `errors[0]` the query-first ordering fails with `assert 422 == 404`.

The description is back in step too — the pagination paragraph credits the `BeforeValidator` and the kept integer schema rather than the removed floor, the path-parameter 404 has a paragraph of its own, and the heading counts the two divergences it lists.

Issue #103, opened 2026-09-01
Unrecognised S3 sub-resources are answered with an object listing instead of being refused


khj809:
_Found by `backlot diff --source s3`, the contract comparison added in #106._

Backlot serves S3 from four catch-all routes that read the query string themselves. A query string it does not recognise falls through to the object listing instead of being refused, so 42 S3 read operations are answered with the body of a different operation under a success status, with no error and no log line: 41 get `200 <ListBucketResult>`, and ListParts gets the object's own bytes.

This is the failure mode the project exists to prevent: a caller gets no signal at all, and parses an object listing as whatever it asked for.

## What a caller sees

Every one of these is the official `boto3` client against a stock Backlot. None of them raises:

```python
s3.get_bucket_versioning(Bucket=b)  # -> {}          real S3: {"Status": "Enabled"|"Suspended"}
s3.get_bucket_acl(Bucket=b)         # -> {}          real S3: {'Owner': ..., 'Grants': [...]}
s3.get_bucket_tagging(Bucket=b)     # -> {}          real S3: raises NoSuchTagSet when unset
s3.get_bucket_policy(Bucket=b)      # -> {'Policy': '<?xml ...><ListBucketResult>...'}
```

The last one is the clearest: botocore maps the whole XML listing into the `Policy` string, so `json.loads(resp["Policy"])` fails on XML — or the listing gets stored as a bucket policy. `get_bucket_tagging` is the quietest: real S3 raises `NoSuchTagSet` when a bucket has no tags, so a caller's error path simply never runs.

## Reproduce

```python
import backlot, boto3, httpx
from botocore.config import Config

with backlot.serve() as m:
    c = httpx.get(f"{m.base_url}/_meta/users").json()
    s3 = boto3.client("s3", endpoint_url=f"{m.base_url}/s3",
                      aws_access_key_id=c["admin_s3_access_key_id"],
                      aws_secret_access_key=c["admin_s3_secret_access_key"],
                      region_name="us-east-1", config=Config(s3={"addressing_style": "path"}))
    b = s3.list_buckets()["Buckets"][0]["Name"]
    print(s3.get_bucket_policy(Bucket=b))
```

## The fix is one decision, in one place

[...]

Issue #101, opened 2026-09-01
Linear's schema has drifted from the one Backlot serves, in 47 breaking ways


khj809:
_Found by `backlot diff --source linear`, the contract comparison added in #106._

`backlot/graphql/linear.graphql` disagrees with the schema `api.linear.app` answers introspection with, in 47 places where Backlot **contradicts** the vendor rather than merely lacking something it has. Code written against the mock compiles, passes, and then behaves differently against the real service, which is the failure this project exists to prevent. Surface Linear has that Backlot simply does not serve is out of scope here and is not listed.

Linear does not version its API, and Backlot's SDL was generated from the documents `@linear/sdk` v88 sends. The two have drifted since. Every line below is one measurement taken on 2026-09-01; re-measure before fixing, because the vendor may have moved again.

Introspect the live schema with a personal API key. Linear takes the key **bare** in the header — a `Bearer` prefix is answered 400, not 401:

```bash
curl -s https://api.linear.app/graphql -H "Authorization: $LINEAR_API_KEY" \
  -H 'Content-Type: application/json' \
  -d '{"query":"{__type(name:\"Issue\"){fields{name type{kind name ofType{kind name}}}}}"}'
```

### Date comparators that reject the duration form Linear accepts (12)

- [x] `DateComparator.eq` — Linear: `DateTimeOrDuration`, Backlot: `DateTime`
- [x] `DateComparator.gt` — Linear: `DateTimeOrDuration`, Backlot: `DateTime`
- [x] `DateComparator.gte` — Linear: `DateTimeOrDuration`, Backlot: `DateTime`
- [x] `DateComparator.lt` — Linear: `DateTimeOrDuration`, Backlot: `DateTime`
- [x] `DateComparator.lte` — Linear: `DateTimeOrDuration`, Backlot: `DateTime`
- [x] `DateComparator.neq` — Linear: `DateTimeOrDuration`, Backlot: `DateTime`
- [x] `NullableDateComparator.eq` — Linear: `DateTimeOrDuration`, Backlot: `DateTime`
- [x] `NullableDateComparator.gt` — Linear: `DateTimeOrDuration`, Backlot: `DateTime`
- [x] `NullableDateComparator.gte` — Linear: `DateTimeOrDuration`, Backlot: `DateTime`
- [x] `NullableDateComparator.lt` — Linear: `DateTimeOrDuration`, Backlot: `DateTime`
[...]

PR #106, opened 2026-09-01
Add `backlot diff` command to check contract divergences against each vendor

khj809:
`backlot diff --source <name>` compares what Backlot serves against what the vendor serves — in **both directions**, and over **arguments as well as fields**. A one-way, fields-only diff run against Linear once reported a clean schema while ten fields and four arguments were missing, all of them on the side it did not walk.

```bash
backlot diff --source slack                    # a published document is public: no credential
backlot diff --source linear                   # reads LINEAR_API_KEY from the environment
backlot diff --source notion --json            # one object on stdout and nothing else there
backlot diff --source jira --update-baseline   # accept the gaps it found
```

## How each vendor's contract is read

The class says how, so the registry carries no `kind` flag that picks a parser:

| Comparison | Sources | Read from | Credential |
|---|---|---|---|
| `GraphQLComparison` | Fireflies, Linear | live introspection | yes |
| `OpenAPIComparison` | Slack, GitHub, Jira, Confluence, Notion, HubSpot | the document the vendor publishes | no |
| `GoogleDiscoveryComparison` | Gmail, Google Drive | Google's Discovery document | no |
| `ProbeComparison` | Amazon S3 | a running server's own answers | no |

S3 is asked rather than read because it selects operations by **query string**, not by path: `?acl`, `?versioning` and ninety more sit at one path, and Backlot serves them from four catch-all routes. A path diff pairs every S3 operation with the same route and reports a clean match forever — a green check that means nothing. So the probe asks a running server instead, and looks for the failure no document can show: an operation neither implemented nor refused, answered 200 with another operation's body.

Comparisons are keyed by the canonical `source_type` from `store.SOURCE_TABLE`, and a test fails if the two sets ever drift apart. `--credential NAME=VALUE` repeats for as many as a comparison declares, and one it does not declare is refused rather than ignored.

## Exit codes

| Code | Meaning |
|---|---|
| `0` | nothing new |
| `1` | the contracts disagree |
| `2` | the contract could not be read |
| `3` | a credential this source declares is not set anywhere |

The split between 1 and 2 is the point. A vendor outage, a CDN challenge page answered 200, an expired key, and a server that will not answer all exit 2, so a scheduled run never files a fidelity bug for something that is not one. 3 is separate from both because it is this repository's misconfiguration rather than anyone's outage: reported as one and left green, the two sources whose contract is introspection would go uncompared night after night, so the workflow fails on it.

## The baseline

`backlot/fidelity/baseline/<source>.json` holds the divergences already read and accepted, so a run reports only what is **new**. Without it the first Linear run lists 782 gaps and nobody reads the output by the third run. Baselines ship inside the package, so an installed copy can be compared without the repository.

- `--update-baseline` acknowledges **gaps** freely and **never** a breaking finding. Acknowledging one of those is a hand edit carrying a note, so the reasoning lands in the diff rather than in a flag nobody kept.
- A gap is acknowledged by what it names. A breaking entry is acknowledged by what it names **and** by its `detail`, because there the detail *is* the contradiction: an entry reading `vendor: Int, Backlot: Float` says nothing about a vendor now serving `String`. Such an entry is reported again and left in the file exactly as written, until someone reads the new shape and rewrites the note.
- A run also reports **resolved** entries — acknowledged divergences the vendor no longer has, which is the baseline's own kind of drift.

## Every claim here is a measurement

Measured 2026-09-01, each recorded beside the code that depends on it.

- **Linear**'s personal API keys go **bare** in `Authorization`; a `Bearer` prefix is answered 400, not 401.
- **HubSpot** addresses each document by release id, and all three releases its index named still serve their own frozen document — so pinning the resolved URL never 404s, it reports no drift forever. The index is re-read on every run.
- **Slack**'s published spec documents one verb per method and, of the search methods, only `search.messages` — `search.all` and `search.files` are absent; measured against slack.com, all fourteen answer `200/ok` over both GET and POST.
- **GitHub**'s spec describes neither `contents` without a path nor the legacy per-sha statuses Backlot reads; both answer 200.
- **Atlassian**'s published Confluence v1 document still describes 65 reads, but no longer the content and space reads Backlot serves, which is exactly what the eight findings say.

That last group is the general lesson, and `docs/fidelity.md` says it plainly: introspection *is* the contract, but a published document only *describes* one. On a REST source `missing_*` is reliable and `extra_*` means *undocumented by the vendor — verify by hand*. Of the 28 `extra_*` findings so far, the 25 `extra_operation` ones were the document being incomplete rather than Backlot being wrong; the three `extra_param` on Jira are Backlot being more permissive than the vendor documents, and Notion's is a legacy path Backlot serves deliberately.

## What it found

- **#101** — Linear's schema has drifted from Backlot's in 47 breaking ways.
- **#103** — 42 unrecognised S3 sub-resources are answered with another operation's body instead of being refused.
- Merging main brought #95's branch and tag listings, and the comparison reported them the same hour: real GitHub paginates both with `page`/`per_page`, and neither of the new listings accepts them.

Those two sources exit 1 until they are fixed; the other nine exit 0.

## CI

`.github/workflows/fidelity.yml` runs every source daily and on `workflow_dispatch`, and **never on a pull request**. Drift is this project's bug, but it is never the bug of whichever pull request happens to be open when a vendor ships a change, so a divergence opens an issue and turns the scheduled run red rather than blocking a merge that has nothing to do with it.

A source has **one issue for the life of the repository**, opened under the `fidelity` label — the vendor API defines the right answer, so closing one needs a measurement against it. A match is found whether it is open or closed and reopened if closed, so triaging one does not produce a fresh duplicate the next morning; it is matched on the exact title and on being the workflow's own, so an issue a person opened under the same name is never rewritten from under them; its body carries the latest run's report and a comment is posted only when that report is not the one already there, so a year of identical comments cannot bury the morning it changed. Each source's job takes a concurrency group, because GitHub indexes issue search asynchronously and a `workflow_dispatch` racing the nightly run would otherwise file the issue twice.

## Verification

- **Tests** — `pytest` on a `.[dev,examples,mcp,llamaindex,mirage,fsspec]` install: 1689 passed, 1 skipped — 1690 collected. The one skip is `llama_index.readers.hubspot`, which is unrelated and pinned out; an install that has it reports 1690 passed and none skipped.
- **Optional surfaces that ran rather than skipped** — `tests/test_s3.py`, `tests/test_sdk.py`, `tests/test_mcp.py`, `tests/test_llamaindex.py`, and the reader, Google and mirage tests in `tests/test_integrations.py`.
- **Against the live vendors** — all eleven sources run against their real contracts: nine exit 0, and Linear (47 breaking, #101) and S3 (42 silent fallthroughs, #103) exit 1. Rewriting a baseline reproduces the committed file byte for byte apart from the measurement date. The comparison caught GitHub publishing two Actions runner reads on 2026-09-02 that its document did not carry the day before; both are acknowledged as gaps.
- **The issue-filing step** — driven end to end against a stubbed tracker over consecutive runs: one issue created and labelled `fidelity`, no comment on an unchanged report, an edit and one comment when the report changed, and a reopen rather than a second issue after the issue was closed.
- **Packaging** — built a wheel, installed it under `/tmp` and loaded all eleven baselines from a cwd the checkout cannot shadow.

- [ ] A new endpoint serving corpus content is ACL-scoped, proved for both an admin and a scoped token — n/a, this adds no endpoint

### DIFF
--- a/.github/workflows/fidelity.yml
+++ b/.github/workflows/fidelity.yml
@@ -10,13 +10,48 @@ on:
   schedule:
     - cron: "17 6 * * *"
   workflow_dispatch:
+    inputs:
+      source:
+        description: "One source to compare. Empty compares every one of them."
+        required: false
+        type: string
 
 permissions:
   contents: read
   issues: write
 
 jobs:
+  # The registry is the list, so adding a source does not also mean editing this file.
+  sources:
+    runs-on: ubuntu-latest
+    outputs:
+      names: ${{ steps.pick.outputs.names }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.13"
+          cache: pip
+          cache-dependency-path: pyproject.toml
+      - name: Install
+        run: pip install -e .
+      - name: Decide which sources to compare
+        id: pick
+        env:
+          CHOSEN: ${{ inputs.source }}
+        run: |
+          python - <<'PY' >> "$GITHUB_OUTPUT"
+          import json, os
+          from backlot.fidelity import COMPARISONS
+
+          chosen = os.environ.get("CHOSEN", "").strip()
+          if chosen and chosen not in COMPARISONS:
+              raise SystemExit(f"no comparison for {chosen!r}; one of {sorted(COMPARISONS)}")
+          print("names=" + json.dumps([chosen] if chosen else sorted(COMPARISONS)))
+          PY
+
   compare:
+    needs: sources
     runs-on: ubuntu-latest
     # Two runs of one source would both create the issue: GitHub indexes issue search asynchronously.
     concurrency:
@@ -25,18 +60,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        source:
-          - slack
-          - gmail
-          - google_drive
-          - github
-          - jira
-          - confluence
-          - hubspot
-          - notion
-          - fireflies
-          - linear
-          - s3
+        source: ${{ fromJSON(needs.sources.outputs.names) }}
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
--- a/.github/workflows/fidelity.yml
+++ b/.github/workflows/fidelity.yml
@@ -26,7 +26,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        source: [slack, gmail, google_drive, github, jira, confluence, hubspot, notion, s3]
+        source: [slack, gmail, google_drive, github, jira, confluence, hubspot, notion]
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
@@ -39,13 +39,16 @@ jobs:
       - name: Compare ${{ matrix.source }} against the contract its vendor publishes
         run: backlot diff --source ${{ matrix.source }}
 
-  schema-diff:
+  # Reported, not gated. Two of these need a credential a fork cannot read; s3 is here instead
+  # because it is red for a filed reason (#103) and a check that is always red stops being read.
+  # Move it back up once that is closed.
+  reported:
     if: github.event_name != 'pull_request'
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        source: [fireflies, linear]
+        source: [fireflies, linear, s3]
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
--- a/.github/workflows/fidelity.yml
+++ b/.github/workflows/fidelity.yml
@@ -118,8 +118,10 @@ jobs:
           # characters and jira's 69k — so the report is capped well under the limit.
           report=$(cat report.txt)
           if [ "${#report}" -gt 55000 ]; then
+            # Sliced, not `head -c`: the test above counts characters, and cutting bytes leaves
+            # half of a multibyte character at the seam — the report is full of `·` and `…`.
             report="$(printf '%s\n… truncated at 55000 of %d characters. The whole report is in the run log: %s' \
-              "$(printf '%s' "$report" | head -c 55000)" "${#report}" "$RUN_URL")"
+              "${report:0:55000}" "${#report}" "$RUN_URL")"
           fi
           body=$(printf '%s\n\n```\n%s\n```\n' \
             "\`backlot diff --source $SOURCE\` reports divergences the baseline does not acknowledge. Each is either a bug in Backlot or a deliberate gap; a gap gets a note and an entry via \`backlot diff --source $SOURCE --update-baseline\`. This body is the latest scheduled run's report, rewritten in place." \
--- a/.github/workflows/fidelity.yml
+++ b/.github/workflows/fidelity.yml
@@ -1,16 +1,12 @@
 # Does the vendor still serve what Backlot claims it does?
 #
-# Two jobs, because the two halves have different costs. The REST sources are compared against
-# contracts their vendors PUBLISH — no credential, no quota, readable from a fork — so that half
-# gates a pull request like any other check. The GraphQL sources need a key, which a fork cannot
-# read and which a vendor outage can take away, so that half runs on a schedule and opens an issue
-# instead of failing a merge that has nothing to do with it.
+# On a schedule, never on a pull request. Drift is this project's bug, but it is never the bug of
+# whichever pull request happens to be open when a vendor ships a change — so a divergence opens an
+# issue rather than failing someone else's merge. Two sources need a credential a fork could not
+# read anyway.
 name: Fidelity
 
 on:
-  pull_request:
-  push:
-    branches: [main]
   schedule:
     - cron: "17 6 * * *"
   workflow_dispatch:
@@ -20,35 +16,23 @@ permissions:
   issues: write
 
 jobs:
-  # No secrets, so this one runs on every pull request, forks included.
-  published-specs:
+  compare:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        source: [slack, gmail, google_drive, github, jira, confluence, hubspot, notion]
-    steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-python@v5
-        with:
-          python-version: "3.13"
-          cache: pip
-          cache-dependency-path: pyproject.toml
-      - name: Install
-        run: pip install -e .
-      - name: Compare ${{ matrix.source }} against the contract its vendor publishes
-        run: backlot diff --source ${{ matrix.source }}
-
-  # Reported, not gated. Two of these need a credential a fork cannot read; s3 is here instead
-  # because it is red for a filed reason (#103) and a check that is always red stops being read.
-  # Move it back up once that is closed.
-  reported:
-    if: github.event_name != 'pull_request'
-    runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        source: [fireflies, linear, s3]
+        source:
+          - slack
+          - gmail
+          - google_drive
+          - github
+          - jira
+          - confluence
+          - hubspot
+          - notion
+          - fireflies
+          - linear
+          - s3
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
@@ -70,11 +54,11 @@ jobs:
           echo "status=${PIPESTATUS[0]}" >> "$GITHUB_OUTPUT"
 
       # Exit 2 is "we could not ask" — an expired key, a vendor outage, a network blip. That is not
-      # a fidelity finding and must not open a bug against the mock.
-      - name: Report an unreadable vendor schema
+      # a fidelity finding and must not open a bug against Backlot.
+      - name: Report an unreadable vendor contract
         if: steps.diff.outputs.status == '2'
         run: |
-          echo "::warning::${{ matrix.source }}'s schema could not be read; no comparison ran."
+          echo "::warning::${{ matrix.source }}'s contract could not be read; no comparison ran."
           cat report.txt
 
       - name: Open an issue for a new divergence
@@ -86,7 +70,7 @@ jobs:
           title="Fidelity: $SOURCE diverges from its vendor"
           existing=$(gh issue list --state open --search "$title in:title" --json number --jq '.[0].number')
           body=$(printf '%s\n\n```\n%s\n```\n' \
-            "\`backlot diff --source $SOURCE\` reported divergences that the baseline does not acknowledge. Each one is either a bug in the mock or a deliberate gap; a deliberate gap gets a note and an entry via \`backlot diff --source $SOURCE --update-baseline\`." \
+            "\`backlot diff --source $SOURCE\` reported divergences the baseline does not acknowledge. Each is either a bug in Backlot or a deliberate gap; a gap gets a note and an entry via \`backlot diff --source $SOURCE --update-baseline\`." \
             "$(cat report.txt)")
           if [ -n "$existing" ]; then
             gh issue comment "$existing" --body "$body"
--- a/.github/workflows/fidelity.yml
+++ b/.github/workflows/fidelity.yml
@@ -18,6 +18,15 @@ permissions:
 jobs:
   compare:
     runs-on: ubuntu-latest
+    # One run per source at a time. A `workflow_dispatch` fired while the nightly run is mid-
+    # comparison would have both halves search for an issue neither has created yet and create it
+    # twice — and GitHub indexes issue search asynchronously, so the second would not see the first
+    # even after it existed. Per job rather than per workflow, because `matrix` is only in context
+    # here; queued rather than cancelled, because a comparison that was going to report something
+    # should finish and report it.
+    concurrency:
+      group: fidelity-${{ matrix.source }}
+      cancel-in-progress: false
     strategy:
       fail-fast: false
       matrix:
@@ -61,21 +70,41 @@ jobs:
           echo "::warning::${{ matrix.source }}'s contract could not be read; no comparison ran."
           cat report.txt
 
-      - name: Open an issue for a new divergence
+      # One issue per source, for the life of the repository, carrying the latest run's report in
+      # its body. Searching open issues alone files a fresh duplicate the morning after someone
+      # triages and closes one, and commenting unconditionally buries the morning a report changed
+      # under a year of identical ones — both of which are the alarm nobody reads by the third run,
+      # arriving through the tracker instead of the terminal.
+      - name: Report the divergence on this source's issue
         if: steps.diff.outputs.status == '1'
         env:
           GH_TOKEN: ${{ github.token }}
           SOURCE: ${{ matrix.source }}
         run: |
-          title="Fidelity: $SOURCE diverges from its vendor"
-          existing=$(gh issue list --state open --search "$title in:title" --json number --jq '.[0].number')
+          TITLE="Fidelity: $SOURCE diverges from its vendor"
+          export TITLE
+          # Matched on the exact title AND on the issue being this workflow's own. `in:title` is a
+          # word search, so it can answer with something that merely shares the words, and an issue
+          # a person opened under this name must never have its body rewritten from under them.
+          existing=$(gh issue list --state all --search "$TITLE in:title" \
+            --json number,title,author,state \
+            --jq 'first(.[] | select(.title == env.TITLE and .author.is_bot)) // empty')
           body=$(printf '%s\n\n```\n%s\n```\n' \
-            "\`backlot diff --source $SOURCE\` reported divergences the baseline does not acknowledge. Each is either a bug in Backlot or a deliberate gap; a gap gets a note and an entry via \`backlot diff --source $SOURCE --update-baseline\`." \
+            "\`backlot diff --source $SOURCE\` reports divergences the baseline does not acknowledge. Each is either a bug in Backlot or a deliberate gap; a gap gets a note and an entry via \`backlot diff --source $SOURCE --update-baseline\`. This body is the latest scheduled run's report, rewritten in place." \
             "$(cat report.txt)")
-          if [ -n "$existing" ]; then
-            gh issue comment "$existing" --body "$body"
-          else
-            gh issue create --title "$title" --body "$body"
+          if [ -z "$existing" ]; then
+            gh issue create --title "$TITLE" --body "$body"
+            exit 0
+          fi
+          number=$(printf '%s' "$existing" | jq -r .number)
+          if [ "$(printf '%s' "$existing" | jq -r .state)" = "CLOSED" ]; then
+            gh issue reopen "$number" --comment "Still diverging on the scheduled run, so this is open again."
+          fi
+          # A comment only when the report is not the one already on the issue: reopening and a
+          # changed report are the two things worth a notification.
+          if [ "$(gh issue view "$number" --json body --jq .body)" != "$body" ]; then
+            gh issue edit "$number" --body "$body"
+            gh issue comment "$number" --body "The divergences on \`$SOURCE\` changed. The issue body now carries the latest report."
           fi
 
       - name: Fail the run so the schedule shows red
--- a/.github/workflows/fidelity.yml
+++ b/.github/workflows/fidelity.yml
@@ -65,11 +65,21 @@ jobs:
           echo "::warning::${{ matrix.source }}'s contract could not be read; no comparison ran."
           cat report.txt
 
+      # Exit 3 is a credential nobody set. Red, because it is this repository's misconfiguration:
+      # warned about and left green, the two sources whose contract is introspection would go
+      # uncompared night after night while the run said nothing was wrong.
+      - name: Fail on a credential this repository has not set
+        if: steps.diff.outputs.status == '3'
+        run: |
+          cat report.txt
+          exit 1
+
       - name: Report the divergence on this source's issue
         if: steps.diff.outputs.status == '1'
         env:
           GH_TOKEN: ${{ github.token }}
           SOURCE: ${{ matrix.source }}
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
         run: |
           TITLE="Fidelity: $SOURCE diverges from its vendor"
           export TITLE
@@ -77,9 +87,19 @@ jobs:
           existing=$(gh issue list --state all --search "$TITLE in:title" \
             --json number,title,author,state \
             --jq 'first(.[] | select(.title == env.TITLE and .author.is_bot)) // empty')
+          # The API refuses a body over 65536 characters, and this step runs under `bash -e`, so
+          # that refusal takes the step down: the job would go red with no issue filed and the
+          # report only in the run log. A vendor restructuring its document turns a whole baseline
+          # into new findings at once — measured against an empty baseline, github's report is 135k
+          # characters and jira's 69k — so the report is capped well under the limit.
+          report=$(cat report.txt)
+          if [ "${#report}" -gt 55000 ]; then
+            report="$(printf '%s\n… truncated at 55000 of %d characters. The whole report is in the run log: %s' \
+              "$(printf '%s' "$report" | head -c 55000)" "${#report}" "$RUN_URL")"
+          fi
           body=$(printf '%s\n\n```\n%s\n```\n' \
             "\`backlot diff --source $SOURCE\` reports divergences the baseline does not acknowledge. Each is either a bug in Backlot or a deliberate gap; a gap gets a note and an entry via \`backlot diff --source $SOURCE --update-baseline\`. This body is the latest scheduled run's report, rewritten in place." \
-            "$(cat report.txt)")
+            "$report")
           if [ -z "$existing" ]; then
             gh issue create --title "$TITLE" --label fidelity --body "$body"
             exit 0
--- a/.github/workflows/fidelity.yml
+++ b/.github/workflows/fidelity.yml
@@ -18,12 +18,7 @@ permissions:
 jobs:
   compare:
     runs-on: ubuntu-latest
-    # One run per source at a time. A `workflow_dispatch` fired while the nightly run is mid-
-    # comparison would have both halves search for an issue neither has created yet and create it
-    # twice — and GitHub indexes issue search asynchronously, so the second would not see the first
-    # even after it existed. Per job rather than per workflow, because `matrix` is only in context
-    # here; queued rather than cancelled, because a comparison that was going to report something
-    # should finish and report it.
+    # Two runs of one source would both create the issue: GitHub indexes issue search asynchronously.
     concurrency:
       group: fidelity-${{ matrix.source }}
       cancel-in-progress: false
@@ -70,11 +65,6 @@ jobs:
           echo "::warning::${{ matrix.source }}'s contract could not be read; no comparison ran."
           cat report.txt
 
-      # One issue per source, for the life of the repository, carrying the latest run's report in
-      # its body. Searching open issues alone files a fresh duplicate the morning after someone
-      # triages and closes one, and commenting unconditionally buries the morning a report changed
-      # under a year of identical ones — both of which are the alarm nobody reads by the third run,
-      # arriving through the tracker instead of the terminal.
       - name: Report the divergence on this source's issue
         if: steps.diff.outputs.status == '1'
         env:
@@ -83,9 +73,7 @@ jobs:
         run: |
           TITLE="Fidelity: $SOURCE diverges from its vendor"
           export TITLE
-          # Matched on the exact title AND on the issue being this workflow's own. `in:title` is a
-          # word search, so it can answer with something that merely shares the words, and an issue
-          # a person opened under this name must never have its body rewritten from under them.
+          # `in:title` is a word search, so the title is matched exactly; `is_bot` leaves a person's own issue alone.
           existing=$(gh issue list --state all --search "$TITLE in:title" \
             --json number,title,author,state \
             --jq 'first(.[] | select(.title == env.TITLE and .author.is_bot)) // empty')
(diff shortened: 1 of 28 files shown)


nuri-yoo (commented):
Reviewed at db5f05a. Ran the suite on an all-extras install (1685 passed, 0 skipped), `ruff check` and `ruff format --check` clean, and all eleven sources against their real contracts: nine exit 0 against their committed baselines, linear exits 1 with 782 acknowledged and the 47 breaking findings of #101 (all 47 paths appear in that issue), and s3 exits 1 with 42/1/41, as the description says. `--update-baseline` reproduces slack, notion and hubspot byte for byte apart from `measured`. A base-only `pip install -e .`, which is what the workflow does, imports the package and runs the notion and s3 comparisons, and a wheel installed into a fresh environment loads all eleven baselines from `/`. Canonical path collisions: none, on either side, for any source.

The workflow's issue logic checks out against live GitHub: `gh issue list --json author` reports `is_bot: true` for a `github-actions[bot]` author, the `in:title` search matches a title containing a colon (tried with #100), and an API-created body round-trips with LF, so the comment-only-when-the-report-changed check holds. Driven end to end against a stubbed `gh` with the real `jq` expressions, the step creates once under the `fidelity` label, leaves a person's issue of the same title alone, posts nothing on an unchanged report, edits and comments once on a changed one, and reopens rather than duplicates after a close. The measured vendor facts hold too: HubSpot's index names three Custom Objects releases and all three still serve (9, 11 and 13 paths); GitHub's document describes neither `contents` without a path nor `statuses/{sha}` and both answer 200; Linear answers a bare key 200 and a `Bearer`-prefixed one 400 with a message saying to drop the prefix; Fireflies refuses a bare key with `Authorization header is missing the 'Bearer' prefix`.

Two findings are about behaviour and I would fix them before merging. `fetch_json` lets a 200 carrying non-object JSON escape as a traceback with exit 1, the status that files an issue (comment on fetch.py). The S3 probe never asks ListParts, because it reads the selector only from `requestUri` and botocore puts that one in a required querystring member; the server answers it with the object body, which is the silent fallthrough the probe exists to catch (comment on s3_probe.py). One coverage gap: union members and interface fields are never compared (comment on graphql_diff.py; Linear's one union matches today). The rest is prose that outlived a change of approach: the registry module docstring still describes a GraphQL-only file, two docstrings promise pull-request gating the workflow deliberately refuses, and one `:class:` reference names a class that never existed.

Three sentences are wider than the data under them, in the description and in the notes that repeat them. "All 25 such findings so far were the document being incomplete" follows a sentence about `extra_*`, but 25 is the count of `extra_operation` alone; `extra_*` is 28 once Jira's three `extra_param` entries are in (docs/fidelity.md uses the narrower `extra_operation` and is right), and the Notion entry's note describes a deliberate choice to keep serving the legacy path rather than an incomplete document. "Atlassian's published Confluence v1 document no longer describes any read at all" (also the eight confluence.json notes and docs/fidelity.md:81): the document describes 65 GET operations, among them `content/search`, `content/{id}/descendant`, `group`, `label` and `user/current`; what it no longer describes are the eight content and space reads Backlot serves, which is exactly what the findings say. "Slack's published spec omits the search methods" (also the fourteen slack.json notes and docs/fidelity.md:77): `/search.messages` is in the spec as GET, and the baseline already treats it that way (`POST /search.messages` is the extra operation, `GET /search.messages?highlight` a missing parameter); it is `search.all` and `search.files` that are absent.

One operational note on the workflow: the repository has no `LINEAR_API_KEY` or `FIREFLIES_API_KEY` secret yet (only `BACKLOT_PYPI_TOKEN`). Until they are set, the two GraphQL jobs exit 2 every night, which the workflow reports as a warning and leaves green, so the two sources whose contract is introspection are the two that never get compared. Worth setting before the first scheduled run, or making a declared credential that is absent fail the job, since that is a repository misconfiguration rather than a vendor outage.

backlot/fidelity/fetch.py
```diff
@@ -0,0 +1,25 @@
...
+
+def fetch_json(url: str, *, timeout: float = 120.0) -> dict:
+    try:
+        response = httpx.get(url, timeout=timeout, follow_redirects=True)
+    except httpx.HTTPError as e:
+        raise FidelityError(f"{url} unreachable: {e}") from e
+    if response.status_code != 200:
+        raise FidelityError(f"{url} answered {response.status_code}")
+    try:
+        return response.json()
```
nuri-yoo: A 200 carrying valid JSON that is not an object escapes here as `AttributeError` and reaches `backlot diff` as a traceback with exit 1, the status the workflow reads as "the schemas disagree". Measured by stubbing `httpx.get` to answer 200 with `["not", "an", "object"]`, `"just a string"` and `42` for notion: all three exit 1 (`AttributeError: 'list' object has no attribute 'get'`, raised in `from_openapi`). The same stub answering `text/html` exits 2 as intended, so the hole is exactly the valid-JSON, non-object case.

`real_schema` in graphql_diff.py guards this with `isinstance(body, dict)` and a comment saying why; the nine sources that read through `fetch_json` have no equivalent. Checking that the parsed value is a dict here, and raising `FidelityError` otherwise, closes it for all of them.

backlot/fidelity/s3_probe.py
```diff
@@ -0,0 +1,243 @@
...
+    rather than a divergence worth listing ninety times.
+    """
+    out = []
+    for name, op in (model.get("operations") or {}).items():
+        http = op.get("http") or {}
+        method = (http.get("method") or "").upper()
+        if method not in ("GET", "HEAD"):
+            continue
+        uri = http.get("requestUri") or "/"
+        path, _, query = urlsplit(uri).path, *uri.partition("?")[1:]
```
nuri-yoo, line 109: The selector is read only from the `?` in `requestUri`, so an operation botocore selects with a querystring member comes out as its bare form and is skipped at `if not op.query`. ListParts is the concrete case. The model declares it with `requestUri: /{Bucket}/{Key+}`, the same as GetObject, `required: [Bucket, Key, UploadId]`, and `UploadId` at `location: querystring, locationName: uploadId`. `operations()` yields `ListParts.query == ''` and `probe()` never asks.

Asking by hand against `backlot.serve()`: `GET /{bucket}/{key}?uploadId=abc123` answers 200 with the object body, byte for byte what the bare GET answers. That is the `silent_fallthrough` shape this module exists to catch, and it is absent from the 42 findings.

Of the 6 read operations whose `requestUri` carries no `?` but whose input shape has querystring members, ListParts is the only one where that member is required; GetObject's `versionId` and ListObjects' `prefix` are optional refinements of the bare operation. Adding the input shape's required querystring members to the selector covers ListParts without dragging those in.

backlot/fidelity/comparisons.py
```diff
@@ -0,0 +1,322 @@
+"""Where each source's two contracts come from: Backlot's own, and the vendor's published one.
+
+Backlot's side reads the SDL file the server itself builds from — ``backlot.graphql.engine.from_sdl``
+loads that same path at import — so the comparison is against what is served, without starting a
+server or importing a corpus.
+
+The vendor side needs a credential. That is the whole reason this runs on a schedule and not on
+every pull request: a fork cannot see the token, and a vendor outage must not block a contributor.
+The no-credential half of drift detection — vendor-published OpenAPI and Google Discovery documents
+— is a separate surface and does not belong to this module.
```
nuri-yoo: This paragraph describes the file from when it held only the GraphQL registry, and each sentence is now false for most of what it contains. Backlot's side is the SDL for the two GraphQL sources only: the eight document sources read `app.openapi()` (openapi_diff.py:75, google_discovery_diff.py:85), and the probe starts a server through `backlot.serve()` (s3_probe.py:242), so "without starting a server" does not hold either. Nine of the eleven entries need no credential, which docs/fidelity.md says in those words. And the OpenAPI and Google Discovery registries are defined below in this same module, so the surface this says "does not belong to this module" is most of the module.

backlot/fidelity/comparisons.py
```diff
@@ -0,0 +1,322 @@
...
+    name: str
+    env: str
+    what: str
+
+
+@dataclass(frozen=True)
+class OpenAPIComparison:
+    """A source compared against the OpenAPI document its vendor publishes.
+
+    Public, which is what lets this kind gate a pull request: no credential, no quota, no account.
```
nuri-yoo: The workflow never runs on a pull request (`on:` is `schedule` and `workflow_dispatch` only) and the description argues that it must not, so this class gates none. What the public document actually buys is no credential in the scheduled job and a run anyone can reproduce locally; suggest saying that instead.

backlot/fidelity/openapi_diff.py
```diff
@@ -0,0 +1,76 @@
+"""Comparing Backlot against an OpenAPI document its vendor publishes.
+
+Six vendors publish one, all of them public, which is what lets this run with no credential, no
+quota and no account — and on a fork's pull request. Where a document is found is not this
```
nuri-yoo: Same as the class docstring in comparisons.py: nothing here runs on a fork's pull request, by design.

backlot/fidelity/comparisons.py
```diff
@@ -0,0 +1,322 @@
...
+        self, credentials: Mapping[str, str] | None = None, *, timeout: float = 120.0
+    ) -> list[Finding]:
+        return google_discovery_diff.divergences(self, timeout=timeout)
+
+
+@dataclass(frozen=True)
+class GraphQLComparison:
+    """A source compared against the schema its vendor answers introspection with.
+
+    Introspection IS the contract, which is what separates this from the published documents the
```
nuri-yoo: The sentence is doubled: "separates this from the published documents the published documents are compared against".

backlot/fidelity/comparisons.py
```diff
@@ -0,0 +1,322 @@
...
+@dataclass(frozen=True)
+class ProbeComparison:
+    """A source whose contract cannot be read off a document, so it is compared behaviourally.
+
+    S3 dispatches on the query string rather than on the path, which no path diff can see. The
+    model is still the vendor's own — botocore's service definition, the same file every AWS SDK
+    is generated from — but what it is compared against is a running Backlot's answers.
+
+    Note that AWS publishes a machine-readable model for S3 too, and this is still not a
+    :class:`SpecComparison`. What separates the two is not whether a document exists but whether it
```
nuri-yoo: `SpecComparison` does not exist. The name appears only on this line; the commit that introduced it (92bf0cd) already had the current five classes, and `git log -S'class SpecComparison'` across all refs finds nothing. "The other class" in the next sentence is also ambiguous now that two classes compare documents; the point being made is about path-shaped contracts, so it is presumably `OpenAPIComparison` and `GoogleDiscoveryComparison` together.

backlot/fidelity/comparisons.py
```diff
@@ -0,0 +1,322 @@
...
+        return self.spec_url
+
+    # None: a probe asks Backlot, not the vendor, and signs with the corpus's own keys. A probe
+    # against a live vendor would declare what it needs here, and nothing else would change.
+    credentials: tuple[Credential, ...] = ()
+
+    def divergences(
+        self, credentials: Mapping[str, str] | None = None, *, timeout: float = 120.0
+    ) -> list[Finding]:
+        return s3_probe.divergences(self, timeout=timeout)
```
nuri-yoo, line 169: Every `ProbeComparison` goes through `s3_probe.divergences`, which fetches the model and unconditionally starts a server before calling `run`. So the `run` hook is per-source but the orchestration is S3's, and the docstring on `s3_probe.divergences` ("a second probe source may need ... no server at all") describes a freedom this dispatch does not give. `test_a_probe_declares_its_prober_and_the_dispatcher_only_hands_over` shows the coupling: a probe named `x` has to patch `s3_probe.fetch_json` and receives a `base_url` regardless. Either the class carries its `divergences` callable alongside `run`, or the docstring should say that the server start is shared by every probe.

backlot/fidelity/graphql_diff.py
```diff
@@ -0,0 +1,267 @@
...
+                Finding("extra_type", BREAKING, name, "the vendor's schema declares no such type")
+            )
+            continue
+        if is_enum_type(backlot_t) and is_enum_type(real_t):
+            out.extend(_diff_enum(backlot_t, real_t, name))
+        elif (is_object_type(backlot_t) and is_object_type(real_t)) or (
+            is_input_object_type(backlot_t) and is_input_object_type(real_t)
+        ):
+            out.extend(_diff_object(backlot_t, real_t, name))
+        elif type(backlot_t) is not type(real_t):
```
nuri-yoo, line 213: Two unions, or two interfaces, with the same name fall through every branch above and land here with `type(backlot_t) is type(real_t)`, so nothing about them is compared. Measured: `union M = A | B | C` on Backlot's side against `union M = A | B` on the vendor's reports `[]`, and an interface carrying an invented field reports only the implementing object's `A.invented`, never `N.invented`.

`linear.graphql` declares one union, `ExternalEntityInfoMetadata`. Checked against Linear's live schema with a real key: its three members match today, so nothing is being missed right now; what the gap costs is that a member added or dropped on either side would go unreported. The interface case is latent (Linear's schema has eight interfaces, Backlot's SDL declares none), but the comment at line 130 reads as if interface fields were walked. Diffing `.types` for unions, and sending interfaces through `_diff_object` with the `is_object_type` guard at line 131 widened to them, would close both.

backlot/fidelity/s3_probe.py
```diff
@@ -0,0 +1,243 @@
...
+What can be asked instead is what the server actually answers. An operation Backlot does not
+implement should be REFUSED. The failure this looks for is the third possibility: answering it with
+some other operation's body. A caller asking for a bucket's versioning configuration and receiving
+an object listing under a 200 gets no error, no log line, and a parse that quietly produces
+nonsense — which is the exact failure this project exists to prevent, and it cannot be seen in a
+document.
+
+An operation is judged indistinguishable when its response has the same status and the same XML
+root element as the same path carrying no query at all. Two operations legitimately answer that
+way — ListObjects and ListObjectsV2 really are what a bare bucket GET means — so those are
```
nuri-yoo: Only one of the two is in the baseline. `s3.json` acknowledges `ListObjectsV2` alone; ListObjects has no query, so `probe()` skips it at line 162 and it is never a finding to acknowledge. Suggest: ListObjectsV2 is acknowledged in the baseline, and ListObjects is the bare form itself and is never asked.

backlot/fidelity/s3_probe.py
```diff
@@ -0,0 +1,243 @@
...
+    # Discovered through the API being probed, rather than out of the corpus: a bucket the probe
+    # cannot reach as a client is a bucket the probe cannot ask questions about either.
+    buckets = re.findall(r"<Name>([^<]+)</Name>", read("/"))
+    if not buckets:
+        raise FidelityError("the corpus serves no S3 bucket, so there is nothing to probe")
+    bucket = buckets[0]
+    keys = re.findall(r"<Key>([^<]+)</Key>", read(f"/{bucket}", "list-type=2"))
+    if not keys:
+        raise FidelityError(f"bucket {bucket!r} holds no object, so the object probes cannot run")
+    return probe(base_url, access_key, secret, operations(model), bucket=bucket, key=keys[0])
```
nuri-yoo: `timeout` is applied to the three set-up reads above but not passed here, so the 46 requests `probe()` makes run on its own default of 20 seconds whatever the caller asked for. Measured with `run(..., timeout=120.0)` and spies on `httpx.get` and `httpx.request`: `get timeout=120.0 x3`, `request timeout=20.0` for every probe. The failing direction is the safe one (a slow server raises `FidelityError` and exits 2), but the signature promises a value the requests that matter never see. `timeout=timeout` here fixes it.

backlot/cli.py
```diff
@@ -508,6 +592,173 @@ def status(data_dir: DataDir = None) -> None:
...
+        bool,
+        typer.Option(
+            "--json",
+            help="print the result as JSON on stdout and nothing else, for piping into something",
+        ),
+    ] = False,
+) -> None:
+    """Compare the schema Backlot serves against the vendor's own, in both directions.
+
+    Fidelity is measured, never assumed — and until this command a measurement only happened when someone went looking. Run on a schedule, it is what notices the vendor changing something in March. Divergences already read and accepted live in `fidelity/<source>.json`, so a run reports what is NEW; accepting one is a file change that goes through review. Exits 1 when anything is unacknowledged, 2 when the vendor's schema could not be read at all — a vendor outage is not a fidelity finding.
```
nuri-yoo: `--help` says the baseline lives in `fidelity/<source>.json`; it is `backlot/fidelity/baseline/<source>.json` (`baseline_path`; docs/fidelity.md has it right).

.github/workflows/fidelity.yml
```diff
@@ -0,0 +1,98 @@
...
+          export TITLE
+          # `in:title` is a word search, so the title is matched exactly; `is_bot` leaves a person's own issue alone.
+          existing=$(gh issue list --state all --search "$TITLE in:title" \
+            --json number,title,author,state \
+            --jq 'first(.[] | select(.title == env.TITLE and .author.is_bot)) // empty')
+          body=$(printf '%s\n\n```\n%s\n```\n' \
+            "\`backlot diff --source $SOURCE\` reports divergences the baseline does not acknowledge. Each is either a bug in Backlot or a deliberate gap; a gap gets a note and an entry via \`backlot diff --source $SOURCE --update-baseline\`. This body is the latest scheduled run's report, rewritten in place." \
+            "$(cat report.txt)")
+          if [ -z "$existing" ]; then
+            gh issue create --title "$TITLE" --label fidelity --body "$body"
```
nuri-yoo, line 130: The report goes into the issue body whole, and a body over 65,536 characters is refused by the API with `body is too long (maximum is 65536 characters)` ([GitHub community discussion #27190](https://github.com/orgs/community/discussions/27190)). This step runs under the default `bash -e {0}`, so the failed `gh issue create` fails the step: the job goes red without an issue, and the report survives only in the run log.

Today's reports are far under it; s3's 41 findings are 12,906 characters. But a document being restructured turns a whole baseline into new findings at once, and measured with an empty `--baseline-dir` that report is 135,477 characters for github (1,254 findings), 93,478 for linear (829, with a real key) and 69,573 for jira (640). A `missing_*` finding costs 95 to 113 characters and an S3 probe finding about 300, so the limit sits near 600 new findings, or 200 for S3. Truncating the fenced block at a fixed size with a line saying so and pointing at the run keeps the issue filed on the morning it matters most.

docs/fidelity.md
```diff
@@ -0,0 +1,211 @@
...
+| Linear | GraphQL introspection | `api_key` — `LINEAR_API_KEY`, sent **bare** |
+| Slack | published OpenAPI | none |
+| Gmail | Google Discovery | none |
+| Google Drive (`google_drive`) | Google Discovery | none |
+| GitHub | published OpenAPI | none |
+| Jira | published OpenAPI (v3) | none |
+| Confluence | published OpenAPI (v1) | none |
+| HubSpot | published OpenAPI, resolved through the API catalog | none |
+| Notion | published OpenAPI | none |
+| Amazon S3 | botocore service model, **probed** — see below | none |
```
nuri-yoo: "see below" points up: the section is "S3 is asked, not read" at line 105, above this table.

docs/fidelity.md
```diff
@@ -0,0 +1,211 @@
...
+
+## A published spec is weaker evidence than introspection
+
+Introspection *is* the contract: a field absent from it is absent from the API. A published OpenAPI
+document only *describes* the contract, and it lags, omits paid tiers, and documents one verb where
+the server accepts two. Every `extra_operation` this has reported so far was the spec being
+incomplete rather than Backlot being wrong:
+
+- Slack's spec documents one verb per method and omits the search methods entirely. Measured against
+  slack.com, all fourteen answer `200/ok` over **both** GET and POST.
```
nuri-yoo: `/search.messages` is in the spec, as GET, and the baseline already treats it that way (`POST /search.messages` is the extra operation, `GET /search.messages?highlight` a missing parameter). What the spec omits is `search.all` and `search.files`. The fourteen slack.json notes and the description carry the same sentence; for the notes, "Slack's published spec documents one verb per method and omits `search.all` and `search.files`, so its silence is not evidence of absence" says what the data says.

```suggestion
- Slack's spec documents one verb per method, and of the search methods only `search.messages`:
  `search.all` and `search.files` are absent. Measured against slack.com, all fourteen answer
  `200/ok` over **both** GET and POST.
```

docs/fidelity.md
```diff
@@ -0,0 +1,211 @@
...
+- GitHub's spec describes neither `contents` without a path nor the legacy per-sha statuses read.
+  Measured against api.github.com, both answer `200`.
+- Atlassian's published Confluence v1 document no longer describes **any** read at all: it retains
+  the write operations and `search`, and nothing else. The v1 reads are deprecated in favour of
+  Confluence REST v2, not documented as removed — so on that source this comparison currently
+  covers almost nothing, which the baseline notes say plainly.
```
nuri-yoo: The document describes 65 GET operations (read on 2026-09-02): `content/search`, `content/{id}/descendant`, `content/{id}/restriction`, `group`, `label`, `user/current`, `space/{key}/settings` and more. What it no longer describes is exactly the eight reads Backlot serves, which is what the eight `extra_operation` findings say. The eight confluence.json notes and the description carry the same "any read" sentence; for the notes, "Atlassian's published v1 document no longer describes the content and space reads Backlot serves; it still describes 65 reads and the writes" would match.

```suggestion
- Atlassian's published Confluence v1 document no longer describes the content and space reads
  Backlot serves: `content`, `content/{id}` with its `child/page`, `child/comment` and `label`,
  `space`, `space/{key}` and its `permission`. It still describes 65 reads (`content/search`,
  `content/{id}/descendant`, `group`, `label`, `user/current` among them) and the writes. The v1
  reads Backlot serves are deprecated in favour of Confluence REST v2, not documented as removed —
  so on that source this comparison currently covers almost nothing, which the baseline notes say.
```

backlot/fidelity/fetch.py
khj809: Fixed. `fetch_json` refuses a parsed value that is not a dict, so all nine document sources exit 2 on it rather than dying deep in a parser. Reproduced all three payloads first (`AttributeError` in `from_openapi` each time).

backlot/fidelity/s3_probe.py
khj809: Fixed. The selector now also comes from the input shape's required querystring members, which covers ListParts and leaves GetObject's `versionId` and ListObjects' `prefix` alone as you noted. Confirmed `?uploadId=abc123` answers 200 with the object body byte for byte, and the probe now reports it: 43 findings, 42 unacknowledged. #103's count is updated.

backlot/fidelity/s3_probe.py
khj809: Fixed — `timeout=timeout`. A test now spies on `httpx.request` and asserts every probe request carries the caller's value.

backlot/fidelity/graphql_diff.py
khj809: Fixed. Unions diff their members (`extra_union_member` breaking, `missing_union_member` gap) and interfaces go through `_diff_object` with the guard widened, so an interface's own fields are walked and `_diff_args` reaches them. Linear's one union still matches, so no new findings there.

.github/workflows/fidelity.yml
khj809: Fixed. The report is capped at 55,000 characters with a line naming the run URL. Ran the step's exact shell against github's 135,493-character empty-baseline report: the body comes out 55,412, under the limit.

backlot/fidelity/comparisons.py
khj809: Rewritten. The docstring now says what the four kinds actually compare — SDL against introspection for the two GraphQL sources, `app.openapi()` against a published document for the eight, and answers from a started server for S3 — and that nine of the eleven need no credential.

backlot/fidelity/comparisons.py
khj809: Rewritten: no credential, no quota, no account, and a finding anyone can reproduce locally. It gates nothing.

backlot/fidelity/openapi_diff.py
khj809: Same fix here.

backlot/fidelity/comparisons.py
khj809: Fixed. That was damage from a bulk `REST`→`SPEC` rename of mine.

backlot/fidelity/comparisons.py
khj809: Fixed. It now names `OpenAPIComparison` and `GoogleDiscoveryComparison` together, and the following sentence says "those two" rather than "the other class".

backlot/fidelity/comparisons.py
khj809: Docstring corrected rather than the design: there is one probe source, so I would rather not add a second dispatch hook for a hypothetical one. It now says that fetching the model and starting the server are `s3_probe.divergences` and shared by every probe, and that a source needing different orchestration would have to route elsewhere rather than only supply a different `run`.

backlot/fidelity/s3_probe.py
khj809: Fixed, using your wording: ListObjectsV2 is what the baseline acknowledges, and ListObjects is the bare form itself so it carries no selector and is never asked.

backlot/cli.py
khj809: Fixed — `backlot/fidelity/baseline/<source>.json`. The help text also documents exit 3 now.

docs/fidelity.md
khj809: Fixed: it links the "S3 is asked, not read" section by anchor instead of saying "below".

docs/fidelity.md
khj809: Applied. The fourteen slack.json notes now say the spec documents one verb per method and, of the search methods, only `search.messages` — `search.all` and `search.files` absent. Confirmed against the spec: `/search.messages` is there as GET, the other two are not.

docs/fidelity.md
khj809: Applied. The eight confluence.json notes now say the document no longer describes the content and space reads Backlot serves, and still describes 65 reads and the writes. Counted 65 GETs; of the ten reads Backlot serves only `content/{id}/restriction/byOperation` and `search` still have one, which is the eight findings exactly.

khj809 (commented):
All seventeen addressed on 6a4f4fd, with a reply on each thread.

The four behaviour findings are fixed and each carries a test that fails without it — verified by reverting each fix in turn. `fetch_json` refuses a non-object document so the nine document sources exit 2 rather than dying in a parser at status 1; ListParts is asked, which makes it 43 findings and 42 unacknowledged (#103's count is updated); `probe()` gets the caller's timeout; unions diff their members and interfaces are walked, with Linear's one union still matching.

Two things from the review body I took as design changes rather than prose. A credential nobody set now exits **3** and the workflow fails on it — reported as a vendor outage and left green, the two introspection sources would go uncompared nightly, and as you found neither key is set on this repository yet, so that would have been the state from the first scheduled run. Setting `FIREFLIES_API_KEY` and `LINEAR_API_KEY` is still needed; the workflow will now say so in red instead of passing. And the issue body is capped at 55,000 characters with a line naming the run — I ran the step's exact shell against github's 135,493-character empty-baseline report and the body comes out 55,412.

On `ProbeComparison`: I corrected the docstring rather than adding a second dispatch hook. There is one probe source, so the freedom the docstring promised is not worth building; it now says the model fetch and server start are shared by every probe.

All three overclaims are narrowed in the description, in docs/fidelity.md and in the twenty-two baseline notes that repeated them. Counted them myself: `extra_*` is 28 (25 `extra_operation` + 3 `extra_param`), Slack's spec has `/search.messages` as GET with `search.all` and `search.files` absent, and Confluence's document has 65 GETs of which only `content/{id}/restriction/byOperation` and `search` are among the ten reads Backlot serves.

nuri-yoo (approved):
Re-reviewed at 3bbbcfc, the two commits since db5f05a; each fix was run, not only read. Suite on an all-extras install: 1690 passed, 0 skipped (the five new tests account for the difference), `ruff check` and `ruff format --check` clean. A 200 carrying a JSON array or string now exits 2 with `answered JSON list, not an object`, and a declared credential that is not set exits 3 naming the variable, which the workflow turns red. The `sources` job's snippet yields all eleven names on an empty input, `["linear"]` on `linear`, and fails the job on a name the registry does not have. Unions and interfaces are compared: the synthetic cases from my first review now report `extra_union_member M.C` and `N.invented`. The report cap, driven through the step's own shell against a stubbed `gh` with the real `jq` expressions, turns the 135,477-character github report into a 55,421-character body carrying the truncation line and the run URL.

All eleven sources against their real contracts, with real keys for the two GraphQL ones: nine exit 0, linear exits 1 with the same 47 (782 acknowledged, and no union finding, since Linear's one union matches), and s3 exits 1 with 42, the 41 from before plus `ListParts: GET /{object}?uploadId` now reported as a silent fallthrough. `--update-baseline` reproduces the rewritten slack, confluence and s3 files byte for byte apart from `measured`. On ProbeComparison, one dispatch path with a docstring that says so is the right call while there is a single probe source; agreed.

Two things before merging, neither in the diff. The description lags the code: the exit-code table has no row for 3, "#103 — 41 unrecognised S3 sub-resources" and "S3 (41 silent fallthroughs)" are 42 now, and "1684 passed, 1 skipped" is 1690; issue #103 says 41 as well and wants the ListParts line. And `head -c 55000` cuts bytes, so a multibyte character straddling the cut leaves an invalid sequence in the body (measured with `'a'*54999 + '—'`); `${report:0:55000}` slices characters, uses the unit `${#report}` already uses, and is shorter. Approving; the threads are resolved.

khj809 (commented):
Both done on c0c0077.

The description now has the row for 3 with the reason it is separate from 2, and 42 in both places. Test line reads 1689 passed, 1 skipped — 1690 collected — with a note that an install carrying the hubspot reader reports your 1690/0; the template asks for the number the named install produced, so neither figure alone was right. Issue #103 was already 42 in its opening but its list still held 41 paths: `ListParts: GET /{object}?uploadId` is added, the object group is 9, and the opening no longer says all 42 answer `<ListBucketResult>` — 41 do and ListParts answers the object's bytes.

On the cut: applied, and I could not reproduce it here. System bash is 3.2 with no UTF-8 locale, so `${report:0:55000}` and `head -c` both leave a truncated `e2` in my test — the slice is character-aware only on bash 4+ under a UTF-8 locale, which is the runner and not this machine. I took it on your measurement plus the argument that stands without one: `[ "${#report}" -gt 55000 ]` already counts characters, so cutting bytes mixed units regardless of what the seam happens to land on.

PR #107, opened 2026-09-02
The README opens with what Backlot does and a 60-second path to using it

nuri-yoo:
The README now says what Backlot does in its first line and gets a reader from `pip install` to an official SDK answering in one screen.

## What changed

The first line is "Run enterprise SaaS APIs locally.", followed by one paragraph naming the services and what is reproduced: response shapes, pagination, auth, errors, per-document ACLs. Then the demo GIF, then the first command — nothing between the title and `pip install` that is not doing work.

"Try it in 60 seconds" shows `pip install`, `backlot import --bundled` and `backlot serve`, then an official SDK pointed at that running server with one changed base URL, then the in-test `backlot.serve()` on a free port. Both are code blocks: every other claim on the page is shown rather than described, and the quickstart is where a reader is most likely to copy rather than read.

A hand-written-mocks-versus-Backlot table states the case for the project, with its columns marked ❌ and ✅. "Bring your own corpus" gains a one-record JSONL sample and the `--dry-run` / `import` / `serve` commands, and says where per-user credentials come from before linking `docs/corpus.md` and `docs/auth.md` for the rest. The paragraph under "What it serves" splits into three: one sentence on the port and the prefixes, one on what the corpus fixes, one on the emulated subset.

Kept: the demo GIF and the architecture figure with their full alt texts, the emoji use cases, the services table with a per-service SDK example, and the bring-your-own-corpus example row. The GIF is re-recorded, because it disagreed with the page in three places: `scripts/readme-demo-recorder` now draws "Run enterprise SaaS APIs locally." on the card that opens and closes the clip, puts `backlot import --bundled` between the install and the serve in both acts, and writes the base URL as `http://127.0.0.1:8000` rather than `localhost`. What the lane shows is the quickstart, typed in the same order. Dropped: the LocalStack comparison and the three framing bullets, which the GIF and the opening paragraph already carry, and the "building before you have credentials" use case, which the first bullet repeats.

`tests/test_docs.py` raises the README ceiling from 130 lines to 180; the page is 142. The reference material it guards against — endpoints, corpus fields, auth schemes, settings — is still in `docs/`, and the ceiling still stops it from coming back.

README.pypi.md is now README.md with two things done to it, and nothing else. The demo GIF and the architecture figure come out, because a relative asset path 404s under pypi.org; every relative link target is written absolute against github.com/brekkylab/backlot, files into /blob/main and directories into /tree/main, for the same reason. Everything the two pages used to differ on for no reason — badges split across five lines, a documentation table that dropped the `docs/` prefix from its link text, "The repo" against "This repo" — is gone. Strip the absolute prefix from one and the images from the other and the files are identical, which is the property the section below checks.

`feat/mcp-subcommand` (#108) adds a line to the agent section of this same file; whichever of the two merges second re-applies a one-line hunk.

No served behavior changes. The one claim about the corpus format is checked below.

## Verification

- **Tests** — `pytest` on a `.[all]` install, rebased onto main's `backlot diff`: 1689 passed, 1 skipped (`llama-index-readers-hubspot`, which the install CI adds with `--no-deps`). `tests/test_docs.py`, `tests/test_skills.py` and `tests/test_packaging.py` on their own: 39 passed. README.md is 143 lines against the 180-line ceiling, states no source count, and every relative link in it resolves. `docs/fidelity.md` is a row in both documentation tables.
- **Optional surfaces that ran rather than skipped** — no code path changes, so nothing here is load-bearing; the suite above ran the `mcp`, `llamaindex`, `mirage`, `fsspec`, s3 and SDK surfaces in full. What was exercised instead is the page itself: both Python blocks were extracted from README.md and run verbatim in a fresh venv outside the checkout, against a locally built wheel, after `backlot import --bundled` and `backlot serve` — the first printed the bundled corpus's channels off the running server, the second the same channels off its own throwaway one. The JSONL sample line passes `backlot import --dry-run` ("OK: 1 records valid").
- **The PyPI page** — README.pypi.md was regenerated from README.md by dropping the two image blocks and prefixing every relative target, and the inverse transformation reproduces README.md byte for byte: no third difference survives. `python -m build` then `twine check` on the wheel and the sdist: both PASSED. The long description in the wheel's METADATA is the 139-line page, with no relative target and no `assets/` reference left in it; `readme_renderer --format md` renders its nine sections, four tables, six code blocks and five badge images. Every `blob/main` and `tree/main` path it links resolves in the tree on main.
- **The GIF** — rebuilt through the recorder's own `build.sh`; rebuilding from the unmodified source first reproduced the committed `assets/demo.gif` byte for byte, so the new file's every difference is attributable to these edits. 319 frames at 900x480, 1,174,531 bytes against 1,172,107. The fourth step squeezed the response card by a measured 43px — it is the flex item that shrinks, and it was losing its last field and its bottom edge — so the step, chip and card spacing gives that back and the card measures 1px of slack at every dense frame. Confluence's base URL, the longest in the second act, ends at x=830 against a checklist rule at 866. The loop seam is 77.3 dB PSNR between the last frame and the first, where the pre-existing build measured 77.2. `npx tsc --noEmit`: clean.
- **Lint** — `ruff check . && ruff format --check .`: clean (Markdown is excluded by config). `python scripts/gen_docs.py --check`: exit 0.

- [ ] A test fails without this change — n/a, documentation. `test_readme_stays_short` is the test that constrains the page, and it is the one this raises.
- [ ] A new endpoint serving corpus content is ACL-scoped, proved for both an admin and a scoped token — n/a.
- [ ] Comments state what was measured, not the history of the fix — n/a, no code comments.


khj809 (approved):
Looks much better. Thanks!

PR #108, opened 2026-09-02
backlot mcp serves every source as MCP tools over stdio, starting a server when none answers

nuri-yoo:
One command an MCP client can run, and every source arrives as tools:

```bash
pip install "backlot[mcp]" && claude mcp add backlot -- backlot mcp
```

`backlot mcp` serves every source as MCP tools over stdio from one process, and when no Backlot server is running it starts one itself, over the data dir's corpus, and stops it when the client disconnects. One flag, `--user <email>`, names who is asking, and every source's credential follows from it. The two bridge scripts that used to live under `examples/using-mcp-with-agents/` are now package code behind that command. #109 (merged into this branch) is part of it: it replaced `--token`/`--username` with `--user` and made S3 a bridged source.

## What changed

**`backlot mcp`** (`backlot/mcp.py`, `backlot/cli.py`). It bridges the server at `--url`; else the one on `127.0.0.1:8000` if a Backlot server answers there (a `/health` body carrying `documents`, so another local dev server on that port is not mistaken for one); else a server it starts as a subprocess through `backlot.server.serve`, over the data dir's corpus or the bundled hello-world corpus when the data dir holds none. An explicit `--url` that does not answer is an error, not a fallback, and it gets the 10s budget `backlot.server._healthy` reserves for a remote hop rather than the 2s speculative probe the default port gets. When the default port is taken and a `--data-dir` was given, the command says on stderr that the dir is not used — that server serves whatever corpus its operator gave it. With no `--source` every source is mounted in one FastMCP server under a `<source>_` namespace (`slack_search_messages`, `atlassian_jira_get_issue`); `--source slack` serves one source under its plain tool names. `--depth` is the GraphQL selection depth (1 for Linear, 2 for Fireflies unless given).

**`--user <email>`** resolves one person's whole credential set — bearer token, email, S3 access-key pair — from the server's own `GET /_meta/users`, matched case-insensitively, and spells it the way each source authenticates: `Bearer` for most, HTTP Basic `email:token` for Atlassian, a SigV4 signature per request for S3 (`backlot.sigv4`, the module the S3 router verifies with). Without `--user` the admin sees everything. An email the corpus does not know is refused before a single tool exists. `BACKLOT_EXPOSE_TOKENS` is gone with its gates on `/_meta/users` and `/_meta/credentials`, since a switch that closed them would take this command's only credential input with it; `tests/test_docs.py` holds `docs/configuration.md` and `.env.example` to `Settings`' field list.

**Tool names are route names** (`backlot/openapi.py`). The MCP-ready spec at `/_meta/openapi/<source>` used to carry FastAPI's suffixed ids, `search_messages_slack_api_search_messages_get`; under a source namespace those crossed the 64-character cap MCP clients apply and were truncated. Each operation is now named for its route (`search_messages`), by `tool_name()`, an exact inverse of `unique_operation_id()`; the raw `/openapi.json` keeps the long deterministic ids. HEAD operations are dropped before renaming (a response with no body cannot answer the question that called it), and S3's `bucket_get` declares the query parameters it honours. Renaming before the dedupe pass lets the fidelity aliases collapse to one tool each, and `test_build_mcp_spec_collapses_exactly_the_known_aliases` is the snapshot of which ids fold how many routes per source: 12 Slack GET/POST pairs; Jira's five v2/v3 ids, `jira_search` folding three; ListBuckets at `/s3` and `/s3/`. A handler that shared a name with another in its source would fail that test rather than silently lose a tool. Under the composite server, Gmail's and Drive's tools no longer repeat the module-local prefix their handlers carry: `mounted_name` folds it into the namespace, so the model reads `gmail_messages_list` and `gdrive_files_list`, not `gmail_gmail_messages_list` and `gdrive_drive_files_list`.

**`backlot.server`.** `serve(data_dir=...)` serves an already-built data dir without importing into a temp dir, and writes its log to scratch rather than into that dir. `meta_users(url)` is the strict reader behind `--user`; `admin_token_for(url)` its lenient wrapper for `serve_or_connect`. `serve_or_connect`'s two messages move from stdout to stderr: a caller may be an MCP stdio server whose stdout is the protocol stream, and one stray line there ends the session. Both children `serve` starts get `stdin=DEVNULL` for the same reason — under `backlot mcp` the inherited handle would be the protocol pipe — and a test asserts it on the calls. When the started server dies before answering (a corrupt `db.sqlite` in the data dir is the common cause), `backlot mcp` exits 1 with the server's own last words rather than a traceback through contextlib.

**Ending.** SIGTERM, SIGINT and SIGHUP are handled alike: the exit closes the `serve()` context the command entered when it started a server (its finally SIGTERMs that server and waits), flushes stderr, and leaves with `os._exit(128 + N)`. It does that directly rather than raising through the stack, because an exception raised out of the handler has to cross fastmcp's stdio transport, where it can be held waiting on the stdin reader thread — non-daemon, blocked in a read nobody will end. Measured: before any of this, SIGTERM with stdin open left the process alive until stdin closed, SIGINT never ended it, and SIGHUP ended it by default disposition with the started server left listening; with the raising version, the SIGHUP case was still alive 20s after the signal on two of four Python versions in CI. The handler also covers the window before `serve()` has yielded: the uvicorn child is spawned several hundred milliseconds before the readiness poll lets the context yield, and a signal there found nothing to close and orphaned the child on its port (9 of 9 runs with SIGTERM at 0.5–0.7s after start) — so `serve()` registers the child right after `Popen`, and the handler stops whatever is still registered through `server.stop_children()` before leaving. That exit runs on a thread of its own, woken through a wakeup fd, rather than in the Python-level signal handler: such a handler runs only on the main thread once it is back in the interpreter, and the main thread here is the event loop asleep in select, so a signal the kernel hands to another thread (the stdin reader, a worker) was never acted on — measured with `pthread_kill`, the process stayed alive indefinitely, which was the intermittent CI timeout on the SIGHUP case. The C-level handler writes the signal number to the wakeup fd from whichever thread it runs on, so the reading thread sees every signal. A SIGKILL can still leave that server running; nothing can unwind after one.

**Startup.** `build_server` fetches every source's spec or introspection concurrently. Serially the default run was ten round trips before the MCP handshake — about 10s against a hosted deployment that answers `/health` in ~1s, more than some MCP clients wait.

**Examples and docs.** The seven bridge launchers (`slack`, `github`, `gmail`, `gdrive`, `hubspot`, `linear`, `fireflies`) run `python -m backlot mcp --source <name> --url ... [--user ...]` instead of a sibling script; `atlassian`, `notion` and `s3` keep their vendor servers. `examples/using-mcp-with-agents/README.md` opens with the command. README.md gains the one-liner in the agent section and reworded MCP row, and `skills/backlot/SKILL.md` names the command in its client table.

## Later: a shorter command

The install line can lose its first half once a release carrying `backlot mcp` is on PyPI: `claude mcp add backlot -- uvx --from "backlot[mcp]" backlot mcp` resolves the package on first launch, the way the awslabs S3 server is run in `examples/using-mcp-with-agents/s3.py`. It works against a local wheel and not against the published `0.0.1`, which has no `mcp` command, so the README shows the `pip install` form until then. What would make that launch quick is a slimmer extra: `[mcp]` carries `anthropic`, `openai` and `openai-agents` for the example agents, none of which the bridge imports — a `mcp` + `fastmcp` + `httpx2` extra is all the command needs.

Served responses do not change. The bridge forwards them untouched (`validate_output=False` for OpenAPI, pinned by a spec that promises an integer over a server that answers a string; the GraphQL envelope as-is, with the MCP error flag only when there is no `data` at all) with the caller's own credential, so what a tool returns is exactly what the server returns to that user. Auth spellings are the ones each client speaks: Bearer, Basic `email:token` for Atlassian as mcp-atlassian sends it, SigV4 for S3 computed by the module the server verifies with. Tool names come from Backlot's route names, not from a vendor spec; the vendor never published MCP tools for these surfaces, so there is nothing to be faithful to there beyond the operations themselves.

## Verification

- **Tests** — `pytest` on a `.[all]` install plus `llama-index-readers-hubspot<0.6 --no-deps`, the install CI uses: **1738 passed, 0 skipped**, 46.60s. Docker (colima), `npx` and `uvx` were present, so the mcp-atlassian, notion-mcp-server and awslabs tests ran rather than skipped. `ruff check . && ruff format --check .` clean; `python scripts/gen_docs.py --check` exit 0.
- **Without the change** — the six changed test files copied onto `main` in a worktree: **22 failed, 97 passed**, and `tests/test_mcp.py` does not collect (`backlot.mcp` does not exist there).
- **Each stated behaviour fails a test when removed** — measured by mutation on this head: the `documents` check in `is_backlot`, the named-server timeout, the `--data-dir` note, SIGINT/SIGHUP registration, the handler leaving on its own, the GraphQL error rule, `validate_output=False`, the namespace fold, concurrent building, and `stdin=DEVNULL` on the served child each fail exactly the test written for them.
- **End to end, outside pytest** — a composite server started by the command over a live corpus: 115 tools across ten sources, longest name 37 characters; `gmail_*` and `gdrive_*` carry no repeated prefix. Over real stdio in the suite: every source namespaced under 64 characters, `--user` scoping a search, an unknown email refused with exit 1, a server started when none answers and gone when the client hangs up or on each of the three signals — including one sent after the server was spawned but before it was ready and one delivered to a thread other than the main one, an unusable corpus reported as an exit.

- [x] A test fails without this change — the 22 above.
- [x] A new endpoint serving corpus content is ACL-scoped, proved for both an admin and a scoped token — `test_backlot_mcp_scopes_every_tool_call_to_the_user` over stdio, and `test_mcp_bridge_enforces_the_acl` / `test_mcp_graphql_bridge_enforces_the_acl` in-process for each of the ten sources.


backlot/cli.py
```diff
@@ -508,6 +513,87 @@ def status(data_dir: DataDir = None) -> None:
...
+    ] = None,
+    username: Annotated[
+        Optional[str],
+        typer.Option(
+            "--username",
+            metavar="USER",
+            help="switch Atlassian to HTTP Basic `USER:token`, the scheme mcp-atlassian speaks; "
+            "Backlot resolves the token and ignores the name",
+        ),
+    ] = None,
```
khj809: These options cannot cover the whole services.
- `--username` is Atlassian-only.
- S3 requires access key and secret key, but there's no way to provide them separately.
- Google services require more credentials related to OAuth.

So I suggest to add just one `--user <email>` option (admin if not provided), then backlot resolves the user's credentials internally and provides them to each service's auth scheme. I measured the blast radius of this task and it seems quite large, so I'd like to add a stacked PR onto this branch if you don't mind. Would it be okay?

nuri-yoo: Agreed and feel free to proceed with the stacked PR

khj809: Merged #109. I'll review the whole changes again thoroughly.

backlot/mcp.py
```diff
@@ -0,0 +1,428 @@
...
+
+def is_backlot(url: str) -> bool:
+    """Whether a *Backlot* server answers at ``url`` — not merely something with a ``/health``.
+
+    The default port is shared with every other local dev server, and a 200 from one of those would
+    send every tool call into 404s. Backlot's ``/health`` body always carries ``documents``
+    (``null`` while the caches warm), which is enough to tell it apart."""
+    try:
+        with urllib.request.urlopen(f"{url.rstrip('/')}/health", timeout=2) as r:
+            return r.status == 200 and "documents" in json.load(r)
```
khj809: Two things here, and the second is the one I'd fix first.

**The 2s timeout is the default-port probe's, and `attach` reuses it for an explicit `--url`.** A named server that is merely slow reads as absent: measured against a `/health` returning a valid Backlot body after 3s, this says False while `server._healthy()` (10s) says True, and the command exits 1 with "no Backlot server answers at …". `_healthy`'s own comment reserves its 10s for "a remote, possibly trans-continental HTTPS hop" — the hop `--url` is for, and `attach` calls `ensure_cert_bundle()` right before it. `backlot.brekkylab.com/health` answers in 0.81–0.97s from here, so there is one second of headroom.

```suggestion
def is_backlot(url: str, timeout: float = 2) -> bool:
    """Whether a *Backlot* server answers at ``url`` — not merely something with a ``/health``.

    The default port is shared with every other local dev server, and a 200 from one of those would
    send every tool call into 404s. Backlot's ``/health`` body always carries ``documents``
    (``null`` while the caches warm), which is enough to tell it apart. The default ``timeout`` suits
    that speculative probe; a caller who NAMED a server should pass the budget a remote hop needs."""
    try:
        with urllib.request.urlopen(f"{url.rstrip('/')}/health", timeout=timeout) as r:
            return r.status == 200 and "documents" in json.load(r)
```

with `attach`'s explicit-`--url` branch passing `timeout=10`.

**And the `documents` check — this function's whole reason to exist — has no test.** Replacing the return with a bare `r.status == 200` leaves test_mcp/test_server/test_openapi/test_cli all green. A stub answering 200 with a body that has no `documents` key would pin it.

backlot/mcp.py
```diff
@@ -0,0 +1,428 @@
...
+    from backlot import server as backlot_server
+
+    url = (url or "").strip().rstrip("/")
+    if url:
+        backlot_server.ensure_cert_bundle()
+        if not is_backlot(url):
+            sys.exit(f"no Backlot server answers at {url}")
+        yield url
+        return
+    if is_backlot(DEFAULT_URL):
```
khj809, line 439: `--data-dir` is silently discarded whenever anything Backlot-shaped holds the default port: this branch yields before `get_settings()` is ever reached. Measured — with `backlot serve --port 8000` over one data dir, `backlot mcp --source slack --data-dir <other>` prints "using the Backlot server at http://127.0.0.1:8000" and serves the first dir's corpus.

That is the case this same function refuses for `--url` ("the caller named a server, and quietly serving a different corpus in its place is the wrong kind of helpful"). A named `--data-dir` is the same statement of intent about which corpus, and `--data-dir`'s help gives no hint it can be dropped. Cheapest honest version is to say so:

```suggestion
    if is_backlot(DEFAULT_URL):
        _stderr(f"using the Backlot server at {DEFAULT_URL}")
        if os.environ.get("BACKLOT_DATA_DIR"):
            _stderr(
                f"(--data-dir is not used: that server serves whatever corpus its own operator "
                f"gave it; pass --url to bridge {os.environ['BACKLOT_DATA_DIR']} instead)"
            )
```

(needs `import os`.) Refusing outright when both are given would be defensible too.

backlot/mcp.py
```diff
@@ -0,0 +1,428 @@
...
+    user: str | None = None,
+    depth: int | None = None,
+) -> None:
+    """Serve ``sources`` (default: all of them) as MCP tools over stdio until the client hangs up."""
+    try:
+        import fastmcp  # noqa: F401 — the check is the import
+    except ImportError:
+        sys.exit('backlot mcp needs the [mcp] extra:  pip install "backlot[mcp]"')
+
+    signal.signal(signal.SIGTERM, _exit_on_signal)
```
khj809: Three notes on this line, all about the same seam.

**It has no test.** Deleting it leaves the suite green — `test_backlot_mcp_starts_a_server_when_none_answers` covers the closed-stdin path only, which is the other path `_exit_on_signal`'s comment names.

**SIGTERM alone does not end the process.** Measured with stdin left open: the started server stops listening within 1s, but `backlot mcp` is still alive at t+30s and exits (143) only once stdin closes — the non-daemon stdin reader keeps the interpreter up. A supervisor that signals and waits hangs.

**SIGINT and SIGHUP orphan the server.** SIGINT alone: neither exits within 30s, and a following SIGKILL leaves the uvicorn listening. SIGHUP alone: the process dies instantly by default disposition, server left listening. A terminal Ctrl-C or hangup is fine, because the group signal reaches uvicorn too — this is `kill -INT`/`kill -HUP` on the pid, and any supervisor that signals only its direct child. So the PR's stated limit ("a closed stdin and SIGTERM both take it down") holds for the child but not for this process.

```suggestion
    for sig in (signal.SIGTERM, signal.SIGINT, signal.SIGHUP):
        signal.signal(sig, _exit_on_signal)
```

backlot/mcp.py
```diff
@@ -0,0 +1,428 @@
...
+            r = await client.post(
+                endpoint, json={"query": self.spec.document, "variables": variables}
+            )
+            body = r.json()
+            # The GraphQL envelope goes through untouched, on the same principle as
+            # `validate_output=False` above. What is added is the MCP error flag, so a caller can
+            # tell a refusal from an answer — raised only when there is no `data` at all (a
+            # rejected document, or a non-null field the ACL hid), not for a field error that came
+            # back beside partial data.
+            failed = bool(body.get("errors")) and body.get("data") is None
```
khj809: The rule this line encodes — errors *beside* partial data are not a tool error — is undocumented by any test: widening it to `failed = bool(body.get("errors"))` leaves test_mcp.py green, because the ACL cases only ever produce the no-data shape, which both versions flag identically.

Worth a case with a field error next to data (a nullable field the ACL hid inside an otherwise-fine result), since under the wider rule an agent throws away a mostly-complete answer.

backlot/mcp.py
```diff
@@ -0,0 +1,428 @@
...
+    client = httpx2.AsyncClient(
+        base_url=base_url,
+        headers={} if signed else _auth_header(creds, source),
+        auth=_SigV4() if signed else None,
+        timeout=30,
+    )
+    # validate_output=False: Backlot's responses are the source of truth; a passthrough must never
+    # reject a real server response for not matching a loose schema.
+    return FastMCP.from_openapi(
+        openapi_spec=spec, client=client, name=f"{source}-bridge", validate_output=False
```
khj809, line 273: `validate_output=False` is a stated fidelity principle and nothing holds it: flipping it to `True` leaves test_mcp.py green, so no tested response currently violates its own schema. Low stakes today, but it means the flag's protection is invisible and a future response shape that does diverge would be rejected by the bridge rather than passed through.

backlot/mcp.py
```diff
@@ -0,0 +1,428 @@
...
+    namespaces = ", ".join(f"{source}_*" for source in servers)
+    instructions = (
+        "Backlot serves these enterprise SaaS APIs over a local corpus. Tools are namespaced "
+        f"by source: {namespaces}."
+    )
+    if "atlassian" in servers:
+        instructions += " atlassian_* covers both Jira and Confluence."
+    root = FastMCP(name="backlot", instructions=instructions)
+    for source, server in servers.items():
+        root.mount(server, namespace=source)
```
khj809: `gmail` and `gdrive` stutter under the namespace, because their handlers already carry a vendor prefix to stay unique inside `routers/google.py`: the model reads `gmail_gmail_messages_list`, `gmail_gmail_labels`, `gdrive_drive_files_list`, `gdrive_drive_about`. The other eight are clean — s3, added here, gives `s3_bucket_get` / `s3_list_buckets` / `s3_object_get`.

```suggestion
    for source, server in servers.items():
        root.mount(server, namespace=source)
    return root


def _mounted_name(source: str, tool: str) -> str:
    """``<source>_<tool>``, without repeating a prefix the route name already carries: the Gmail
    and Drive handlers are named ``gmail_*`` / ``drive_*`` to stay unique inside one module, and
    ``gmail_gmail_messages_list`` says it twice to a model that reads the name on every call."""
    head = tool.split("_", 1)[0]
    return tool if head in (source, source.removeprefix("g")) else f"{source}_{tool}"
```

as a sketch — the mount call is FastMCP's, so the real fix is either renaming those handlers or a `tool_name_fn`/transform at mount time, whichever that version exposes.

backlot/mcp.py
```diff
@@ -0,0 +1,428 @@
...
+):
+    """One FastMCP server over ``sources``.
+
+    A single source is served as itself, so its tools keep the names Backlot's own routes give them
+    (``search_messages``, ``jira_get_issue``). Several are mounted under the source's own name as a
+    namespace — ``slack_search_messages``, ``github_get_issue`` — because two vendors can and do
+    name an operation the same way."""
+    from fastmcp import FastMCP
+
+    servers = {
```
khj809: This builds every source serially, so the default all-source run makes 11 sequential requests before the handshake completes: 1 `meta_users` + 8 spec fetches + 2 introspections. Locally `initialize()` returns in 0.74s, so it does not show. Against the hosted deployment, whose `/health` I measured at 0.81–0.97s per round trip, the same startup is roughly 10s of serial waiting — inside some MCP clients' startup budget and outside others'.

Gathering them concurrently, or building each source's server lazily on first use, would take it back to roughly one round trip.

backlot/openapi.py
```diff
@@ -135,12 +194,13 @@ def _duplicate_operation_ids(spec: dict) -> list[str]:
...
-    """The MCP-ready spec for ``source``: sliced to its paths, fidelity aliases collapsed.
+    """The MCP-ready spec for ``source``: sliced to its paths, HEAD dropped, operations named for
+    their routes, fidelity aliases collapsed.
 
     Raises ``KeyError`` for an unknown source and ``ValueError`` if any operationId collision
     survives (an invariant — dedupe should always resolve them)."""
     prefixes = SOURCE_PREFIXES[source]
-    spec = dedupe_operations(slice_spec(full_spec, prefixes))
+    spec = dedupe_operations(name_operations(drop_head_operations(slice_spec(full_spec, prefixes))))
     dupes = _duplicate_operation_ids(spec)
```
khj809, line 204: This check cannot fire. `dedupe_operations` has already resolved every collision by the time it runs, so `_duplicate_operation_ids` always returns `[]` — which matters because renaming to route names removed the only thing that kept two *distinct* routes from collapsing into one tool. On this head the collapse set is exactly the intended one (12 Slack GET/POST pairs, 5 Jira v2/v3), but that is a snapshot: add a route whose handler shares a name with an existing one in the same source and one of the two tools silently disappears, with no error and no failing test.

The invariant worth asserting is a count rather than a duplicate — per source, tools == distinct routes minus the aliases you expect. `grep -c` for any such assertion in tests/test_openapi.py returns 0 today.

backlot/server.py
```diff
@@ -268,6 +272,9 @@ def serve(records: list[dict] | None = None, host: str = "127.0.0.1"):
                     "warning",
                 ],
                 env=env,
+                # The server reads nothing from stdin, and must not inherit the caller's: when
+                # the caller is `backlot mcp`, that is the MCP protocol pipe.
+                stdin=subprocess.DEVNULL,
```
khj809, line 297: This is one of the two things the branch's last commit is titled after, and it has no test: deleting the line leaves the suite green. The comment states the stake exactly — under `backlot mcp` the inherited handle is the MCP protocol pipe — so a regression corrupts a live session while CI stays quiet. `proc.stdin is None`, or the child's fd 0 resolving to /dev/null, would pin it. Same for the `import` subprocess above.

khj809 (commented):
Went back over the whole of this PR from `main` after #109 landed, with the parts I wrote in it. Two docs drifts are pushed to this branch as a37ea26 — `head_bucket`'s docstring claimed 404 where the signature check answers 403/400, and `meta_users`' paragraph said "Strict either way" after one of the two failure modes it contrasted had moved out. Both were mine from #109; shout if you'd rather they came as a separate PR.

The rest are notes, none blocking. The pattern worth naming: several deliberate, carefully-commented behaviours have no test holding them in place, and I found them by mutation rather than by reading — `is_backlot`'s `documents` check, `stdin=DEVNULL` on the served subprocess, the SIGTERM registration, the GraphQL partial-data rule and `validate_output=False` can each be deleted or inverted with the suite still green. The three previously-raised ones I did not fix (the 2s probe on an explicit `--url`, `--data-dir` losing to :8000, the signal paths) are re-verified on this head with numbers.

backlot/cli.py
nuri-yoo: Resolving: `--user` landed with #109, merged into this branch.

backlot/mcp.py
nuri-yoo: ae210ef. `is_backlot(url, timeout=PROBE_TIMEOUT)`, and `attach`'s `--url` branch passes `NAMED_SERVER_TIMEOUT = 10`. `test_a_named_server_gets_the_remote_budget_the_default_port_probe_does_not` runs a `/health` that answers after 3s: False at the probe's budget, True at the named-server one, and `attach` takes it. `test_is_backlot_wants_the_documents_key_not_just_a_200` pins the `documents` check with a 200 whose body lacks it; `return r.status == 200` fails that test.

backlot/mcp.py
nuri-yoo: Went with saying so, in ae210ef: when the default port answers and `BACKLOT_DATA_DIR` is set, `attach` prints on stderr that the dir is not used, names it, and points at `--url` or freeing the port. Refusing would also have caught a `BACKLOT_DATA_DIR` from a `.env` the caller did not type for this run, which is the common way it arrives. `test_attach_says_when_a_data_dir_cannot_be_honoured` — and asserts silence without the var.

backlot/mcp.py
nuri-yoo: ae210ef registers all three, and after the `with` unwinds `run` leaves through `os._exit`: the cleanup had run, and what kept the process alive was fastmcp's non-daemon stdio reader blocked on stdin. `test_backlot_mcp_exits_on_a_signal_with_stdin_still_open[SIGTERM|SIGINT|SIGHUP]` holds stdin open past the handshake and asserts 128+N; `test_backlot_mcp_takes_its_server_down_on_sighup` checks the started server's port closes. On the previous head the SIGTERM case did not return within 20s and SIGHUP left the server listening — both as you measured. Registration alone was not enough: without the `os._exit`, the SIGTERM case still hangs.

backlot/mcp.py
nuri-yoo: ae210ef: the rule is `graphql_outcome(body) -> (content, is_error)`, a pure function, and `test_graphql_outcome_flags_a_refusal_but_not_a_field_error_beside_data` states the three shapes — no data, data beside a field error, clean. Widening to `bool(body.get("errors"))` fails it.

backlot/mcp.py
nuri-yoo: ae210ef: `test_openapi_bridge_passes_a_response_through_even_when_it_breaks_its_own_schema` — a spec that promises an integer over an `httpx2.MockTransport` that answers a string. The bridge passes it through; with `validate_output=True` the call comes back as an error, so the flag is visible now.

backlot/mcp.py
nuri-yoo: ae210ef, folded rather than dropped: `mounted_name` turns `drive_files_list` into `gdrive_files_list` and leaves `gmail_messages_list` as it is, so every tool still begins with its namespace and the instructions line (`gdrive_*`) stays true. Done at mount time through fastmcp's `tool_names`; measured, its values get the namespace prefixed, so it takes the tail. Under the composite: `gmail_messages_list`, `gmail_labels`, `gdrive_files_list`, `gdrive_about`; 115 tools, longest still 37. `test_backlot_mcp_serves_every_source_namespaced` maps the served ids through `mounted_name` and asserts no name repeats its first segment.

backlot/mcp.py
nuri-yoo: ae210ef: `build_server` fetches each source's spec or introspection in a thread pool (`openapi_server` now accepts a spec it may already have been handed). `test_build_server_fetches_every_source_concurrently` gives each fetch 0.5s and asserts the eight OpenAPI sources build in under 2s, where serial is 4s. Lazy building would have kept the tool list unknown at `initialize`, which the namespace renaming above needs.

backlot/openapi.py
nuri-yoo: Kept the check — it is still the invariant the function documents — and added the snapshot in ae210ef: `test_build_mcp_spec_collapses_exactly_the_known_aliases` asserts per source exactly which ids fold and how many routes each absorbs. Measured: the 12 Slack GET/POST pairs; Jira's five v2/v3 ids, with `jira_search` folding three routes (GET+POST under v2 and v3); ListBuckets at `/s3` and `/s3/`. Your five was the id count; the routes removed are seven. A handler that shares a name with another in its source now fails that test rather than losing a tool silently.

backlot/server.py
nuri-yoo: ae210ef: `test_served_subprocesses_do_not_inherit_stdin` wraps `Popen` and `run` in `backlot.server` and asserts `stdin is subprocess.DEVNULL` on both children. Removing it from the serve `Popen` fails the test.

backlot/mcp.py
nuri-yoo: Follow-up in b7fbccb. CI on ae210ef caught what my machine did not: with stdin held open, the SIGHUP case was still alive 20s after the signal on 3.11 and 3.14 (3.12 and 3.13 passed) — the raised exception has to cross fastmcp's stdio transport, where it can wait on the stdin reader thread. The handler now closes the `serve()` context `attach` entered (its finally SIGTERMs the server and waits), flushes stderr and calls `os._exit` itself, so nothing above it is involved. Same tests, 3× locally green; CI on b7fbccb is the real check.

backlot/mcp.py
```diff
@@ -0,0 +1,529 @@
...
+    else:
+        _stderr(
+            f"no corpus in {settings.data_dir}; starting Backlot over the bundled hello-world "
+            "corpus (`backlot import <corpus.jsonl>` to serve your own)"
+        )
+        started = backlot_server.serve()
+    global _started
+    with started as s:
+        _stderr(f"Backlot is up at {s.base_url}")
+        _started = started
```
khj809, line 469: `_started` is set only after `with started as s:` has succeeded, and `serve()` spawns the uvicorn child several hundred milliseconds before it yields — it still has the readiness poll to get through. A terminating signal inside that window finds `_started is None`, so the handler goes straight to `os._exit` and the child is orphaned.

Measured on this commit, bundled corpus, SIGTERM to the parent only:

| signal at | `Backlot is up` printed | orphaned `backlot serve` |
|---|---|---|
| 0.3s | no | 0 (child not spawned yet) |
| **0.6s** | **no** | **1–2, holding their ports** |
| 0.9s | yes | 0 |
| 1.5s | yes | 0 |

3/3 reproducible at 0.6s; one run left `-m backlot serve --port 52541` and `--port 52555` alive with no parent. `attach` takes 0.9s to yield over the bundled corpus, and longer over a real one, where `serve(data_dir=…)` skips the import but still waits for `/health` to report `documents` — so the window widens exactly as the corpus grows.

It also lands where it is most likely to be hit: a client that gives up on a slow start kills the process during startup, which is this window.

`ExitStack` does not close it — the child is created inside `serve()`'s generator before it yields, so nothing is entered yet and there is nothing for the stack to unwind. What the handler needs is the child itself, not the context manager. Something like a registry `serve()` appends to right after `Popen` returns:

```python
# backlot/server.py, immediately after `proc = subprocess.Popen(...)`
_CHILDREN.append(proc)          # module-level list
# ... and in serve()'s existing finally, after _terminate(proc):
_CHILDREN.remove(proc)
```

with `_exit_on_signal` draining it before `os._exit`, so the normal path stays the context manager and the window is covered by the same `_terminate` the CM would have run. Whichever shape you prefer — the point is that between `Popen` and the yield there is a live child nothing can reach.

khj809 (commented):
Re-ran the nine, this time by mutation rather than by reading: all seven I could mutate now fail without their behaviour, including the two I could only describe before (`is_backlot`'s `documents` key, `stdin=DEVNULL` on both children). Signals measured on this head with stdin held open and the process in its own group, so only the parent is signalled — SIGTERM 143, SIGINT 130, SIGHUP 129, each in 0.3s with the started server down and no orphan. `--url` gets the 10s budget (a 3s-latency server is accepted now), the `--data-dir` notice prints, `gmail_messages_list`/`gdrive_files_list` no longer stutter, initialize is 0.61s, and adding a route whose handler shares a name with another in its source fails `test_build_mcp_spec_collapses_exactly_the_known_aliases`. 1734 passed, 0 skipped.

One correction to my earlier note: I claimed `validate_output=False` had no test after your commit too. That was my mutation being wrong, not your test — I replaced the first `validate_output=False` in the file, which is the comment on the line above, so the argument never changed. Targeting the argument, `test_openapi_bridge_passes_a_response_through_even_when_it_breaks_its_own_schema` fails exactly as intended, on the client session's structured-content check.

One new thing, in the handler itself — inline.

backlot/mcp.py
nuri-yoo: Reproduced on b7fbccb before changing anything: SIGTERM to the parent only, stdin held open, own process group, bundled corpus. At 0.5s, 0.6s and 0.7s the uvicorn child was orphaned in 9 of 9 runs (ppid 1, port held), at 0.8s in 1 of 3, at 0.3s and from 0.9s in none; here the child appeared at about 0.5s and `Backlot is up` at about 0.9s. The scratch dir with the imported corpus leaked in those runs as well, since `TemporaryDirectory` is never exited either.

c5dd072 takes your shape. `serve()` appends the child to `server._CHILDREN` right after `Popen` and removes it in its own `finally`; `server.stop_children()` terminates and reaps whatever is still there; the handler calls it after closing `_started`. It reaches the module through `sys.modules` rather than an import, so a signal that lands while the main thread is itself importing `backlot.server` cannot meet that import's lock, and with the module unimported no child can exist. Terminating an exited child is a no-op, so the handler overlapping `serve()`'s `finally` is harmless.

Test: `test_backlot_mcp_takes_down_a_server_it_spawned_but_had_not_yet_yielded` runs the real `mcp.run` and `serve()` in a driver that holds `_warm` false for ten seconds and prints the server's pid from `Popen`, signals the parent as soon as the pid appears, and asserts exit 143, no `Backlot is up`, and the child gone within 10s. With the `stop_children` call removed it fails with the orphan's pid. The sweep above re-run on this commit: 0 orphans in 21 of 21. 1735 passed, 0 skipped; ruff clean.

Left as is: the few instructions between `Popen` returning and the append. A signal exactly there still misses the child; closing it would mean masking signals around the spawn, which is not available on Windows and buys microseconds.

nuri-yoo:
CI on c5dd072 failed `test_backlot_mcp_exits_on_a_signal_with_stdin_still_open[SIGHUP]` on 3.11 only, process still alive 20s after the signal — the same symptom as ae210ef on 3.11 and 3.14. That case takes `--url`, so nothing is spawned and the handler had nothing to do but `os._exit`. It never ran.

Cause, measured rather than guessed: a Python-level signal handler runs on the main thread, and only once that thread is back in the interpreter. Here the main thread is fastmcp's event loop, asleep in select with nothing to wake it — asyncio sets a wakeup fd only when a loop signal handler is registered, and nothing in this process does (checked by spying `signal.set_wakeup_fd`: zero calls). When the kernel hands a process-directed `kill` to another thread, the C-level handler runs there and sets the flag, and the main thread sleeps on until stdin closes. Reproduced with `pthread_kill`: SIGTERM and SIGHUP to the main thread ended the process at once (143, 129); the same two to a sidecar thread left it alive past 12s, every run. Which thread receives a `kill` is the kernel's choice, so this is not version-specific — 3.11 and 3.14 are just the runs that drew the other thread. It is not from the `_CHILDREN` change either; that path is a no-op with `--url`.

286c3a9: `run` sets a wakeup fd (a socket pair held at module level — a collected socket would close the fd), installs a no-op Python handler for the side effect of the C-level handler writing the signal number there, and a daemon thread reads that fd and does the exit: close the started server, stop any registered child, flush, `os._exit(128 + N)`. New test delivers each of the three signals to a non-main thread with `pthread_kill` and requires exit within 20s; without the `set_wakeup_fd` call it fails on the timeout. 1738 passed, 0 skipped, ruff clean.

khj809 (approved):
Everything seems been addressed well. Thanks for the huge work!

PR #109, opened 2026-09-03
`backlot mcp` takes one `--user`, and S3 is a bridged source signed per request

khj809:
Stacked on #108 — base is `feat/mcp-subcommand`, so the diff here is only this change.

One flag names who is asking, and every source's credential follows from it:

```bash
backlot mcp --user ava.chen@acme.com
```

## The problem it removes

`backlot mcp` took credentials through `--token` and `--username`, and that surface leaked one special case per service. `--username` was meaningful only for Atlassian — the value a dummy Backlot discards once the token resolves — and given with any other `--source` it was accepted and silently ignored. S3 authenticates with an access-key/secret pair, and no flag could carry two values, so `--source s3` was refused outright. A caller who wanted "what does Ava see?" had to read `GET /_meta/users`, copy a token and paste it, when the thing they named was a person.

## What changed

**`--user <email>`** (`backlot/mcp.py`, `backlot/cli.py`). `--token` and `--username` are gone. `Credentials` is one caller's whole set — token, email, and the S3 access-key pair — resolved from the server's own `GET /_meta/users` and spelled per source in one place: `Bearer` for most, HTTP Basic `email:token` for Atlassian (the caller's real email now, not a placeholder), a SigV4 signature per request for S3. Without `--user` the admin sees everything, as before; the admin is the one caller the directory does not list as a person, and having no email is also what makes Atlassian fall back to Bearer, exactly as omitting `--username` did. An email the corpus does not know is refused before a single tool exists, rather than quietly resolving to the admin's unfiltered view — which a caller who named a person would have no way to notice.

**S3 is a bridged source** (`backlot/openapi.py`, `backlot/mcp.py`). `SOURCE_PREFIXES` gains `"s3": ["/s3"]`, and the bridge signs each request instead of holding a fixed header. `backlot/sigv4.py` already computes the canonical request, the signing key and the signature — the S3 router verifies inbound signatures with it — so signing calls that same `expected_signature`, and the two sides cannot be built two different ways. No new dependency: that module is standard library only. Tools are `list_buckets`, `bucket_get`, `object_get`, named for Backlot's own routes rather than the `aws s3api …` surface a vendor CLI gives. `examples/using-mcp-with-agents/s3.py` still drives awslabs' server, because pointing a real vendor client at Backlot is its own thing to show.

**S3's tools say what they take and what they do** (`backlot/routers/s3.py`). The router reads its query params off the raw request, so each has to be declared by hand (`openapi.qp`) — and none were. `bucket_get` now declares `prefix`, `delimiter`, `start-after`, `max-keys`, `continuation-token` and `location`, every one of which `_list_objects_v2` already honoured and none of which a tool could reach: a bucket holding more than `_MAX_KEYS` objects came back `IsTruncated` with no way to ask for page two. `list-type` is deliberately absent, since Backlot answers the V2 shape whether or not a caller asks for it and advertising a parameter that changes nothing is worse than not offering it. The four handlers also gained docstrings, which are what a bridge turns into tool descriptions — `bucket_get` was reaching a model described as "Bucket Get".

**Config docs are held to `Settings`** (`tests/test_docs.py`). Removing a setting left `.env.example` naming `BACKLOT_EXPOSE_TOKENS` and `docs/configuration.md` claiming "There are ten", and nothing in a repo that otherwise enforces its docs noticed. Two guards close that class: the configuration page must name exactly `Settings`' fields, with a matching table row count and count sentence, and `.env.example` may omit a setting but never name one that does not exist.

**HEAD leaves the MCP slice** (`backlot/openapi.py`). `drop_head_operations` runs before anything is renamed. A HEAD response carries no body, so a tool derived from one returns an empty string on every call — measured, before it was dropped. It also closes a hole that only S3 could reach: `_METHODS` has no `head`, so a HEAD operation is renamed by neither pass, and S3's object route — GET and HEAD under one operationId — would have shipped as `object_get` beside an un-renamed `object_get_s3__bucket___key__get` hitting the same route, the redundant pair that module exists to prevent. Dropping it rather than ranking it leaves `unique_operation_id` untouched, so the raw `/openapi.json` ids provably do not move.

**`GET /_meta/users` always answers.** `BACKLOT_EXPOSE_TOKENS` is removed, with its gates on `/_meta/users` and `/_meta/credentials`: a switch that closed them would take this command's only credential input with it. `backlot.server.meta_users` is the honest reader behind `--user`, raising on anything that is not a usable directory, and `admin_token_for` is its lenient wrapper returning the package-default guess. The asymmetry is deliberate and commented: naming a person cannot be satisfied by a guess, while asking who the admin is can, and refusing to connect over a credential the caller may never use would break a working session.

**`_trusted_for_token_fetch` is removed.** It gated the `GET /_meta/users` token fetch behind https-or-loopback, on the grounds that treating an unauthenticated plaintext response as a credential is the wrong default. The fetched value is only ever sent back to the server that supplied it — every `Server.token` consumer pairs it with the same `Server.base_url`, and it is never written to disk, logged, or sent elsewhere — so an attacker able to rewrite that response already controls the channel to that server and gains nothing by choosing the token. What the gate did cost was real: against any plain-HTTP non-loopback server it suppressed the fetch and fell back to guessing `admin-service-token`, so a deployment whose operator set `BACKLOT_ADMIN_TOKEN` failed every call with HTTP 200 / `{"ok": false}` — the failure `Server.token`'s own docstring exists to avoid.

**Instructions match what is mounted** (`backlot/mcp.py`). The composite server's `instructions` — which an MCP client feeds the model as the server's description — is built from the sources actually mounted, the note about Atlassian's two products included. A `--source slack --source s3` run no longer tells the model that `gmail_*`, `notion_*` and the rest are there to call.

**Examples and docs.** The seven launchers take `--user` in place of `--token`, and `build_params(base_url, user)` passes nothing when the caller wants the admin. The vendor-server launchers keep their vendor's own credential shape, since Backlot is not the one parsing those flags: `atlassian.py` and `notion.py` still take `--token`, `s3.py` still takes `--access-key`/`--secret-key`. `examples/using-mcp-with-agents/README.md` states the `--url` rule the way the code behaves (a named server that does not answer is an error, not a fallback) and now says `httpx2` where the bridge uses `httpx2`.

Served responses do not change. S3's signature is computed by the module the server verifies with, so a request the bridge signs is valid for exactly the reasons a boto3 request is; the region comes from the credential scope, which is how Backlot's verifier already reads it, and `us-east-1` is boto3's own default. Atlassian's Basic spelling is unchanged except that the username is now the caller's real email, which is what mcp-atlassian sends. Tool names come from Backlot's route names, not from a vendor spec — the vendor never published MCP tools for these surfaces.

## Verification

- **Tests** — `pytest` on a `.[all]` install plus `llama-index-readers-hubspot<0.6 --no-deps`: **1659 passed, 0 skipped**, 23.5s. Docker (colima), `npx` and `uvx` all present, so the mcp-atlassian, notion-mcp-server and awslabs tests ran rather than skipped.
- **Without the change** — the six changed test files copied onto this branch's parent in a worktree: **27 failed, 114 passed**.
- **End to end, outside pytest** — `backlot mcp` with no server running, over real stdio: **115 tools across all ten sources**, longest name 37 characters. One real tool call per source, 10 of 10 answering with corpus data, and the same ten calls under `--user ava.chen@acme.com` narrowed on eight of them (`gdrive` 18,593 → 15,109 bytes, `atlassian` 38,341 → 22,485, `s3` 418 → 323; `github_list_repos` and `linear_teams` are unrestricted for that person in the bundled corpus). All seven launchers' `build_params` driven through a real stdio session to a tool list, as the admin and as a scoped person, without an LLM. S3 keys containing slashes round-trip through the signed client (`builds/pipeline/2026-02-17/manifest.json`), and a bucket paged at `max-keys=2` walks its 7 keys over 4 pages through `continuation-token`, all distinct — the query string signing case as well as the paging one.
- **Lint** — `ruff check . && ruff format --check .`: clean. `python scripts/gen_docs.py --check`: exit 0 (the generated source table picked up S3's new MCP column on its own).

- [x] A test fails without this change — the 22 above.
- [x] A new endpoint serving corpus content is ACL-scoped, proved for both an admin and a scoped token — S3 joins `test_mcp_bridge_enforces_the_acl`, which resolves credentials through `backlot.mcp.resolve` and reads a restricted object as the admin and not as a scoped user.


tests/test_mcp.py
```diff
@@ -353,17 +348,28 @@ async def _go():
         lambda n: n.startswith("jira_get_issue"),
         lambda row, st: {"key": row["key"]},
         lambda t, row: '"key"' in t and '"fields"' in t,
-        "svc@example.com",  # this bridge authenticates with basic auth
+        # nothing to pass: Basic for the resolved user, Bearer for the admin, both run below
```
nuri-yoo, line 353: `reads(resolve(base, email))` only asserts that the scoped user is blocked, so the Basic spelling itself goes untested: with `_auth_header` mutated to send `creds.token[::-1]` for Atlassian, `tests/test_mcp.py` still passes 24/24 (measured on this head merged with `4a02c1b`). It is the same shape as the S3 gap `test_mcp_s3_bridge_signs_for_the_resolved_user` closes below. A positive read through Basic of an issue the user can see would make "both run below" true here too.

backlot/mcp.py
```diff
@@ -59,17 +69,107 @@
...
+    try:
+        if user is None:
+            return Credentials(
+                token=directory["admin_token"],
+                email=None,
+                s3_access_key_id=directory["admin_s3_access_key_id"],
+                s3_secret_access_key=directory["admin_s3_secret_access_key"],
+            )
+        for entry in directory.get("users", ()):
+            if entry.get("email") == user:
```
nuri-yoo: Exact match. `--user AVA.CHEN@ACME.COM` exits 1 with "is not one of the 10 users" against the bundled corpus. Atlassian and Google both treat the address case-insensitively, so an email copied in the shape a Jira page displays it can be refused for a person the corpus does have. `casefold()` on both sides would do it; the entry's own `email` is what goes into `Credentials`, so the Basic header keeps the corpus spelling.

docs/auth.md
```diff
@@ -33,8 +33,9 @@ curl -s localhost:8000/_meta/credentials   # for a Google client that wants a co
   "service_account": { "type": "service_account", "client_email": "…", "private_key": "…" } }
 ```
 
-Both serve working credentials in the clear, so `BACKLOT_EXPOSE_TOKENS=false` closes them (404).
-The same values are in `data/tokens.yaml`.
+Both serve working credentials in the clear, and always answer: `backlot mcp --user <email>`
+resolves a person to their credentials through `/_meta/users`, so a switch that closed it would
+take that command's only credential input with it. The same values are in `data/tokens.yaml`.
```
nuri-yoo, line 38: Does the switch have to go, rather than stay with `backlot mcp --user` exiting on the 404? Two cases still want it: a shared deployment that follows configuration.md's advice and sets `BACKLOT_ADMIN_TOKEN` now hands that token to anyone through `/_meta/users` with nothing to close, and one that already runs `BACKLOT_EXPOSE_TOKENS=false` comes back open without a word, since `Settings` ignores a var it no longer knows. If the answer is what supported-sources.md now says, that the corpus is a fixture and the credentials guard nothing, that settles it for me. Asking now because this is the last point at which it is free to change.

nuri-yoo (approved):
Approving. Verified on this head merged with `feat/mcp-subcommand` at `4a02c1b` (the branch now carries main): 1718 passed, 0 skipped, on a `.[all]` install with Docker, `npx` and `uvx` present; `ruff check` and `ruff format --check` clean; `gen_docs.py --check` exit 0. Over real stdio from that tree: 115 tools across the ten sources, longest name 37 characters; `--user` with an email the corpus does not have exits 1 and the server the command started is gone afterwards (its port refuses); S3 signing through the bridge round-trips keys and prefixes containing spaces, `+`, `%`, `=`, `~`, `#` and non-ASCII, each absent key answering `NoSuchKey` rather than `SignatureDoesNotMatch`.

Two things outside the diff:

- `scripts/gen_docs.py` still says `SOURCE_PREFIXES` "omits S3 deliberately" (module docstring, line 17) and that "S3 has no entry at all" (`_slice_key`, line 118). The generated table is right, and the prose that explains why is now the opposite of the code.
- A key containing `?` is refused with `SignatureDoesNotMatch` through this bridge. That is the verifier's doing rather than the bridge's, and it predates this PR: boto3 hits it on main too. Filed as #113 with the cause (`auth.py:230` takes the canonical query from `request.url.query`, which Starlette rebuilds from the decoded path).

Inline: one coverage gap in the Atlassian Basic path, one case-sensitivity nit on `--user`, and one question on the removed switch. None of them blocking.

docs/auth.md
khj809: Decided to leave it removed. Two reasons, and they are about where the project is rather than about the two cases being wrong.

Backlot is early enough that the switch was not carrying much: it closed two `/_meta/*` endpoints on a server whose whole corpus is a fixture, and it had no test, no deployment using it that we know of, and no story for the silent-reopen you point out. Weighed against that, the cost of keeping it is a server mode in which `backlot mcp --user` cannot resolve anybody — the command's only credential input, failing on a configuration the operator may not remember setting. Building that state on purpose to preserve a switch of small effect is the trade I did not want to make.

On the shared-deployment case specifically: closing `/_meta/users` would not have protected the Atlassian surface anyway. Measured just now against a real Atlassian Cloud site with a user API token, `email:token` answers 200 while `email:<empty>`, `email:<wrong>` and `<unknown-email>:token` all answer 401 — and Backlot answers 200 to all four, because `auth.resolve_basic` accepts the username alone when the password does not resolve and ignores it when it does. So on that surface an email is already enough, and emails are all over the corpus content. Both divergences are the router's, and the PR stacked on this one closes them with those numbers as its regression test.

If a deployment does end up wanting the endpoints shut, the honest shape for it is probably auth on `/_meta/*` rather than a switch that turns a working command into a broken one.

khj809:
Both inline findings and both notes outside the diff are addressed on `a30c377`. Suite there: 1721 passed, 0 skipped, on a `.[all]` install with Docker, `npx` and `uvx` present; `ruff` and `gen_docs.py --check` clean.

**`--user` was case-sensitive.** Matched with `casefold()` on both sides now, and the directory entry's own spelling is what goes into `Credentials`, so Atlassian's Basic header still carries the corpus's. `AVA.CHEN@ACME.COM` and `Ava.chen@acme.com` both resolve to `ava.chen@acme.com`.

**`scripts/gen_docs.py` no longer says `SOURCE_PREFIXES` omits S3.** Line 17 describes only the Jira/Confluence merge and the `gdrive` rename, and `_slice_key`'s docstring drops the clause.

**The Atlassian Basic gap is closed, but not the way you proposed — the proposal does not work here, and that turned out to be the interesting part.** I added the positive read you described, then measured that it cannot close the gap: `auth.resolve_basic` accepts the username alone when the password does not resolve, so a read as the resolved user succeeds whether the token is intact or reversed. Even an empty password answers 200. A positive read therefore cannot say which half of `email:token` authenticated.

So the guarantee moved to the header itself, in `test_auth_header_spells_each_source_the_way_its_client_speaks`: mangling the token, swapping the two fields, and dropping the Basic branch for Bearer are each caught there, and your `creds.token[::-1]` mutation now fails the file. The positive read stays, since it does show the path works end to end for a resolved user. I briefly had a stronger integration assertion — a read under an email the corpus does not have, which only the token could have authenticated — and then removed it, because it made a divergence its own expected answer.

That divergence is the follow-up. Measured against a real Atlassian Cloud site (`GET /rest/api/3/myself`) with a user API token: `email:token` answers 200, while an empty password, a wrong password, and a valid token under someone else's address all answer 401. Backlot answers 200 to all four. #114 (draft, stacked on this one) makes `resolve_basic` require the pair, with the admin/service token as the single exception, since `Acl.resolve` gives it no address to match a username against.

**Thanks for filing #113.** Chasing it to `auth.py:230` and confirming boto3 hits it on `main` is what made it clear it is the verifier's and not this bridge's, so I left it there rather than folding it in.

PR #110, opened 2026-09-03
github: the branch listing holds the refs its pulls advertise, and a name it omits is not a ref

### LINKED ISSUE #96: github: the branch listing denies the head refs its own pulls advertise
**Source and commit** — `github`, at `5c6dab1` (the branch listing added in #95, closing #92).

`GET /repos/{owner}/{repo}/branches` answers a fixed one-element list naming `main`, while every pull the same server serves advertises a head ref in the **same** repo that the listing does not contain.

|---|---|---|
| an open pull's same-repo `head.ref` | appears in `/branches` | absent — the listing is always exactly `["main"]` |
| a `base.ref` a corpus states as something other than `main` | appears in `/branches` | absent |
| `/branches/{name}` for a name the listing omits | 404 | 200 for any single-segment name, 404 for one containing a slash |

The third row is the older half of the contradiction and is why this is one issue rather than two: the single-branch route answers 200 for a name nobody ever stated, so "which branches does this repo have" gets three different answers depending on which route is asked.

**How you know** — measured against api.github.com on 2026-09-01, on three repos with an open pull whose head is a branch of the repo itself:

```
pydantic/pydantic: PR #13686 head.ref='codex/add-docs-navigation-dispatch' -> in /branches: True (65 branches listed)
fastapi/fastapi:   PR #16152 head.ref='optimize-workflow-cache'            -> in /branches: True (22 branches listed)
psf/requests:      PR #5693  head.ref='bug/5671'                           -> in /branches: True (6 branches listed)
```

A merged pull is the exception that a fix has to respect rather than ignore: GitHub's "automatically delete head branches" setting means a merged pull's head ref is routinely gone from `/branches` while the pull still reports it.

```python
import backlot, requests

with backlot.serve() as m:
    p = requests.get(f"{api}/pulls", headers=h, params={"state": "all"}).json()[0]
    p = requests.get(f"{api}/pulls/{p['number']}", headers=h).json()
    print(p["head"]["ref"], p["head"]["repo"]["full_name"])   # raj/streaming-export acme/pipeline
    print([b["name"] for b in requests.get(f"{api}/branches", headers=h).json()])  # ['main']
    print(requests.get(f"{api}/branches/raj/streaming-export", headers=h).status_code)  # 404
    print(requests.get(f"{api}/branches/totally-made-up", headers=h).status_code)       # 200
```

**What it costs a client** — one that reads a pull and then resolves its head against the branch listing (to build a ref picker, or to check the ref still exists before reading files at it) concludes every open pull here comes from a deleted or forked branch. `head.repo.full_name` says otherwise in the same response.

**Where it lands** — `list_branches` in `backlot/routers/github.py`, which today returns a literal `"main"`. The material for a real listing is already in the store: `head_ref` and `base_ref` per pull row (`store.py`, and `importer/byo.py` maps them from a record's `head`/`base`).

**What a fix has to settle**, none of which #92 answered:

- Which pulls contribute a head ref — open only, or merged ones too. Real deletes the head branch on merge often enough that including them is its own divergence.
- Whether `base_ref` contributes. A base branch outlives the pull, so probably yes.
- ACL. Pulls are filtered per caller, so a listing built from them varies by token — defensible (the repo listing already does), but it should be a decision rather than a side effect.
- Ordering, and the commit each entry reports. Every ref resolves to the one snapshot commit (`_repo_commit_sha`), which is what `/branches/{branch}` and `git/ref` already do.
- **The any-name-answers-200 behaviour is five surfaces, not the one in the table.** Measured on the bundled `pipeline` at `5c6dab1`, with `totally-made-up` as the ref: `/branches/{name}` 200, `git/ref/heads/{name}` 200 (the snapshot commit), `git/trees/{name}` 200 (the root tree), `/commits/{sha}` 200 (the string echoed back as the sha), `/contents/{path}?ref=` 200 (HEAD). Once the branch set is knowable, "does this name exist" is one decision in all of them, and 404ing unknown names in `/branches/{name}` alone moves the contradiction rather than closes it.
- **…but one of those five is deliberate and documented, so "404 everything unknown" is not the answer.** `?ref=` falling back to HEAD is the no-history tolerance `test_github_contents_serves_a_snapshot_by_its_stated_ref` states outright, and a corpus names snapshots with a `ref` field (`pr-1`, `pr-3` in `history-repo`) that are refs without being branches. A fix has to keep snapshot refs addressable while the branch listing stays a list of branches.
- **`?protected=` gets entangled the moment the listing has more than one entry.** Real is three-valued — a truthy value selects the protected branches, `false`/`0` the unprotected ones, an absent or empty parameter all of them (measured on fastapi/fastapi: 22 branches, one protected, answering 1 / 21 / 22). `_truthy` gives the right answers today only because the single entry is `protected: false`, and that stays coincidental with every branch a fix adds, since nothing in a corpus states protection. The docstring on `list_branches` already states real's three cases correctly (#95); a fix should keep the collapse deliberate rather than let it drift back into looking like the whole rule.

khj809:
Closes #96.

## What changed

**A repository can state its own refs.** A `subtype: "repo"` record carries `default_branch`, `branches` (each `{"name", "protected"}`) and `tags` — facts about the repository rather than about a document in it. Stated branches are the repo answering for itself: `/branches` serves exactly them, `?protected=` selects on the flag, and the pulls are not consulted. The record has no author, body or ACL of its own; it writes the container's own row (`github_repos`), the way `jira_projects.key` and `linear_teams.served_key` already do, and the repo stays visible exactly when one of its documents is. `/tags` serves what it states instead of always `[]`.

**Without such a record the branch set is inferred from the pulls** — the default branch, every pull's `base_ref`, and an **open, same-repo** pull's `head_ref` — instead of the fixed `["main"]` that contradicted every pull the same server served. A merged pull's head is left out. Inferred that way the listing is built from pulls, which are ACL-filtered, so it is scoped to the caller: the answer `/user/repos` already gives for repos. A stated listing is the same for everyone.

**A pull says when its head lives in a fork.** `head_repo` takes the full `owner/name` — the one place a corpus writes an owner, since a fork's owner is what differs. Its `head.repo.full_name` and `head.label` carry that owner, and its head is a branch of the fork, so `/branches` does not list it. Without it every pull looked same-repo. A `head_repo` naming this repo is not a fork, and its head is listed like any other.

**A name the listing omits is no longer a ref anywhere.** `_branch_rows` and `_commit_ish` are what all six routes ask, so `/branches/{name}`, `git/ref/heads/{name}`, `git/trees/{ref}`, `/statuses/{ref}` and `/contents/{path}?ref=` 404 it and `/commits/{ref}` 422s it. `git/ref` also resolves `tags/{name}` for a stated tag and `pull/{n}/head`/`…/merge` for a pull that exists, and no longer accepts the fully-qualified `refs/heads/…` spelling it used to strip — real 404s that, so accepting it let a client sending the git spelling pass here and fail in production. Four of those routes took the ref as one path segment and so 404'd a branch whose name contains a slash; they take a trailing path now, as `git/ref` already did.

The branch names and the commit shas behind those answers come from one read of the repo's pulls rather than two, which a query-counting test pins — that ACL-joined scan ran twice on every route that asks whether a ref exists.

A corpus also names snapshots of a file with a `ref`, and those remain addressable on `?ref=` — refs without being branches, so they resolve there and are absent from `/branches`.

**A repo record states the repo and nothing else.** A document field on one — a title, an author, a `visibility` — is refused at import rather than dropped, since the record registers its container and returns, so such a field is not even stored. And a stated listing must hold the base of every open pull: deleting a branch with an open pull on it closes that pull, so a corpus stating both describes a repo real cannot be in. Both the load and `--dry-run` refuse it, from one rule.

**What a record must state moved with it.** `repo` and `source_type` are all any github record needs; `author_email`, `content`, `created` and `title` are now required of a document — `issue`, `pull_request`, `file` — rather than of every record, because a repo record has none of them. The `state` rule keeps its existing shape and gains `repo` beside `file` as the subtype it does not apply to.

Both corpora carry the new record. The bundled one gains an open pull on `pipeline` and a repo record for it, so the demo shows each half: `pipeline` states its branches and tags, and `docs-site` still has them inferred. The BYO example corpus — the one whose job is to show every field a record may carry — gains a repo record and a pull from a fork, and its README says what each is for.

Growing `hello.jsonl` is what makes a documented count of it wrong, and this branch grows it: the three that existed — in `CONTRIBUTING.md`, the skill and `backlot/server.py`'s module docstring — were correct against `origin/main` and are not against this head. They now say what the corpus covers, as `docs/corpus.md` already did, and a test refuses both a count on a line that names the bundled corpus and today's count anywhere, in `.py` as well as `.md`, so the number cannot be written back in at the one moment it is still correct.

Measured against api.github.com with `gh api` on 2026-09-03; each measurement sits in the docstring of the route it decides.

- **An open pull's head is a branch, a merged pull's is not.** 27 of 27 open same-repo `head.ref`s appear in `/branches` (pydantic/pydantic, fastapi/fastapi, psf/requests, pallets/flask, sqlalchemy/alembic, encode/httpx, jupyter/notebook); 0 of 54 merged ones do, across those plus django/django, psf/black and numpy/numpy. Every `base.ref` appears, including a non-default base — pydantic/pydantic lists `pure-annotation-schema-cache`. Closed-but-unmerged heads split 6 of 28, which is why a corpus that wants the set exact states it rather than leaving it inferred.
- **A fork's head is not a branch of the base repo**, and the pull says so twice: an outside pull on pydantic/pydantic reports `head.repo.full_name` `chenlichao/pydantic` and `head.label` `chenlichao:fix/…` — the fork's owner, where `base.label` keeps `pydantic:`.
- **`?protected=` is three-valued**: a truthy value selects the protected branches, `false`/`0` the unprotected ones, an absent or empty value all of them — 1 / 21 / 22 on fastapi/fastapi, whose 22 branches include one protected.
- **Ordering** is ascending by name: `3.0`, `bug/5671`, `init_level_types`, `main`, … on psf/requests.
- **A tag item** carries `name`, `commit: {sha, url}`, `zipball_url`, `tarball_url` and `node_id`, with the archive urls spelling the ref out as `refs/tags/{name}` (psf/requests).
- **A name no branch holds**, on psf/requests: `/branches/totally-made-up`, `git/ref/heads/…`, `git/trees/…`, `/statuses/…` and `/contents/README.md?ref=…` are 404, and `/commits/…` is **422** carrying `No commit found for SHA: totally-made-up`. A 40-hex sha naming no object answers the same, so a well-formed sha is not a way in.
- **A slashed branch name** resolves on all four: `/branches/bug/5671`, `/commits/bug/5671`, `git/trees/bug/5671` and `/statuses/bug/5671` are each 200.
- **Pull refs**: `git/ref/pull/{n}/head` is 200 for a pull that exists and 404 for one that does not (pydantic/pydantic #13686 and #999999), and `…/merge` exists only while the pull is OPEN — psf/requests #7616 (merged) and #7589 (closed) answer 404 there where open #7586 answers 200.
- **A ref standing for a commit resolves to it**: `/commits/main` answers the branch's own sha on psf/requests, the value `/branches/main` reports, and a tag resolves wherever a branch does — `v2.34.2` is 200 on `/commits`, `git/trees`, `/statuses` and `?ref=`.
- **An empty `?ref=`** is an absent one (`contents/README.md?ref=` answers the file), and the 404 for a ref with no object carries `No commit found for the ref {ref}` against the contents documentation, not the generic body.
- **`git/ref` takes `heads/main`, not `refs/heads/main`**: the qualified spelling answers 404 on psf/requests with the get-a-reference endpoint's own body, so it is a missing ref rather than a missing route. The `ref` this route *answers* with stays fully qualified, as real's does.

Every sha this server advertises still resolves — a pull's `head.sha` and `base.sha`, the `commits_url` it prints, and `/pulls/{n}/commits` are all 200 on `/commits` and `git/trees` — so the new 422 breaks none of its own links.

**Still deferred**, and not a divergence a corpus can close: every ref resolves to the repo's one snapshot commit, because Backlot keeps no commit history. That is the documented no-history simplification (`get_tree`), and it is what a branch-per-commit head would need.

## Verification

- **Tests** — `pytest` on a `uv sync --all-extras` install (the `[all]` extra), with `main` merged in: 1858 tests, 1857 passed, 1 skipped, 0 failed, 0 errors.
- **Optional surfaces that ran rather than skipped** — `tests/test_s3.py` (26), `tests/test_sdk.py` (5), `tests/test_mcp.py` (42), `tests/test_llamaindex.py` (12) and `tests/test_integrations.py` (22), the last including the fsspec GitHub tests, one of which asserts the stated branches and tags through `GithubFileSystem.branches`/`.tags`/`.refs`. The one skip is `tests/test_llamaindex.py:119`, which needs `llama_index.readers.hubspot`.
- **Lint** — `ruff check . && ruff format --check .`: all checks passed, 139 files already formatted.

Paging on `/branches` and `/tags` is left out and filed as brekkylab/backlot#117: real sends `Link` on both and neither goes through `clamp_page`, which was invisible while the listing had one entry.


backlot/routers/github.py
```diff
@@ -1321,12 +1405,23 @@ async def get_branch(owner: str, repo: str, branch: str, request: Request):
...
+    route used to echo any string back as a sha with a 200, so a client that pinned to a name it
+    had misspelled read a commit that does not exist.
+    """
     conn = auth.conn(request)
     caller = _require(request)
     ids = auth.visible_ids(request, caller)
     _require_repo(conn, repo, ids)
+    sha = sha.strip("/")
+    if sha not in _commit_ish(conn, repo, ids):
+        raise _no_commit_for_sha(sha)
```
nuri-yoo, line 1450: A branch name passes this check and is then echoed as the commit: at this head `/commits/main` answers `{"sha": "main", "node_id": "MDEyOkNvbW1pdG1haW4", "url": ".../commits/main"}`. Real answers the commit's sha (psf/requests: `dae7ef63…` for `/commits/main`, the same value `/branches/main` reports under `commit.sha`), and `url`/`node_id` carry that sha. The docstring above says `/commits/main` and `git/trees/main` report the same sha; here they report `main` and the tree sha.

Every branch (and every tag, see `_commit_ish`) is the one snapshot commit, so resolving is one line before the body is built:

```python
    if not re.fullmatch(r"[0-9a-f]{40}", sha):  # a branch or tag name stands for the snapshot commit
        sha = _repo_commit_sha(repo)
```

with a pull's head/base sha (already 40-hex) left as it is. The query-counting test is unaffected.

backlot/routers/github.py
```diff
@@ -1476,6 +1578,114 @@ def _repo_commit_sha(repo: str) -> str:
...
+    """What may stand in for a commit: a branch name or a commit sha, real GitHub's own rule for
+    the `{ref}` of `/commits` and `git/trees`.
+
+    One read of the repo's pulls, shared by both halves: the branch names and the commit shas come
+    from the same rows, and fetching them separately ran that ACL-joined scan twice on every route
+    that asks whether a ref exists.
+    """
+    pulls = store.github_pull_refs(conn, repo, ids)
+    branches = _branch_rows(conn, repo, ids, pulls=pulls)
+    return {b["name"] for b in branches} | _commit_shas(repo, pulls)
```
nuri-yoo: Stated tags are missing from this set, so they are refs for `/tags` and `git/ref/tags/{name}` and for nothing else. Real takes a tag wherever it takes a branch, measured on psf/requests with `v2.34.2`: `/commits/{tag}` 200 (with the tagged commit's sha), `git/trees/{tag}` 200, `/statuses/{tag}` 200 `[]`, `/contents/README.md?ref={tag}` 200.

This head, repo record with `tags: ["v1.0"]`: `/tags` lists it and `git/ref/tags/v1.0` is 200, then `/commits/v1.0` 422 `No commit found for SHA: v1.0`, `git/trees/v1.0` 404, `/statuses/v1.0` 404, `?ref=v1.0` 404. The fsspec client this PR's integration test is written for meets it on the bundled corpus: `GithubFileSystem.refs` offers `v1.0.0`, `fs.ls("", sha="v1.0.0")` raises `FileNotFoundError`, `sha="release/2026-03"` lists the tree.

```python
    return {b["name"] for b in branches} | set(_repo_tags(conn, repo)) | _commit_shas(repo, pulls)
```

`_repo_tags` is the primary-key lookup, not the pull scan, so `test_github_a_ref_check_reads_the_repos_pulls_once` still counts one. `_ref_exists` already reads `_repo_tags` for `tags/`, so the two would agree.

nuri-yoo (commented):
Read at `e84c4e3` (base `4ea200a`); every claim below was run against that head in a worktree, and the real-side numbers against api.github.com with `gh api` today. Suite on this machine: 1702 passed (no skip; `llama_index.readers.hubspot` is installed here), `ruff check` and `ruff format --check` clean on the seven touched `.py` files.

What reproduces: the ghost bodies (`Branch not found` with the get-a-branch url, `Not Found` with the per-route urls on `/statuses`, `git/trees` and `git/ref`, the 422 with the ref echoed on `/commits`, `git/ref/refs/heads/main` 404) match this head's byte for byte; psf/requests' listing is sorted by name; `?protected=` splits 1 / 21 / 22; pydantic has 22 open same-repo pulls today and all 22 heads are in `/branches`, with `pure-annotation-schema-cache` listed; psf/requests' 16 merged same-repo heads are all gone from its listing. Every 40-hex sha this head prints across `/pulls/{n}`, `/pulls/{n}/commits`, `/pulls/{n}/files`, `/pulls/{n}/reviews`, `/pulls/{n}/comments`, `/branches`, `/tags` and `git/ref` answers 200 on both `/commits` and `git/trees` (the tree and blob shas 422 on `/commits`, as on real). `backlot import --bundled` does print the per-source breakdown the skill now points at, and `--id-map` is fine with a repo record in the corpus. Appending a shard that carries the repo record restates the refs correctly.

Three places the body's rule does not hold yet, then things worth a decision, then prose.

**A branch name given to `/commits/{ref}` comes back as the `sha`.** Real answers `/commits/main` with the commit's sha (psf/requests: `dae7ef63…`, the value `/branches/main` reports under `commit.sha`), and `url` and `node_id` carry that sha. This head answers `{"sha": "main", "node_id": "MDEyOkNvbW1pdG1haW4", "url": ".../commits/main"}`. The route's docstring says `/commits/main` and `git/trees/main` report the same sha; here one reports the tree sha and the other the string `main`. Inline.

**A stated tag is a ref for `/tags` and `git/ref/tags/{name}` and for nothing else.** Real takes a tag wherever it takes a branch: on psf/requests `v2.34.2` is 200 on `/commits/{tag}`, `git/trees/{tag}`, `/statuses/{tag}` and `/contents/README.md?ref={tag}`. On this head a repo record stating `tags: ["v1.0"]` gets `/tags` and `git/ref/tags/v1.0` at 200 and then 422 / 404 / 404 / 404 on those four. The client the fsspec test is written for meets it on the bundled corpus: `GithubFileSystem.refs` offers `v1.0.0`, and `fs.ls("", sha="v1.0.0")` raises `FileNotFoundError` where `sha="release/2026-03"` lists the tree. `_commit_ish` is the one place to add them. Inline.

**`--append` with a shard that has no repo record erases the stated refs.** Loaded a corpus whose repo record states `main` protected, `feat/x`, `tags: ["v1.0"]`: `/branches` `[(feat/x, false), (main, true)]`, `/tags` `[v1.0]`. Appended one numbered issue for the same repo and nothing else: `/branches` `[(feat/x, false), (main, false)]`, `/tags` `[]`. `write_containers` upserts `self.repo_meta.get(name, {})` for every container the shard touched, so the NULL the store comment reserves for "the corpus did not say" is also what "a later shard did not repeat it" produces. Inline with the one-line fix.

Decisions rather than fixes from me:

- **A `repo` record carrying document fields is accepted and every one of them dropped.** `{"subtype": "repo", "doc_id": "r1", "title": …, "content": …, "author_email": …, "visibility": "private", "readers": […], "state": "open", "created": …, "number": 7}` imports without a word and serves none of it. Three lines above the new check, `ref` on a non-file is refused because it "would be stored and serve nothing"; this is the same no-op, not even stored. Inline.
- **An open pull whose `base` the repo record omits.** Record states `[main, feat/x]`, an open pull says `base: develop`: `/pulls/{n}` advertises `base.ref: develop`, `/branches/develop` is 404. "Stated wins" is right for a head, since a head deleted under an open pull is a state real produces. For the base of an open pull it is not: "If the branch is associated with at least one open pull request, deleting the branch closes the pull requests." (docs.github.com, *Creating and deleting branches within your repository*). Inline.
- **`git/ref/pull/{n}/merge` for a pull that is no longer open.** Real: psf/requests #7616 (merged) and #7589 (closed, unmerged) answer 404 on `…/merge` and 200 on `…/head`; open #7586 answers 200 on both. This head answers 200 on `…/merge` for a merged pull. The docstring's measurement (pydantic #13686) is an open pull. Inline.
- **`/branches` and `/tags` do not page.** Real sends `Link` on both (`?per_page=2` on psf/requests: `rel="next"`, `rel="last"`), and this repo's own `/pulls?per_page=1` answers one item with a `Link`. `/branches?per_page=1` here answers all three, `/tags?per_page=1` both, no `Link`. Harmless while the listing had one entry; now it is the one list route that ignores `per_page`. Inline. Fine to leave out of this PR; say so and I file it as an issue.
- **`?ref=` with an empty value** is 200 on real (treated as absent, psf/requests) and 404 here. `list_branches` already reads an empty `?protected=` as absent; `_require_ref` could read `""` the same way. Inline.

Prose that no longer matches the code, or the body:

- `backlot/server.py:12` and `:17-18` say "136 records", "`source_documents` is 136" and "`documents` is 167". At this head the bundled corpus loads as `source_documents` 138 and `documents` 168 (`backlot import --bundled` prints "Loaded 168 documents"). The new docs test scans `*.md` only, so this is a third count it cannot see. The body says the two `.md` counts "were already stale before this branch": `origin/main`'s `hello.jsonl` is 136 lines and loads as `source_documents` 136, so they were current until this branch, which the paragraph's first sentence already says. And `source_documents` now counts the repo record (a 6-document corpus with a repo record reports 7 offered, 6 served): the two counts "differ on purpose" for a stated reason, parsing promotion, and a record that is not a document is a new reason.
- `backlot/schemas/README.md:131` "`created` is required everywhere" no longer holds; `:120` lists github's subtypes as `issue|pull_request`.
- "the repo stays visible exactly when one of its documents is" (schema description, store docstring, BYO README) holds for a scoped token and not for the admin: a repo record for `pipeline` plus one issue in another repo gives the admin `/user/repos` `['other', 'pipeline']` and `/repos/{o}/pipeline/branches` 200. `_repo_visible` answers true for `ids is None` whenever the container row exists, and a repo record now creates that row without a document, which nothing could before this branch. Inline on the store docstring.
- `head_repo` naming this repo itself, reachable only by a corpus that writes its own org as the owner, so latent rather than likely: `"head_repo": "acme/pipeline"` on an open pull of `acme/pipeline` yields `head.repo.full_name` `acme/pipeline`, `head.label` `acme:feat/own`, and `/branches/feat/own` 404, the contradiction in the title, from a record the schema accepts. Inline.
- `?ref=totally-made-up` on real is `{"message": "No commit found for the ref totally-made-up", "documentation_url": "https://docs.github.com/v3/repos/contents/"}`; this head answers `Not Found` with the get-repository-content url. Inline.

backlot/importer/byo.py
```diff
@@ -2455,6 +2479,24 @@ def write_containers(self) -> None:
                     "served_key=excluded.served_key",
                     (name, group_id, synth.linear_team_id(name), self._linear_team_key(name)),
                 )
+            elif src == "github":
+                # The repo's own refs, from the `subtype: "repo"` record if the corpus wrote one.
+                # `_j(None)` stays NULL, which is what tells "the corpus did not say" (infer the
+                # branches from the pulls) apart from a stated empty list (a repo with none).
+                meta = self.repo_meta.get(name, {})
```
nuri-yoo, line 2568: `--append` with a shard that does not repeat the repo record writes NULL over what an earlier load stated. Measured: first load, repo record states `main` protected + `feat/x` + `tags: ["v1.0"]` → `/branches` `[(feat/x, false), (main, true)]`, `/tags` `[v1.0]`. Append one numbered issue for the same repo, nothing else → `/branches` `[(feat/x, false), (main, false)]`, `/tags` `[]`. The stated listing silently became the inferred one. (A shard that DOES carry the record restates correctly; measured too.)

The comment above reserves NULL for "the corpus did not say"; after this, "a later shard did not repeat it" produces the same NULL. `group_id` has the same last-write rule, but an appended record always states a group (or its slug), whereas a shard states nothing about refs unless it carries the record. Keeping the stored value when the shard is silent:

```sql
default_branch=COALESCE(excluded.default_branch, default_branch),
branches=COALESCE(excluded.branches, branches), tags=COALESCE(excluded.tags, tags)
```

A repo record that wants to clear a stated set can still say `"branches": []`, since `_j([])` is `"[]"`, not NULL.

backlot/importer/byo.py
```diff
@@ -1564,6 +1573,31 @@ def add(self, rec: dict, where: str = "record") -> None:
...
+        containers[(src, container)] = group
+        if group:
+            groups.add(group)
+
+        # A `subtype: "repo"` record is a statement ABOUT the repo, not a document in it: no
+        # author, no body, no grant, and no `doc_id` to hash out of a body it does not have. It
+        # writes the container's own row in `write_containers`, with every other container, and
+        # stops here. Its `group` is read like any record's, so a corpus can name the repo's owning
+        # group without also having to state a document in it.
+        if src == "github" and rec.get("subtype") == "repo":
```
nuri-yoo, line 1665: This returns before anything else in the record is read, so a `repo` record carrying document fields is accepted and every one of them dropped. Measured: `{"subtype": "repo", "repo": "pipeline", "doc_id": "r1", "title": "t", "content": "c", "author_email": "ava@acme.com", "visibility": "private", "readers": ["user:ava@acme.com"], "state": "open", "created": "2026-01-01T00:00:00Z", "number": 7}` imports without a word; `visibility`, `readers`, `number` and the rest serve nothing. The `ref`-on-a-non-file check a few lines up refuses the same kind of no-op with "would be stored and serve nothing", and this one is not even stored.

Fits `_github_pairing_errors`: for `subtype == "repo"`, refuse any key outside `{source_type, subtype, repo, group, default_branch, branches, tags}`.

backlot/routers/github.py
```diff
@@ -1476,6 +1578,114 @@ def _repo_commit_sha(repo: str) -> str:
...
+
+    Scoped to `ids` when inferred: pulls are ACL-filtered, so the branches drawn from them are too.
+    Deliberate, and the same answer `/user/repos` already gives — a caller who cannot see the pull
+    is not told its branch exists — rather than a side effect of where the material happens to
+    live. A STATED listing is the same for every caller, because the record it comes from carries
+    no ACL of its own (see `store.github_repo_meta`).
+    """
+    meta = store.github_repo_meta(conn, repo)
+    default = (meta["default_branch"] if meta else None) or _DEFAULT_BRANCH
+    if meta is not None and meta["branches"] is not None:  # stated: the pulls are not read at all
```
nuri-yoo, line 1671: "Stated wins" is right for a head: a head deleted under an open pull is a state real produces. For the BASE of an open pull it is not. Measured here: record states `[main, feat/x]`, an open pull states `base: develop` → `/pulls/{n}` advertises `base.ref: develop`, `/branches/develop` is 404, the listing omits it. Real cannot be in that state: "If the branch is associated with at least one open pull request, deleting the branch closes the pull requests." (docs.github.com, *Creating and deleting branches within your repository*).

Two ways out, your call: the importer refuses a stated `branches` that omits an open pull's `base` (a cross-record check, as `resolve_jira_keys` already does for keys), or a stated listing gains the bases of open pulls. Either way this docstring should also say that a closed-unmerged head is omitted from the inferred listing; today only `test_github_a_stated_branch_listing_replaces_the_inferred_one`'s docstring carries the 6-of-28 that decided it.

backlot/routers/github.py
```diff
@@ -1074,7 +1090,25 @@ async def get_git_ref(owner: str, repo: str, ref: str, request: Request):
...
+    can see, so a ref is not a way to learn a restricted pull exists."""
+    head, _, name = ref.partition("/")
+    if head == "heads":
+        return bool(name) and name in _branch_names(conn, repo, ids)
+    if head == "tags":
+        return bool(name) and name in _repo_tags(conn, repo)
+    if head == "pull":
+        number, _, kind = name.partition("/")
+        if kind not in ("head", "merge") or not number.isdigit():
+            return False
```
nuri-yoo, line 1107: `…/merge` exists on real only while the pull is open. psf/requests: #7616 (merged) and #7589 (closed, unmerged) answer 404 on `git/ref/pull/{n}/merge` and 200 on `…/head`; open #7586 answers 200 on both. This head answers 200 on `…/merge` for a merged pull (measured on a corpus pull with `merged_at` set). The pydantic #13686 in the docstring is open, which is why it did not show. `row["state"] == "open"` for `kind == "merge"` is the whole fix; `github_by_number` returns `state`.

backlot/routers/github.py
```diff
@@ -1251,68 +1307,96 @@ async def get_blob(owner: str, repo: str, sha: str, request: Request):
...
-            },
-            "protected": False,
-        }
+        {"name": b["name"], "commit": {"sha": sha, "url": url}, "protected": b["protected"]}
+        for b in rows
     ]
 
 
 @router.get("/repos/{owner}/{repo}/tags")
 async def list_tags(owner: str, repo: str, request: Request):
```
nuri-yoo, line 1374: Real pages this listing and `/branches`: `?per_page=2` on psf/requests answers two items with `Link: …; rel="next", …; rel="last"` on both routes. This head: `/branches?per_page=1` answers all three of `pipeline`'s branches, `/tags?per_page=1` both tags, no `Link`, while `/pulls?per_page=1` on the same server answers one with a `Link`. Ignoring `per_page` was invisible while the listing had one entry; these are now the two list routes on the router that do not go through `clamp_page` and the link header. Fine to keep out of this PR; say so and I file it as an issue.

backlot/store.py
```diff
@@ -2505,6 +2513,51 @@ def github_by_number(conn, repo, number, visible_ids=None) -> sqlite3.Row | None
...
+    ).fetchall()
+
+
+def github_repo_meta(conn, repo) -> sqlite3.Row | None:
+    """The repo's own row — its stated `default_branch`, `branches` and `tags`, any of which may
+    be NULL for a corpus that stated none.
+
+    Not ACL-scoped, and it does not need to be: a `subtype: "repo"` record carries no document and
+    no grant, so this row reveals nothing a caller could not already see. Whether the repo exists
+    FOR a caller is decided by its documents, which is `routers.github._require_repo` and is
```
nuri-yoo: Holds for a scoped token, not for the admin. Measured: a repo record for `pipeline` plus one issue in `other`, and nothing else in `pipeline` → admin `/user/repos` lists `['other', 'pipeline']` and `/repos/{o}/pipeline/branches` answers 200 with the stated listing. `_repo_visible` returns true for `ids is None` as soon as the container row exists, and a repo record now creates that row with no document behind it, which no record could do before this branch. Either this sentence (and the schema description and the BYO README, which say the same) gets "for a scoped caller", or `_repo_visible` counts documents for the admin as well.

backlot/routers/github.py
```diff
@@ -1860,6 +2070,11 @@ def _pr_obj(
     src = repo_files if repo_files is not None else _RepoFiles(conn, repo, ids)
     if files is None:
         files = _pr_files(conn, owner, repo, row, api_base, ids, src)
+    # `head_repo` is a full `owner/name` precisely because a fork's owner is what differs; an
+    # unstated one means the head is a branch of this repo, under this owner.
+    head_repo = row["head_repo"] or f"{owner}/{repo}"
```
nuri-yoo, line 2127: `head_repo` naming this repo itself. Measured: `"head_repo": "acme/pipeline"` on an open pull of `acme/pipeline` → the pull says `head.repo.full_name` `acme/pipeline` and `head.label` `acme:feat/own`, and `/branches/feat/own` is 404, since `_branch_rows` reads any non-null `head_repo` as a fork. That is the contradiction in the PR title, produced by a record the schema accepts. Latent rather than likely — a corpus has to write its own org as the owner to reach it — but the BYO README calls `head_repo` "the one place a corpus writes an owner", so someone will. Cheapest fix is here or in `_branch_rows`: treat a `head_repo` equal to the served `owner/repo` as same-repo; or have the importer refuse one whose name is `repo` and whose owner is the org.

backlot/routers/github.py
```diff
@@ -1198,6 +1240,20 @@ async def _contents_response(
...
+    """404 unless `ref` names something on this repo, which real does for a ref it has no object
+    for (`?ref=totally-made-up` on psf/requests).
+
+    Wider than the branch listing by one set: a corpus NAMES the snapshots of a file it states
+    (`store.github_file_refs`), and those refs are how an older revision is addressable at all.
+    They are refs without being branches, so they resolve here and are absent from `/branches`.
+    """
+    if ref is None or ref in _commit_ish(conn, repo, ids):
+        return
+    if ref not in store.github_file_refs(conn, repo, ids):
```
nuri-yoo, line 1273: Two small things on this check. `?ref=` with an EMPTY value is 200 on real (psf/requests `contents/README.md?ref=`, the file as if `ref` were absent) and 404 here, since `""` is neither a commit-ish nor a file ref; `list_branches` already reads an empty `?protected=` as absent. And real's body for a ref it has no object for is `{"message": "No commit found for the ref totally-made-up", "documentation_url": "https://docs.github.com/v3/repos/contents/", "status": "404"}`; this head answers `Not Found` with the get-repository-content url. `_no_commit_for_sha` already builds the `/commits` body the same way, so this is one more `github_body` on the exception.

backlot/store.py
```diff
@@ -760,7 +760,15 @@ def id_seed_scope(source_type: str) -> str | None:
 CREATE TABLE IF NOT EXISTS slack_channels    (channel TEXT PRIMARY KEY, group_id TEXT);
 CREATE TABLE IF NOT EXISTS gmail_mailboxes   (mailbox TEXT PRIMARY KEY, group_id TEXT);
 CREATE TABLE IF NOT EXISTS gdrive_folders    (folder  TEXT PRIMARY KEY, group_id TEXT);
-CREATE TABLE IF NOT EXISTS github_repos      (repo    TEXT PRIMARY KEY, group_id TEXT);
+-- `default_branch`, `branches` and `tags` are the repo's own refs, stated by a
+-- `subtype: "repo"` record. Container-level facts on the container's own row, for the same reason
+-- `jira_projects.key` and `linear_teams.served_key` are: one row per repo, and no document owns
+-- them. NULL is "the corpus did not say", which is not the same as an empty list — an unstated
+-- branch set is inferred from the repo's pulls, a stated empty one is a repo with no branches.
```
nuri-yoo, line 767: "a stated empty one is a repo with no branches" is not what `_branch_rows` does: `branches: []` lists the default branch (measured: `[main]`), and the schema description says so ("The default branch is listed even when this omits it"). One of the two sentences.

examples/bring-your-own-corpus/README.md
```diff
@@ -98,6 +98,8 @@ See `sample_corpus.jsonl` for a fully-populated record of every source type.
   things, and only the latter has recipients and a Message-ID of its own. A message's `content`
   may be empty — a header-only auto-ack is still a message, and dropping it would renumber the
   rest of the thread.
+- **GitHub repositories:** a `subtype: "repo"` record states facts about the repository rather than about a document in it — `default_branch`, `branches` (each `{"name", "protected"}`) and `tags`. Stated `branches` are the repo's answer for itself: `GET /branches` serves exactly them, `?protected=` selects on the flag, and the pulls are not consulted. Without such a record the branch set is inferred from the pulls — the default branch, every pull's `base`, and an open pull's `head` — which is right for an open pull and cannot know that a closed pull's branch was deleted. The record carries no author, body or ACL: the repo stays visible exactly when one of its documents is.
```
nuri-yoo: "cannot know that a closed pull's branch was deleted" reads as though a closed pull's head is listed and might be stale. The inferred listing omits every closed pull's head, merged or not; what it cannot know is whether that branch survived. Something like: "…and an open pull's `head`; a closed pull's head is left out, since whether that branch survived is not knowable from the pull."

khj809:
Thanks — every one of these reproduced. Addressed at `346d1f9`, in three commits: `6a6d347` the bugs, `2edd700` the decisions and the prose, `346d1f9` the docstring line below. Suite 1706 passed, 1 skipped (`llama_index.readers.hubspot` not installed here), `ruff check` and `ruff format --check` clean.

**The three bugs.**

- **A branch name came back as the `sha`.** A ref that is not already 40-hex now resolves to the repo's snapshot commit before the body is built, so `/commits/main` reports what `/branches/main` reports under `commit.sha`, and `url` and `node_id` carry it. Your one-liner, at `get_commit`.
- **A stated tag was a ref for two routes.** `_commit_ish` returns the tags too, so `/commits/{tag}`, `git/trees/{tag}`, `/statuses/{tag}` and `?ref={tag}` all answer. Your fsspec case is the check I kept: on the bundled corpus `fs.ls("", sha="v1.0.0")` now lists `['README.md', 'src']` where it raised. `_repo_tags` is the primary-key lookup, so the pull-scan count is still one.
- **`--append` erased the stated refs.** `COALESCE` on each of the three columns, as you wrote it. A record clearing a stated set with `"branches": []` still works, since `_j([])` is `"[]"`.

**The decisions.**

- **A `repo` record carrying document fields** is refused: anything outside `{source_type, subtype, repo, group, default_branch, branches, tags}`. `group` stays — the container's owning group is a fact about the repo.
- **An open pull whose base a stated listing omits** is refused at import, the first of your two options. One rule (`_unlisted_open_bases`) serves the load and `--dry-run`, since a corpus must not pass the check that spares the author a failed import and then fail it; under `--append` the listing is read from this run's records first and the stored row second, which works because `resolve_cross_references` runs ahead of `write_containers`.
- **`pull/{n}/merge`** is 200 only while the pull is open, with your psf/requests measurement in the docstring beside pydantic's.
- **`?ref=`** reads an empty value as absent, and its 404 carries `No commit found for the ref {ref}` against the contents documentation.
- **Paging** is out of this PR, as you offered — filed as #117, with the `?protected=`-filters-before-the-slice measurement in it.

**The prose.**

- `backlot/server.py`'s three counts are gone, and the docs test scans `.py` as well as `.md` — thank you for catching that the sweep was markdown-only. It also now says that `source_documents` counts a record that is not a document, which is the second reason the two totals differ.
- You are right that the two markdown counts were **current on `origin/main`** and that this branch made them stale. The PR body said "already stale before this branch" and the test docstring said the same; both are corrected.
- `created` is required of a document rather than everywhere, and the subtype list in `backlot/schemas/README.md` is `issue|pull_request|file|repo`.
- **Repo visibility**: I took the prose half of your either/or rather than changing `_repo_visible`. The schema description, the store docstring and the BYO README now say visible to a scoped caller when a document is, and to the admin as soon as the record exists — a repo stated with nothing in it being the case where those differ. Happy to make it count documents for the admin as well if you would rather the two agree; that is a behaviour change on a path no corpus could reach before this branch, so I did not make it unasked.
- **`head_repo` naming this repo** is no longer read as a fork: `_branch_rows` takes the `owner` and compares, the way `_pr_obj` already resolved the same field, so the two cannot disagree. Adding that record to the fixture turned the existing listing test red, which is how it is pinned.
- The BYO README no longer implies a closed pull's head is listed, and `_branch_rows`' docstring now carries the closed-unmerged case and the 6-of-28 that decided it — that line was the last thing outstanding.

One thing your review changed beyond the code: the docs test I added did not fail when I mutated a doc, because it filtered `.git`/`.venv`/`.claude` on each file's **absolute** parts, and this worktree lives under `.claude/worktrees/`, so it skipped every file and passed by finding nothing. It filters on the repo-relative path now, and both directions are mutation-checked.

backlot/importer/byo.py
```diff
@@ -2588,6 +2714,31 @@ def resolve(src: str, seed, what: str, whose: str):
...
+        if self.gh_open_bases:
+
+            def _listed(repo: str) -> set[str] | None:
+                stated = self.repo_meta.get(repo, {}).get("branches")
+                if stated is None:
+                    row = store.github_repo_meta(conn, repo)
+                    if row is None or row["branches"] is None:
+                        return None
+                    stated = json.loads(row["branches"])
+                default = self.repo_meta.get(repo, {}).get("default_branch") or "main"
```
nuri-yoo: One case where this check and the served listing disagree, reached only through `--append`. `_branch_rows` lists the stated `default_branch` even when `branches` omits it, but when the listing comes from the STORED row this falls back to `"main"` rather than `row["default_branch"]`, since `self.repo_meta` is empty for a shard that carries no repo record.

Measured at `346d1f9`. Repo record `default_branch: "trunk"`, `branches: [{"name": "feat/x"}]`. As one corpus with an open pull `base: "trunk"`: load OK, `--dry-run` `OK: 3 records valid`, and `/branches` serves `trunk`. Same repo record loaded first, then the pull alone under `--append` (with `number`): `line 1: base 'trunk' is not a branch the subtype: "repo" record for 'pipeline' lists, and an open pull's base must exist`. The stored row reads `pipeline|trunk|[{"name": "feat/x"}]`, so the branch is there; only the default the check unions in is wrong.

Read the default from wherever the listing came from: `row["default_branch"]` when `stated` was loaded from the row, since the two travel together on the record and on the row alike.

nuri-yoo (commented):
Re-ran every item at `346d1f9`; all twelve hold, and the suite here is `1706 passed` with no skips, `ruff check` and `ruff format --check` clean.

Routes: `/commits/main` now answers `3625fc3f…`, the same value `/branches/main` reports under `commit.sha`, and `url`/`node_id` carry it. The stated tag `v1.0` answers 200 on `/commits`, `git/trees`, `/statuses` and `?ref=`, with `/commits/v1.0` resolving to the same sha as `/commits/main`. `git/ref/pull/{n}/merge` is 200 for the three open pulls and 404 for the merged one, `…/head` 200 for all four. `?ref=` empty is 200; `?ref=ghost` on `/contents` and `/readme` both carry `No commit found for the ref ghost` against the contents documentation URL. With no repo record, an open pull whose `head_repo` is `acme/pipeline` lists `feat/own` and one from `ext/pipeline` does not, a closed-unmerged head is omitted and its base is kept.

Importer: `--append` of a numbered issue leaves the stored row at `main | [main(protected), feat/x] | [v1.0]`; a record with `branches: []`, `tags: []` clears both to `[]`. A repo record carrying document fields is refused on load and on `--dry-run` with the field list. An open pull whose base the stated listing omits is refused on load, on `--dry-run`, and under `--append` against the stored listing; one whose base is listed appends fine.

Prose: the mutation is what I checked. Putting `136 records` back into `backlot/server.py`'s docstring fails `test_no_prose_states_the_bundled_corpus_size` with `backlot/server.py:12: '136 records'`, and reverting passes it. The prose half of the visibility either/or is fine by me; leaving `_repo_visible` alone on a path no corpus reached before is the right call unasked.

One thing left, inline: the `--append` base check falls back to `"main"` for the default when the listing comes from the stored row, so a repo whose default is not `main` has a valid open pull refused on `--append` that the same corpus passes as one load and as a `--dry-run`. Approving as soon as that line reads the row's default.

khj809:
Fixed at `3335fa8`. Your diagnosis was exact: `_listed` read `branches` from the stored row and the default from `self.repo_meta`, which is empty for a shard carrying no repo record, so the union got `main` instead of the row's `trunk`.

The default now comes from wherever the listing came from — the record's when the record supplied `branches`, the row's when the row did. Your case, on all three paths:

```
one corpus      : OK
--dry-run       : OK
--append        : OK
--dry-run (bad) : REFUSED     # base 'nope', still refused
one corpus (bad): REFUSED
```

The test that pins it loads the same two records as one corpus and as an append and requires both to reach the same verdict, which is the property that was broken rather than the symptom.

One thing beyond the line you marked: the fallback was written `or "main"` here and `_DEFAULT_BRANCH = "main"` in the router, and this bug is what two copies of it look like. It is `store.GITHUB_DEFAULT_BRANCH` now, next to the column, and both the router and the importer read it — the serving default and the default the check unions in cannot disagree again. The `--dry-run` branch used the same literal and now reads it too.

Suite 1707 passed, 1 skipped, `ruff check` and `ruff format --check` clean.

nuri-yoo (approved):
Re-ran the `trunk` case at `3335fa8`: a repo record with `default_branch: "trunk"` and `branches: [{"name": "feat/x"}]`, plus an open pull with `base: "trunk"`, loads as one corpus, passes `--dry-run`, and now appends after the record was loaded alone — the three paths agree. The refusal still holds where it should: `base: "develop"` is refused both as one corpus and under `--append`. Suite here is `1707 passed`, no skips, `ruff check` and `ruff format --check` clean.

Folding the two `"main"` literals into `store.GITHUB_DEFAULT_BRANCH` is the right shape for it — the router's serving default and the importer's check now read one value. Approving.

PR #111, opened 2026-09-03
linear: the served schema declares Linear's own types on the 47 fields that contradicted it

### LINKED ISSUE #101: Linear's schema has drifted from the one Backlot serves, in 47 breaking ways

nuri-yoo:
Closes #101.

## What changed

`backlot/graphql/linear.graphql` now declares each of the 47 fields #101 lists at the type `api.linear.app` reports for it, and the resolvers produce that type rather than only declaring it. Grouped the way the issue groups them:

**Date comparators (12).** `DateComparator` and `NullableDateComparator` take `DateTimeOrDuration`, and `IssueFilter.dueDate` is a `NullableTimelessDateComparator` over `TimelessDateOrDuration`. Both scalars are declared with Linear's grammar bound onto them: extended-form ISO 8601 timestamps (`T` or a space before a time of at least `HH:MM`), the year and year-month shortcuts (`"2021"`, `"2026-02"`), and signed ISO 8601 durations relative to now (`"-P2W1D"`, `"PT1H"`, every component through `"-P1Y2M3W4DT5H6M7S"`). `backlot.graphql.engine.Engine` gained a `scalars` argument for this, so the coercion runs where graphql-core runs it, before any resolver: a malformed operand is a 400 validation error naming the scalar, which is where Linear rejects it, and each scalar names itself the way Linear's does (`DateTimeOrDuration`, `TimelessDate`). A bare epoch (`"1700000000"`), the hour-only time `"2021-03-05T10"` and `"PT"` are rejected because Linear rejects them; Backlot used to accept an epoch. The basic and week forms Python's parser would take (`"20210305T100000"`, `"2021-W10"`) are refused on the measured rejection of `"20210305"`, and a duration that leaves the calendar (`"P9999Y"`) answers the scalar's message rather than the interpreter's. A `TimelessDateOrDuration` operand that carries a time is read down to its UTC day. `in` / `nin` are added to both date comparators while the operand type was being changed, which closes four baselined gaps. `null` on a date comparator is also fixed along the way: the old code ran the Boolean through the date coercion, so `completedAt: {null: true}` was an error.

**Null rows under `neq` and `nin`.** Measured while checking the comparators above, and the opposite of what the compiler did. On a field of the issue Linear drops null rows under `neq` and keeps them under `nin`; on a relation (`project: {name: {neq: …}}`, `assignee: {name: {neq: …}}`) it keeps the issues without the relation under both. `_Comparator` now renders those two rules (`relation=True` for the columns a nested filter maps onto), where it used to keep nulls under `neq` everywhere and drop them under `nin` everywhere. The existing test that pinned the old `neq` rule is replaced by one pinning the measured one.

**Null rows in the in-memory string matcher.** The attachment filters are evaluated in Python over one issue's attachments, and that matcher had its own rules: it treated an absent value as `""` (so `subtitle: {eq: ""}` and `subtitle: {neq: "zzz"}` both matched an attachment with no subtitle) and did not know the `null` operator at all (`subtitle: {null: true}` was `unsupported comparator 'null'`, on a field whose comparator declares it). Measured against an attachment created with no subtitle and no source: `null: true` selects it, `nin` keeps it, and `eq` (even `eq: ""`), `neq`, `neqIgnoreCase` and `notContains` drop it, the same field rule as a null number or date. Its `sourceType` is served as `"unknown"` there, and no `sourceType` predicate matches it, `nin` included. Backlot now serves `"unknown"` for an attachment whose corpus records no source, filters `sourceType` on what the corpus recorded, and applies the two rules above.

**An unset priority, and an unrecorded edit.** `Issue.priority` is served as `0` when the column is NULL (Linear's "no priority"; the field is non-null there), but the filter and the sort read the raw column, so `priority: {eq: 0}` and `lt: 1` missed the issue a client had just read `priority: 0` from, `nin: [0]` returned it, `null: true` answered a field that is never null, and an ascending sort put it last. Both now go through `store.LINEAR_PRIORITY_EXPR` (`COALESCE(priority, 0)`). `Issue.updatedAt` had the same split: served and sorted as `COALESCE(updated_ts, created_ts)` for an issue with no recorded edit, filtered on the raw column, so `eq` on the served value, `gte` and `lt` missed that issue, and once this branch made `neq` compare a field directly, `neq` dropped it too, where Linear's non-null `updatedAt` cannot drop a row (the review's reproduction: `{updatedAt: {neq: "2020-01-01T00:00:00Z"}}` answered one issue of two on this branch and both on `main`). The filter and both sort maps now read one `store.LINEAR_UPDATED_EXPR`.

**Enums (11 fields, 10 enums).** `ExternalSyncService`, `IntegrationService`, `IssueSharedAccessDisallowedField`, `FrequencyResolutionType`, `ProjectUpdateHealthType`, `DateResolutionType`, `Day`, `ReleaseNoteGenerationStatus`, `TeamVisibility` are declared with the vendor's full member lists, and `ExternalEntityInfo.id` is `String!`. `Project.frequencyResolution` serves `weekly`, which is what a freshly created project answers; the string `"week"` it served before is not a member of the enum.

**Counts (8).** `Int!` in the SDL and integers from the resolvers, so the wire carries `0` and not `0.0`.

**Filter and sort inputs (7).** `EstimateComparator` with `and` / `or` implemented in the SQL comparator (an empty list constrains nothing, as measured). `IssueIDComparator`, whose values resolve as `issue(id:)` resolves them: a served UUID or an identifier whose key is in either case (`eng-1` finds `ENG-1`, as `bre-1` finds `BRE-1` on Linear); the suffix is compared as written, since the corpus schema lets it be any text where Linear's is digits, so `eng-1a` reaches a stored `ENG-1a` and `ENG-1A` names nothing. A value that names no issue and is neither a UUID nor Linear's `KEY-digits` shape answers Linear's `Argument Validation Error` (code `INVALID_INPUT`, a 200 with `errors`). The lookup comes before the shape check because the corpus schema lets a served identifier carry a non-numeric suffix, so nothing Backlot serves is refused. `EntityIdentifierIDComparator` matches a project's served UUID and answers an empty page for anything else, as Linear did for `"PROJ-1"`, `"not-a-uuid"` and an unknown UUID. On both comparators a UUID counts only with an RFC 4122 variant nibble (8-b): the nil UUID, `1111…1111` and `ffff…ffff` are `Argument Validation Error` on Linear while v1, v4 and v5 UUIDs with a variant of 8-b are looked up, so the variant is checked and the version is not (`linear_filters.UUID_SHAPE`), and on `project.id` at any depth of `and` / `or`, as the issue-id side already did. `IssueFilter.creator: NullableUserFilter`, whose `null` the compiler already handled. `SourceTypeComparator` declared in full, with the six operators Backlot's string comparators did not have (`notContains`, `notContainsIgnoreCase`, `startsWithIgnoreCase`, `notStartsWith`, `notEndsWith`, `containsIgnoreCaseAndAccent`) implemented in the in-memory matcher that evaluates it. `UserSortInput.name: UserNameSort` with `nulls` defaulting to `last` as upstream, and `users(sort:)` now applied over the served `name`; it was declared and ignored, so a client asking for Z to A was answered A to Z.

**Surface the vendor does not have (5).** `Issue.attachments(url)`, `IssueFilter.branchName`, `ReleaseFilter.slugId`, `UserSortInput.createdAt` and `UserSortInput.updatedAt` are gone; each is a 400 now, as it is on Linear. `store.linear_attachments` loses its `url` parameter, whose docstring attributed it to Linear ("Linear's own exact-match argument on this connection"). That attribution was wrong, and the docstring now records the measurement instead.

**Nullable where Linear is non-null (4).** `Issue.sharedAccess!` (the resolver already produced it), `Release.stage!`, `Release.pipeline!`, `ReleaseNote.pipeline!`. A release always sits in one stage of one pipeline, so `_release` serves an id-only stub for each, keyed off the release name through two new `synth` helpers, and the by-id `release` root answers the same ids as the connection.

**Baseline.** `backlot/fidelity/baseline/linear.json` drops the four resolved gaps and carries `measured: 2026-09-03`. 778 acknowledged, 0 new.

Introspection of `https://api.linear.app/graphql` on 2026-09-03 with a personal API key, bare in the header. The 47 types and 10 member lists in `tests/test_linear.py::LINEAR_TYPES` / `LINEAR_ENUMS` are copied from that response, so the suite pins them offline and `backlot diff --source linear` re-measures them live.

Behaviour measured against the live API the same day, each quoted in the comment beside the code that reproduces it: the accepted and rejected operand table for `createdAt: {gt: X}` (accepted `"2021"`, `"2021-03"`, `"2021-03-05 10:00"`, `"2021-03-05T10:00:00+09:00"`, `"1"`, `" 2021 "`, `"P1D"`, `"-P1D"`, `"+P1D"`, `"PT1H"`, `"P1M"`, `"-P1Y2M3W4DT5H6M7S"`; rejected `"1700000000"`, `"20210305"`, `"2021-03-05T10"`, `"P"`, `"PT"`, `"2021-13-01"`, `""`, and any non-string literal or variable), with the error text `Expected value of type "DateTimeOrDuration", found "P"; Unable to parse value 'P' into a valid date`; `Field "createdAt" is not defined by type "UserSortInput"`; `Unknown argument "url" on field "Issue.attachments"`; `Field "branchName" is not defined by type "IssueFilter"`; a real issue answering `customerTicketCount: 0`, `priority: 0` and the `sharedAccess` object; a real team answering `issueCount: 4`, `ledInitiativeCount: 0`, `visibility: public`; `users(sort: [{name: {order: Descending}}])` ordering by name. On `IssueFilter.id`: `"BRE-1"`, `"bre-1"` (both find BRE-1), `"B2E-1"`, `"ZZZ-404"` and an unknown UUID are looked up and answer a page; `"not-a-uuid"`, `"BRE-x"`, `"BRE-"`, `"BRE1"` and the nil UUID answer `Argument Validation Error` with code `INVALID_INPUT`, as does a list holding one such value or `neq` over one. On `project: {id: …}`: `"PROJ-1"`, `"not-a-uuid"`, `"proj-1"` and an unknown UUID each answer an empty page; only the nil UUID is refused.

The workspace the key reaches had no project and no issue with a due date or an estimate, so those were measured against a throwaway project and a throwaway issue (due 2026-03-15, estimate 3) created with `projectCreate` / `issueCreate` and deleted right after (both deletes answered `success: true`; the workspace lists neither now). From that: a new project answers `frequencyResolution: weekly`, `health: null`, `priority: 0`, `startDateResolution: null`, `updateRemindersDay: null`. On `dueDate`, `eq` matched `"2026-03-15"`, `"2026-03-15T00:00:00Z"`, `"2026-03-15T23:59:59Z"` and `"2026-03-16T02:00:00+09:00"` (17:00Z on the 15th) and did not match `"2026-03-15T23:59:59-05:00"` (04:59Z on the 16th) or `"2026-03-15T02:00:00+09:00"`; `gte: "2026-03-15T12:00:00Z"` matched and `gt: "2026-03-15T00:00:00Z"` did not; `eq: "2026-03"` missed and `gte: "2026-03"` hit; `lt: "P0D"` hit the past due date. `estimate: {neq: 99}` over four null estimates answered nothing, `estimate: {nin: [99]}` answered all four, `dueDate: {nin: ["2026-03-15"]}` answered exactly the issues with no due date, `completedAt: {neq: "2020-01-01"}` answered nothing and `completedAt: {nin: ["2020-01-01"]}` everything; `project: {name: {neq: "zzz"}}`, `project: {name: {nin: ["zzz"]}}`, `assignee: {name: {neq: "zzz"}}` and `assignee: {id: {neq: …}}` all answered the issues with no project / assignee. `estimate: {and: []}` and `{or: []}` answered every issue. The int literal on `dueDate` answered `Expected value of type "TimelessDateOrDuration", found 1; Unable to parse literal value of kind 'IntValue'. TimelessDate supports only 'StringValue' ones`, and the int variable `Unable to parse value '1'. TimelessDate supports only string values`.

The string rule was measured last, against a second throwaway issue created with no description and two attachments, one with no subtitle, all deleted after (`issueDelete` answered `success: true`; the workspace lists no issue titled for the measurement). On the attachment with no subtitle: `subtitle: {null: true}` answered it, `subtitle: {nin: ["zzz"]}` answered both attachments, `subtitle: {neq: "zzz"}`, `{neqIgnoreCase: "zzz"}`, `{notContains: "zzz"}` and `{eq: ""}` answered only the one with a subtitle. Both attachments were served `sourceType: "unknown"`, and `sourceType: {neq: "zzz"}`, `{nin: ["zzz"]}`, `{notContains: "zzz"}` and `{eq: ""}` each answered nothing. `Issue.description` behaved differently again (`description: {null: true}` answered nothing for the issue whose description is null, while `neq` / `nin` / `notContains` all answered it) and is not reproduced: Backlot's `content` column is NOT NULL, so no served description is ever null.

Measured and NOT changed here: the labels filter over an issue with no labels. Linear answered `labels: {every: {name: {eq: "zzz"}}}` with nothing (no vacuous truth) and `labels: {some: {name: {neq: "zzz"}}}`, `{some: {name: {nin: ["zzz"]}}}`, `{some: {name: {notContains: "zzz"}}}` with every such issue; Backlot's `every` (a `NOT EXISTS … NOT`) answers all and its `some` (an `EXISTS`) answers none, the opposite on both counts. The reading that fits — the negation pushed outside the quantifier — cannot be told apart from "some label is not named X" without an issue that carries labels, which the reachable workspace has none of, so this is written up as #112 with the measured table rather than redesigned on half a measurement.

`backlot diff --source linear` before this change: 829 divergences, 47 new, all breaking. After: 778 divergences, 0 new, 4 resolved.

## Verification

- **Tests** — `pytest` on the full-extras venv (Python 3.14): `1790 passed in 23.43s`. With `-rs`, no skips reported. `tests/test_linear.py` and `tests/test_graphql.py` also run under Python 3.11 (`uv run --python 3.11 --with-editable '.[dev]'`): `252 passed`.
- **Optional surfaces that ran rather than skipped** — every optional test file ran (0 skipped). The official `@linear/sdk` v88 example typechecks and runs end to end against a server it starts itself (`npx tsc --noEmit && PYTHON=… npx tsx index.ts`), so the SDK's own generated documents still validate against the corrected schema.
- **Lint** — `ruff check . && ruff format --check .`: `All checks passed!` / `139 files already formatted`.
- **Live re-measurement** — `backlot diff --source linear`: `778 divergence(s) · 778 acknowledged · 0 new`.
- **Fails without the fix** — each commit, with its `backlot/` changes stashed and its tests kept: first `78 failed, 1 passed, 2 errors`; second `5 failed`; third `7 failed, 1 passed`; fourth `3 failed, 6 passed`; sixth `1 failed`; seventh `6 failed`; the three review commits (`updatedAt` on the served expression, the nested `project.id` refusal, the identifier key case) `1 failed` each, every one the review's own reproduction as a test. Each pass is a regression guard meant to hold on both sides. The fourth commit went up with one red test of its own — widening the identifier shape to either case also let `not-a-uuid` through — and the fifth commit is the fix (look up first, judge the shape on a miss); the full run above is at head.

### DIFF
--- a/backlot/graphql/linear_filters.py
+++ b/backlot/graphql/linear_filters.py
@@ -211,6 +211,34 @@ def _to_timeless(value) -> str | None:
         raise GraphQLError(str(e)) from None
 
 
+# A UUID as Linear's id comparators validate it: 8-4-4-4-12 hex with an RFC 4122 variant nibble
+# (8, 9, a or b). Measured 2026-09-03 on both `IssueFilter.id` and `NullableProjectFilter.id`: the nil
+# UUID, `1111…1111` (variant 1) and `ffff…ffff` (variant f) answer `Argument Validation Error`
+# (code INVALID_INPUT), while `…-4111-8111-…` (v4), `…-1111-8111-…` (v1) and `…-5111-9111-…` (v5)
+# are looked up -- so the version nibble is not checked, the variant is. Every id Backlot serves is
+# a v4 with variant 8-b (`synth._uuid_from`), so nothing of its own is refused.
+UUID_SHAPE = re.compile(
+    r"^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[89abAB][0-9a-fA-F]{3}-[0-9a-fA-F]{12}$"
+)
+_UUID_LIKE = re.compile(r"^[0-9a-fA-F]{8}(-[0-9a-fA-F]{4}){3}-[0-9a-fA-F]{12}$")
+
+
+def argument_validation_error() -> GraphQLError:
+    """The error Linear's id comparators answer for a value they refuse: a 200 whose `errors` carry
+    `Argument Validation Error` with code INVALID_INPUT (measured 2026-09-03)."""
+    return GraphQLError("Argument Validation Error", extensions={"code": "INVALID_INPUT"})
+
+
+def _refuse_malformed_uuids(spec: dict | None) -> None:
+    """`EntityIdentifierIDComparator` (``project.id``): any string is looked up -- `"PROJ-1"`,
+    `"not-a-uuid"` and an unknown UUID each answered an empty page -- EXCEPT something shaped like a
+    UUID with a variant Linear's validator rejects, which is refused before the lookup."""
+    for op, raw in (spec or {}).items():
+        for value in raw if isinstance(raw, list) else [raw]:
+            if isinstance(value, str) and _UUID_LIKE.match(value) and not UUID_SHAPE.match(value):
+                raise argument_validation_error()
+
+
 # A comparator key -> how it becomes SQL, given a column expression and the value.
 # `%` / `_` in a LIKE needle are escaped so a user-supplied value stays literal.
 _LIKE_ESCAPE = str.maketrans({"\\": "\\\\", "%": "\\%", "_": "\\_"})
@@ -570,6 +598,7 @@ def _issue_filter(conn, flt: dict, team_keys: dict | None = None) -> tuple[str, 
                 )
             )
         elif key == "project":
+            _refuse_malformed_uuids(spec.get("id"))
             add(
                 *_sub_filter(
                     conn,
--- a/backlot/graphql/linear_filters.py
+++ b/backlot/graphql/linear_filters.py
@@ -239,6 +239,25 @@ def _refuse_malformed_uuids(spec: dict | None) -> None:
                 raise argument_validation_error()
 
 
+def _refuse_malformed_project_ids(spec: dict | None) -> None:
+    """:func:`_refuse_malformed_uuids` over every ``id`` a ``ProjectFilter`` reaches, the nested ones
+    included: ``_sub_filter`` follows ``and`` / ``or`` down to the same lookup, so a value refused at
+    the top level was quietly looked up (and matched nothing) one level down. The issue-id side,
+    ``_resolve_issue_ids``, already recurses this way."""
+    for sub in _through_and_or(spec):
+        _refuse_malformed_uuids(sub.get("id"))
+
+
+def _through_and_or(spec):
+    """``spec`` and every filter object nested under its ``and`` / ``or``, depth first."""
+    if not isinstance(spec, dict):
+        return
+    yield spec
+    for key in ("and", "or"):
+        for sub in spec.get(key) or []:
+            yield from _through_and_or(sub)
+
+
 # A comparator key -> how it becomes SQL, given a column expression and the value.
 # `%` / `_` in a LIKE needle are escaped so a user-supplied value stays literal.
 _LIKE_ESCAPE = str.maketrans({"\\": "\\\\", "%": "\\%", "_": "\\_"})
@@ -603,7 +622,7 @@ def _issue_filter(conn, flt: dict, team_keys: dict | None = None) -> tuple[str, 
                 )
             )
         elif key == "project":
-            _refuse_malformed_uuids(spec.get("id"))
+            _refuse_malformed_project_ids(spec)
             add(
                 *_sub_filter(
                     conn,
--- a/backlot/graphql/linear_filters.py
+++ b/backlot/graphql/linear_filters.py
@@ -56,6 +56,18 @@ from backlot import synth
 # graphql-core (like graphql-js, which Linear runs) reports a GraphQLError from a scalar verbatim
 # but wraps any other exception as `Expected value of type 'DateTimeOrDuration', found "P"; <message>`
 # -- the envelope the measurement above shows.
+#
+# `TimelessDateOrDuration` (2026-09-03, against an issue due 2026-03-15 created for the purpose and
+# deleted after): the same grammar, read down to the UTC day. `eq` matched for "2026-03-15",
+# "2026-03-15T00:00:00Z", "2026-03-15T23:59:59Z" and "2026-03-16T02:00:00+09:00" (17:00Z on the
+# 15th); it did NOT match for "2026-03-15T23:59:59-05:00" (04:59Z on the 16th) or
+# "2026-03-15T02:00:00+09:00" (17:00Z on the 14th). `gte: "2026-03-15T12:00:00Z"` matched and
+# `gt: "2026-03-15T00:00:00Z"` did not, so the operand is truncated before the comparison rather
+# than compared as an instant. "2026-03" is the first of the month (`eq` misses, `gte` hits), a
+# duration is relative to today (`lt: "P0D"` hits a past due date). The scalar names itself
+# `TimelessDate` in its errors: `Unable to parse literal value of kind 'IntValue'. TimelessDate
+# supports only 'StringValue' ones`, and for a variable `Unable to parse value '1'. TimelessDate
+# supports only string values` (quoted, where DateTimeOrDuration prints the value bare).
 
 _DURATION = re.compile(
     r"^(?P<sign>[+-])?P"
@@ -92,13 +104,15 @@ def _from_duration(m: re.Match) -> _dt.datetime:
     )
 
 
-def parse_datetime_or_duration(value) -> _dt.datetime:
+def parse_datetime_or_duration(value, scalar: str = "DateTimeOrDuration") -> _dt.datetime:
     """One ``DateTimeOrDuration`` operand as an aware ``datetime``, or a ``ValueError`` whose
-    message is the one Linear's scalar produces (see the measurements above)."""
+    message is the one Linear's scalar produces (see the measurements above). ``scalar`` is the
+    name the message carries: ``TimelessDateOrDuration`` calls itself ``TimelessDate`` there."""
     if not isinstance(value, str):
-        raise ValueError(
-            f"Unable to parse value {value!r}. DateTimeOrDuration supports only string values"
-        )
+        # Measured: DateTimeOrDuration prints the value bare (`value 1700000000.`), TimelessDate
+        # quotes it (`value '1'.`).
+        shown = repr(value) if scalar == "DateTimeOrDuration" else f"'{value}'"
+        raise ValueError(f"Unable to parse value {shown}. {scalar} supports only string values")
     s = value.strip()
     m = _DURATION.match(s)
     if m is not None:
@@ -122,22 +136,21 @@ def parse_datetime_or_duration(value) -> _dt.datetime:
 
 def parse_timeless_date_or_duration(value) -> str:
     """One ``TimelessDateOrDuration`` operand as the bare ``YYYY-MM-DD`` the ``due_date`` column
-    holds. Linear documents the scalar with the same text as ``DateTimeOrDuration`` and accepts the
-    same spellings (measured: "2026", "2026-03", "2026-03-15", "2026-03-15T10:00:00Z" and "-P1D"
-    all validate on ``dueDate``), so the grammar is shared and the time of day is dropped. The day
-    is read in UTC: a timeless date has no zone of its own, and UTC is the zone every ``DateTime``
-    this schema serves is written in."""
-    return parse_datetime_or_duration(value).astimezone(_dt.timezone.utc).date().isoformat()
+    holds. Linear documents the scalar with the same text as ``DateTimeOrDuration``, accepts the
+    same spellings, and reads an operand that carries a time down to its UTC day (measured, see the
+    module comment: "2026-03-16T02:00:00+09:00" is the 15th, "2026-03-15T23:59:59-05:00" is not)."""
+    when = parse_datetime_or_duration(value, "TimelessDate")
+    return when.astimezone(_dt.timezone.utc).date().isoformat()
 
 
-def _parse_literal(parse, node, _variables=None):
+def _parse_literal(parse, scalar: str, node, _variables=None):
     """The literal half of a scalar's coercion. Linear's scalars take a string literal only, and
     say so: `Unable to parse literal value of kind 'IntValue'. DateTimeOrDuration supports only
-    'StringValue' ones` (measured 2026-09-03)."""
+    'StringValue' ones` (measured 2026-09-03; the other scalar says `TimelessDate`)."""
     if not isinstance(node, StringValueNode):
         raise ValueError(
             f"Unable to parse literal value of kind '{type(node).__name__.removesuffix('Node')}'. "
-            "DateTimeOrDuration supports only 'StringValue' ones"
+            f"{scalar} supports only 'StringValue' ones"
         )
     return parse(node.value)
 
@@ -149,13 +162,13 @@ SCALARS = {
     "DateTimeOrDuration": {
         "parse_value": parse_datetime_or_duration,
         "parse_literal": lambda node, variables=None: _parse_literal(
-            parse_datetime_or_duration, node, variables
+            parse_datetime_or_duration, "DateTimeOrDuration", node, variables
         ),
     },
     "TimelessDateOrDuration": {
         "parse_value": parse_timeless_date_or_duration,
         "parse_literal": lambda node, variables=None: _parse_literal(
-            parse_timeless_date_or_duration, node, variables
+            parse_timeless_date_or_duration, "TimelessDate", node, variables
         ),
     },
 }
@@ -195,12 +208,37 @@ def _like(value: str) -> str:
 
 
 class _Comparator:
-    """Renders one comparator object (``{eq: …, contains: …}``) against a column."""
+    """Renders one comparator object (``{eq: …, contains: …}``) against a column.
+
+    NULL under a negative operator follows what api.linear.app answered on 2026-09-03, over issues
+    whose ``estimate``, ``dueDate`` and ``completedAt`` were null:
+
+    * on a FIELD of the issue, ``neq`` DROPS the null rows (`estimate: {neq: 99}` over four null
+      estimates answered nothing) and ``nin`` KEEPS them (`estimate: {nin: [99]}` answered all four;
+      `dueDate: {nin: ["2026-03-15"]}` answered exactly the issues with no due date);
+    * on a RELATION (``project: {name: {neq: …}}``, ``assignee: {name: {neq: …}}``,
+      ``assignee: {id: {neq: …}}``, and ``nin`` alike) issues WITHOUT the relation are kept.
+
+    ``relation=True`` is the second rule, for the columns ``_sub_filter`` maps a nested filter onto.
+    Neither rule is SQL's own three-valued one (`<>` and `NOT IN` both drop nulls), so both are
+    spelled out here rather than left to the engine.
+    """
 
     def __init__(
-        self, col: str, *, text: bool = False, epoch: bool = False, timeless: bool = False
+        self,
+        col: str,
+        *,
+        text: bool = False,
+        epoch: bool = False,
+        timeless: bool = False,
+        relation: bool = False,
     ):
         self.col, self.text, self.epoch, self.timeless = col, text, epoch, timeless
+        self.relation = relation
+
+    def _not_equal(self, expr: str) -> str:
+        """``expr`` is a ``<>`` test; on a relation the absent rows pass it too."""
+        return f"({self.col} IS NULL OR {expr})" if self.relation else expr
 
     def _value(self, v):
         # A DateComparator's operand is a DateTimeOrDuration but the column is unix seconds; a
@@ -235,9 +273,7 @@ class _Comparator:
                 parts.append(f"{self.col} = ?")
                 params.append(v)
             elif op == "neq":
-                # NULL never equals anything, so a plain `<> ?` would silently drop NULL rows a
-                # caller asking "not X" expects to see.
-                parts.append(f"({self.col} IS NULL OR {self.col} <> ?)")
+                parts.append(self._not_equal(f"{self.col} <> ?"))
                 params.append(v)
             elif op == "lt":
                 parts.append(f"{self.col} < ?")
@@ -257,7 +293,12 @@ class _Comparator:
                     parts.append("0" if op == "in" else "1")
                     continue
                 marks = ",".join("?" for _ in vals)
-                parts.append(f"{self.col} {'IN' if op == 'in' else 'NOT IN'} ({marks})")
+                if op == "in":
+                    parts.append(f"{self.col} IN ({marks})")
+                else:
+                    # Null rows pass `nin` on a field and on a relation alike (measured; see the
+                    # class docstring), which a bare `NOT IN` would drop.
+                    parts.append(f"({self.col} IS NULL OR {self.col} NOT IN ({marks}))")
                 params += vals
             elif op == "contains":
                 parts.append(f"{self.col} LIKE ? ESCAPE '\\'")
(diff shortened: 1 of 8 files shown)


khj809 (commented):
Verified the 47 gaps against the live API: `backlot diff --source linear` reports 829 total / 47 new on `main` and 778 / 0 new here, and the 47 it lists on `main` match #101 item for item. The 51-divergence drop is exactly the 47 breaking plus the 4 baselined `in` / `nin` gaps, so no new gap was introduced — which, since `graphql_diff` compares enum members individually, is also what proves the ten new member lists (the 45-member `IntegrationService` included) match introspection. Full suite green on all extras.

Three things found by re-running the measured behaviour, one of them a regression this PR introduces.

backlot/graphql/linear_filters.py
```diff
@@ -283,14 +526,20 @@ def add(frag, p):
...
+            # `EstimateComparator`: the number comparators plus `and` / `or` over them.
             add(*_Comparator("estimate").render(spec))
         elif key == "dueDate":
-            add(*_Comparator("due_date").render(spec))
+            # `NullableTimelessDateComparator`: operands are TimelessDateOrDuration, the column a
+            # bare YYYY-MM-DD, so the comparison is lexical over the ISO date.
+            add(*_Comparator("due_date", timeless=True).render(spec))
         elif key in ("createdAt", "updatedAt", "completedAt", "canceledAt"):
             col = {
                 "createdAt": "created_ts",
```
khj809, line 564: `Issue.updatedAt` is `DateTime!`, served as `COALESCE(updated_ts, created_ts)` — but this mapping points the filter at the raw `updated_ts`, so the treatment `priority` just got a few lines up is missing here, on the one other column `LINEAR_SORT_COLUMNS` and `LINEAR_ORDER_COLUMNS` already COALESCE.

The new field-mode `neq` makes that a regression rather than a standing miss. Two issues, ENG-1 with `updated: 2026-01-05` and ENG-2 with no `updated` (so `updated_ts` IS NULL and `Issue.updatedAt` serves `2026-02-01T00:00:00Z` off `created_ts`):

```
{updatedAt: {neq: "2020-01-01T00:00:00Z"}}
  main -> ["ENG-1", "ENG-2"]
  head -> ["ENG-1"]
```

ENG-2 vanishes because `_not_equal` now emits a bare `updated_ts <> ?` for a field-mode comparator and SQL's `<>` drops the NULL row. Linear's `updatedAt` is never null, so `neq` cannot drop a row there — which is also why `DateComparator` is the declared type and carries no `null` operator. This is the same divergence the `priority` change closes.

Same root cause, standing rather than new: `{updatedAt: {eq: "2026-02-01T00:00:00Z"}}` answers `[]` for the very issue a client just read that value from, and `{updatedAt: {gte: "2026-01-01T00:00:00Z"}}` / `{lt: "2030-01-01T00:00:00Z"}` both omit ENG-2.

A `LINEAR_UPDATED_EXPR = "COALESCE(updated_ts, created_ts)"` in `store.py`, read by the two sort maps and by this dict, would close all four. There is no test over the `updatedAt` filter at the moment, which is why the suite stays green.

backlot/graphql/linear_filters.py
```diff
@@ -349,6 +598,7 @@ def add(frag, p):
                 )
             )
         elif key == "project":
+            _refuse_malformed_uuids(spec.get("id"))
```
khj809: This refusal only sees the top-level `id` of a project spec, so a value Linear rejects is refused un-nested and quietly accepted one level down:

```
{project: {id: {eq: "00000000-0000-0000-0000-000000000000"}}}
  -> 200, errors: [Argument Validation Error, code INVALID_INPUT]   (the measured answer)
{project: {and: [{id: {eq: "00000000-0000-0000-0000-000000000000"}}]}}
  -> 200, data.issues.nodes: []
```

`_sub_filter` recurses through `and` / `or` without re-running `_refuse_malformed_uuids`, so the nil UUID (and `1111...1111`, `ffff...ffff`) reaches `_derived_in` and merely fails to match. The sibling path already gets this right: `_resolve_issue_ids` recurses, so `{and: [{id: {eq: "not-a-uuid"}}]}` is refused as measured. Moving the call into `_sub_filter`'s `id` branch, or recursing over the spec's `and` / `or` here, would make the two agree.

backlot/store.py
```diff
@@ -1305,11 +1310,20 @@ def linear_by_id(conn, issue_id, visible_ids=None) -> sqlite3.Row | None:
...
+    Case-insensitive the way Linear's is: ``bre-1`` resolved to ``BRE-1`` there (measured
+    2026-09-03). Two indexed equality probes, the value as written and then upper-cased, rather than
+    a ``COLLATE NOCASE`` that would put the identifier index aside on every lookup. A corpus may
+    write a mixed-case suffix (the schema leaves everything after the hyphen opaque), which is why
+    the as-written probe comes first."""
     sql = "SELECT * FROM linear_issues WHERE identifier = ?"
-    params: list = [identifier]
     clause, cparams = _acl_clause("linear", visible_ids=visible_ids)
-    return conn.execute(sql + clause + " ORDER BY id LIMIT 1", params + cparams).fetchone()
+    for candidate in dict.fromkeys((identifier, str(identifier).upper())):
```
khj809: The docstring says case-insensitive the way Linear's is, but the two probes miss exactly the case the paragraph names as the reason the as-written probe comes first. With `identifier: "ENG-1a"` in the corpus (the schema leaves everything after the hyphen opaque):

```
issue(id: "ENG-1a") -> found
issue(id: "eng-1a") -> Entity not found     (probes "eng-1a", then "ENG-1A")
```

`issues(filter: {id: {eq: "eng-1a"}})` splits the same way, then falls through to `_LINEAR_IDENTIFIER_SHAPE` and answers the empty-page sentinel. Either `identifier = ? COLLATE NOCASE` — paying the index this comment is protecting — or narrowing the claim to an upper-cased key would keep the spellings in agreement.

backlot/graphql/linear_filters.py
nuri-yoo: Fixed in 26cc993. Reproduced first: on this branch `{updatedAt: {neq: "2020-01-01T00:00:00Z"}}` answered ENG-1 alone where `main` answered both, and `eq` on the served `2026-02-01T00:00:00Z`, `gte: "2026-01-01T00:00:00Z"` and `lt: "2030-01-01T00:00:00Z"` each omitted ENG-2 on both.

One `store.LINEAR_UPDATED_EXPR` now, read by `LINEAR_ORDER_COLUMNS`, `LINEAR_SORT_COLUMNS` and the filter map, as you sketched; `completedAt` / `canceledAt` stay on their raw columns, being nullable on Linear as well. The DDL index at `store.py:647` was already on that expression. The test states the served `updatedAt` first and asserts `eq`, `neq`, `gte`, `lt`, `gt` and a descending sort against it; with the filter back on the raw column it fails at the first `eq`.

backlot/graphql/linear_filters.py
nuri-yoo: Fixed in b038f27. Both shapes reproduced: `[]` for the nil UUID under `and` and under `or`, `Argument Validation Error` at the top level.

`_refuse_malformed_project_ids` walks the spec through `and` / `or` and applies the existing check to every `id` it reaches, so it agrees with `_resolve_issue_ids`. Kept on `project` only, since that is the comparator the rule was measured on (`EntityIdentifierIDComparator`); `team.id` and `assignee.id` are unmeasured and unchanged. The test nests the nil UUID under `and`, under `or` beside a name predicate, and two levels down, and keeps a well-formed unknown UUID an empty page at depth; without the walk it fails on the first shape with no `errors` in the response.

backlot/store.py
nuri-yoo: Fixed in 126f359, the key-folding option. Reproduced: against a stored `ENG-1a`, `issue(id: "eng-1a")` and `issue(id: "ENG-1A")` were both `Entity not found`.

One correction on the filter half: `issues(filter: {id: {eq: "eng-1a"}})` did not fall through to the empty-page sentinel. The lookup missed, and then `_LINEAR_IDENTIFIER_SHAPE` (digits after the hyphen) refused the value as `Argument Validation Error`. Same root cause, different answer.

Three indexed probes now: as written, key upper-cased, whole value upper-cased. Linear's identifier carries letters only in the key, so that is the half folded; the suffix is compared as written, and the docstring says so instead of claiming whole-value insensitivity, so `ENG-1A` is a different string from `ENG-1a` and names nothing. `COLLATE NOCASE` passed over for the index reason you named. The test writes `ENG-1a`, reaches it as `ENG-1a` and `eng-1a` through both `issue(id:)` and the `id` filter, and pins `ENG-1A` as a miss; with two probes it fails on `eng-1a`.

Head `126f359`: 1790 passed, `backlot diff --source linear` 778 / 778 acknowledged / 0 new, `ruff check .` and `ruff format --check .` clean.

khj809 (approved):
Looks good

PR #114, opened 2026-09-03
Atlassian Basic authenticates the pair, the way the real service does

khj809:
Atlassian's Basic scheme authenticates the pair, and Backlot authenticated either half.

## Measured

Against a real Atlassian Cloud site, `GET /rest/api/3/myself`, with a user API token:

| Basic username | Basic password | Real Atlassian | Backlot before | Backlot now |
|---|---|---|---|---|
| the account's email | its API token | 200 | 200 | 200 |
| the account's email | empty | 401 | **200** | 401 |
| the account's email | wrong | 401 | **200** | 401 |
| an unknown email | a valid token | 401 | **200** | 401 |
| the address upper-cased | its API token | 200 | 200 | 200 |
| the address mixed-case | its API token | 200 | 200 | 200 |

Knowing an address was enough to read as that person, and addresses are all over the corpus content — authors, recipients, commenters. A valid token under a placeholder address was accepted where the real service refuses it.

## What changed

`auth.resolve_basic` requires the password to resolve **and** the username to be that account's own address, matched case-insensitively as Atlassian treats an address. It no longer falls back to the username, and no longer ignores it.

The admin/service token is the one exception, and the only one: `Acl.resolve` gives it `Caller(email=None)`, so there is no address to match against. It has no vendor analogue to be faithful to — it is Backlot's own full-crawl identity, and it is what lets an Atlassian client send the placeholder username its config demands, so `svc@example.com:admin-service-token` still works.

Callers that paired a user's token with a placeholder now pass the address. `examples/using-official-sdk/{jira,confluence}.py` and `examples/using-mcp-with-agents/atlassian.py` refuse either half alone: a `--token` with no matching `--username`, and a `--username` with no token, which used to send the admin's token under that address while printing that responses were filtered to that user. The admin token is exempt both ways, which is the case `docs/auth.md` calls out. `docs/auth.md` carries the four cases above in place of the claim that the username is ignored, and `examples/using-official-sdk/README.md`'s Atlassian rows take the address `T` belongs to — that table's own note says `T` may be a per-user token, which the placeholder would now refuse.

## The test that would have stayed green

`tests/test_mcp.py`'s mcp-atlassian test drives the vendor server with `JIRA_USERNAME` fixed to the placeholder. Under this change its scoped read fails on auth, which reaches the test exactly as an ACL miss does — so it would have kept passing with the ACL doing nothing. It now passes each token's own address and adds a third read, of an issue that user *can* see, as the control that tells auth from ACL. The control is the load-bearing line: putting the placeholder back on it fails with "should read an issue they can see, or the miss below is an auth failure", while putting it on the scoped read leaves the test green, since a read that fails on auth reaches `assert not reads(...)` exactly as an ACL miss does.

## Verification

- `pytest` on a `.[all]` install plus `llama-index-readers-hubspot<0.6 --no-deps`: **1864 passed, 0 skipped**. Docker (colima), `npx` and `uvx` present, so the mcp-atlassian, notion-mcp-server and awslabs tests ran rather than skipped.
- **Without the change** — the two changed test files copied onto `main`: **3 failed, 59 passed**. Reverting `resolve_basic` to its previous body on this branch fails the same three, and nothing else.
- The five credential forms, driven against a live server: the account's own address with its token 200, empty password 401, wrong password 401, an unknown address with a valid token 401, the placeholder with the admin token 200 — matching the real service on all four of its cases.
- The three launchers, four cases each: a user's `--token` alone refused, `--username` alone refused, the admin token accepted under the placeholder, the matching pair accepted and printing that user's name. Their no-argument default paths still answer.
- `ruff check . && ruff format --check .`: clean. `python scripts/gen_docs.py --check`: exit 0.

- [x] A test fails without this change — the 3 above.

## Not in this PR

On the routes Backlot actually serves, real Jira answers a failed pair as an *anonymous* request (400 on `search/jql`, 200 with empty `values` on `project/search`, 404 on an issue) with `X-Seraph-LoginReason: AUTHENTICATED_FAILED`, and Confluence answers 403 in its own envelope; Backlot answers 401. The table above is scoped to `GET /rest/api/3/myself`, which Backlot does not serve, and where 401 is what the real service gives. That 401-versus-anonymous gap predates this PR (#44 shaped those errors for client parsing) and needs an anonymous caller to close, so it is its own issue rather than part of this one — measurements courtesy of @nuri-yoo's review.

### DIFF
--- a/backlot/auth.py
+++ b/backlot/auth.py
@@ -150,19 +150,26 @@ def resolve_api_key(request: Request) -> Caller | None:
 
 
 def resolve_basic(request: Request) -> Caller | None:
-    """Atlassian: resolve by the api_token (password); fall back to the username email."""
-    a = acl(request)
-    user, pw = basic_password(request)
-    caller = a.resolve(pw)
-    if caller is not None:
-        return caller
-    # allow username=email as an identity shortcut (a Backlot convenience)
-    if user and "@" in user:
-        from backlot import store
+    """Atlassian: the api_token is the password, and the username has to be its own account.
 
-        if store.get_user(conn(request), user):
-            return Caller(email=user, is_admin=False)
-    return None
+    Both halves, because that is what the real service requires. Measured against a real Atlassian
+    Cloud site (``GET /rest/api/3/myself``) with a user API token: ``email:token`` answers 200,
+    while an empty password, a wrong password, and a valid token under someone else's address all
+    answer 401. Matched case-insensitively, which is also measured: the same token under
+    ``AVA.CHEN@…`` and ``Ava.chen@…`` answers 200 as that same account.
+
+    The admin/service token is the one caller with no address — ``Acl.resolve`` gives it
+    ``Caller(email=None)`` — so there is nothing to match a username against and any username is
+    taken. It has no vendor analogue to be faithful to: it is Backlot's own full-crawl identity,
+    and it is what lets an Atlassian client send the placeholder username its config demands.
+    """
+    user, pw = basic_password(request)
+    caller = acl(request).resolve(pw)
+    if caller is None:
+        return None
+    if caller.email is None:
+        return caller
+    return caller if (user or "").casefold() == caller.email.casefold() else None
 
 
 def visible_ids(request: Request, caller: Caller) -> set[str] | None:
--- a/examples/using-official-sdk/jira.py
+++ b/examples/using-official-sdk/jira.py
@@ -8,6 +8,7 @@
 """
 
 import argparse
+import sys
 
 from atlassian import Jira
 
@@ -46,7 +47,7 @@ _p.add_argument(
 _p.add_argument(
     "--username",
     default="svc@example.com",
-    help="Atlassian Basic-auth username (email); Backlot resolves the caller by the token/password",
+    help="Atlassian Basic-auth username: the address the token belongs to, which the real service requires to match (the placeholder works only for the admin token, which has none)",
 )
 _p.add_argument(
     "--password",
@@ -60,7 +61,24 @@ args = _p.parse_args()
 with serve_or_connect(CORPUS, url=args.url) as s:
     username = args.username
     password = args.password or args.token or s.token
-    if args.username != "svc@example.com" or args.password or args.token:
+    # Atlassian authenticates the PAIR: a user's api_token under someone else's address is a 401
+    # on the real service, and Backlot answers the same. Both halves have to name one identity, so
+    # either alone is refused rather than sent. The admin/service token is the exception both ways
+    # — it has no address, so any username carries it, which is what `docs/auth.md` documents.
+    named_token = (args.password or args.token) not in (None, s.token)
+    named_user = args.username != "svc@example.com"
+    if named_token and not named_user:
+        sys.exit(
+            "a --token/--password names one user, so pass --username with that user's own email "
+            f"too (GET {s.base_url}/_meta/users lists both)"
+        )
+    if named_user and not named_token:
+        sys.exit(
+            f"--username alone would send the admin token under {args.username}, which Backlot "
+            "takes as the admin: pass --token with that user's own token too "
+            f"(GET {s.base_url}/_meta/users lists both)"
+        )
+    if named_user or named_token:
         print(f"authenticating as {username} → responses are ACL-filtered to that user")
     jira = Jira(url=f"{s.base_url}/atlassian", username=username, password=password)
 
--- a/tests/test_mcp.py
+++ b/tests/test_mcp.py
@@ -61,12 +61,15 @@ def _docker_available() -> bool:
         return False
 
 
-def _atlassian_params(base: str, token: str):
+def _atlassian_params(base: str, token: str, username: str = "svc@example.com"):
     """`docker run` args pointing mcp-atlassian at a local server (see examples/.../atlassian.py).
 
     mcp-atlassian only classifies a host as Atlassian *Cloud* when it ends in `.atlassian.net`, so
     use a fake `backlot.atlassian.net` mapped to the host via `--add-host`, and Basic auth where the
-    api_token is a Backlot token (Backlot resolves it to a user and enforces that user's ACL)."""
+    api_token is a Backlot token (Backlot resolves it to a user and enforces that user's ACL).
+
+    ``username`` has to be that token's own address, as the real service requires — the default
+    placeholder works only for the admin token, which has no address to match."""
     from mcp import StdioServerParameters
 
     port = base.rsplit(":", 1)[1]  # Docker reaches Backlot on the host via host-gateway
@@ -81,13 +84,13 @@ def _atlassian_params(base: str, token: str):
             "-e",
             f"JIRA_URL={url}/atlassian",
             "-e",
-            "JIRA_USERNAME=svc@example.com",
+            f"JIRA_USERNAME={username}",
             "-e",
             f"JIRA_API_TOKEN={token}",
             "-e",
             f"CONFLUENCE_URL={url}/atlassian/wiki",
             "-e",
-            "CONFLUENCE_USERNAME=svc@example.com",
+            f"CONFLUENCE_USERNAME={username}",
             "-e",
             f"CONFLUENCE_API_TOKEN={token}",
             "-e",
@@ -150,24 +153,41 @@ async def _call(params, tool_pred, args, ok_pred) -> bool:
 
 @pytest.mark.skipif(not _docker_available(), reason="Docker not available")
 def test_mcp_atlassian_acl_enforced(live_server):
+    """A Jira issue the admin reads through mcp-atlassian is not-found for a scoped user.
+
+    Both calls carry a credential the real service would accept — the placeholder username for the
+    admin, who has no address, and the user's own address for the user — so the scoped miss is the
+    ACL's. Without the matching username the read would fail on auth instead, which reaches this
+    test the same way and would leave it green with the ACL doing nothing; the third read is the
+    control that rules that out."""
     base, settings = live_server
     user = yaml.safe_load(settings.tokens_path.read_text())["users"][0]
     row, email = _restricted_doc(settings, user["token"], "jira")
     assert row is not None, f"no Jira issue is ACL-restricted from {email} in the sample corpus"
     key = row["key"]  # the stored column: jira is probed, so a re-derived hash can disagree
+    seen, _ = _visible_doc(settings, user["token"], "jira")
+    assert seen is not None, f"no Jira issue is readable by {email} in the sample corpus"
+    visible = seen["key"]
 
-    def reads(token):
+    def reads(token, username, issue_key):
         return asyncio.run(
             _call(
-                _atlassian_params(base, token),
+                _atlassian_params(base, token, username),
                 tool_pred=lambda n: n == "jira_get_issue",
-                args={"issue_key": key},
+                args={"issue_key": issue_key},
                 ok_pred=lambda t: t.strip().startswith("{") and '"key"' in t,
             )
         )
 
-    assert reads(settings.admin_token), "admin should read the issue through mcp-atlassian"
-    assert not reads(user["token"]), f"{email} should be blocked from the issue via mcp-atlassian"
+    assert reads(settings.admin_token, "svc@example.com", key), (
+        "admin should read the issue through mcp-atlassian"
+    )
+    assert reads(user["token"], email, visible), (
+        f"{email} should read an issue they can see, or the miss below is an auth failure"
+    )
+    assert not reads(user["token"], email, key), (
+        f"{email} should be blocked from the restricted issue via mcp-atlassian"
+    )
 
 
 # --------------------------------------------------------------------------- Notion
@@ -257,6 +277,22 @@ def test_mcp_s3_lists_objects(live_server):
 
 
 # ------------------------------------------------------ the OpenAPI→MCP bridge (backlot.mcp)
+
+
+def _visible_doc(settings, user_token: str, source: str):
+    """A doc of ``source`` this user CAN read — the mirror of :func:`_restricted_doc`.
+
+    Returns ``(row, email)`` with ``row`` None when the corpus has none, so a caller says what is
+    missing the way the restricted-doc callers do rather than raising ``StopIteration`` bare."""
+    conn = store.connect_ro(settings.db_path)
+    acl = Acl.load(settings.tokens_path, settings.admin_token, settings.org_name)
+    caller = acl.resolve(user_token)
+    vids = acl.visible_ids(conn, caller)
+    for row in conn.execute(f"SELECT * FROM {store.table(source)}"):
+        key = tuple(row[c] for c in store.id_columns(source))
+        if store.get_document(conn, source, *key, visible_ids=vids) is not None:
+            return row, caller.email
+    return None, caller.email
 
 
 def _bridge_call(base, source, creds, *, tool_pred, args, ok_pred) -> bool:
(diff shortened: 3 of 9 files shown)


backlot/auth.py
```diff
@@ -150,19 +150,25 @@ def resolve_api_key(request: Request) -> Caller | None:
...
 
 def resolve_basic(request: Request) -> Caller | None:
-    """Atlassian: resolve by the api_token (password); fall back to the username email."""
-    a = acl(request)
+    """Atlassian: the api_token is the password, and the username has to be its own account.
+
+    Both halves, because that is what the real service requires. Measured against a real Atlassian
+    Cloud site (``GET /rest/api/3/myself``) with a user API token: ``email:token`` answers 200,
+    while an empty password, a wrong password, and a valid token under someone else's address all
+    answer 401. Matched case-insensitively, the way Atlassian treats an address.
```
nuri-yoo: The four measured rows in the body are all lowercase, so "the way Atlassian treats an address" is asserted here without a row for it. Is `AVA.CHEN@…:<token>` a measured 200 on the real site? If so it is a fifth row for the table and this sentence stands; if not, the honest wording is that Backlot matches case-insensitively so an address copied from a page pairs with its token, without attributing it to Atlassian. (`backlot/mcp.py:135` already says the same about Atlassian and Google, so the wording predates this PR, but this is the line that puts it into auth.) One curl on your side, since you have the token.

examples/using-official-sdk/jira.py
```diff
@@ -60,6 +61,14 @@
...
     password = args.password or args.token or s.token
+    if (args.password or args.token) and args.username == "svc@example.com":
+        # Atlassian authenticates the PAIR: a user's api_token under someone else's address is a
+        # 401 on the real service, and Backlot answers the same. Only the admin token, which has
+        # no address, goes through under the placeholder.
+        sys.exit(
+            "a --token/--password names one user, so pass --username with that user's own email "
+            f"too (GET {s.base_url}/_meta/users lists both)"
+        )
     if args.username != "svc@example.com" or args.password or args.token:
```
nuri-yoo: This is the mirror of the guard above it. `--username <email>` with no `--token`/`--password` leaves `password = s.token`, the admin token, which the server now takes under any address, and this line then prints that responses are ACL-filtered to that user. Measured against a live server:

```
$ python examples/using-official-sdk/jira.py --url $BASE --username nobody@example.invalid
authenticating as nobody@example.invalid → responses are ACL-filtered to that user
2 issues; first issue:
```

That is the admin's view under a name the corpus does not know. `README.md:97` recommends exactly this form ("`--username <email>` alone identifies the user"). `confluence.py:67` is the same line, and `examples/using-mcp-with-agents/atlassian.py` has the same shape without the print (`build_params(s.token, args.username)` sends the admin token under the address, silently).

```suggestion
    if args.username != "svc@example.com" and not (args.password or args.token):
        sys.exit(
            f"--username alone would send the admin token under {args.username}, which Backlot "
            "takes as the admin: pass --token with that user's own token too "
            f"(GET {s.base_url}/_meta/users lists both)"
        )
    if args.password or args.token:
```

examples/using-official-sdk/jira.py
```diff
@@ -60,6 +61,14 @@
 with serve_or_connect(CORPUS, url=args.url) as s:
     username = args.username
     password = args.password or args.token or s.token
+    if (args.password or args.token) and args.username == "svc@example.com":
```
nuri-yoo: Small: `docs/auth.md` now says `svc@example.com:admin-service-token` works, and this refuses it: `jira.py --url $BASE --token admin-service-token` exits with "a --token/--password names one user", which is not true of that token. Either exempt it (`(args.password or args.token) not in (None, s.token)`) or let the message say the admin token is the exception. Not worth holding this up, but the message and the doc currently disagree about the one case the doc calls out.

tests/test_mcp.py
```diff
@@ -150,24 +153,46 @@ async def _call(params, tool_pred, args, ok_pred) -> bool:
...
+    control that rules that out."""
     base, settings = live_server
     user = yaml.safe_load(settings.tokens_path.read_text())["users"][0]
     row, email = _restricted_doc(settings, user["token"], "jira")
     assert row is not None, f"no Jira issue is ACL-restricted from {email} in the sample corpus"
     key = row["key"]  # the stored column: jira is probed, so a re-derived hash can disagree
+    conn = store.connect_ro(settings.db_path)
+    acl = Acl.load(settings.tokens_path, settings.admin_token, settings.org_name)
+    vids = acl.visible_ids(conn, acl.resolve(user["token"]))
+    visible = next(
```
nuri-yoo: Two small things in 168-175. A bare `next()` here raises `StopIteration` with no message if the corpus has no issue this user can see, where the `assert row is not None` three lines up says what is missing. And the block is the same seven lines as 452-460 in `test_mcp_atlassian_bridge_authenticates_basic_for_the_resolved_user`. A `_visible_doc(settings, user_token, source)` beside `_restricted_doc`, returning `None` rather than raising, would serve both and keep the assertion message pattern.

tests/test_mcp.py
```diff
@@ -150,24 +153,46 @@ async def _call(params, tool_pred, args, ok_pred) -> bool:
...
 
-    assert reads(settings.admin_token), "admin should read the issue through mcp-atlassian"
-    assert not reads(user["token"]), f"{email} should be blocked from the issue via mcp-atlassian"
+    assert reads(settings.admin_token, "svc@example.com", key), (
+        "admin should read the issue through mcp-atlassian"
+    )
+    assert reads(user["token"], email, visible), (
+        f"{email} should read an issue they can see, or the miss below is an auth failure"
+    )
+    assert not reads(user["token"], email, key), (
```
nuri-yoo, line 188: The body says "putting the placeholder back on the scoped read fails the test". I did, on this line only (`reads(user["token"], "svc@example.com", key)`), and the test passes: 1 passed with the Docker test executed. The scoped read fails on auth, `reads` returns False, and `assert not` holds, which is the failure mode the docstring above describes. The placeholder on the control read at 190 is what fails: "ava@acme.com should read an issue they can see, or the miss below is an auth failure". So the load-bearing line is 190, and the body sentence should name the control rather than the scoped read.

nuri-yoo (commented):
Reviewed at 5446ea2 in its own worktree. The core change is right and I could not break it by mutation; what is left is prose the change made false, one sentence in the body that the test contradicts, and one measurement the docstring asserts without a row for it.

**Verified.** Full suite on the `.[all]` env with Docker, `npx` and `uvx` present: 1864 ran, 1863 passed, 0 skipped. The one failure is `test_generated_docs_are_current`, and it is my worktree not the PR: the test runs `scripts/gen_docs.py --check` as a subprocess, which imports `backlot` from the editable install (a different checkout) rather than from the worktree; the same command with `PYTHONPATH=.` exits 0. `ruff check` and `ruff format --check` clean. Mutations on `resolve_basic`, each restored from a copy: the previous body back in fails exactly the three parametrized `refuses_a_half_credential` cases and nothing else (the Docker test stays green, as the old code also took `email:token`); dropping `.casefold()` fails the case test; dropping the `email is None` return fails the admin test. Against a live server: `nobody@example.invalid:<ava token>` is 401 on both `/rest/api/3/search/jql` and `/wiki/rest/api/space`, `:<ava token>` with an empty username is 401, `AVA@ACME.COM:<ava token>` is 200, `serverInfo` still answers unauthenticated. All three launchers refuse a bare `--token` with the `/_meta/users` pointer.

**Three sentences the change made false, all outside the hunks:**

1. `docs/auth.md:48`: "The token is what authenticates — never the username, even where a scheme carries one." This is now the opposite of what the section 100 lines below it says for Atlassian. Something like "The token is what authenticates. Atlassian's Basic is the one scheme that also checks the username: it has to be the token's own address (below)."

2. `examples/using-official-sdk/README.md:97-98`: "For Jira/Confluence either `--password <token>` or `--username <email>` alone identifies the user (Backlot resolves by the api token, falling back to the username email)." Both halves are now wrong. `--password` alone is what the launcher refuses; `--username` alone sends the admin token under that address, which the server takes as the admin. Measured: `jira.py --url … --username nobody@example.invalid` prints `authenticating as nobody@example.invalid → responses are ACL-filtered to that user` and returns the admin's two issues. The parenthetical describes the code this PR deletes. Inline on `jira.py:72` for the launcher side.

3. `tests/test_mcp.py:447-451`, the docstring of `test_mcp_atlassian_bridge_authenticates_basic_for_the_resolved_user`: "``auth.resolve_basic`` accepts the username alone when the password does not resolve — measured, an empty password answers 200, which real Atlassian refuses." True when it merged in #108; this PR is what makes it false. The paragraph's conclusion flips too: with the pair required, this test's positive read does show both halves authenticated, so "what it cannot show is WHICH half" can go, leaving the pointer to the header test if you want to keep it.

**PR body.** "Verified load-bearing: putting the placeholder back on the scoped read fails the test." I ran that mutation and the test passes (1 passed, Docker test executed). With the placeholder on the scoped read the read fails on auth, `reads` returns False and `assert not reads(...)` holds, which is exactly the failure mode the test's own docstring describes. Putting the placeholder on the *control* read (line 190) is what fails, with the message "ava@acme.com should read an issue they can see, or the miss below is an auth failure". So the control is the load-bearing line, and the sentence should name it. Inline on `test_mcp.py:193`.

**One measurement.** `auth.py:158` asserts the case-insensitive match is "the way Atlassian treats an address", but the four measured rows are all lowercase. Inline.

Small, not blocking: the duplicated visible-issue lookup and bare `next()` in `test_mcp.py:168-175`; the launcher refusing the admin token under the placeholder that `docs/auth.md` now says works (`jira.py:64`); `tests/test_auth.py:4-6` still says the file covers "currently the API-key scheme" only.

Outside this PR, but it bears on the table's "Real Atlassian" column. I sent a wrong pair (`nobody@example.com:wrongtoken`, and the same address with an empty password) to a public Cloud site and to brekkylab.atlassian.net; both answered identically. `GET /rest/api/3/myself` is a 401 (text/html "Client must be authenticated to access this resource.", `WWW-Authenticate: OAuth realm=…`). The endpoints Backlot serves are not: Jira treats the failed pair as an anonymous request and answers as it would to nobody, with only an `X-Seraph-LoginReason: AUTHENTICATED_FAILED` header to say the credential was rejected; Confluence is a 403 in its own envelope.

| Endpoint | Real, wrong or empty password | Backlot |
|---|---|---|
| `rest/api/3/myself` | 401, text/html, `WWW-Authenticate: OAuth realm=…` | not served |
| `rest/api/3/search/jql` | 400 as anonymous: "Unbounded JQL queries are not allowed here" | 401 |
| `rest/api/3/project/search` | 200 as anonymous: empty `values` | 401 |
| `rest/api/3/issue/ABC-1` | 404 as anonymous: "Issue does not exist or you do not have permission to see it" | 401 |
| `wiki/rest/api/space`, `wiki/rest/api/content/1` | 403: "Request rejected because caller cannot access Confluence" | 401 |

The Jira rows come in Jira's `{"errorMessages": [...], "errors": {}}` envelope with `X-Seraph-LoginReason: AUTHENTICATED_FAILED`; the Confluence row in Confluence's `{"statusCode": 403, "message": "…"}`.

So the 401 rows hold for `myself`, which Backlot does not serve; on the routes it does serve, the real answer to a failed pair is an anonymous view (Jira) or a 403 (Confluence), and Backlot's 401 with the `errors/atlassian.py` envelope predates this PR (#44, a client-parsing shape rather than a measured one). The valid-token-under-another-address row I could not measure without a token. None of this changes what this PR does, which is to stop authenticating the pair as that user, and the 401 versus anonymous question needs an anonymous caller to fix, so it is its own issue. I will file it with this table unless you would rather it stayed here.

backlot/auth.py
khj809: Measured on the real site with the token: `AVA.CHEN@BREKKYLAB.COM:<token>` and `Ava.chen@…:<token>` both answer 200 as that same account. So it is a fifth and sixth row, now in the table, and `auth.py` says what was measured rather than attributing it: "Matched case-insensitively, which is also measured: the same token under `AVA.CHEN@…` and `Ava.chen@…` answers 200 as that same account."

examples/using-official-sdk/jira.py
khj809: Confirmed by running it — `--username nobody@example.invalid` with no token printed that responses were filtered to that user and returned the admin s two issues. Both launchers now refuse either half alone, and `examples/using-mcp-with-agents/atlassian.py` too, which had the same shape without the print. Took the guard rather than the suggestion verbatim so the two conditions read as one rule: `named_token` and `named_user` are computed once, and each is refused without the other. `README.md:97` no longer recommends the single-flag form.

examples/using-official-sdk/jira.py
khj809: Fixed with the exemption rather than the message: `named_token` is `(args.password or args.token) not in (None, s.token)`, so `--token admin-service-token` goes through under the placeholder, which is what `docs/auth.md` says. Verified all four cases per launcher — user token alone refused, username alone refused, admin token accepted, matching pair accepted.

tests/test_mcp.py
khj809: Added `_visible_doc(settings, user_token, source)` beside `_restricted_doc`, same shape and returning `(None, email)` rather than raising. Both call sites use it and assert with the message pattern.

tests/test_mcp.py
khj809: Ran both mutations: placeholder on the scoped read leaves the test green, placeholder on the control at 190 fails with "should read an issue they can see, or the miss below is an auth failure". My verification did mutate the control; the body sentence named the wrong line. Corrected, and it now says why the scoped read cannot carry that weight.

khj809:
All of it is in `e769601`, and the three prose items travel with the change now rather than trailing it.

Two things I had wrong. The load-bearing line is the control, not the scoped read — I ran both mutations and yours is what happens: the placeholder on the scoped read leaves the test green because a read that fails on auth reaches `assert not reads(...)` exactly as an ACL miss does. My verification had mutated the control and the sentence named the other line. And `--username` alone was worse than I had it: with no token the launcher sent the admin's under that address while printing that responses were filtered to that user, so both halves are refused now, in all three launchers, with the admin token exempt both ways since that is the case `docs/auth.md` calls out.

On the case-insensitivity row you asked for: measured on the real site, `AVA.CHEN@…:<token>` and `Ava.chen@…:<token>` both answer 200 as that same account. Two more rows in the table, and `auth.py` states the measurement instead of attributing the behaviour.

`docs/auth.md:48`, `examples/using-official-sdk/README.md:97` and the Basic test's docstring are all rewritten; `_visible_doc` replaces the duplicated lookup and the bare `next()`; `tests/test_auth.py`'s module docstring names Basic alongside the API-key scheme.

Your 401-versus-anonymous table is the more interesting find and I would rather it were its own issue — please file it. The gap predates this PR (#44 shaped those errors for client parsing rather than from a measurement), closing it needs an anonymous caller on those routes, and it changes the answer for every failed credential rather than the pair specifically. I have added a short section to the body scoping the table to `myself` and pointing at it, so the claim there does not read wider than what was measured.

On your `test_generated_docs_are_current` failure: `PYTHONPATH=.` clears it here too, and it passes in this worktree, so I have taken it as the editable-install path rather than the PR.

examples/using-official-sdk/jira.py
nuri-yoo: Confirmed on `e769601` against a live server: the four cases per launcher come out as you list them for `jira.py` and `confluence.py`, and both mutations reproduce (placeholder on the scoped read 1 passed, on the control fails with the "should read an issue they can see" message).

One place the admin exemption did not reach: `examples/using-mcp-with-agents/atlassian.py` tests `args.token` rather than `args.token not in (None, s.token)`, so it differs from the other two on the two admin-token inputs.

```
$ atlassian.py --url $BASE --token admin-service-token
--token names one user, so pass --username with that user's own email too: ...
[exit 1]

$ atlassian.py --url $BASE --username ava@acme.com --token admin-service-token
authenticating as ava@acme.com → retrieval is ACL-filtered to that user
(proceeds, sending the admin token under ava's address)
```

`jira.py` and `confluence.py` accept the first and refuse the second. The body's "The admin token is exempt both ways" covers all three launchers, so this launcher is the one it does not yet describe. Mirroring `named_token` there (`args.token not in (None, s.token)`) would close both. Not blocking.

nuri-yoo (approved):
Re-ran on `e769601`: 1870 passed, 0 skipped with every extra installed; both test mutations reproduce as the body now says; the curl matrix matches the six-row table, including the two case rows; `ruff` and `gen_docs.py --check` clean. The one remaining item is the `atlassian.py` admin-token exemption in the thread above, not blocking. The 401-versus-anonymous table is filed as #123.

PR #115, opened 2026-09-03
s3: the canonical query is the wire query string, so a key containing `?` verifies

### LINKED ISSUE #113: s3: a key containing `?` fails SigV4 verification, because the canonical query is read from the rebuilt URL
**What you ran** — boto3 against a Backlot server, asking for an object whose key contains a `?`. The key does not exist in the corpus, so the expected answer is `NoSuchKey`; the point is what the signature check says before that.

```python
import json, urllib.request, backlot, boto3
from botocore.config import Config
from botocore.exceptions import ClientError

with backlot.serve() as s:
    d = json.load(urllib.request.urlopen(f"{s.base_url}/_meta/users"))
    c = boto3.client(
        "s3", endpoint_url=f"{s.base_url}/s3", region_name="us-east-1",
        aws_access_key_id=d["admin_s3_access_key_id"],
        aws_secret_access_key=d["admin_s3_secret_access_key"],
        config=Config(signature_version="s3v4", s3={"addressing_style": "path"}, retries={"max_attempts": 0}),
    )
    for key in ["a b.txt", "a#b.txt", "x=y&z.txt", "q?x.txt"]:
        try:
            c.get_object(Bucket="acme-artifacts", Key=key)
        except ClientError as e:
            print(f"{key!r:12} -> {e.response['Error']['Code']}")
```

**What happened**

```
'a b.txt'    -> NoSuchKey
'a#b.txt'    -> NoSuchKey
'x=y&z.txt'  -> NoSuchKey
'q?x.txt'    -> SignatureDoesNotMatch
```

The same for a SigV4 presigned URL (`generate_presigned_url("get_object", ...)` with `signature_version="s3v4"`): `a b.txt` answers `NoSuchKey`, `q?x.txt` answers `SignatureDoesNotMatch`. boto3 puts `/s3/acme-artifacts/q%3Fx.txt` on the wire in both cases, and a request signed by hand for exactly that path with `backlot.sigv4.expected_signature` is rejected the same way, so the client is not the one that disagrees.

**What you expected instead** — `NoSuchKey`, as for every other key. `?` is a legal character in an S3 object key; the S3 User Guide's "Object key naming guidelines" lists it among the "characters that might require special handling", which "most likely must be URL encoded", and `%3F` is that encoding. Real S3 signs the encoded path and verifies it.

**Where it is** — `backlot/auth.py:224-230`, in `resolve_sigv4`. The canonical URI is taken from `scope["raw_path"]`, which is right (S3 signs the wire path verbatim, and uvicorn hands it over untouched as `/s3/acme-artifacts/q%3Fx.txt`). The canonical query string is taken from `request.url.query`, which is not: Starlette builds `request.url` from the *decoded* `scope["path"]`, so for this request it assembles `http://host/s3/acme-artifacts/q?x.txt` and then reads it back as path `/s3/acme-artifacts/q` with query `x.txt`. The verifier canonicalises `x.txt=` where the client signed an empty query, and the two strings-to-sign differ. Measured on Starlette 1.6.0:

```python
>>> scope = {..., "path": "/s3/b/q?x.txt", "raw_path": b"/s3/b/q%3Fx.txt", "query_string": b""}
>>> Request(scope).url.query
'x.txt'
>>> Request(scope).url.path
'/s3/b/q'
```

`#` does not trip it because the rebuilt URL treats everything after it as a fragment, which changes `url.path` (unused here) and leaves `url.query` empty. `request.query_params` is built from `scope["query_string"]` directly, which is why the presigned branch still finds `X-Amz-Signature` and fails at the signature rather than at `MissingSecurityHeader`.

**The fix** is to read the query the same way the path is read: `request.scope.get("query_string", b"").decode()` in place of `request.url.query` at `auth.py:230`. Nothing else in the module rebuilds a URL from the decoded path. A regression test is one signed `GET` for a key with `?` asserting `NoSuchKey` rather than `SignatureDoesNotMatch`, beside the existing SigV4 cases in `tests/test_s3.py`.

**Where you ran it** — main at `4ea200a` (the file is identical on `feat/mcp-subcommand` and on #109's branch), Python 3.14.4, macOS 26.5.2, boto3 1.43.56, uvicorn 0.52.4 with httptools, Starlette 1.6.0. Not behind any optional dependency. #109 makes this reachable from an agent as well — its S3 bridge signs each `object_get` tool call with the same module — but the defect is on the verifying side and predates it.

nuri-yoo:
Closes #113.

A key containing `?` is refused with `SignatureDoesNotMatch` where every other absent key answers `NoSuchKey`. `?` is legal in an S3 key; the S3 User Guide lists it among the "characters that might require special handling" that "most likely must be URL encoded", and `%3F` is that encoding, which is what boto3 puts on the wire.

## Where it was

`resolve_sigv4` took the canonical URI from `scope["raw_path"]`, which is right, and the canonical query from `request.url.query`, which is not. Starlette assembles `request.url` from the decoded `scope["path"]`, so `/s3/acme-artifacts/q%3Fx.txt` becomes `http://host/s3/acme-artifacts/q?x.txt` and reads back as path `/s3/acme-artifacts/q` with query `x.txt`. The verifier canonicalised `x.txt=` where the client had signed an empty query. `#` did not trip it because the rebuilt URL treats the rest as a fragment, which leaves `url.query` empty; `=` and `&` in a key do not split a path. Measured on Starlette 1.6.0 with uvicorn 0.52.4 (httptools), which hands `raw_path` over verbatim.

## What changed

`backlot/auth.py` reads the query from `scope["query_string"]`, the bytes uvicorn received, the same way the path already comes from `raw_path`. One line of behaviour; the comment above it says why `request.url` is the wrong source for either half.

`tests/test_s3.py`'s `_request` helper now decodes `path` into `scope["path"]` the way uvicorn does, so the unit tests build the same request the server sees. The two existing unit tests for a valid header signature and a valid presigned URL run over a key containing `?` as well as the plain path (`SIGNED_PATHS`), and one live test drives the server with a botocore-signed GET for such a key.

S3 verifies a request against the canonical request the client built, and the client builds it from the encoded path and the query string it sent. boto3 sends `q%3Fx.txt` with no query, header-signed or presigned, and real S3 answers about the key, not about the signature.

## Verification

- `pytest` on a `.[all]` install plus `llama-index-readers-hubspot<0.6 --no-deps`, Docker (colima), `npx` and `uvx` present: **1693 passed, 0 skipped**, 22.23s. `ruff check . && ruff format --check .` clean; `python scripts/gen_docs.py --check` exit 0.
- **Without the change** — `auth.py`'s one line reverted on this branch: **3 failed, 23 passed** in `tests/test_s3.py` — the live `?`-key test and the two `q%3Fx.txt` parameters; both plain-path parameters still pass.
- boto3 1.43.56 against a served corpus, `signature_version="s3v4"`, path-style: `get_object` for `a b.txt`, `a#b.txt`, `x=y&z.txt` and `q?x.txt` all answer `NoSuchKey`; a presigned URL for `q?x.txt` answers `NoSuchKey`; the one key the corpus has answers 200. Before the change the two `q?x.txt` cases answered `SignatureDoesNotMatch`.

- [x] A test fails without this change — the 3 above.
- [ ] A new endpoint serving corpus content is ACL-scoped — no new endpoint; the ACL path is untouched.

### DIFF
--- a/tests/test_s3.py
+++ b/tests/test_s3.py
@@ -686,4 +686,3 @@ def test_presigned_unexpired_ok(path):
     caller, err = auth.resolve_sigv4(req)
     assert err is None
     assert caller == Caller(email="ava@acme.com", is_admin=False)
-
--- a/tests/test_s3.py
+++ b/tests/test_s3.py
@@ -11,7 +11,7 @@ import urllib.request
 import xml.etree.ElementTree as ET
 from datetime import datetime, timedelta, timezone
 from types import SimpleNamespace
-from urllib.parse import quote, urlencode
+from urllib.parse import quote, unquote, urlencode
 
 import pytest
 import yaml
@@ -100,6 +100,21 @@ def test_s3_missing_key_is_nosuchkey(live_server):
 
     base_url, settings = live_server
     url, headers = _sign_get(base_url, "/s3/eng-artifacts/does/not/exist.md", settings.admin_token)
+    with pytest.raises(urllib.error.HTTPError) as e:
+        urllib.request.urlopen(urllib.request.Request(url, headers=headers))
+    assert e.value.code == 404 and b"NoSuchKey" in e.value.read()
+
+
+def test_s3_key_containing_a_question_mark_verifies(live_server):
+    """`?` is a legal character in a key, and a client sends it as `%3F`. Starlette rebuilds
+    `request.url` from the DECODED path, so for `/q%3Fx.txt` its `.query` is `x.txt` — a query the
+    client never sent or signed. The verifier canonicalises the wire query string instead, and an
+    absent key with a `?` in it answers NoSuchKey like every other absent key, not
+    SignatureDoesNotMatch. Signed by botocore, the way boto3 sends it."""
+    import urllib.request
+
+    base_url, settings = live_server
+    url, headers = _sign_get(base_url, "/s3/eng-artifacts/q%3Fx.txt", settings.admin_token)
     with pytest.raises(urllib.error.HTTPError) as e:
         urllib.request.urlopen(urllib.request.Request(url, headers=headers))
     assert e.value.code == 404 and b"NoSuchKey" in e.value.read()
@@ -527,12 +542,13 @@ def _acl():
 
 def _request(method, path, query, headers) -> Request:
     """A minimal Starlette Request mirroring what `resolve_sigv4` reads: headers,
-    query_params, method, url.query, and scope['raw_path'] — plus a fake app.state.acl
-    so `auth.acl(request)` resolves without a real ASGI app."""
+    query_params, method, scope['query_string'] and scope['raw_path'] — plus a fake app.state.acl
+    so `auth.acl(request)` resolves without a real ASGI app. `path` is the wire path; uvicorn
+    hands it over verbatim as `raw_path` and percent-decoded as `path`, and so does this."""
     scope = {
         "type": "http",
         "method": method,
-        "path": path,
+        "path": unquote(path),
         "raw_path": path.encode("ascii"),
         "query_string": query.encode("ascii"),
         "headers": [(k.lower().encode("ascii"), v.encode("ascii")) for k, v in headers.items()],
@@ -661,5 +677,21 @@ def test_presigned_unexpired_ok():
     current = datetime.now(timezone.utc).strftime(AMZ_DATE_FORMAT)
     req = _presigned_request(current, expires=3600)
     caller, err = auth.resolve_sigv4(req)
+    assert err is None
+    assert caller == Caller(email="ava@acme.com", is_admin=False)
+
+
+def test_canonical_query_is_the_wire_query_string_not_the_rebuilt_urls():
+    """A `%3F` in the path decodes to `?`, and Starlette's `request.url` is assembled from the
+    decoded path — so `url.query` for `/s3/eng-artifacts/q%3Fx.txt` is `x.txt`, a query the client
+    never sent or signed. The verifier has to canonicalise `scope["query_string"]`, the bytes that
+    were actually on the wire, the same way it already takes the path from `raw_path`. Both auth
+    forms, since both go through the same canonical request."""
+    current = datetime.now(timezone.utc).strftime(AMZ_DATE_FORMAT)
+    path = "/s3/eng-artifacts/q%3Fx.txt"
+    caller, err = auth.resolve_sigv4(_header_auth_request(current, path=path, query=""))
+    assert err is None
+    assert caller == Caller(email="ava@acme.com", is_admin=False)
+    caller, err = auth.resolve_sigv4(_presigned_request(current, expires=3600, path=path))
     assert err is None
     assert caller == Caller(email="ava@acme.com", is_admin=False)
--- a/tests/test_s3.py
+++ b/tests/test_s3.py
@@ -657,9 +657,15 @@ def test_header_auth_skew_check_precedes_signature_check():
     assert err == "RequestTimeTooSkewed"
 
 
-def test_header_auth_accepts_current_date():
+# A `%3F` in the key decodes to a `?` that splits Starlette's rebuilt `request.url`, so the
+# canonical request has to come off the wire — see the comment in `resolve_sigv4`.
+SIGNED_PATHS = ["/s3/eng-artifacts", "/s3/eng-artifacts/q%3Fx.txt"]
+
+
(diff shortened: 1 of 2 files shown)


tests/test_s3.py
```diff
@@ -663,3 +679,19 @@ def test_presigned_unexpired_ok():
...
+    were actually on the wire, the same way it already takes the path from `raw_path`. Both auth
+    forms, since both go through the same canonical request."""
+    current = datetime.now(timezone.utc).strftime(AMZ_DATE_FORMAT)
+    path = "/s3/eng-artifacts/q%3Fx.txt"
+    caller, err = auth.resolve_sigv4(_header_auth_request(current, path=path, query=""))
+    assert err is None
+    assert caller == Caller(email="ava@acme.com", is_admin=False)
+    caller, err = auth.resolve_sigv4(_presigned_request(current, expires=3600, path=path))
+    assert err is None
+    assert caller == Caller(email="ava@acme.com", is_admin=False)
```
khj809: This function is `test_header_auth_accepts_current_date` and `test_presigned_unexpired_ok` with a different `path` — same setup, same assertions, and both helpers already take `path=`. Parametrizing those two over the path drops this function without losing anything:

```diff
-def test_header_auth_accepts_current_date():
+# A `%3F` in the key decodes to a `?` that splits Starlette's rebuilt `request.url`, so the
+# canonical request has to come off the wire — see the comment in `resolve_sigv4`.
+SIGNED_PATHS = ["/s3/eng-artifacts", "/s3/eng-artifacts/q%3Fx.txt"]
+
+
+@pytest.mark.parametrize("path", SIGNED_PATHS)
+def test_header_auth_accepts_current_date(path):
     current = datetime.now(timezone.utc).strftime(AMZ_DATE_FORMAT)
-    req = _header_auth_request(current)
+    req = _header_auth_request(current, path=path)

-def test_presigned_unexpired_ok():
+@pytest.mark.parametrize("path", SIGNED_PATHS)
+def test_presigned_unexpired_ok(path):
     current = datetime.now(timezone.utc).strftime(AMZ_DATE_FORMAT)
-    req = _presigned_request(current, expires=3600)
+    req = _presigned_request(current, expires=3600, path=path)
```

Measured on this branch with that edit: `tests/test_s3.py` 26 passed, `ruff check` and `ruff format --check` clean. Reverting only the `auth.py` line fails exactly three — the live `?`-key test plus the two `[/s3/eng-artifacts/q%3Fx.txt]` parameters — while both plain-path parameters still pass. Identical regression coverage, one function fewer.

It also gets a little stronger: the parametrized header case keeps the helper's default `query="list-type=2"`, so it catches a query the client actually signed being canonicalised as `x.txt?list-type=2`, where `query=""` only compares an empty query against `x.txt`.

Side benefit: the Starlette explanation appears three times right now (the `resolve_sigv4` comment, the live test's docstring, this docstring). Folding it in leaves the `auth.py` comment as the single source.

To be clear, `test_s3_key_containing_a_question_mark_verifies` should stay exactly as it is — it is the only test that proves uvicorn hands `raw_path` and `query_string` over verbatim, which is the premise of the fix, and `_sign_get`'s `importorskip("botocore")` means it would skip the regression by itself wherever botocore is absent. The redundancy is between this unit test and the two unit tests above it, not between the layers.

tests/test_s3.py
```diff
@@ -663,3 +679,19 @@ def test_presigned_unexpired_ok():
     caller, err = auth.resolve_sigv4(req)
     assert err is None
     assert caller == Caller(email="ava@acme.com", is_admin=False)
+
+
+def test_canonical_query_is_the_wire_query_string_not_the_rebuilt_urls():
```
khj809: Nit on the name: `_not_the_rebuilt_urls` is about a single `request.url`, so the plural reads as a typo — `..._not_the_rebuilt_url`. Moot if this folds into the parametrize.

tests/test_s3.py
nuri-yoo: Taken as written, in 131d18d (and e7ac6c3 for the formatter). `SIGNED_PATHS` parametrizes both unit tests and the third is gone; the live test stays. Re-measured: `tests/test_s3.py` 26 passed; with the `auth.py` line reverted, exactly the three you named fail and both plain-path parameters pass. Body updated to match.

tests/test_s3.py
nuri-yoo: Moot now — the function is gone with the parametrize.

khj809 (approved):
Looks good

PR #116, opened 2026-09-03
linear: the labels filter answers a label-less issue the way Linear does

### LINKED ISSUE #112: Linear's labels filter answers the opposite of Backlot's over an issue with no labels
_Found while measuring for #111; out of that PR's scope because the fix needs data the reachable workspace does not have._

`IssueFilter.labels` is compiled as `some` = `EXISTS (label WHERE predicate)` and `every` = `NOT EXISTS (label WHERE NOT predicate)` in `backlot/graphql/linear_filters.py::_labels_predicate`. Over an issue with **no labels**, api.linear.app answers the opposite of that on two counts (measured 2026-09-03, four issues in a workspace with no labels):

| filter | Linear | Backlot |
|---|---|---|
| `labels: {some: {name: {eq: "zzz"}}}` | none | none |
| `labels: {every: {name: {eq: "zzz"}}}` | none | **all** (vacuous truth) |
| `labels: {some: {name: {neq: "zzz"}}}` | **all** | none |
| `labels: {every: {name: {neq: "zzz"}}}` | all | all |
| `labels: {some: {name: {nin: ["zzz"]}}}` | **all** | none |
| `labels: {some: {name: {notContains: "zzz"}}}` | **all** | none |
| `labels: {some: {name: {contains: ""}}}` | none | none |
| `labels: {every: {name: {contains: ""}}}` | none | **all** |
| `labels: {length: {eq: 0}}` | all | not declared (a `gap` in the baseline) |

The pattern is consistent with Linear pushing a negation outside the quantifier — `some: {name: {neq: X}}` reading as "no label is named X" and `every: {name: {eq: X}}` as "has labels, and all are X" — but a workspace with no labelled issue cannot tell that reading from "some label is not named X". The two disagree on an issue labelled `["a", "zzz"]`, so the measurement that decides the fix needs an issue with two labels (an `issueLabelCreate` on a team, an `issueCreate` with `labelIds`, both deleted after), then the eight filters above plus `some`/`every` over `{eq: "a"}`, `{neq: "a"}`, `{in: ["a", "b"]}`, `{nin: ["a"]}`.

### Done when

- [ ] The table above is re-measured with a labelled issue present and the reading is written into `_labels_predicate`'s docstring as the vendor's rule
- [ ] `some` and `every` over an empty label set answer what Linear answers for a positive and for a negative predicate
- [ ] `labels: {length: …}` is either served or stays an acknowledged gap with a note
- [ ] Each change brings a test that fails without it, per `AGENTS.md`

nuri-yoo:
Closes #112.

## What changed

**`some` and `every` answer a label-less issue the way Linear does.** Backlot compiled `labels: {some: P}` as `EXISTS (label WHERE P)` and `labels: {every: P}` as `NOT EXISTS (label WHERE NOT P)`, the textbook quantifiers, and #112 measured them answering the opposite of api.linear.app on an issue with no labels. With labelled issues present the rule is knowable, and it is this: the polarity of the predicate decides what an issue with no labels answers. `every` over a positive predicate (`eq`, `in`, `contains`, …) needs at least one label and every label satisfying it. `some` over a negative predicate (`neq`, `nin`, `neqIgnoreCase`, …) is the complement of that: no labels, or one label satisfying it. `every` over a negative predicate is the vacuous "no label fails it", which a label-less issue passes; `some` over a positive one is the plain EXISTS. Each negative form is the negation of the other quantifier's positive form, so Linear is pushing the `not` outside the quantifier, which is the reading #112 guessed at and could not confirm without a labelled issue. A compound predicate has a polarity too, measured on mixed shapes: `and` is negative only when every branch is, `or` when any branch is, and a comparator when every operator in it is (`{eq: A, neq: B}` reads positive). "Every operator of none" is true, so a branch with nothing in it reads as negative, which is what the `or` rule below turns on. `_labels_predicate` compiles the four cases, `_reads_as_negation` decides the polarity, and the docstring on `_labels_predicate` carries the rule with the measurement.

**The rest of `IssueLabelCollectionFilter`, declared and served, with its keys read in Linear's order.** The same throwaway issues settled the fields the baseline listed as gaps. `length` is a `NumberComparator` over the label count (`{eq: 0}` answered the label-less issues, `{eq: 2}` the two-label one, `{in: [0, 2]}` both, `{gte: 1, lte: 1}` the one-label one). `null: true` answered nothing at all, an issue's label collection being never null, not even when empty; `null: false` answered everything. A bare `name` at the collection level answered as `some` does. The keys do not AND together: `and` answers if present, else `or`, else the first present of `length`, `every`, `some`, `name`, and the others are ignored whatever their order. `{and: [{length: {eq: 2}}], or: [{length: {eq: 1}}]}` answered the two-label issue, where an AND is empty; `{or: [{length: {eq: 0}}], length: {eq: 2}}` the label-less issues; `{length: {eq: 1}, some: {name: {eq: z}}}` the one-label issue; `{some: {name: {eq: z}}, every: {name: {eq: a}}}` the one-label issue; `{name: {eq: z}, some: {name: {eq: a}}}` both labelled issues. `null` is the one key that ANDs, and only beside those four: `{or: [{length: {eq: 0}}], null: true}` answered the label-less issues, the `or` alone. The SDL declared `and` / `or` and the compiler refused them with "not supported by Backlot"; `_labels_filter` now reads the keys in `_LABELS_PRECEDENCE`. The baseline drops the three gaps this closes (`length`, `name`, `null`). The served description of the type says the precedence too.

**A predicate that constrains nothing, and the `or` branch that is not the same thing.** `labels: {}`, `labels: {some: {}}`, `labels: {every: {}}`, `labels: {some: {name: {}}}`, `labels: {some: {and: []}}` and `labels: {some: {or: [{}]}}` each answered every issue on Linear, the label-less ones included. Backlot refused the first two ("a labels filter must constrain something"), on the reasoning that an empty fragment dropping the whole filter was a silent wrong answer. The vendor's answer is that an empty predicate is no predicate, so they compile to nothing now, and the test that pinned the refusal pins that instead. A literally empty `or` is the exception at the top of a predicate: `some: {or: []}` and `every: {or: []}` each answered exactly the issues with at least one label, so that `or` is a predicate every label satisfies and the quantifier still asks for a label. Inside an `or`, though, the empty fragment is where dropping it turns a refusal into a quietly narrower answer, and the three levels answer it differently, each measured with labelled issues present:

- At the top of `IssueFilter`, a branch with a key that constrains nothing makes the WHOLE `or` constrain nothing, its other keys included: `{or: [{labels: {}}, {title: {eq: T}}]}`, `{or: [{labels: {some: {}}}, …]}`, `{or: [{title: {}}, …]}`, `{or: [{state: {name: {}}}, …]}` and `{or: [{labels: {}, title: {eq: "no such title"}}, {title: {eq: T}}]}` each answered every issue, where the title branch alone answers one. A branch with nothing in it is dropped: `{or: [{}, {title: {eq: T}}]}`, `{or: [{and: []}, …]}`, `{or: [{or: []}, …]}` and `{or: [{title: null}, …]}` each answered the title match alone. Inside an `and` a vacuous branch is dropped: `{and: [{labels: {}}, {title: {eq: T}}]}` answered the title match. `_issue_parts` returns a `vacuous` flag beside the fragment, which is how the empty fragment of a dropped branch is told from the empty fragment of a vacuous one.
- In an `IssueLabelCollectionFilter`'s `or`, any branch compiling to nothing makes the filter constrain nothing: `{or: [{}, {length: {eq: 2}}]}` and `{or: [{some: {}}, {length: {eq: 2}}]}` each answered every issue; in its `and` the branch is dropped, `{and: [{}, {length: {eq: 2}}]}` answered the two-label issue.
- In an `IssueLabelFilter`'s `or`, a branch with nothing in it (`{}`, `{and: []}`) is dropped and the `or` reads as a negative predicate: `some: {or: [{}, {name: {eq: z}}]}` answered `[a, z]` plus the label-less issues and `every: {or: [{}, {name: {eq: z}}]}` the label-less issues alone, the answers `some` / `every: {name: {neq: a}}` give. A `{name: {}}` branch makes the whole predicate constrain nothing (`every: {or: [{name: {}}, {name: {eq: z}}]}` answered every issue); a `{or: []}` branch is dropped and reads positive (`some: {or: [{or: []}, {name: {eq: z}}]}` answered `[a, z]` alone). `_label_predicate` returns one of four kinds with the fragment (none, quant, vacuous, real) and `_label_branches` acts on it.

**A null operand is a condition.** `eq`, `neq`, `lt`, `lte`, `gt` and `gte` with `null` match nothing, on a nullable field too: `title: {eq: null}`, `dueDate: {eq: null}` over issues with no due date, `estimate: {gte: null}`, `labels: {length: {eq: null}}` and `labels: {some: {name: {eq: null}}}` each answered none. A string or list operator with `null` is a condition every row passes: `title: {contains: null}` answered every issue, `labels: {some: {name: {contains: null}}}` and `{in: null}` the labelled issues only, and `labels: {some: {name: {neqIgnoreCase: null}}}` every issue, a negative predicate every label passes. Backlot skipped a null operand, which answered every issue for all of them, and under the `or` rule above the skip would have read as a branch that constrains nothing. `_Comparator.render` renders it as `0` or `1` by operator. `null: null` is the one operand that is not, and it reads three ways: on a comparator as `null: true` (`dueDate: {null: null}` answered every issue with no due date, `priority: {null: null, eq: 0}` none over four priority-0 issues, as `{null: true, eq: 0}` does), on a relation as no key at all (with one issue assigned beside three unassigned, `assignee: {null: null}` answered all four, `{null: null, name: {eq: ME}}` the assigned one, `{null: null, name: {eq: "nobody"}}` none), and on the label collection as no key at all (`labels: {null: null}` every issue, `{null: null, length: {eq: 0}}` the label-less ones). `_Comparator.render` reads it as `true`; `_sub_filter` and `_labels_filter` drop it.

Also measured and pinned: an empty `in` matches no label and an empty `nin` every label, on `length` and inside both quantifiers; a bare `name` with a negative comparator answers as `some` does; a predicate mixing a comparator with a nested `and` / `or` is one AND, polarity included.

Not touched: the operators Linear's `StringComparator` has and Backlot's SQL comparator does not (`notContains`, `notStartsWith`, …), which measured as negative predicates and are named in `_NEGATIVE_OPS` for when they are declared; and `IssueLabelFilter.id` (`some: {id: {eq: <label id>}}` answered the issue carrying that label), a separate gap this does not claim.

Measured against api.linear.app with a personal API key, in a workspace whose four issues carry no labels. Two labels (`zz-a`, `zz-zzz`) and two issues (`[zz-a, zz-zzz]`, `[zz-a]`) were created with `issueLabelCreate` / `issueCreate(labelIds:)`, the filters were run, and all four were deleted (`issueDelete` and `issueLabelDelete` each answered `success: true`; the workspace lists 4 issues and its 3 original labels after) — 93 filters over four runs for the first table, and for the second, after review, 60 read-only filters over the label-less workspace plus 112 with the same two labelled issues re-created and deleted the same way. Every cell below is what came back; `L2` is the two-label issue, `L1` the one-label issue, `U` the four label-less ones, `T` the title of one of them.

| filter | Linear |
|---|---|
| `some: {name: {eq: "zz-a"}}` / `{eq: "zz-zzz"}` / `{eq: "nope"}` | L2 L1 / L2 / none |
| `some: {name: {neq: "zz-a"}}` / `{neq: "zz-zzz"}` / `{neq: "nope"}` | L2 U / L2 L1 U / L2 L1 U |
| `some: {name: {nin: ["zz-a"]}}` / `{notContains: "zz-a"}` / `{neqIgnoreCase: "ZZ-A"}` / `{notStartsWith: "zz-a"}` | L2 U, all four |
| `some: {name: {in: ["zz-a", "nope"]}}` / `{contains: ""}` / `{eqIgnoreCase: "ZZ-A"}` | L2 L1, all three |
| `every: {name: {eq: "zz-a"}}` / `{eq: "zz-zzz"}` / `{eq: "nope"}` | L1 / none / none |
| `every: {name: {neq: "zz-a"}}` / `{neq: "zz-zzz"}` / `{neq: "nope"}` | U / L1 U / L2 L1 U |
| `every: {name: {nin: ["zz-a"]}}` / `{notContains: "zz-a"}` / `{neqIgnoreCase: "ZZ-A"}` | U, all three |
| `every: {name: {in: ["zz-a", "zz-zzz"]}}` / `{contains: ""}` / `{startsWith: "zz"}` | L2 L1, all three |
| `some: {and: [{neq: "zz-a"}, {neq: "nope"}]}` / `every: {and: [same]}` | L2 U / U |
| `some: {and: [{eq: "zz-a"}, {neq: "nope"}]}` / `{eq: "zz-a", neq: "nope"}` / `every: {and: [{eq: "zz-a"}, {neq: "nope"}]}` | L2 L1 / L2 L1 / L1 |
| `some: {or: [{neq: "zz-a"}, {eq: "nope"}]}` / `every: {or: [{eq: "zz-zzz"}, {neq: "zz-a"}]}` | L2 U / U |
| `length: {eq: 0}` / `{eq: 1}` / `{eq: 2}` / `{gt: 0}` / `{lt: 2}` / `{neq: 0}` / `{in: [0, 2]}` / `{nin: [0]}` / `{gte: 1, lte: 1}` | U / L1 / L2 / L2 L1 / L1 U / L2 L1 / L2 U / L2 L1 / L1 |
| `{or: [{length: {eq: 0}}, {every: {name: {eq: "zz-a"}}}]}` | L1 U |
| `{some: {name: {eq: "zz-a"}}, every: {name: {eq: "zz-a"}}}` | L1 |
| `{name: {eq: "zz-a"}}` (no quantifier) | L2 L1 |
| `{null: true}` / `{null: false}` / `{null: null}` | none / L2 L1 U / L2 L1 U |
| `{}` / `{some: {}}` / `{every: {}}` / `{some: {name: {}}}` / `{some: {and: []}}` / `{some: {or: [{}]}}` / `{and: []}` / `{or: []}` | L2 L1 U, all eight |
| `{some: {or: []}}` / `{every: {or: []}}` | L2 L1, both |
| `{some: {or: [], name: {neq: "zz-a"}}}` / `{some: {name: {neq: "zz-a"}, and: [{name: {eq: "zz-zzz"}}]}}` / `{every: {name: {neq: "nope"}, or: [{name: {eq: "zz-a"}}]}}` | L2 / L2 / L1 |
| `{name: {neq: "zz-a"}}` / `{name: {nin: ["zz-a"]}}` / `{name: {eq: "zz-a", neq: "nope"}}` (no quantifier) | L2 U / L2 U / L2 L1 |
| `{length: {in: []}}` / `{some: {name: {in: []}}}` / `{every: {name: {in: []}}}` | none, all three |
| `{length: {nin: []}}` / `{some: {name: {nin: []}}}` / `{every: {name: {nin: []}}}` | L2 L1 U, all three |
| `{id: {eq: <zz-zzz id>}}` / `{some: {id: {eq: <zz-zzz id>}}}` | L2 (not implemented here) |

The review found that an `or` branch compiling to nothing was dropped, and with it that the collection filter's keys are not an AND. Measured after it, with the same labelled issues present:

| filter | Linear |
|---|---|
| `{some: {name: {eq: "zz-zzz"}}, every: {name: {eq: "zz-a"}}}`, either order | L1 |
| `{length: {eq: 1}, some: {name: {eq: "zz-zzz"}}}` / `{length: {eq: 2}, every: {name: {eq: "zz-a"}}}` / `{length: {eq: 1}, name: {eq: "zz-zzz"}}`, either order | L1 / L2 / L1 |
| `{name: {eq: "zz-zzz"}, some: {name: {eq: "zz-a"}}}` / `{every: {name: {eq: "zz-a"}}, name: {eq: "zz-zzz"}}`, either order | L2 L1 / L1 |
| `{null: false, some: {name: {eq: "zz-zzz"}}}` / `{null: true, some: …}` / `{length: {eq: 1}, null: false}` | L2 / none / L1 |
| `{and: [{length: {eq: 2}}], or: [{length: {eq: 1}}]}` / `{and: [{length: {eq: 1}}], or: [{length: {eq: 2}}]}`, either order | L2 / L1 |
| `{or: [{length: {eq: 1}}], some: {name: {eq: "zz-zzz"}}}` / `{and: [{length: {eq: 2}}], every: {name: {eq: "zz-a"}}}` | L1 / L2 |
| `{or: [{length: {eq: 0}}], length: {eq: 2}}` / `{or: [{length: {eq: 99}}], length: {eq: 0}}` / `{or: [{length: {eq: 0}}], null: true}` / `{and: [{length: {eq: 2}}], null: true}` | U / none / U / L2 |
| `{and: [{}], length: {eq: 2}}` / `{and: [{length: {eq: 2}}], or: [{}]}` / `{and: [{}], or: [{length: {eq: 2}}]}` / `{or: [{length: {eq: 2}}], and: []}` | L2 L1 U / L2 / L2 L1 U / L2 L1 U |
| `{or: [{}, {length: {eq: 2}}]}` / `{or: [{some: {}}, {length: {eq: 2}}]}` / `{and: [{}, {length: {eq: 2}}]}` | L2 L1 U / L2 L1 U / L2 |
| `{some: {or: [{}, {name: {eq: "zz-zzz"}}]}}` / `{every: {or: [{}, …]}}` / `{some: {or: [{and: []}, …]}}` / `{every: {and: [{or: [{}, …]}]}}` | L2 U / U / L2 U / U |
| `{some: {or: [{}, {name: {neq: "zz-zzz"}}]}}` / `{every: {or: [{}, {name: {neq: "zz-zzz"}}]}}` | L2 L1 U / L1 U |
| `{some: {or: [{or: []}, {name: {eq: "zz-zzz"}}]}}` / `{some: {or: [{name: {}}, …]}}` / `{every: {or: [{name: {}}, …]}}` / `{some: {or: [{and: [{}]}, …]}}` | L2 / L2 L1 U / L2 L1 U / L2 L1 U |
| `{some: {and: [{}, {name: {eq: "zz-zzz"}}]}}` / `{some: {and: [{name: {}}, …]}}` / `{some: {and: [{and: []}, …]}}` / `{some: {and: [{}]}}` | L2 / L2 / L2 / L2 L1 U |
| `{some: {name: {eq: "zz-zzz"}, or: [{}]}}` / `{some: {name: {eq: "zz-zzz"}, or: [{}, {name: {eq: "zz-a"}}]}}` | L2 / none |
| `{some: {name: {eq: null}}}` / `{every: {name: {eq: null}}}` / `{some: {name: {neq: null}}}` / `{every: {name: {neq: null}}}` | none / none / U / U |
| `{some: {name: {contains: null}}}` / `{every: {name: {contains: null}}}` / `{startsWith: null}` / `{endsWith: null}` / `{eqIgnoreCase: null}` / `{containsIgnoreCase: null}` / `{in: null}` under `some` | L2 L1, all seven |
| `{some: {name: {neqIgnoreCase: null}}}` / `{every: {name: {nin: null}}}` / `{some: {name: {eq: "zz-a", contains: null}}}` / `{length: {eq: 2, lt: null}}` / `{length: {eq: null}}` | L2 L1 U / L2 L1 U / L2 L1 / none / none |
| `{or: [{}, {title: {eq: T}}]}` / `{or: [{and: []}, …]}` / `{or: [{or: []}, …]}` / `{or: [{title: null}, …]}` | the T issue, all four |
| `{or: [{labels: {}}, {title: {eq: T}}]}` / `{labels: {some: {}}}` / `{labels: {every: {}}}` / `{labels: {null: false}}` / `{labels: {and: []}}` / `{labels: {some: {name: {}}}}` / `{title: {}}` / `{priority: {}}` / `{estimate: {}}` / `{state: {}}` / `{state: {name: {}}}` / `{or: [{}]}` / `{and: [{}]}` in that branch | every issue, all thirteen |
| `{or: [{labels: {}, title: {eq: "no such title"}}, {title: {eq: T}}]}` / `{or: [{or: [{labels: {}}, {title: {eq: "no such title"}}]}, {title: {eq: T}}]}` | every issue, both |
| `{and: [{labels: {}}, {title: {eq: T}}]}` / `{or: [{and: [{labels: {}}, {title: {eq: "no such title"}}]}, {title: {eq: T}}]}` / `{labels: {}, title: {eq: "no such title"}}` | the T issue / the T issue / none |
| `{or: [{labels: {length: {eq: 1}}}, {title: {eq: T}}]}` / `{or: [{labels: {some: {name: {eq: "zz-zzz"}}}, title: {eq: "no such title"}}, {title: {eq: T}}]}` | L1 + the T issue / L2 + the T issue |
| `{title: {eq: null}}` / `{neq: null}` / `{dueDate: {eq: null}}` / `{estimate: {gte: null}}` / `{priority: {eq: null}}` | none, all five |
| `{title: {contains: null}}` / `{in: null}` / `{nin: null}` / `{startsWith: null}` / `{eqIgnoreCase: null}` / `{neqIgnoreCase: null}` | every issue, all six |
| `{or: [{title: {eq: null}}, {title: {eq: T}}]}` / `{or: [{title: {contains: null}}, {title: {eq: "no such title"}}]}` / `{or: [{dueDate: {eq: null}}, {title: {eq: "no such title"}}]}` | the T issue / every issue / none |
| `{dueDate: {null: null}}` / `{estimate: {null: null}}` / `{completedAt: {null: null}}` / `{canceledAt: {null: null}}` (no issue has any of the four) / `{priority: {null: null}}` / `{priority: {null: null, eq: 0}}` (every issue carries priority 0) / `{or: [{priority: {null: null}}, {title: {eq: T}}]}` | every issue, all four / none, both / the T issue |
| `{assignee: {null: null}}` / `{null: true}` / `{null: false}` / `{null: null, name: {eq: ME}}` / `{null: null, name: {eq: "nobody"}}` / `{or: [{assignee: {null: null}}, {title: {eq: "no such title"}}]}`, with one issue assigned to ME beside three | every issue / the three / the one / the one / none / every issue |
| `{labels: {null: null}}` / `{labels: {null: null, length: {eq: 0}}}` / `{project: {null: null}}` / `{creator: {null: null}}` (nothing has a project or a creator) | every issue, all four |

The `some`/`every` cells over `neq`, `nin`, `notContains` and `eq` are the eight #112 tabulated on label-less issues, now with the labelled rows that decide the reading. The fixture in `tests/test_linear.py` has the same shape as the measurement (`ENG-1` `[bug, gateway]`, `ENG-2` `[bug]`, `DES-1` none), so every cell above that Backlot's `StringComparator` can express is pinned: 159 collection-level cells in `test_labels_quantifiers_answer_as_linear` and 81 whole-filter cells in `test_issue_filter_or_answers_a_vacuous_branch_as_linear`.

`backlot diff --source linear` on `6bcd0c0`, re-run 2026-09-04: `776 divergence(s) · 775 acknowledged · 1 new`. The one new finding is `missing_enum_value IntegrationService.origin`, a member Linear added after 14:02Z on 2026-09-03; `main` at `ee8be69` answers `779 · 778 · 1` with the same finding, so it is not this branch's and is fixed against `main` separately. The difference of three is the `IssueLabelCollectionFilter` gaps this closes.

## Verification

- **Tests** — `pytest -rs` on the full-extras venv, Docker, `npx` and `uvx` present: `2104 passed`, no skips reported.
- **Lint** — `ruff check . && ruff format --check .`: `All checks passed!` / `138 files already formatted`. `python scripts/gen_docs.py --check`: exit 0.
- **Fails without the fix** — with `linear_filters.py` at `main` and the tests kept: `191 failed, 220 passed`; with only the SDL at `main`: `119 failed, 292 passed`; with the compiler as it stood after the two applied review suggestions: `101 failed, 310 passed`. One change at a time, each restored from a copy: skipping null operands again fails 29 cells, reading an empty `or` branch as positive 5, the applied first suggestion's `or` arm 9, ANDing the collection keys 13, reading `or` before `and` 9, dropping a vacuous `or` branch at the top of `IssueFilter` 22, rendering a relation's `null: null` as `IS NOT NULL` again 3, reading a comparator's as no operand 8.
- **Live re-measurement** — `backlot diff --source linear`: above, `776 · 775 · 1` on `6bcd0c0`, the one shared with `main`.


khj809 (commented):
Reviewed against `03ebe29` with a live re-measurement. The quantifier-polarity rule and its 75-cell matrix hold up: `pytest -q tests/test_linear.py` is green, reverting `linear_filters.py` to `main` fails most of the new matrix, and `backlot diff --source linear` really does close the three `IssueLabelCollectionFilter` gaps.

Four findings, all one area. Every input that used to be a 400 here (`must constrain something`, `needs \`some\` or \`every\``, or an undeclared field) now compiles to an empty fragment, and an empty fragment is *dropped from an `or`* — so a filter Linear answers with everything is answered narrowed. Measured 2026-09-03 against api.linear.app, read-only queries over a workspace of four label-less issues; the right column is this branch on its own fixture (ENG-1 `[bug, gateway]`, ENG-2 `[bug]`, DES-1 none):

| filter | Linear | this branch |
|---|---|---|
| `{labels: {some: {or: [{}, {name: {eq: A}}]}}}` | every issue | `[ENG-1]` |
| `{labels: {every: {or: [{}, {name: {eq: A}}]}}}` | every issue | `[]` |
| `{labels: {or: [{}, {length: {eq: N}}]}}` | every issue | `[ENG-1]` |
| `{or: [{labels: {some: {}}}, {title: {eq: T}}]}` | every issue | `[DES-1]` |
| `{labels: {or: [{length: {eq: 0}}], length: {eq: N}}}` | `[DES-1]`, the `or` alone | `[]` |
| `{labels: {and: [{}], length: {eq: N}}}` | every issue | `[ENG-1]` |

The two suggestions below are verified together: `ruff check` and `ruff format --check` clean, the whole of `tests/test_linear.py` still passes, and every `or` row above then answers what Linear answers.

One thing on the description: the `backlot diff --source linear` paragraph is stale. #111 is merged into this branch now (`b987c94`), so the head answers `775 divergence(s) · 775 acknowledged · 0 new` — the number the description still frames as a hypothetical local merge with #111's head, rather than `826 · 779 · 47`.

backlot/graphql/linear_filters.py
```diff
@@ -471,6 +471,13 @@ def _label_predicate(spec: dict) -> tuple[str, list]:
...
+            if key == "or" and sub == []:
+                # Measured 2026-09-03: a literally empty `or` is a predicate every label satisfies,
+                # and the quantifier still applies -- `some: {or: []}` and `every: {or: []}` both
+                # answered exactly the issues with at least one label. An empty `and`, and an `or`
+                # whose branches are themselves empty (`or: [{}]`), are no predicate at all: every
+                # issue answered, label-less ones included, which the empty fragment below gives.
+                frags = ["1"]
             frag, p = (
                 (("(" + (" AND " if key == "and" else " OR ").join(frags) + ")") if frags else ""),
                 [],
```
khj809: An `or` branch that compiles to nothing is dropped from the join here; Linear collapses the whole `or` instead. Measured 2026-09-03: `{labels: {some: {or: [{}, {name: {eq: A}}]}}}` and `{labels: {every: {or: [{}, {name: {eq: A}}]}}}` each answered every issue, label-less ones included — this answers only the labelled ones, and `[]` for `every`. `and` does behave the way the code already does (`{some: {and: [{}, {name: {eq: A}}]}}` answered none), so only the `or` arm moves.

The suggestion also routes the sub-params through `p` instead of extending `params` up front: once an `or` can throw its fragment away, extending `params` before the fragment is known would orphan those placeholder values.

While here — the `_label_predicate` docstring just above still says an empty fragment is "the silent wrong answer this module exists to prevent", which now reads against `_labels_predicate` returning one on purpose. Worth saying which positions an empty fragment is safe in, since that is exactly the distinction this bug turns on.

```suggestion
            frags = [f for f, _ in subs if f]
            if key == "or" and sub == []:
                # Measured 2026-09-03: a literally empty `or` is a predicate every label satisfies,
                # and the quantifier still applies -- `some: {or: []}` and `every: {or: []}` both
                # answered exactly the issues with at least one label.
                frags, sub_params = ["1"], []
            elif key == "or" and len(frags) != len(sub):
                # Measured 2026-09-03: an `or` BRANCH that constrains nothing makes the WHOLE `or`
                # constrain nothing -- `some: {or: [{}, {name: {eq: A}}]}` and
                # `every: {or: [{}, {name: {eq: A}}]}` each answered every issue, label-less ones
                # included, where the surviving branch alone answers only the labelled ones. An
                # empty `and` branch is dropped instead: `some: {and: [{}, {name: {eq: A}}]}`
                # answered none.
                frags, sub_params = [], []
            else:
                sub_params = [x for _, sp in subs for x in sp]
            frag, p = (
                (("(" + (" AND " if key == "and" else " OR ").join(frags) + ")") if frags else ""),
                sub_params,
```

backlot/graphql/linear_filters.py
```diff
@@ -483,22 +490,116 @@ def _label_predicate(spec: dict) -> tuple[str, list]:
...
+            continue
+        if key in ("some", "every"):
+            add(*_labels_predicate(sub, every=key == "every"))
+        elif key in ("and", "or"):
+            subs = [_labels_filter(x) for x in sub]
+            frags = [f for f, _ in subs if f]
+            for _, p in subs:
+                params.extend(p)
+            if frags:
+                parts.append("(" + (" AND " if key == "and" else " OR ").join(frags) + ")")
```
khj809: Same defect one level up. Measured 2026-09-03: `{labels: {or: [{}, {length: {eq: 99}}]}}`, `{or: [{some: {}}, {length: {eq: 99}}]}` and `{or: [{null: false}, {length: {eq: 99}}]}` each answered every issue, where `length: {eq: 99}` alone answers none. `params` also only wants extending when the fragment is kept.

```suggestion
        elif key in ("and", "or"):
            subs = [_labels_filter(x) for x in sub]
            frags = [f for f, _ in subs if f]
            if key == "or" and len(frags) != len(sub):
                # Measured 2026-09-03: as in `_label_predicate`, an `or` branch that constrains
                # nothing makes the whole `or` constrain nothing -- `{or: [{}, {length: {eq: 99}}]}`
                # answered every issue, where `length: {eq: 99}` alone answered none.
                return "", []
            if frags:
                parts.append("(" + (" AND " if key == "and" else " OR ").join(frags) + ")")
                params.extend(x for _, sp in subs for x in sp)
```

backlot/graphql/linear_filters.py
```diff
@@ -483,22 +490,116 @@ def _label_predicate(spec: dict) -> tuple[str, list]:
...
+    pushes the ``not`` outside the quantifier -- and :func:`_reads_as_negation` says which form a
+    compound predicate takes. An empty predicate (``some: {}``, ``some: {and: []}``) constrains
+    nothing there: every issue answered, label-less ones included, so it compiles to nothing here
+    too. The one exception is a literally empty ``or`` (see :func:`_label_predicate`)."""
     inner, params = _label_predicate(spec)
     if not inner:
-        # An empty inner predicate must NOT silently vanish: dropping it turns a narrowing filter
-        # into a full-corpus answer. Nothing legitimate produces one, so it is a client error.
-        raise GraphQLError("a labels filter must constrain something (e.g. labels.some.name)")
+        return "", []
```
khj809: The other half of the `or` problem, one level above this: with `_labels_predicate` returning `""`, `_issue_filter`'s `or` drops the branch. Measured 2026-09-03, `{or: [{labels: {}}, {title: {eq: T}}]}` answered every issue, and so did the same query with `{labels: {some: {}}}`, `{labels: {every: {}}}`, `{labels: {null: false}}` and `{labels: {and: []}}` in that branch; this branch answers only the title match. Each of those was a 400 on `main`, so it is a refusal turned into a quietly narrower answer.

No suggestion, because the fix needs a state `("", [])` cannot carry. Linear drops a *literally empty* `or` branch — `{or: [{}, {title: {eq: T}}]}` answered only the title match — but treats a branch whose key constrains nothing as true. Distinguishing "no keys" from "keys that constrain nothing", and compiling the second to a true fragment rather than to `""`, separates the two.

backlot/graphql/linear_filters.py
```diff
@@ -483,22 +490,116 @@ def _label_predicate(spec: dict) -> tuple[str, list]:
...
-    return (f"EXISTS (SELECT 1 FROM json_each(COALESCE(labels, '[]')) WHERE {inner})", params)
+            no_label_fails if negative else f"({_LABEL_COUNT} > 0 AND {no_label_fails})"
+        ), params
+    a_label_matches = f"EXISTS (SELECT 1 FROM {_LABEL_EACH} WHERE {inner})"
+    return (f"({_LABEL_COUNT} = 0 OR {a_label_matches})" if negative else a_label_matches), params
+
+
+def _labels_filter(spec: dict) -> tuple[str, list]:
+    """One ``IssueLabelCollectionFilter``. Besides the two quantifiers, measured on 2026-09-03:
+    ``and`` / ``or`` compose collection filters; ``length`` is a ``NumberComparator`` over the
```
khj809: "`and` / `or` compose collection filters" holds only while no sibling key is present. Measured 2026-09-03, order-independent: `{or: [{length: {eq: 0}}], length: {eq: 99}}` and `{length: {eq: 99}, or: [{length: {eq: 0}}]}` both answered every issue — the `or` branch alone, where an AND answers none; `{and: [{}], length: {eq: 99}}` answered every issue; and `{or: [{length: {eq: 0}}], null: true}` answered every issue too, though `null: true` alone answers nothing. `and` and `or` do AND with each other (`{and: [{length: {eq: 99}}], or: [{length: {eq: 0}}]}` answered none).

So an `IssueLabelCollectionFilter` carrying `and` or `or` ignores `some`, `every`, `name`, `length` and `null` beside it. Both keys are new here — `and`/`or` were refused and `length`/`null` undeclared on `main` — so this is newly reachable rather than pre-existing. Left without a suggestion: it wants `and`/`or` read before the loop rather than inside it, and its own rows in `_LABEL_CELLS`.

backlot/graphql/linear_filters.py
nuri-yoo: Applied as suggested in a6f2da1, then re-measured with labelled issues present, and the labelled answer is not the label-less one. `some: {or: [{}, {name: {eq: "zz-zzz"}}]}` answered `[zz-a, zz-zzz]` plus the four label-less issues, and `every: {or: [{}, {name: {eq: "zz-zzz"}}]}` the label-less four alone. On a label-less workspace both read as "every issue", which is what the suggestion compiles to; with labels it is the surviving branch read as a NEGATIVE predicate, the answers `some` / `every: {name: {neq: "zz-a"}}` give on this data. Two neighbours separate the cases: `some: {or: [{or: []}, {name: {eq: "zz-zzz"}}]}` answered `[zz-a, zz-zzz]` alone (dropped, and positive), and `every: {or: [{name: {}}, {name: {eq: "zz-zzz"}}]}` every issue (that branch does make the whole predicate constrain nothing). `some: {and: [{}, {name: {eq: "zz-zzz"}}]}` answered `[zz-a, zz-zzz]`, so `and` drops the branch as the code already did.

3e7986a replaces the arm: `_label_predicate` returns a kind with the fragment (none / quant / vacuous / real) and `_label_branches` acts on it, and `_reads_as_negation` reads "every operator of none" as true, which is what makes `{}` and `{and: []}` negative and `{or: []}` positive. All of these are cells in `_LABEL_CELLS`; putting the applied arm back fails nine of them. The `_label_predicate` docstring is rewritten around which positions an empty fragment is right in, as you asked.

backlot/graphql/linear_filters.py
nuri-yoo: Applied in d94ac37 and it stands with labels present: `{or: [{}, {length: {eq: 2}}]}` and `{or: [{some: {}}, {length: {eq: 2}}]}` each answered every issue, `{and: [{}, {length: {eq: 2}}]}` the two-label issue. In 3e7986a it moved into `_labels_branches` because of the precedence in the thread below, and `params` is extended for kept branches only, as here.

backlot/graphql/linear_filters.py
nuri-yoo: Fixed in 3e7986a. `_issue_parts` returns `(fragment, params, vacuous)`, where vacuous is "a key of this object constrains nothing", and an `or` with a vacuous branch is vacuous itself. Re-measured with labels present: your five branches each answer every issue; `{or: [{}, T]}`, `{or: [{and: []}, T]}`, `{or: [{or: []}, T]}` and `{or: [{title: null}, T]}` answer the title match alone. Two more that shape the fix: a vacuous key collapses the `or` even beside a false sibling in the same branch, `{or: [{labels: {}, title: {eq: "no such title"}}, T]}` answered every issue, and inside an `and` a vacuous branch is dropped, `{or: [{and: [{labels: {}}, {title: {eq: "no such title"}}]}, T]}` answered the title match.

One more state than "no keys" versus "keys that constrain nothing" was needed. `{title: {eq: null}}` compiles to nothing on `main` too, and the flag would have read it as vacuous, but Linear answers it as a condition nothing passes: `{or: [{title: {eq: null}}, T]}` answered the title match alone, `dueDate: {eq: null}` none over issues with no due date, `estimate: {gte: null}` none. A string or list operator with null is the opposite, a condition every row passes: `{title: {contains: null}}` answered every issue, `labels: {some: {name: {contains: null}}}` the labelled ones only, `labels: {some: {name: {neqIgnoreCase: null}}}` every issue (negative, and every label passes it). `_Comparator.render` renders a null operand as `0` or `1` by operator instead of skipping it. 64 cells in `_ISSUE_CELLS`; deleting the two vacuous-`or` lines fails 22 of them, skipping null operands again fails 29 across both matrices.

backlot/graphql/linear_filters.py
nuri-yoo: Fixed in 3e7986a, and the rule is wider than `and` / `or`. With labelled issues present the collection filter's keys do not AND at all: `and` answers if present, else `or`, else the first present of `length`, `every`, `some`, `name`, whatever their order. `{and: [{length: {eq: 2}}], or: [{length: {eq: 1}}]}` answered `[zz-a, zz-zzz]` and `{and: [{length: {eq: 1}}], or: [{length: {eq: 2}}]}` `[zz-a]`, both orders, so `and` and `or` do not AND with each other either; on label-less issues that cell is empty under both readings, which is what `{and: [{length: {eq: 99}}], or: [{length: {eq: 0}}]}` measured. Down the chain: `{length: {eq: 1}, some: {name: {eq: "zz-zzz"}}}` answered `[zz-a]` (an AND is empty), `{some: {name: {eq: "zz-zzz"}}, every: {name: {eq: "zz-a"}}}` `[zz-a]`, `{name: {eq: "zz-zzz"}, some: {name: {eq: "zz-a"}}}` both labelled issues. `null` is the one key that ANDs, and only beside those four: `{or: [{length: {eq: 0}}], null: true}` answered the label-less issues (the `or` alone, not everything: `{or: [{length: {eq: 99}}], length: {eq: 0}}` answered none), `{null: true, some: {name: {eq: "zz-zzz"}}}` none, `{null: false, length: {eq: 1}}` `[zz-a]`.

The existing cell `{some: {name: {eq: "bug"}}, every: {name: {eq: "bug"}}}` → ENG-2 could not tell an AND from `every` winning; the new ones can. `_labels_filter` reads the keys in `_LABELS_PRECEDENCE` and its docstring carries the rule; removing the `break` fails 13 cells, reading `or` before `and` fails 9.

nuri-yoo:
Thanks — the label-less workspace hid two of these from me and one from the review. All four are addressed in `3e7986a`, each re-measured with the two labelled issues re-created (and deleted after; the workspace lists 4 issues and 3 labels, as before), and the measurement changed two of the answers:

| finding | what the labelled measurement says | where |
|---|---|---|
| 1. `_label_predicate` drops an empty `or` branch | Applied suggestion re-measured: `some: {or: [{}, {name: {eq: z}}]}` is not "every issue" but `[a, z]` plus the label-less issues, and `every` the label-less alone — the branch is dropped and the `or` reads as a negative predicate. `{or: []}` as a branch drops and reads positive; `{name: {}}` as a branch makes the predicate constrain nothing. | `_label_predicate` returns a kind, `_reads_as_negation` reads "every operator of none" as true; 9 cells fail with the applied arm put back |
| 2. `_labels_filter` drops an empty `or` branch | Stands with labels present. Moved into `_labels_branches`. | — |
| 3. `_issue_filter`'s `or` drops a vacuous branch | Fixed: a branch with a key that constrains nothing makes the whole `or` constrain nothing, its other keys included; a branch with nothing in it is dropped; inside an `and` a vacuous branch is dropped. | `_issue_parts` carries a `vacuous` flag; 64 cells in `_ISSUE_CELLS`, 22 fail without the two lines |
| 4. `and` / `or` ignore their siblings | Fixed, and wider: the keys do not AND at all — `and`, else `or`, else the first of `length`, `every`, `some`, `name`, order-independent; `null` ANDs only beside those four. `and` and `or` do not AND with each other either (`{and: [{length: {eq: 2}}], or: [{length: {eq: 1}}]}` answers `[a, z]`), which the label-less cell could not tell. | `_LABELS_PRECEDENCE`; 13 cells fail with the `break` removed, 9 with `or` read before `and` |

One thing the fix for 3 needed that the review did not predict: `{title: {eq: null}}` compiles to nothing on `main` too and would have read as a vacuous branch, but Linear answers a comparison with null as a condition nothing passes (`{or: [{title: {eq: null}}, T]}` answers the title match; `dueDate: {eq: null}` none over issues with no due date), and a string or list operator with null as one every row passes (`title: {contains: null}` every issue; `labels: {some: {name: {contains: null}}}` the labelled ones). `_Comparator.render` renders a null operand as `0` or `1` by operator; skipping it again fails 29 cells.

Nothing not taken. The description is rewritten: the `backlot diff` paragraph now says what this head answers (`775 · 775 acknowledged · 0 new`, #111 merged in), the second measurement table carries every cell above, and the cell counts are 157 collection-level and 64 whole-filter. Verification on the new head: `2079 passed`, no skips, on the full-extras venv with Docker, `npx` and `uvx` present; `ruff check` / `ruff format --check` clean; `gen_docs.py --check` exit 0. With `linear_filters.py` at `main`: `175 failed, 217 passed`; as it stood after the two applied suggestions: `90 failed, 302 passed`.

khj809 (commented):
Re-verified `0cc0113`: the new reading holds. I re-measured 38 cells against api.linear.app — the 19 of your claims decidable on a label-less workspace, 13 the code asserts but no comment cites (`in`/`nin`/`neq`/`startsWith` with null, `length: {eq: null}`, `or: [{or: [{}]}]`, `or: [{and: [{}]}]`, `estimate: {and: []}`, `length` vs `every`, `every` vs `name`, an empty `and` beside `or`, and the polarity of `neq: null` / `eq: null`), and the six that were divergent in my last pass — and the served answer matches on every one. `backlot diff --source linear` is `775 · 775 · 0`, lint is clean, and the suite is green.

You were right about the `or` and I was not: on a label-less workspace "the whole `or` is unconstrained" and "the surviving branch read negatively" are the same answer, so my measurement could not tell them apart and I suggested the wrong one. The labelled re-measurement is what decides it.

Three small things left, two of them documentation.

The measurement dates went missing from the prose this commit edited or added — `_labels_predicate`'s docstring lost its `2026-09-03`, and the new `_label_predicate` / `_labels_filter` docstrings, the `_NEGATIVE_OPS` comment and the null-operand comment carry none. The file is dated everywhere else (`_Comparator`, `_derived_in`, the module header), so the newest rules are now the only undated claims in it — 4 dated mentions where `03ebe29` had 6.

backlot/graphql/linear_filters.py
```diff
@@ -313,7 +313,18 @@ def render(self, spec: dict) -> tuple[str, list]:
         parts: list[str] = []
         params: list = []
         for op, raw in spec.items():
+            if raw is None and op in ("and", "or"):
+                continue
             if raw is None and op != "null":
```
khj809, line 330: One operator is outside the new rule: `null` itself. An explicit `{null: null}` skips both new guards and falls through to `elif op == "null"`, where `raw` is falsy, so it renders `IS NOT NULL` — exactly `null: false` inverted.

Measured 2026-09-03, and it is neither of the other two readings: `dueDate: {null: null}` answered every issue over four with no due date, so it is not the "nothing passes" that `eq: null` is; `priority: {null: null}` answered none over four that all carry a priority, and `{or: [{priority: {null: null}}, {title: {eq: T}}]}` the title match alone, so it is not a vacuous key either. It reads as `null: true`. On that same workspace Backlot's `IS NOT NULL` answers none where Linear answers all four; on the fixture `{dueDate: {null: null}}` gives `[ENG-1, ENG-2]` where the rule gives `[DES-1]`, and `estimate` / `completedAt` / `canceledAt` take the same path.

Normalising the operand keeps it in one place. Verified: `ruff` clean, `tests/test_linear.py` green, and the five `null` cells then answer what Linear answers.

```suggestion
            if raw is None and op in ("and", "or"):
                continue
            if raw is None and op == "null":
                # Measured 2026-09-03: an explicit null operand on `null` itself reads as
                # `null: true`, not as `false` and not as the condition-nothing-passes below.
                # `dueDate: {null: null}` answered every issue over four with no due date (so not
                # "nothing passes"), `priority: {null: null}` none over four that all carry a
                # priority, and `{or: [{priority: {null: null}}, {title: {eq: T}}]}` the title
                # match alone (so not a vacuous key either).
                raw = True
            if raw is None and op != "null":
```

backlot/graphql/linear.graphql
```diff
@@ -282,9 +282,19 @@ input IssueLabelFilter {
   or: [IssueLabelFilter!]
 }
 
+"""
+Linear's quantifiers over a collection are not the textbook ones: `every` with a positive predicate
+needs at least one label, and `some` with a negative one (`neq`, `nin`, …) also answers an issue with
+no labels -- see `linear_filters._labels_predicate` for the measured rule. `length` counts the
+labels; `null: true` matches nothing, since an issue's label collection is never null; a bare
+`name` reads as `some`.
+"""
```
khj809, line 294: This description was not touched by the last two commits, and it is now the thinnest part of the type's documentation: the most surprising thing about `IssueLabelCollectionFilter` is that its keys do not AND — `and` wins, else `or`, else the first of `length`, `every`, `some`, `name` — and that is not mentioned here. It also still points at `_labels_predicate`, where the precedence rule now lives in `_labels_filter`.

Worth remembering this text is served: introspection on Backlot returns it, while api.linear.app returns `"Issue label filtering options."` for the same type. `AttachmentFilter` sets the precedent for Backlot prose here, so this is not new — but a Python symbol (`linear_filters._labels_predicate`) going out on the wire is, and a reader of the served schema cannot follow it. A one-line pointer to the precedence, with the module reference dropped to a `#` comment above the type, would carry more and leak less.

backlot/graphql/linear_filters.py
nuri-yoo: Taken, and it is the comparator's rule only. Re-measured 2026-09-03: beside a sibling it still reads as `true` (`priority: {null: null, eq: 0}` answered none over four priority-0 issues, as `{null: true, eq: 0}` does and `{null: false, eq: 0}`, all four, does not). On a relation the same key is dropped: with one issue assigned beside three unassigned, `assignee: {null: null}` answered all four where `null: true` answered the three and `null: false` the one, `{null: null, name: {eq: ME}}` the assigned one alone, `{null: null, name: {eq: "nobody"}}` none. `_sub_filter` let `null: null` through to `IS NOT NULL`, so on the fixture `{assignee: {null: null}}` answered ENG-1 where Linear answers every issue; `project` and `creator` took the same path. The label collection drops it too (`labels: {null: null}` all four, `{null: null, length: {eq: 0}}` the label-less ones), which `_labels_filter` already did through `spec.get` returning `None`; it now drops the key with the other null-valued ones.

702bc6d: `_sub_filter` skips every null-valued key, and 19 cells pin the three readings (17 in `_ISSUE_CELLS`, 2 in `_LABEL_CELLS`). With the `_sub_filter` line put back 3 fail; without `raw = True` 8. Your comment stays and says which of the three readings it is (7a5edb2). One thing beside it that this PR does not take: `null: true` on a relation ignores its siblings on api.linear.app (`{assignee: {null: true, name: {eq: ME}}}` answered the three unassigned) where `_sub_filter` ANDs them, on `main` too. That is #124.

backlot/graphql/linear.graphql
nuri-yoo: Rewritten in 7a5edb2. The served text carries the precedence now (`and` if present, else `or`, else the first present of `length`, `every`, `some`, `name`; `null` ANDs only beside those four) and no Python symbol; the pointers to `_labels_filter` and `_labels_predicate` are a `#` comment above the type, which introspection does not serve, the same place this file already names `linear_filters.SCALARS`. `backlot diff --source linear` on the new head: `775 · 775 · 0`.

nuri-yoo:
Two commits on top of the applied suggestion. 702bc6d is the behaviour: `null: null` on a relation is dropped rather than read as `null: true` (the thread above has the re-measurement, with one issue assigned for it), plus 19 cells. 7a5edb2 is prose only. The dates are back: `_labels_predicate` has its `2026-09-03` again, and `_label_predicate`, `_labels_filter`, the `_NEGATIVE_OPS` comment and both null-operand comments say when they were measured, 16 dated mentions in the file against 6 at `03ebe29`. The served description of `IssueLabelCollectionFilter` names the key precedence and no Python symbol.

Verification on 7a5edb2: `2104 passed`, no skips, on the full-extras venv with Docker, `npx` and `uvx` present; `ruff check` / `ruff format --check` clean; `gen_docs.py --check` exit 0; `backlot diff --source linear` `775 · 775 · 0`. With `linear_filters.py` at `main`: `191 failed, 220 passed`; at `794a807`, the applied suggestion before this round: `3 failed, 408 passed`. The description is updated to match. The relation `null: true` divergence found on the way is #124, not this PR.

khj809:
One more, and not from this branch: `backlot diff --source linear` has picked up a live drift since you last ran it. On `6bcd0c0` it now answers `776 divergence(s) · 775 acknowledged · 1 new`, the one being `missing_enum_value IntegrationService.origin` — the vendor's enum carries a member the served SDL does not.

It is not yours. `origin/main` answers `779 · 778 · 1` with that same single finding, and the difference of 3 is exactly the gaps this PR closes. Linear added the member some time after the baseline was measured this morning, which is the case `backlot diff` exists for.

Worth knowing `pytest` stays green through it: `LINEAR_ENUMS` pins the member list against Backlot's own SDL rather than against the live API, so the fidelity job is the only thing that turns red — and it turns red on `main` too.

So two things. The `backlot diff` line in the description reads `775 · 775 · 0`, which no longer reproduces. And the member itself looks like main's to take rather than this PR's, unless you would rather fold it in here. Could you re-run it and see whether you get the same?

nuri-yoo:
Reproduced. On `6bcd0c0`: `776 divergence(s) · 775 acknowledged · 1 new`, the one `missing_enum_value IntegrationService.origin`. On `origin/main` at `ee8be69`: `779 · 778 · 1`, the same finding. Introspecting the enum directly gives 45 members against the 44 the served SDL declares, `origin` the only difference, no description, not deprecated. The scheduled run at 11:20Z yesterday was green for linear and my `775 · 775 · 0` was at 14:02Z, so the member landed after that.

Agreed it is main's. The fix is the one member on the enum in `linear.graphql` plus the same on `LINEAR_ENUMS`, and no baseline change, since the finding is new rather than acknowledged. I will open that as its own PR against main rather than fold it in here, and update the `backlot diff` line in this description to the `6bcd0c0` result with the drift named.

khj809 (approved):

PR #118, opened 2026-09-03
s3: a sub-resource Backlot does not implement is refused, not answered with the listing

### LINKED ISSUE #103: Unrecognised S3 sub-resources are answered with an object listing instead of being refused

nuri-yoo:
Closes #103.

**What changed.** `bucket_get` and `object_get` in `backlot/routers/s3.py` recognised `?location` and otherwise fell through to the listing or to the object's bytes. They now read the sub-resource selectors botocore's service model declares at each path and refuse the ones Backlot does not implement (26 at a bucket's path, 8 at an object's) with `NotImplemented`, HTTP 501, in the `<Error>` envelope. The listing, `?list-type=2` and `?location` are untouched. The S3 baseline acknowledges the 41 refusals as `missing_operation` gaps, each with what real S3 answers written beside it.

**What real S3 does, measured.** Every claim below was measured on 2026-09-03 in two regions: against an existing general purpose bucket in ap-northeast-2, and against a throwaway one created in us-east-1 (the region Backlot reports) for this purpose and deleted afterwards. Nothing was configured on either, requests were signed with botocore, one per selector. Every answer was the same in both regions, down to status, root element and error code, with one exception noted below.

- Every one of the 42 operations is answered; none is refused with NotImplemented. 21 answer 200 with a root of their own (`?versioning` is an empty `<VersioningConfiguration/>`, `?acl` an `<AccessControlPolicy>`, `?uploads` a `<ListMultipartUploadsResult>`), 15 answer a 404 of their own (`NoSuchTagSet`, `NoSuchBucketPolicy`, `NoSuchCORSConfiguration`, `NoSuchLifecycleConfiguration`, `NoSuchUpload`, and `NoSuchConfiguration` for a Get*Configuration whose `id` is not configured), 3 answer 400 `InvalidRequest` (`?attributes` without its header, `?legal-hold` and `?retention` on a bucket without Object Lock), 2 answer 405 (`?metadataTable` is `V1APIsNotAllowed`; `?torrent` is the one region-dependent answer, `MethodNotAllowed` in us-east-1 and `RequestTorrentNotSupportedInRegion` in ap-northeast-2, both 405), and 1, `?session`, answers the listing.
- An unknown query key is ignored: `?foo=bar` and `?x-id=ListObjects` answer the listing (the AWS SDK for JavaScript adds an `x-id` naming the operation wherever several share a path, `?x-id=GetObject` and `?analytics&x-id=GetBucketAnalyticsConfiguration` among them, so it has to stay ignorable), `?Versioning` answers the listing, and `?versioning=1` is still GetBucketVersioning. So the refusal is keyed on the set of selectors, not on an allow-list of the listing's own parameters, which would refuse what real S3 ignores.
- Two selectors at once are 400 `InvalidArgument`, `Conflicting query string parameters: acl, versioning`, named alphabetically whatever order they were sent in, with `<ArgumentName>ResourceType</ArgumentName>` and the first as `<ArgumentValue>`. Backlot answers the same, `location` included.
- A conflict is reported before existence, and existence before the sub-resource: `?acl&versioning` on a bucket that does not exist is `InvalidArgument`, a missing bucket with `?versioning` is `NoSuchBucket`, a missing key with `?acl` is `NoSuchKey`, but a missing key with `?uploadId` is `NoSuchUpload`. Backlot keeps that order: the conflict is checked first, ListParts is refused before the key is looked up, and the other object sub-resources after it.
- HEAD with any selector is a 405 with an empty `application/xml` body, on buckets and on objects, for a bucket or a key that exists and for one that does not, so the refusal comes before either is looked up; a HEAD carrying two selectors is the conflict's 400 with the same empty body. Backlot's HEAD routes answer the same, through one shared `_head_refusal`. (Real S3 also sends an `Allow` header listing the methods the sub-resource does take, `GET` or `GET, PUT`; Backlot takes none of them, so it sends none.)

**Why NotImplemented.** The issue asked for the status and body to be measured before settling on them. Real S3 never refuses these operations, so there is no measured refusal to copy. The S3 API reference's Error code table documents `NotImplemented` as "A header you provided implies functionality that is not implemented", HTTP 501, which is the one code that says the service is not performing the operation asked for, and 501 is not a status botocore retries. Each refusal is therefore a divergence from real S3, recorded in the baseline as a gap with the measured real answer in its note, so implementing any of them later starts from the shape real S3 actually gives.

**What a caller saw, and sees.** The issue's reproduction with boto3 against a stock Backlot: `get_bucket_versioning` returned `{}` and `get_bucket_policy` returned the XML listing as the policy string. Object sub-resources were worse than the issue recorded: `get_object_tagging` and `list_parts` surfaced as a 500 `Internal Server Error` after four retries. The server answered 200 with the object's bytes, and botocore's `_handle_200_error` handler, which inspects 200 bodies for the in-body errors S3 sends on some operations, could not parse them as XML and rewrote the status to 500, which the retry handler then retried. Now every one of them raises `ClientError` with code `NotImplemented`, HTTP 501 and `RetryAttempts` 0, while `get_bucket_location`, `list_objects`, `list_objects_v2` and `get_object` are unchanged.

**`?session` is deliberately not refused.** CreateSession exists for directory buckets only, and real S3 answers `GET /{bucket}?session` on a general purpose bucket with the listing; `?session&versioning` is answered as GetBucketVersioning with no conflict, so `session` is not a selector there at all. The listing is the faithful answer. The probe still classifies it as a fallthrough, so it is acknowledged in the baseline by hand, next to ListObjectsV2, with that measurement as its note.

**Docs.** The S3 section of `docs/supported-sources.md` now says what happens to a sub-resource outside its table.

**Tests** (`tests/test_s3.py`): each of the 26 bucket and 8 object selectors is refused with 501 and neither the listing nor the object's text is in the body; the bare listing, `?list-type=2`, `?location`, `?foo=bar`, `?x-id=`, `?Versioning` and `?session` still answer as before; two selectors conflict as real S3 conflicts them, before existence is checked; existence precedes the sub-resource except for ListParts; HEAD with a selector is 405 with an `application/xml` content type before the bucket or the key is looked up, HEAD with two selectors is an empty 400, and a bare HEAD still answers; and the boto3 reproduction raises once per call with no retry. With main's router in place, 39 of the module's 66 tests fail.

**Verification** (on f2a9b1f, main merged, #114 and #115 included). `--update-baseline` rewrites `s3.json` byte-identical; `ruff check .` and `ruff format --check .` clean; `scripts/gen_docs.py --check` exit 0; `pytest -rs`: 1904 passed, 0 skipped; `backlot diff --source s3`: 43 divergences, 43 acknowledged, 0 new, exit 0.


backlot/routers/s3.py
```diff
@@ -188,37 +260,37 @@ async def list_buckets(request: Request):
...
     caller, visible, err = _auth(request)
     if err:
         return Response(status_code=err.status_code)
     conn = auth.conn(request)
     if not _bucket_visible(conn, bucket, visible):
         return Response(status_code=404)
+    if _selected(request.query_params, _BUCKET_SELECTORS):
+        # No sub-resource has a HEAD form: real S3 answers `HEAD /{bucket}?versioning` 405 with an
+        # empty application/xml body (measured).
+        return Response(status_code=405, media_type="application/xml")
```
khj809: `HEAD /{bucket}?versioning` on a bucket that does not exist answers 404 here, while the object route answers 405 before the key is looked up. The description claims 405 "on buckets and on objects, for a key that exists and for one that does not", so one of the two is wrong, and no test covers the missing-bucket case.

If the measurement holds for a missing bucket:

```suggestion
    if _selected(request.query_params, _BUCKET_SELECTORS):
        # No sub-resource has a HEAD form: real S3 answers `HEAD /{bucket}?versioning` 405 with an
        # empty application/xml body, whether or not the bucket exists (measured).
        return Response(status_code=405, media_type="application/xml")
    if not _bucket_visible(conn, bucket, visible):
        return Response(status_code=404)
```

If it does not, narrow the sentence in the description instead, and either way add the missing-bucket case next to the missing-key one in `test_head_with_a_subresource_is_405_and_a_bare_head_still_answers`.

backlot/routers/s3.py
```diff
@@ -345,20 +417,31 @@ def _list_objects_v2(request: Request, conn, bucket: str, visible) -> Response:
...
-    object the caller cannot read is NoSuchKey, not AccessDenied, so a listing and a read agree
-    about what exists."""
     caller, visible, err = _auth(request)
     if err:
         return err if request.method == "GET" else Response(status_code=err.status_code)
     conn = auth.conn(request)
+    selected = _selected(request.query_params, _OBJECT_SELECTORS)
+    if selected and request.method == "HEAD":
+        # No sub-resource has a HEAD form: real S3 answers `HEAD /{key}?acl` 405 with an empty
+        # application/xml body, whether or not the key exists (measured).
```
khj809: Real S3 sends an `Allow` with these 405s and RFC 9110 requires one; Backlot sends none. The probe only issues GET, so `backlot diff` can never see this divergence and the reason for it lives only in the PR body. Worth keeping where the next reader will find it:

```suggestion
        # No sub-resource has a HEAD form: real S3 answers `HEAD /{key}?acl` 405 with an empty
        # application/xml body, whether or not the key exists (measured). Real S3 also sends an
        # `Allow` naming the methods that sub-resource takes; Backlot takes none of them, so it
        # sends none.
```

Same for the comment in `head_bucket`.

backlot/routers/s3.py
```diff
@@ -98,6 +128,52 @@ def _error(
...
+    reference's Error code table), and 501 is not a status botocore retries, so a boto3 caller gets one ``ClientError`` straight away.
+    What the catch-all used to answer was worse in both directions: a bucket sub-resource got the
+    listing, which botocore parsed as an empty result (``get_bucket_versioning`` -> ``{}``), and an
+    object sub-resource got the object's bytes, which botocore could not parse as XML and reported
+    as a 500 after retrying.
+
+    Real S3 answers each of these operations rather than refusing it — measured against a general
+    purpose bucket, ``?versioning`` is an empty ``<VersioningConfiguration>`` and
+    ``?tagging`` is ``NoSuchTagSet`` — so every refusal here is a gap the S3 baseline records, with
+    the answer real S3 gives written beside it.
```
khj809: This PR's own baseline records an object's `?tagging` as `200 <Tagging>` with an empty TagSet — only a bucket's is `NoSuchTagSet` — and the docstring covers both paths.

```suggestion
    Real S3 answers each of these operations rather than refusing it — measured against a general
    purpose bucket, ``?versioning`` is an empty ``<VersioningConfiguration>``, a bucket's
    ``?tagging`` is ``NoSuchTagSet`` and an object's an empty ``<Tagging>`` — so every refusal here
    is a gap the S3 baseline records, with the answer real S3 gives written beside it.
```

docs/supported-sources.md
```diff
@@ -259,6 +259,12 @@ virtual-hosted client looks for `acme-artifacts.localhost:8000` and finds nothin
 | `GetObject` | `Range` |
 | `HeadObject` | |
 
+Any other sub-resource — `?versioning`, `?acl`, `?tagging`, `?uploadId` and the rest of what botocore
+declares at a bucket's or an object's path — is refused with `NotImplemented` (501), so a client gets
+an error to handle rather than the listing or the object's bytes parsed as something else. Two
+sub-resources at once are `InvalidArgument`, as on real S3, and an unknown query key is ignored, as
+on real S3.
```
khj809: `?session` is not refused, so "the rest of what botocore declares" overstates it here. The router comment and the baseline both say why; this is the only place a client author reads that does not.

```suggestion
Any other sub-resource — `?versioning`, `?acl`, `?tagging`, `?uploadId` and the rest of what botocore
declares at a bucket's or an object's path — is refused with `NotImplemented` (501), so a client gets
an error to handle rather than the listing or the object's bytes parsed as something else. The one
exception is `?session`: CreateSession is for directory buckets only and real S3 answers it with
the listing on a general purpose bucket, so Backlot does too. Two sub-resources at once are
`InvalidArgument`, as on real S3, and an unknown query key is ignored, as on real S3.
```

backlot/routers/s3.py
nuri-yoo: Measured now: `HEAD /{missing-bucket}?versioning`, `?location` and `?acl`, in us-east-1 and ap-northeast-2, are all 405 with the empty `application/xml` body, the same as on a bucket that exists, so the description's claim holds and the route was wrong. Applied as suggested, with the selector check hoisted above the bucket lookup, and the missing-bucket case added next to the missing-key one in the test.

The same measurement showed one more thing the HEAD branches got wrong: a HEAD carrying two selectors (`?versioning&acl`, `?acl&tagging`) is the conflict's 400 with an empty body on real S3, not a 405, whether or not the bucket or the key exists. Both routes now answer that too, through one `_head_refusal` helper they share, and `test_head_with_two_subresources_is_the_conflicts_400_with_an_empty_body` covers bucket, object, missing bucket and missing key. Against the previous router both HEAD tests fail (`2 failed, 64 deselected`); on the new one `tests/test_s3.py` is 66 passed, the full suite 1898 passed with 0 skipped, and `backlot diff --source s3` is still 43 acknowledged, 0 new (the probe issues no HEAD, so the baseline does not move).

backlot/routers/s3.py
nuri-yoo: Applied. The sentence sits in the docstring of the `_head_refusal` helper both HEAD branches now call, so it is in one place rather than two. The measured `Allow` values were `GET` for `?location` and `GET, PUT` for `?versioning` and `?acl`, which is what "the methods that sub-resource takes" describes.

backlot/routers/s3.py
nuri-yoo: Applied as suggested; the baseline's GetObjectTagging entry is the source, and the docstring now says which path answers which.

docs/supported-sources.md
nuri-yoo: Applied as suggested.

khj809 (approved):
All four addressed, and verified here on 007477b: `HEAD /{missing-bucket}?versioning` is now 405, `_head_refusal` keeps the `Allow` decision in one place, the `?tagging` attribution names which path answers which, and the docs carry the `?session` exception. The HEAD-carrying-two-selectors 400 you found while re-measuring is a good catch beyond what was asked.

Reproduced: `tests/test_s3.py` 66 passed, 39 of them failing against main's router, full suite green, `ruff` clean, and `backlot diff --source s3` 43 acknowledged / 0 new.

One thing left, in the description only: the Verification line still reads `pytest: 1729 passed, 0 skipped` and still mentions merging #115, which has since landed. Please refresh that line — no code change needed.

nuri-yoo:
Verification line refreshed, no code change. It now reads against f2a9b1f (main merged, #114 and #115 in): `ruff` clean, `scripts/gen_docs.py --check` exit 0, `pytest -rs` 1904 passed / 0 skipped, `backlot diff --source s3` 43 acknowledged / 0 new, and `--update-baseline` still byte-identical. The #115 merge sentence is gone since it landed.

PR #121, opened 2026-09-03
github: the branch and tag listings page, and a page url carries the filter the caller sent

### LINKED ISSUE #117: github: /branches and /tags ignore per_page and send no Link
**Source and commit** — `github`, at `2edd700` (on the branch for #96).

`GET /repos/{owner}/{repo}/branches` and `GET /repos/{owner}/{repo}/tags` ignore `per_page` and send no `Link` header. They are the only two list routes on this router that do not go through `clamp_page` and the link header the rest share.

|---|---|---|
| `/branches?per_page=2` on a repo with more | two items, `Link: …; rel="next", …; rel="last"` | every branch, no `Link` |
| `/tags?per_page=2` on a repo with more | two items, with the same `Link` | every tag, no `Link` |
| `/pulls?per_page=1` (same server) | one item with a `Link` | one item with a `Link` |

**How you know** — measured against api.github.com on 2026-09-03, `?per_page=2` on psf/requests answers two items on both routes with `rel="next"` and `rel="last"`.

**Reproduction** — the bundled corpus, repo `pipeline`, which states three branches and two tags:

```python
import backlot, requests

with backlot.serve() as m:
    for route in ("branches", "tags"):
        r = requests.get(f"{api}/{route}", headers=h, params={"per_page": 1})
        print(route, len(r.json()), r.headers.get("Link"))   # 3 None / 2 None
```

**What it costs a client** — one that pages a listing the way it pages every other one reads the first page and stops, and never learns there was no second page to ask for. It was invisible while the branch listing was a fixed one-element list; #96 made both listings as long as the corpus says, so a repo with more branches than a page now silently ignores the page size it was given.

**Where it lands** — `list_branches` and `list_tags` in `backlot/routers/github.py`. `list_pulls` in the same module is the shape to follow: `clamp_page` against `default_page_size`/`max_page_size`, then `_paged`.

**What a fix has to settle**

- Whether `?protected=` filters before or after the page is cut. Real filters first — `?protected=false&per_page=1` on fastapi/fastapi answers one of the 21 unprotected branches, not "page 1 of 22, minus the protected ones" — so the filter belongs ahead of the slice.
- Whether the `Link` carries `?protected=` back. `_paged` takes the query it should echo (`list_pulls` passes `{"state": state}`), so this is a decision about which parameters a next-page url must preserve, not new machinery.
- Ordering is already settled and must not move: ascending by name, which is what `_branch_rows` returns and what real does.

khj809:
Closes #117.

## What changed

`/repos/{owner}/{repo}/branches` and `/repos/{owner}/{repo}/tags` page like the rest of the GitHub router: `clamp_page` against `default_page_size`/`max_page_size`, then `_paged`, which is the path `list_pulls` already takes. Both ignored `per_page` outright, so a client that pages a listing by following `rel="next"` read the first page of these two and never learned there was no second page to ask for. It stayed invisible while the branch listing was a fixed one-element list; #96 made both listings as long as the corpus says.

On `/branches`, `?protected=` selects ahead of the page cut and the `Link` counts the pages of the selection, not of the whole listing.

Which parameters a page url carries back is now one rule in one place, `_echo`: the ones the caller actually sent. That fixed a second divergence in the two listings the issue named as the shape to follow — `state` defaults to `open`, and `list_issues`/`list_pulls` echoed that default, so `/issues?per_page=1` linked a `next` saying `state=open` where real links `per_page` and `page` alone. A client that asked for every issue and walked the links lost the closed ones from page 2 on.

Ordering does not move: ascending by name for branches, the order the repo record stated for tags, which is the order `rel="next"` walks.

The github fidelity baseline drops the four acknowledgements this closes — `branches?page`, `branches?per_page`, `tags?page` and `tags?per_page`, each recorded as "the vendor accepts it; Backlot does not". `backlot diff --source github` reports them as no longer diverging; pytest never would.

Measured against api.github.com on 2026-09-03:

- `psf/requests/branches?per_page=2` and `…/tags?per_page=2` each answer two items with `rel="next"` and `rel="last"`, and `last` counts the whole listing — page 3 for the branches, page 81 for the 161 tags.
- `fastapi/fastapi/branches?protected=false&per_page=1` answers one of the 21 unprotected branches and reports `rel="last"` page 21. Twenty-one is the size of the filtered listing, so the filter runs before the slice; 22 pages minus the protected one would have been the other reading.
- The same call spells `protected=false` into both urls. An empty value is echoed identically, `protected=&per_page=1&page=2` with `rel="last"` page 22, so what decides is whether the parameter arrived rather than whether it narrows anything.
- `?protected=banana` answers its one protected branch with no `Link` at all, which is `github_link_header`'s existing answer for a single page.
- `psf/requests/pulls?per_page=1` links `per_page` and `page` only; `?state=open&per_page=1` links `state=open` beside them. Same split on `/issues`, with `state=all` echoed when sent. A parameter the handler defaulted is not one the caller sent.

Each measurement is stated once, in the router beside the rule it decided — `_echo` for which filters a page url carries, `list_branches` and `list_tags` for their own paging. The tests assert the behaviour and point at those rather than restating them.

## Verification

- **Tests** — `pytest` on a `.[all]` install (`uv sync --all-extras`, plus `uv pip install --no-deps "llama-index-readers-hubspot<0.6"`): 1908 passed, 0 skipped. Before the hubspot reader went in, `-rs` reported that one skip and nothing else.
- **Optional surfaces that ran rather than skipped** — the `fsspec` GitHub filesystem tests in `tests/test_integrations.py` (7 passed), which are the real client of these two routes: `GithubFileSystem.branches`/`.tags`/`.refs` are these listings and nothing else, and `test_github_filesystem_enumerates_the_repos_refs` still reads all three branches and both tags of `pipeline` because that client sends no `per_page`. Also all of `tests/test_sdk.py`, `tests/test_mcp.py`, `tests/test_llamaindex.py`, and the mirage shim, Google and reader tests in `tests/test_integrations.py`.
- **Lint** — `ruff check . && ruff format --check .`: all checks passed, 138 files already formatted.

- [x] A test fails without this change — both parametrizations of `test_github_ref_listings_page_like_every_other_listing` fail on the first assertion with the whole listing where one item was asked for (`assert ['v1.0', 'v1.1'] == ['v1.0']` for tags); the `?protected=` echo assertions folded into `test_github_stated_protection_gives_protected_all_three_answers` fail on the absent header; and both parametrizations of `test_github_a_page_url_echoes_only_the_filters_the_caller_sent` fail on `state=open` appearing in a url the caller never sent it to — one per listing that defaults `state`, since reverting either handler alone leaves the other's assertion green.
- [ ] A new endpoint serving corpus content is ACL-scoped, proved for both an admin and a scoped token — no new endpoint. Every route touched already scopes through `auth.visible_ids`, and paging a list does not reach that.

## Not in this PR

Eight more list routes on this router take no `page`/`per_page` and send no `Link`: `/issues/{n}/comments`, `/pulls/{n}/reviews`, `/pulls/{n}/comments`, `/pulls/{n}/commits`, `/statuses/{sha}`, `/collaborators`, `/orgs/{org}/teams` and `/repos/{owner}/{repo}/teams`. Seven of them carry the same divergence this PR closes for branches and tags, including the ones Backlot can only ever answer with a single item: real answers `[]` past the last page of a one-item listing (`psf/requests/pulls/7616/commits?per_page=1&page=99`, measured), where a route that ignores `page` serves that one item on every page a client asks for — a walk with no end. Only `/statuses/{sha}` cannot show it, being `[]` on every page either way. That is its own issue, filed as #126, which also carries two divergences in `github_link_header` that every listing inherits: past the last page real points `rel="prev"` at the last page holding rows and still sends `rel="last"`, and a page url omits `per_page` for a caller who sent none.

`backlot diff --source github` also reports one gap this branch did not introduce and does not close: `GET /repos/{owner}/{repo}/stargazers/history`, an operation that has appeared in GitHub's published OpenAPI since the baseline was measured on 2026-09-02. Backlot serves no stargazers route at all, so the gap reads the same on `main`; acknowledging someone else's surface here would bury it.

### DIFF
--- a/backlot/routers/github.py
+++ b/backlot/routers/github.py
@@ -255,17 +255,13 @@ def _paged(
 
 
 def _echo(request: Request, **params) -> dict:
-    """The filters a next-page url must spell out: the ones the CALLER sent, in `_paged`'s shape.
+    """The filters a next-page url spells out: the ones the CALLER sent, in `_paged`'s shape.
 
-    Real GitHub carries a listing's filter into its `Link` only when the request carried it.
-    Measured on psf/requests: `/pulls?per_page=1` links `per_page` and `page` alone where
-    `?state=open&per_page=1` links `state=open` beside them, the same on `/issues`, and
-    `/branches?protected=` on fastapi/fastapi echoes even an empty value. So what decides is
-    whether the parameter arrived, not whether it narrows anything.
-
-    A default the handler applied is not something the caller asked for: `state` defaults to
-    `open`, and echoing that gave a client paging every issue a `next` url saying `state=open`,
-    which drops the closed ones from page 2 on — a listing that changes shape as it is walked.
+    Real echoes a listing's filter only when the request carried it — `/pulls?per_page=1` links
+    `per_page` and `page` alone where `?state=open&per_page=1` links `state=open` too, and an empty
+    `?protected=` is echoed as well (measured on psf/requests and fastapi/fastapi). A default the
+    handler applied is not one the caller asked for: echoing `state`'s `open` narrows the url a
+    paginator follows, dropping the rows a `state=all` walk asked for.
     """
     return {k: v for k, v in params.items() if k in request.query_params}
 
@@ -1376,10 +1372,9 @@ async def list_branches(
     are protected. For one that does not, every branch is unprotected and real's last two coincide
     — a pull cannot imply protection, so an inferred listing has none (see :func:`_branch_rows`).
 
-    `?protected=` selects AHEAD of the page cut, so a page holds `per_page` of the branches that
-    survived it and the `Link` counts the pages of the selection: measured on fastapi/fastapi,
-    `?protected=false&per_page=1` answers one of the 21 unprotected branches and reports
-    `rel="last"` page 21, not the 22 pages of the whole listing.
+    `?protected=` selects AHEAD of the page cut, so the `Link` counts the pages of the selection:
+    `?protected=false&per_page=1` on fastapi/fastapi answers one of the 21 unprotected branches and
+    reports `rel="last"` page 21, not the 22 of the whole listing.
     """
     conn = auth.conn(request)
     caller = _require(request)
@@ -1422,9 +1417,8 @@ async def list_tags(
     — Backlot keeps no commit history (see :func:`get_tree`), so a tag cannot point at a different
     one.
 
-    Paged like every other listing on this router: `?per_page=2` on psf/requests answers two tags
-    with `rel="next"` and a `rel="last"` counting all 161 (measured). The order is the one the repo
-    record stated, which is the order `rel="next"` walks.
+    Paged like the rest of the router, and real pages it too: `?per_page=2` there answers two tags
+    with a `rel="last"` counting all 161.
     """
     conn = auth.conn(request)
     caller = _require(request)
--- a/backlot/routers/github.py
+++ b/backlot/routers/github.py
@@ -255,13 +255,16 @@ def _paged(
 
 
 def _echo(request: Request, **params) -> dict:
-    """The filters a next-page url spells out: the ones the CALLER sent, in `_paged`'s shape.
+    """The FILTERS a next-page url spells out: the ones the caller sent, in `_paged`'s shape.
 
     Real echoes a listing's filter only when the request carried it — `/pulls?per_page=1` links
     `per_page` and `page` alone where `?state=open&per_page=1` links `state=open` too, and an empty
     `?protected=` is echoed as well (measured on psf/requests and fastapi/fastapi). A default the
     handler applied is not one the caller asked for: echoing `state`'s `open` narrows the url a
     paginator follows, dropping the rows a `state=all` walk asked for.
+
+    Filters only. `per_page` is `github_link_header`'s to write, and it writes the effective size
+    whether or not the caller named one, where real omits it for a caller who did not (#126).
     """
     return {k: v for k, v in params.items() if k in request.query_params}
 
--- a/backlot/routers/github.py
+++ b/backlot/routers/github.py
@@ -1331,7 +1331,12 @@ async def get_blob(owner: str, repo: str, sha: str, request: Request):
 
 @router.get("/repos/{owner}/{repo}/branches")
 async def list_branches(
-    owner: str, repo: str, request: Request, protected: str | None = Query(None)
+    owner: str,
+    repo: str,
+    request: Request,
+    protected: str | None = Query(None),
+    page: PageParam = None,
+    per_page: PageParam = None,
 ):
     """The repo's branches: the default branch, plus the refs its pulls advertise.
 
@@ -1354,6 +1359,11 @@ async def list_branches(
     All three answers are distinct for a repo whose `subtype: "repo"` record states which branches
     are protected. For one that does not, every branch is unprotected and real's last two coincide
     — a pull cannot imply protection, so an inferred listing has none (see :func:`_branch_rows`).
+
+    `?protected=` selects AHEAD of the page cut, so a page holds `per_page` of the branches that
+    survived it and the `Link` counts the pages of the selection: measured on fastapi/fastapi,
+    `?protected=false&per_page=1` answers one of the 21 unprotected branches and reports
+    `rel="last"` page 21, not the 22 pages of the whole listing.
     """
     conn = auth.conn(request)
     caller = _require(request)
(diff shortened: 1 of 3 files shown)


tests/test_github.py
```diff
@@ -866,6 +869,59 @@ def names(params):
...
+    nxt = _link_rels(first.headers["Link"])["next"]
+    second = c.get(nxt.split("testserver", 1)[1], headers=gh_admin_h)
+    assert [x["name"] for x in second.json()] == names[1:]
+    assert {"prev", "first"} <= set(_link_rels(second.headers["Link"]))
+
+    # one page carries no Link at all, as real sends none
+    assert "Link" not in c.get(url, headers=gh_admin_h).headers
+
+
+def test_github_a_page_url_echoes_only_the_filters_the_caller_sent(gh_client, gh_admin_h, gh_org):
```
nuri-yoo: Non-blocking. This guards the `/pulls` half of the `_echo` change only. With `list_issues` put back to `{"state": state}` (`backlot/routers/github.py:755`) and everything else at this head, `tests/test_github.py` is still 123 passed, so the `/issues` line the body describes (`/issues?per_page=1` linking `state=open`) can regress without a test going red. `/issues` on `diffable` lists the same pulls as rows, so a `@pytest.mark.parametrize("route", ["pulls", "issues"])` over the url here should cover both without a new fixture.

backlot/routers/github.py
```diff
@@ -254,6 +254,18 @@ def _paged(
...
+def _echo(request: Request, **params) -> dict:
+    """The filters a next-page url spells out: the ones the CALLER sent, in `_paged`'s shape.
+
+    Real echoes a listing's filter only when the request carried it — `/pulls?per_page=1` links
+    `per_page` and `page` alone where `?state=open&per_page=1` links `state=open` too, and an empty
+    `?protected=` is echoed as well (measured on psf/requests and fastapi/fastapi). A default the
+    handler applied is not one the caller asked for: echoing `state`'s `open` narrows the url a
+    paginator follows, dropping the rows a `state=all` walk asked for.
+    """
+    return {k: v for k, v in params.items() if k in request.query_params}
```
nuri-yoo, line 269: Non-blocking, prose only. "The ones the CALLER sent" is true of the filters, and `per_page` is the one parameter this rule does not reach: `github_link_header` puts `per_page` into every url whether or not the caller sent it, where real omits it when absent. `psf/requests/tags` with no query links `?page=2` and `?page=6` (measured 2026-09-04); Backlot links `per_page=100&page=2`. Harmless to a follower, so nothing to change in this PR, but the docstring (and the body's "one rule in one place") read a little wider than the code. Might be worth a parenthetical here, or a note wherever `github_link_header` gets its next visit.

nuri-yoo (approved):
Approving. Every claim in the body I could reach I re-ran from this head (`3704070`).

**Measured against api.github.com on 2026-09-04**, all as the body states: `psf/requests/branches?per_page=2` answers two with `rel="next"` and `rel="last"` page 3; `tags?per_page=2` has `last` page 81; `fastapi/fastapi/branches?protected=false&per_page=1` has `last` page 21; `?protected=&per_page=1` echoes `protected=` and has `last` page 22; `?protected=banana` answers one branch with no `Link`; `psf/requests/pulls?per_page=1` links `per_page` and `page` only while `?state=open&per_page=1` adds `state=open`; `/issues?state=all&per_page=1` echoes `state=all`.

**Run at this head** (a `.[all]` venv on 3.14): full suite 1907 passed, 0 skipped; `tests/test_github.py` 123 passed; `ruff check` and `ruff format --check` on the two touched files clean; `gen_docs.py --check` exit 0; `backlot diff --source github` reports exactly one new gap, `GET /repos/{}/{}/stargazers/history`, and the baseline lost exactly the four `branches?page`, `branches?per_page`, `tags?page`, `tags?per_page` entries (1254 to 1250, nothing added). The issue's reproduction on `pipeline` now prints one branch with `rel="next"` page 2 and `rel="last"` page 3, and one tag with `rel="last"` page 2. The eight routes named under Not in this PR are exactly the router's `@router.get` handlers that take no `page` parameter and answer a list (walked the AST; nothing else). fsspec 2026.7.0's `GithubFileSystem.tags` and `.branches` send no `per_page`, as stated.

One non-blocking inline on the test coverage of the `/issues` half of `_echo`, and one on `_echo`'s docstring.

**Outside this diff**, in `github_link_header` and therefore inherited by every listing rather than introduced here: past the last page, real's `rel="prev"` points at the last page and `rel="last"` is still sent. `psf/requests/branches?per_page=1&page=99` answers `[]` with `prev` page 6, `last` page 6, `first` page 1, and `/pulls?per_page=1&page=99` has the same shape with page 86. Backlot at `pipeline/branches?per_page=1&page=99` answers `[]` with `prev` page 98 and `first` only. A client following `next` never lands there, but one that overshoots and follows `prev` is sent to another empty page. This sits next to what #126 already turns on (the empty page past the end), so it may belong as a line under that issue's "what a fix has to settle" rather than its own.

tests/test_github.py
khj809: Confirmed and fixed in 3e68478. Reverting `list_issues` to `{"state": state}` at the previous head left `tests/test_github.py` green, so the `/issues` line was untested exactly as you describe. The test is now parametrized over `["pulls", "issues"]` on `diffable` — `/issues` serves the repo's pulls as rows, so no new fixture — and the same mutation now fails it on `le/issues?state=open&per_page=1&page=2`.

backlot/routers/github.py
khj809: Re-measured and it holds: `psf/requests/tags` with no query links `?page=2` and `?page=6` with no `per_page`, where Backlot writes `per_page=100&page=2`. The docstring claimed more than the code, so it now says filters only, and names `github_link_header` as the owner of `per_page` with a pointer to #126. Left the behaviour alone here — it is one line in shared machinery that every listing reads, and the bundled corpus cannot even show it (it needs a listing longer than `default_page_size` answered without a `per_page`, and nothing in the corpus has 100 of anything). Filed as divergence 10 on #126, with that testing constraint written down.

PR #122, opened 2026-09-03
github: a branch carries the protection object, protection_url and _links, and every url names the corpus's own spelling

### LINKED ISSUE #120: github: /branches/{branch} omits the protection object, protection_url and _links real answers on every branch
**Source and commit** — `github`, at `3335fa8` (on the branch for #96).

`GET /repos/{owner}/{repo}/branches/{branch}` answers `name`, `commit` and `protected`, and stops. Real answers three more members on every branch, protected or not: a `protection` object, a `protection_url`, and `_links`. The list route is not affected — real's `/branches` items carry exactly `commit`, `name`, `protected`, which is what Backlot's do.

|---|---|---|
| `/branches/{branch}` keys | `_links`, `commit`, `name`, `protected`, `protection`, `protection_url` | `commit`, `name`, `protected` |
| `protection` on an unprotected branch | `{"enabled": false, "required_status_checks": {"checks": [], "contexts": [], "enforcement_level": "off"}}` | absent |
| `protection_url` | `https://api.github.com/repos/{owner}/{repo}/branches/{branch}/protection` | absent |
| `_links` | `{"self": ".../repos/{owner}/{repo}/branches/{branch}", "html": "https://github.com/{owner}/{repo}/tree/{branch}"}` | absent |
| `/branches?per_page=1` item keys | `commit`, `name`, `protected` | `commit`, `name`, `protected` |

**How you know** — measured against api.github.com on 2026-09-03. psf/requests `3.0` (`protected: false`) and brekkylab/backlot `main` (`protected: true`) both answer the six keys, and both answer the same `protection` object with `enabled: false`. That second measurement is the one a fix has to read carefully: brekkylab/backlot `main` is protected by rulesets (`/rules/branches/main` lists `deletion`, `non_fast_forward`, `required_status_checks`), its classic protection endpoint `/branches/main/protection` is 404, and the `protection` object reports the classic mechanism only — so `protection.enabled` is not a function of `protected`.

**Reproduction** — the bundled corpus, repo `pipeline`, which states `main` as protected:

```python
import backlot, requests

with backlot.serve() as m:
    for name in ("main", "release/2026-03"):
        b = requests.get(f"{api}/branches/{name}", headers=h).json()
        print(name, sorted(b))                    # ['commit', 'name', 'protected'] for both
        print(b["protection"]["enabled"])         # KeyError: 'protection'
```

**What it costs a client** — one written against real reads `branch["protection"]["enabled"]` or follows `protection_url` without checking, because on real those are always present, and it raises on the first branch it fetches. PyGithub's `Branch` object declares all three attributes, so a client through that library sees the gap too.

**Where it lands** — `get_branch` in `backlot/routers/github.py`, which builds the body by hand. `list_branches` next to it needs nothing, since real's list items do not carry these.

**What a fix has to settle**

- What `protection.enabled` says for a stated `protected: true`. The corpus states one bit, and real's object describes classic protection specifically. Two readings are defensible: `enabled: false` with the empty `required_status_checks` for every branch, which is what real answers for a ruleset-protected branch and is byte-identical across the two measurements above; or `enabled: true` for a stated `protected: true`, which is what a classic-protected branch would answer. Whichever it is, the first has a measurement behind it and the second still needs one against a classic-protected branch.
- Whether `protection_url` should point at a route that exists. Real answers `/branches/{branch}/protection` with 404 for a branch that has no classic protection (both branches above), so the cheapest honest shape is to serve the url and let the route 404 the way real does, rather than to serve the url and nothing behind it.
- `_links.html` names a github.com url. Every other `html_url` this router serves has already settled how the host is spelt; this one should follow it, not invent a second answer.

khj809:
Closes #120.

## What changed

**The branch object.** `GET /repos/{owner}/{repo}/branches/{branch}` answers the six members real answers on every branch, protected or not: `name`, `commit`, `_links`, `protected`, `protection` and `protection_url`, in real's own order. It answered the first, second and fourth and stopped, so a client that reads `branch["protection"]["enabled"]` or follows `protection_url` without checking — which is how one written against real is written, and what PyGithub's `Branch` declares — raised on the first branch it fetched. `protection.enabled` follows the corpus's stated `protected` bit, over real's empty `required_status_checks` block; `_links.html` is github.com, the host every other `html_url` here is already spelt against. A slash in a branch name stays a slash in all three urls.

`protection_url` now points at a route rather than at nothing: `GET .../branches/{branch}/protection` answers real's 404 for a caller without repo-admin rights. It is declared above `get_branch` so the trailing `/protection` beats a name that could swallow it, which is real's precedence too. Its handler resolves nothing, so the credential, the owner and the API version are answered by the router-wide dependencies alone — 401, 404 and 400, each asserted.

The list route is untouched — real's items carry exactly `commit`, `name` and `protected`, which is what Backlot's carry, and a new assertion pins that they stay that. Dropping `GET /repos/{}/{}/branches/{}/protection` from `backlot/fidelity/baseline/github.json` is the acknowledgement that route no longer needs.

**The owner and repo spelling.** Both path segments now match in any case and answer in the corpus's spelling. The `{owner}`/`{org}` segment already matched case-insensitively, as GitHub logins do, and was then interpolated everywhere: `/repos/ACME/codebase` answered a `full_name`, an `html_url`, eleven url templates and a git/ssh/clone triple all naming `ACME`. `{repo}` did not match at all — container lookup was a case-sensitive `=`, so `/repos/acme/CODEBASE` was a 404 where real answers the repository. Each is canonicalized once, in a router-wide dependency, so the ~30 handlers that interpolate them and everything `synth` derives from them cannot disagree; `/orgs/{org}` builds its `url` and `repos_url` from that name rather than from the request's own path. The repo half resolves a NAME only — whether the caller may see it is still `_require_repo`'s answer, taken afterwards on the canonical name, so a repo hidden from a scoped token stays a 404 in every spelling.

A `repo:` search qualifier resolves the same way, where the old behaviour was worse than a 404: an unresolved name left the issue search unnarrowed, so `repo:GATEWAY` answered with every repo's issues instead of that repo's two. The case-insensitive resolve is a new store function rather than a widening of `get_container`, which every source shares — nothing measured here says a Slack channel or a Drive folder behaves this way, and `jira_project_by_key` is the existing precedent for keeping exact and case-insensitive resolution apart.

The owner half came out of the self-review of the two new `_links` fields — fixing it only for them would have made them disagree with the `commit.url` built two lines above.

Measured against api.github.com on 2026-09-03, over 30 branches of 28 repos. Every one answered those six keys and no seventh, and none omitted the `protection` object.

`protection.enabled` is not `protected`, which is the reading the fix had to get right: `enabled` reports CLASSIC branch protection where `protected` covers any mechanism. A ruleset-protected branch answers `protected: true` with `enabled: false` (fastapi/fastapi `master`, pallets/flask `main`, astral-sh/uv `main`, and brekkylab/backlot `main`, whose `/rules/branches/main` lists three rules while its classic endpoint has nothing); a classic-protected one answers `enabled: true` — 15 of the 22 protected branches measured, psf/requests `main` among them; an unprotected branch answers false either way (git/git and torvalds/linux `master`, psf/requests `3.0`). A corpus states one bit and no mechanism, and Backlot serves no rulesets route, so the classic reading is the one that leaves nothing unexplained: the ruleset reading would call a branch protected while the same object says `enabled: false` with nothing on the served surface to say why.

`required_status_checks` is `enforcement_level: "off"` over two empty lists for both values of `enabled`. A corpus records no CI — the reason `/statuses/{sha}` answers `[]` — and that block is byte-identical to what real answers both for an unprotected branch (psf/requests `3.0`) and for a classic-protected one requiring no check (django/django, numpy/numpy, facebook/react, psf/black, apache/airflow `main`).

The protection endpoint's 404 is measured as a non-admin: psf/requests `main`, which reports `enabled: true`, and `3.0`, which reports false, both answer 404 `Not Found` with the branch-protection anchor. An admin gets one of two other answers — 404 `Branch not protected` for a branch with no classic protection (brekkylab/backlot `main`), or a 200 carrying the whole classic configuration, a dozen settings a corpus states nothing about — so Backlot answers as the caller it can account for rather than inventing the settings for the caller it cannot. Its precedence over a slashed branch name is measured rather than assumed from declaration order: `/branches/bug/5671/protection` on psf/requests, where `bug/5671` IS a branch, answers this endpoint's anchor and not get-a-branch's, so real read the suffix as the route and `bug/5671` as the branch. Both sides therefore leave a branch actually named `…/protection` unreachable here.

For the spellings: `/repos/PSF/REQUESTS` and `/repos/PSF/Requests` both answer `name: requests`, `full_name: psf/requests`, `owner.login: psf` and every url, template and git url lowercase; `/orgs/PSF` answers `login: psf`, `url: https://api.github.com/orgs/psf` and `repos_url` to match; an issue item under the same mixed-case path answers `url`, `repository_url`, `comments_url` and `html_url` lowercase; and `search/issues?q=repo:PSF/Requests` answers the same 4,173 issues as the lowercase qualifier, with `repository_url` canonical. Every one is a 200 with no `Location`, so real normalizes in the body rather than redirecting — which is what makes canonicalizing the path parameter the whole fix.

The pagination `Link` header is deliberately left naming the path the request arrived on: real paginates by numeric repository id there (`/repositories/1362490/issues`), so there is no named url to match. One divergence the measurement turned up is NOT in scope and wants its own issue: a repo object's `owner.url` points at `/users/{login}`, a route Backlot does not serve, which `_repo_obj`'s own "a template iff the resource" rule says it should not do.

## Verification

- **Tests** — `pytest` on a `uv sync --all-extras` install, with `main` merged in: 1904 passed, 1 skipped.
- **Optional surfaces that ran rather than skipped** — `tests/test_s3.py`, `tests/test_sdk.py`, `tests/test_mcp.py`, `tests/test_llamaindex.py` and `tests/test_integrations.py` together: 146 passed, 1 skipped (`llama_index.readers.hubspot` is not published). The PyGithub path was also driven directly against a served throwaway corpus: `Branch.protection_url` and `Branch.raw_data["_links"]` read back, following `protection_url` gives the 404 above, and a request to `/repos/ACME/pipeline/issues` answers item urls spelling the org `acme`.
- **Lint** — `ruff check . && ruff format --check .`: clean, 138 files already formatted.
- **Fidelity** — `backlot diff --source github`: 1253 acknowledged, no stale entries (it named the dropped one before this change; the spellings are body fields, which that comparison does not read). It now also reports one NEW gap that is not from this branch — GitHub's published spec has gained `GET /repos/{}/{}/stargazers/history` since the baseline's `2026-09-02`, a sibling of the two stargazer operations already acknowledged. Adding a served route can only remove `missing_operation` findings, and `ci.yml` does not run this comparison; the scheduled fidelity workflow should carry that one.

- [x] A test fails without this change — `test_github_stated_protection_decides_the_filter_and_the_protection_object` fails on the three missing members, and `test_github_answers_the_corpus_spelling_whatever_case_was_asked_for` fails three ways: 22 fields of the repo object echoing the caller's org, a 404 on the mixed-case repo, and the `repo:` qualifier widening the issue search to 11 hits instead of 2.
- [x] A new endpoint serving corpus content is ACL-scoped, proved for both an admin and a scoped token — the protection route serves no corpus content and resolves nothing, and its 401/404/400 answers for a missing credential, a wrong owner and an unsupported version are asserted; the branch object it hangs off is already scoped, proved in `test_github_branch_listing_is_scoped_to_the_caller` for an admin, a member who sees the pull behind a branch, and one who does not.

### DIFF
--- a/backlot/routers/github.py
+++ b/backlot/routers/github.py
@@ -387,9 +387,10 @@ async def search_issues(
     free, quals = _parse_q(q, _GH_ISSUE_QUALS)
     container = None  # a repo: qualifier narrows to one repo
     for v in quals.get("repo", []):
-        # Any case, as real takes it: `repo:PSF/Requests` answers the same 4,173 issues as
-        # `repo:psf/requests` (measured 2026-09-03). An unresolved name leaves `container` None,
-        # which widens to the whole corpus — a spelling this refused answered other repos' issues.
+        # Any case, as real takes it: `repo:PSF/Requests is:issue` answers the same 4,173 as the
+        # lowercase spelling (measured 2026-09-04). A name that resolves in NO case still leaves
+        # `container` None, which WIDENS this search to the whole corpus, where `_code_repos`
+        # beside it narrows to nothing and real refuses the query with a 422.
         spelled = store.container_spelling(conn, "github", v.split("/")[-1])
         if spelled is not None:
             container = spelled
--- a/backlot/routers/github.py
+++ b/backlot/routers/github.py
@@ -168,12 +168,36 @@ async def _validate_path_owner(request: Request) -> None:
     request.path_params[key] = canonical
 
 
+async def _canonical_path_repo(request: Request) -> None:
+    """Hand the handlers the corpus's spelling of ``{repo}``, whatever case it was asked for.
+
+    Real resolves the name case-insensitively and answers in its own spelling — `/repos/PSF/Requests`
+    is 200 with `name: requests` (measured 2026-09-03) — where a case-sensitive container lookup
+    404'd it. Canonicalized, not merely matched, for the reason :func:`_validate_path_owner` gives.
+
+    A NAME and nothing else: whether the caller may SEE the repo stays :func:`_require_repo`'s
+    answer, taken afterwards on this name, so a repo hidden from a scoped token is a 404 in every
+    spelling.
+    """
+    repo = request.path_params.get("repo")
+    if repo is None:
+        return
+    spelled = store.container_spelling(auth.conn(request), "github", repo)
+    if spelled is not None:
+        request.path_params["repo"] = spelled
+
+
 router = APIRouter(
     prefix="/github",
     tags=["github"],
     # Order is the answering order: an unsupported API version is a malformed request and real
-    # refuses it before authenticating or routing, so it is declared first.
-    dependencies=[Depends(_validate_api_version), Depends(_validate_path_owner)],
+    # refuses it before authenticating or routing, so it is declared first. The repo's spelling is
+    # resolved last, and so never for a request that fails the version or the credential.
+    dependencies=[
+        Depends(_validate_api_version),
+        Depends(_validate_path_owner),
+        Depends(_canonical_path_repo),
+    ],
 )
 
 
@@ -363,9 +387,12 @@ async def search_issues(
     free, quals = _parse_q(q, _GH_ISSUE_QUALS)
     container = None  # a repo: qualifier narrows to one repo
     for v in quals.get("repo", []):
-        name = v.split("/")[-1]
-        if store.get_container(conn, "github", name) is not None:
-            container = name
+        # Any case, as real takes it: `repo:PSF/Requests` answers the same 4,173 issues as
+        # `repo:psf/requests` (measured 2026-09-03). An unresolved name leaves `container` None,
+        # which widens to the whole corpus — a spelling this refused answered other repos' issues.
+        spelled = store.container_spelling(conn, "github", v.split("/")[-1])
+        if spelled is not None:
+            container = spelled
     if free:
         cand = store.search_documents(conn, free, "github", ids, limit=10_000, container=container)
     else:
@@ -469,8 +496,11 @@ def _code_repos(conn, quals: dict, org: str) -> set[str] | None:
         owner, _, name = v.rpartition("/")
         if owner and owner.lower() != org.lower():
             continue
-        if store.get_container(conn, "github", name) is not None:
-            names.add(name)
+        # The corpus's spelling, from a qualifier in any case — as the owner half is already
+        # compared (see :func:`store.container_spelling` for what real answers).
+        spelled = store.container_spelling(conn, "github", name)
+        if spelled is not None:
+            names.add(spelled)
     return names
 
 
--- a/backlot/routers/github.py
+++ b/backlot/routers/github.py
@@ -1401,6 +1401,23 @@ async def list_tags(owner: str, repo: str, request: Request):
     ]
 
 
+@router.get("/repos/{owner}/{repo}/branches/{branch:path}/protection")
+async def get_branch_protection(owner: str, repo: str, branch: str, request: Request):
+    """Where `protection_url` points: real's 404 for a caller without repo-admin rights.
+
+    Real gates this on those rights and answers everyone else `Not Found`, protected branch or not
+    (psf/requests `main` and `3.0`, 2026-09-03). The admin answers need what a corpus does not
+    state — `Branch not protected`, or a 200 carrying the classic configuration — so nothing here
+    resolves; the credential and the owner are still the router-wide dependencies'.
+
+    Declared above :func:`get_branch` so the trailing `/protection` beats a name that could
+    swallow it, which is real's precedence: `/branches/bug/5671/protection` on psf/requests, where
+    `bug/5671` IS a branch, answers this anchor and not get-a-branch's. A branch actually named
+    `…/protection` is therefore unreachable on both.
+    """
+    raise HTTPException(status_code=404, detail="Not Found")
+
+
 @router.get("/repos/{owner}/{repo}/branches/{branch:path}")
 async def get_branch(owner: str, repo: str, branch: str, request: Request):
     """One branch, if the listing holds it — 404 otherwise, as real answers for a name no branch
@@ -1408,7 +1425,24 @@ async def get_branch(owner: str, repo: str, branch: str, request: Request):
 
     The name is the trailing PATH because a branch name may contain a slash and real serves it
     whole: `/branches/bug/5671` on psf/requests answers that branch, where a single path segment
-    could only 404 it.
+    could only 404 it. A slash stays a slash in the urls below, unescaped, as real spells them.
+
+    Six members on every branch, protected or not — measured 2026-09-03 across 30 branches of 28
+    repos, none omitting one or carrying a seventh. `_links.html` is github.com, the host every
+    `html_url` here is already spelt against; `self` and `protection_url` are Backlot's own, built
+    from the branch this route resolved, like the `commit.url` beside them.
+
+    **`protection.enabled` is not `protected`.** It reports CLASSIC protection where `protected`
+    covers any mechanism, so real answers `protected: true` with `enabled: false` for a
+    ruleset-protected branch (fastapi/fastapi `master`, brekkylab/backlot `main`) and
+    `enabled: true` for a classic one (psf/requests `main`, 15 of the 22 protected branches
+    measured). A corpus states the one bit and no mechanism, and Backlot serves no rulesets route,
+    so it reads as classic — the other reading calls a branch protected with nothing served to say
+    why.
+
+    `required_status_checks` is real's empty block either way, measured on an unprotected branch
+    (psf/requests `3.0`) and a classic-protected one requiring no check (django/django `main`): a
+    corpus records no CI, which is also why `/statuses/{sha}` answers `[]`.
     """
     conn = auth.conn(request)
     caller = _require(request)
@@ -1420,14 +1454,28 @@ async def get_branch(owner: str, repo: str, branch: str, request: Request):
         raise HTTPException(status_code=404, detail="Branch not found")
     ab = _api_base(request)
     commit_sha, tree_sha = _repo_commit_sha(repo), _repo_tree_sha(repo)
+    self_url = f"{ab}/repos/{owner}/{repo}/branches/{branch}"
     return {
         "name": branch,
-        "protected": found["protected"],
         "commit": {
             "sha": commit_sha,
             "commit": {"tree": {"sha": tree_sha}},
             "url": f"{ab}/repos/{owner}/{repo}/commits/{commit_sha}",
         },
+        "_links": {
+            "self": self_url,
+            "html": f"https://github.com/{owner}/{repo}/tree/{branch}",
+        },
+        "protected": found["protected"],
+        "protection": {
+            "enabled": found["protected"],
+            "required_status_checks": {
+                "enforcement_level": "off",
+                "contexts": [],
+                "checks": [],
+            },
+        },
+        "protection_url": f"{self_url}/protection",
     }
 
 
--- a/backlot/routers/github.py
+++ b/backlot/routers/github.py
@@ -133,18 +133,39 @@ async def _validate_api_version(request: Request) -> None:
 
 
 async def _validate_path_owner(request: Request) -> None:
-    """404 a request whose ``{owner}``/``{org}`` segment is not the org we serve.
+    """404 a request whose ``{owner}``/``{org}`` segment is not the org we serve, and hand the
+    handlers the org's OWN spelling of it rather than the caller's.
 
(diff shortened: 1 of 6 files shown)


backlot/store.py
```diff
@@ -2881,6 +2881,29 @@ def get_container(conn, source_type, name) -> sqlite3.Row | None:
...
+    """The corpus's spelling of a container named in any case, or None for one it does not hold.
+
+    Only the GitHub router resolves a container this way, because only GitHub is measured to:
+    `/repos/PSF/Requests` answers 200 with `name: requests` (2026-09-03). A Slack channel or a
+    Drive folder would each need their own measurement, which is why this is separate from
+    :func:`get_container` the way :func:`jira_project_by_key` is separate from exact key lookup.
+
+    An EXACT match wins, and is tried first so the common request stays on the column's index.
+    With no exact match and more than one candidate there is nothing to prefer, so this reports
+    None and the caller 404s — real cannot present that case, a BYO corpus can. ``NOCASE`` folds
```
nuri-yoo: "the caller 404s" is true of `_canonical_path_repo` and not of the two search callers: `search_issues` leaves `container = None` for a `None` here and widens to the whole corpus, `_code_repos` narrows to no hits. Measured on a corpus stating `Gateway` and `gateway`: `search/issues?q=repo:GATEWAY` answers both repositories' items. The behaviour is #130's; this docstring could say "the path dependency leaves the name as asked and `_require_repo` 404s it" so it does not promise what the search routes do not do.

backlot/routers/github.py
```diff
@@ -342,9 +387,12 @@ async def search_issues(
     free, quals = _parse_q(q, _GH_ISSUE_QUALS)
     container = None  # a repo: qualifier narrows to one repo
     for v in quals.get("repo", []):
-        name = v.split("/")[-1]
-        if store.get_container(conn, "github", name) is not None:
-            container = name
+        # Any case, as real takes it: `repo:PSF/Requests` answers the same 4,173 issues as
+        # `repo:psf/requests` (measured 2026-09-03). An unresolved name leaves `container` None,
+        # which widens to the whole corpus — a spelling this refused answered other repos' issues.
```
nuri-yoo: This still describes the current behaviour for a name that resolves in no case: `repo:ghost` answers 11 items on the bundled corpus, `repo:acme/ghost` the same, and the owner half is not read, so `repo:someone-else/docs-site` answers `docs-site`'s 3. Real answers 422 `Validation Failed` for all three (measured today), and `search/code` beside this narrows an unresolved name to 0 as real does on that route. Out of this PR's scope, so filed as #130; mentioning it here because the comment reads as if the widening were past tense.

tests/test_github.py
```diff
@@ -2068,6 +2125,82 @@ def test_github_validates_the_owner_segment(gh_client, gh_admin_h, gh_org):
...
+
+    # `/orgs/{org}` is held to the same rule, and the synthesized `id` is the sharp end of it:
+    # derived from the name, it forked into two values for one org.
+    org = c.get(f"/github/orgs/{shout}", headers=gh_admin_h).json()
+    assert org == c.get(f"/github/orgs/{gh_org}", headers=gh_admin_h).json()
+    assert org["login"] == gh_org and org["html_url"] == f"https://github.com/{gh_org}"
+    assert org["url"].endswith(f"/orgs/{gh_org}") and org["repos_url"].endswith("/repos")
+
+    # A `repo:` qualifier resolves the same way, on both search routes: real answers the same
+    # 4,173 issues for `repo:PSF/Requests` as for `repo:psf/requests`, with `repository_url`
```
nuri-yoo: Nit: `search/issues?q=repo:PSF/Requests` totals 7,234 today; 4,173 is the `is:issue` count of the same repository. The equality the test asserts holds either way, so only the number wants changing (same figure in the `search_issues` comment at `github.py:390`).

nuri-yoo (approved):
Approving at `4be7506`. Everything below was run against that head, in a worktree, with `PYTHONPATH=. .venv/bin/python` on an all-extras install.

**Reproduced**

- `pytest`: 1905 passed, 0 skipped. `ruff check . && ruff format --check .` clean, `scripts/gen_docs.py --check` clean. With `backlot/routers/github.py` swapped for `main`'s, `test_github_stated_protection_decides_the_filter_and_the_protection_object` fails on the three missing members and `test_github_answers_the_corpus_spelling_whatever_case_was_asked_for` on the mixed-case 404, as the body says; the list-route assertion passes on both, as it should.
- `backlot diff --source github`: 1253 acknowledged, the dropped `/protection` entry not reported stale, one new `stargazers/history` gap. `main`'s baseline lacks that operation too, so it is the spec moving and not this branch; `fidelity.yml` runs daily and will carry it.
- Against api.github.com today: `bug/5671` keeps its slash in `_links.self` and `protection_url`; `/branches/bug/5671/protection` answers the branch-protection anchor; `/protection` is 404 `Not Found` for a non-admin on `main`, on `3.0`, and on a branch that does not exist; `/repos/PSF/Requests` and `/orgs/PSF` answer lowercase everywhere; the `Link` header names `/repositories/1362490/issues`; the six keys arrive in the order the handler builds them; django/django and numpy/numpy `main` answer `enabled: true` over `enforcement_level: "off"`. psf/requests `main` answers `enforcement_level: "everyone"`, which the body does not claim otherwise, so nothing here contradicts it.
- Every url-bearing route under `/repos/ACME/PIPELINE` (repo, branches, branch, tags, issues, issue, comments, pulls, pull, files, commits, reviews, commit, tree, contents, readme, ref, statuses, collaborators, teams) and `/orgs/ACME/repos`, `/user/repos`, both searches: no string in any body carries the caller's spelling. `_canonical_path_repo` runs after `_validate_path_owner`'s `_require`, so 401 still precedes it.
- ACL: on a corpus with a repository hidden from `bob`, nineteen routes asked for as `ACME/VAULT` answer byte-identically to the lowercase request (404 `Not Found` each), and `repo:VAULT` in search answers 0. The name is canonicalized, the visibility is not widened.
- Clients: PyGithub reads `Branch.protection_url` and `raw_data["_links"]` and `get_protection()` raises the 404 above; fsspec's `GithubFileSystem` with `org="ACME", repo="PIPELINE"` lists, reads and enumerates refs; llama-index's `GitHubRepositoryIssuesReader` with the same spelling returns the same eight documents as the lowercase one; the OpenAPI-to-MCP bridge now lists 32 tools, `get_branch` answers the mixed-case slashed request, and `get_branch_protection` raises a tool error carrying the 404 body, which is what a tool derived from real's spec does for a non-admin. Percent-encoded slashes (`release%2F2026-03`, with and without `/protection`) and a percent-encoded owner answer as real does. Branch names stay case-sensitive (`/branches/MAIN` is 404 `Branch not found`), which is also real.

**Not this PR's** (pre-existing, noted so they are not read as regressions): a trailing slash on `/branches/main` answers 200 where real 404s; HEAD is 405 here and 200/404 on real; OPTIONS is 405 against real's 204.

Three inline notes, none blocking. The first is a follow-up I opened as #130 rather than something for this branch.

backlot/store.py
khj809: Confirmed on your scenario — a corpus stating both `Gateway` and `gateway`, where `repo:GATEWAY` matches neither exactly: `search/issues` answers both repositories' items (2), `search/code` answers none (0), and `/repos/{org}/GATEWAY` is the only 404. The docstring now reports None and says what None MEANS is each caller's own — the path dependency leaves the name as spelt so `_require_repo` 404s it, an unresolved qualifier is each search route's own business. Fixed in 3257497.

backlot/routers/github.py
khj809: Right on both counts — `repo:ghost` still answers every repository's issues on the bundled corpus, so the widening is present tense and only the CASE half of it is this branch's. I measured real's side too: `repo:ghost-does-not-exist-xyz` and `repo:someone-else/requests` both answer 422 `Validation Failed` ("The listed users and repositories cannot be searched..."). The comment is now present tense and names all three answers — this route widens, `_code_repos` narrows to nothing, real 422s. Fixed in 3257497; the behaviour stays #130's.

tests/test_github.py
khj809: Measured again today: bare `repo:PSF/Requests` is 7,234 and `repo:PSF/Requests is:issue` is 4,173, so the figure was right for the query I ran and wrong for the query I wrote. Both comments now name the `is:issue` qualifier the number came from, with the date. Fixed in 3257497.

Issue #128, opened 2026-09-04
linear: or inside a relation filter ORs the keys of one branch, and an empty or null-only branch beside a comparator adds the issues without the relation


nuri-yoo:
**Source and commit** — `linear`, at `d6025bf` (on the branch for #129).

[...]

PR #129, opened 2026-09-04
linear: `null: true` on a relation filter reads nothing else in the object, and `IntegrationService` carries `origin`

### LINKED ISSUE #124: linear: `null: true` on a relation filter ignores its sibling keys; Backlot ANDs them
_Found while re-measuring `null: null` for #116; out of that PR's scope because it is on `main` too and the fix needs a reading #116's measurement could not settle._

A nested relation filter (`assignee`, `creator`, `project`) is compiled in `backlot/graphql/linear_filters.py::_sub_filter` as an `AND` of its keys, `null` included: `{null: true, name: {eq: X}}` becomes `col IS NULL AND name = X`. Over an issue set with one assigned issue, api.linear.app answers `null: true` as if it were the only key, while `null: false` does AND with its siblings (measured 2026-09-03, one issue assigned to the viewer (`ME`) beside three unassigned; the Backlot column is the fixture in `tests/test_linear.py`, where `ENG-1` is assigned to Bob Stone and `ENG-2`, `DES-1` are not):

| filter | Linear | Backlot |
|---|---|---|
| `assignee: {null: true}` | the three unassigned | the two unassigned |
| `assignee: {null: true, name: {eq: "ME"}}` | **the three unassigned** | none |
| `assignee: {null: true, name: {eq: "nobody"}}` | **the three unassigned** | none |
| `assignee: {null: true, name: {neq: "ME"}}` | the three unassigned | the two unassigned (agrees, because `neq` keeps the issues without the relation) |
| `assignee: {null: false, name: {eq: "ME"}}` | the one | the one |
| `assignee: {null: false, name: {neq: "ME"}}` | none | none |
| `or: [{assignee: {null: true}}, {title: {eq: T}}]` | all four | all four |

Two readings fit every row: `null: true` **overrides** its siblings (the issues without the relation, full stop), or `null: true` is **ORed** with them (the issues without the relation, plus those the siblings match). The workspace cannot tell them apart because every assigned issue also matches or fails the sibling as a block; the row that decides it is `{null: true, name: {eq: "ME"}}` over two assigned issues, one to `ME` and one to someone else, where override answers the unassigned alone and OR adds the `ME` issue. `project` and `creator` take the same code path and were not measured with a relation present (the workspace has no project, and its four issues have no creator).

`_labels_filter` is not affected: on `IssueLabelCollectionFilter` the keys already do not AND, and `null` there ANDs only beside `length`, `every`, `some`, `name` (#116).

### Done when

- [x] The deciding row is measured with two assigned issues and the reading (override or OR) is written into `_sub_filter`'s docstring as the vendor's rule, with the date
- [x] `null: true` beside a sibling key on `assignee`, `creator` and `project` answers what Linear answers; `null: false` keeps ANDing
- [x] Each change brings a test that fails without it, per `AGENTS.md`

### LINKED ISSUE #127: linear: `IntegrationService` is missing `origin`, a member Linear added after the 2026-09-03 baseline
_Found by khj809 re-running `backlot diff --source linear` on #116; out of that PR's scope because `main` answers the same finding and the enum has nothing to do with the labels filter._

`backlot/graphql/linear.graphql` declares `IntegrationService` with the 44 members introspected on 2026-09-03, and `tests/test_linear.py::LINEAR_ENUMS` pins the served list to those same 44, so `pytest` stays green while `backlot diff` answers `779 divergence(s) · 778 acknowledged · 1 new` on `main` at `ee8be69`. Backlot serves `null` for `Issue.integrationSourceType`, so no answer changes; what a client sees is the enum's member set on introspection, which the enum's own description promises is the vendor's whole one.

Linear does not version its API. The scheduled fidelity run at 11:20Z on 2026-09-03 was green for linear and a run at 14:02Z the same day still answered `0 new`, so the member landed after that. The line below is one measurement taken on 2026-09-04; re-measure before fixing, because the vendor may have moved again.

Introspect the live enum with a personal API key, bare in the header:

```bash
curl -s https://api.linear.app/graphql -H "Authorization: $LINEAR_API_KEY" \
  -H 'Content-Type: application/json' \
  -d '{"query":"{ __type(name: \"IntegrationService\") { enumValues { name } } }"}'
```

### An enum missing a member the vendor added (1)

- [x] `IntegrationService.origin` — Linear: present, between `gitlab` and `googleCalendarPersonal`, 45 members in all; Backlot: absent, 44

### Done when

- [x] `IntegrationService` carries `origin` in the vendor's position and `LINEAR_ENUMS` matches it
- [x] `backlot diff --source linear` on `main` answers `0 new` with `backlot/fidelity/baseline/linear.json` unchanged, since the finding is new rather than acknowledged
- [x] Each change brings a test that fails without it, per `AGENTS.md`

nuri-yoo:
Closes #124. Closes #127. The `or` rows measured on the way are #128, not this.

## `null: true` on a relation reads nothing else in the object

`backlot/graphql/linear_filters.py::_sub_filter` compiled a nullable relation's `null: true` as one AND term among the object's keys, so `assignee: {null: true, name: {eq: X}}` was `col IS NULL AND name = X` and answered none. api.linear.app answers the issues without the relation and reads nothing else in the object. The row #124 could not settle, `{null: true, name: {eq: ME}}` over two issues with the relation, one matching the sibling and one not, is settled: measured 2026-09-04 on `project` with `BRE-1` in a throwaway project `P1` and `BRE-2` in `P2` beside two in none, every sibling tried (`name`, `id`, both, an `and` or `or` of them) answered the two without; on `assignee` with one issue assigned to the viewer beside three unassigned, the same; `creator` takes the same path and answered every issue of a workspace whose issues have no creator, where the AND answers none. The workspace is back to four issues, no project, no assignee.

The rule, as `_sub_filter` now reads it. `and` branches are keys of the same object, so `{and: [{null: true}], name: {eq: X}}` and `{and: [{null: true}, {name: {eq: X}}]}` are `null: true` too. The object's own `null` wins over one inside its `and` branches (`{null: false, and: [{null: true}]}` is the issues with the relation); between branches `true` wins. `null: false` does AND with its siblings. An `or` is the union of its branches, each read by this rule, so `{or: [{null: true}, {name: {eq: X}}]}` is the unassigned issues plus X's. Two relation filters at the level of `IssueFilter` are two filters and AND as `IssueFilter` ANDs: `{and: [{assignee: {null: true}}, {assignee: {name: {eq: X}}}]}` is none, as measured.

| filter (workspace as above) | Linear | Backlot before |
|---|---|---|
| `project: {null: true, name: {eq: "P1"}}` / `{null: true, id: {eq: P1}}` / `{null: true, name: {neq: "P1"}}` / `{null: true, name: {eq: "nobody"}}` | the two without, all four rows | **none** |
| `project: {and: [{null: true}], name: {eq: "P1"}}` / `{and: [{null: true}, {name: {eq: "P1"}}]}` / `{and: [{and: [{null: true}]}, {name: {eq: "P1"}}]}` | the two without | **none** |
| `project: {and: [{null: true}, {null: false}]}` / `{and: [{null: false}, {null: true}]}` / `{null: true, and: [{null: false}]}` | the two without | **none** |
| `project: {null: false, and: [{null: true}]}` | the two with | **none** |
| `project: {null: false, name: {eq: "P1"}}` / `{null: false, name: {neq: "P1"}}` / `{null: false, name: {eq: "nobody"}}` | P1's / P2's / none | P1's / P2's / none |
| `project: {null: false, or: [{null: true}, {name: {eq: "P1"}}]}` | P1's | P1's |
| `project: {or: [{null: true}, {name: {eq: "P1"}}]}` / `{or: [{null: true, name: {eq: "P1"}}]}` | P1's and the two without / the two without | P1's and the two without / **none** |
| `{and: [{project: {null: true}}, {project: {name: {eq: "P1"}}}]}` | none | none |
| `{project: {null: true, name: {eq: "P1"}}, assignee: {null: true}}` (BRE-1 assigned to the viewer) | the two without | **none** |
| `assignee: {null: true, name: {eq: "ME"}}` / `{null: true, name: {neq: "ME"}}` / `{and: [{null: true}, {name: {eq: "ME"}}]}` | the three unassigned | **none** |
| `creator: {null: true, name: {eq: "ME"}}` (no issue has a creator there) | all four | **none** |

The 27 cells in `_RELATION_NULL_CELLS` pin these on the fixture (`ENG-1` Bob Stone's and in `runtime`, `ENG-2` in `runtime` unassigned, `DES-1` neither, every issue with a creator) in `test_relation_null_true_reads_nothing_else_in_the_object_as_linear`; 18 of them fail with the compiler at `main`; the other 9 are the five rows in the `null: false` does AND group, `{or: [{null: true}, {name: {eq: X}}]}`, the `IssueFilter`-level `and`, `{project: {null: true, name: {eq: "runtime"}}, assignee: {name: {eq: "Bob Stone"}}}`, and `{null: true, name: {neq: X}}`, which the relation `neq` rule (issues without the relation pass it) already answered right. The served `null` on `NullableUserFilter` and `NullableProjectFilter` now carries a description saying what `true` reads, scoped to the object's own keys and its `and` branches so it claims nothing about an `or` branch. What an `or` inside a relation filter does is not the union: an empty `or`, an empty branch, a branch carrying more than one key, and an `or` sitting beside another key of the object each diverge. Those rows are #128 with the measurements, #133 compiles the rule that fits them on top of this branch, and this PR leaves `_sub_filter`'s `or` as the union and says in the docstring that the union is Backlot's and which rows it loses.

## `IntegrationService` carries `origin`

`IntegrationService` in `backlot/graphql/linear.graphql` carried the 44 members introspected on 2026-09-03. Linear added a 45th, `origin`, between `gitlab` and `googleCalendarPersonal`, some time after 14:02Z that day (the scheduled fidelity run at 11:20Z was green and a run at 14:02Z still answered `0 new`; khj809 found it on #116 the next morning). This adds the member in the vendor's position, the same on `LINEAR_ENUMS`, and moves the enum description's date to 2026-09-04 with a note on when the member arrived. The baseline is untouched: the finding was `new`, not acknowledged, so closing it needs no entry. Nothing answered changes; Backlot serves `null` for `Issue.integrationSourceType`.

## Verification

- **Tests** — `pytest tests/test_linear.py`: `438 passed`. Full suite, `pytest -rs` on the full-extras venv with Docker, `npx` and `uvx` present: `2171 passed`, no skips reported.
- **Lint** — `ruff check . && ruff format --check .`: `All checks passed!` / `138 files already formatted`. `python scripts/gen_docs.py --check`: exit 0.
- **Fails without the fix** — `linear_filters.py` at `main`, tests kept: `18 failed, 420 passed`, all 18 in `test_relation_null_true_reads_nothing_else_in_the_object_as_linear`. `linear.graphql` at `main`, tests kept: `1 failed, 437 passed`, the failure `test_enum_carries_linears_full_member_list[IntegrationService-members4]`.
- **Live re-measurement** — `backlot diff --source linear` on this branch, 2026-09-04: `775 divergence(s) · 775 acknowledged · 0 new`; `main` at `bd0e2c9`: `776 · 775 · 1`, the one `missing_enum_value IntegrationService.origin`. The relation rows: five measurement rounds against api.linear.app, 119 filters, each round creating the two projects (and, from the second round, assigning one issue to the viewer) and removing them after; the workspace ended each round as it began.

### DIFF
--- a/backlot/graphql/linear_filters.py
+++ b/backlot/graphql/linear_filters.py
@@ -923,13 +923,14 @@ def _sub_filter(conn, spec: dict, mapping: dict) -> tuple[str, list]:
         null = any(inner)
     else:
         null = None
-    # Which column carries "no such relation" is mapping-specific, so use the first mapped column.
-    col = next(m[1] for m in mapping.values())
-    if null is True:
-        return f"{col} IS NULL", []
     parts: list[str] = []
     params: list = []
-    if null is False:
+    if null is not None:
+        # Which column carries "no such relation" is mapping-specific, so use the first mapped
+        # column. Only a nullable relation reaches here; `state` and `team` carry no `null` key.
+        col = next(m[1] for m in mapping.values())
+        if null is True:
+            return _join([f"{col} IS NULL"], "AND"), []
         parts.append(f"{col} IS NOT NULL")
     for key, sub, _ in items:
         if key == "null":
--- a/backlot/graphql/linear_filters.py
+++ b/backlot/graphql/linear_filters.py
@@ -867,33 +867,76 @@ def _issue_parts(conn, flt: dict, team_keys: dict | None) -> tuple[str, list, bo
     return _join(parts, "AND"), params, vacuous
 
 
-def _sub_filter(conn, spec: dict, mapping: dict) -> tuple[str, list]:
-    """A nested object filter (``state``, ``assignee``, ``team``, ``project``)."""
-    parts: list[str] = []
-    params: list = []
+def _relation_items(spec: dict, nested: bool = False) -> list[tuple[str, object, bool]]:
+    """The keys of a nested object filter with its ``and`` branches read as more keys of the same
+    object, each tagged with whether it came from a branch. ``null: null`` is no key at all, unlike
+    on a comparator (see `_Comparator.render`): measured 2026-09-03 with one assigned issue beside
+    three unassigned, `assignee: {null: null}` answered all four where `null: true` answered the
+    three and `null: false` the one, `{null: null, name: {eq: ME}}` the assigned one alone,
+    `{null: null, name: {eq: "nobody"}}` none, and `{or: [{assignee: {null: null}}, {title: {eq:
+    "no such title"}}]}` all four -- the branch left as `{}` is a key that constrains nothing, as
+    `{assignee: {}}` is."""
+    out: list[tuple[str, object, bool]] = []
     for key, sub in (spec or {}).items():
         if sub is None:
-            # `null: null` included: on a relation it is no key at all, unlike on a comparator
-            # (see `_Comparator.render`). Measured 2026-09-03 with one assigned issue beside three
-            # unassigned: `assignee: {null: null}` answered all four where `null: true` answered
-            # the three and `null: false` the one, `{null: null, name: {eq: ME}}` the assigned one
-            # alone, `{null: null, name: {eq: "nobody"}}` none, and `{or: [{assignee: {null:
-            # null}}, {title: {eq: "no such title"}}]}` all four -- the branch left as `{}` is a
-            # key that constrains nothing, as `{assignee: {}}` is.
             continue
-        if key in ("and", "or"):
+        if key == "and":
+            for branch in sub:
+                out.extend(_relation_items(branch, nested=True))
+        else:
+            out.append((key, sub, nested))
+    return out
+
+
+def _sub_filter(conn, spec: dict, mapping: dict) -> tuple[str, list]:
+    """A nested object filter (``state``, ``assignee``, ``team``, ``project``).
+
+    On a nullable relation ``null: true`` is not one condition among the keys: api.linear.app
+    answers the issues without the relation and reads nothing else in the object, so
+    ``assignee: {null: true, name: {eq: X}}`` is the unassigned issues, not none. Measured
+    2026-09-04 on ``project`` with two issues in two throwaway projects beside two in none (every
+    sibling tried -- ``name``, ``id``, both, an ``and`` or ``or`` of them -- answered the two
+    without a project) and on ``assignee`` with one issue assigned to the viewer beside three
+    unassigned; ``creator`` takes the same path and answered every issue of a workspace whose
+    issues have no creator, where an AND would have answered none. ``null: false`` does AND with
+    its siblings: ``{null: false, name: {eq: X}}`` is the issues with the relation named X.
+
+    ``and`` branches are read as more keys of the same object, so ``{and: [{null: true}], name:
+    {eq: X}}`` and ``{and: [{null: true}, {name: {eq: X}}]}`` are ``null: true`` too. A ``null``
+    written at the object's own level wins over one inside its ``and`` branches (``{null: false,
+    and: [{null: true}]}`` is the issues with the relation); between branches ``true`` wins
+    (``{and: [{null: false}, {null: true}]}`` is the issues without it). An ``or`` is the union
+    of its branches, each read by this rule, so ``{or: [{null: true}, {name: {eq: X}}]}`` is the
+    unassigned issues plus X's. An ``or`` at the level of ``IssueFilter`` is not this object:
+    ``{and: [{assignee: {null: true}}, {assignee: {name: {eq: X}}}]}`` is two relation filters
+    ANDed and answers none, as measured."""
+    items = _relation_items(spec)
+    own = [sub for key, sub, nested in items if key == "null" and not nested]
+    inner = [sub for key, sub, nested in items if key == "null" and nested]
+    if own:
+        null = own[0]
+    elif inner:
+        null = any(inner)
+    else:
+        null = None
+    # Which column carries "no such relation" is mapping-specific, so use the first mapped column.
+    col = next(m[1] for m in mapping.values())
+    if null is True:
+        return f"{col} IS NULL", []
+    parts: list[str] = []
+    params: list = []
+    if null is False:
+        parts.append(f"{col} IS NOT NULL")
+    for key, sub, _ in items:
+        if key == "null":
+            continue
+        if key == "or":
             subs = [_sub_filter(conn, s, mapping) for s in sub]
             frags = [f for f, _ in subs if f]
             for _, p in subs:
                 params.extend(p)
             if frags:
-                parts.append("(" + (" AND " if key == "and" else " OR ").join(frags) + ")")
-            continue
-        if key == "null":
-            # `null: true` on a nullable relation means "no such relation on this issue". Which
-            # column carries that is mapping-specific, so use the first mapped column.
-            col = next(m[1] for m in mapping.values())
-            parts.append(f"{col} IS {'NULL' if sub else 'NOT NULL'}")
+                parts.append("(" + " OR ".join(frags) + ")")
             continue
         target = mapping.get(key)
         if target is None:
--- a/backlot/graphql/linear_filters.py
+++ b/backlot/graphql/linear_filters.py
@@ -905,11 +905,15 @@ def _sub_filter(conn, spec: dict, mapping: dict) -> tuple[str, list]:
     {eq: X}}`` and ``{and: [{null: true}, {name: {eq: X}}]}`` are ``null: true`` too. A ``null``
     written at the object's own level wins over one inside its ``and`` branches (``{null: false,
     and: [{null: true}]}`` is the issues with the relation); between branches ``true`` wins
-    (``{and: [{null: false}, {null: true}]}`` is the issues without it). An ``or`` is the union
-    of its branches, each read by this rule, so ``{or: [{null: true}, {name: {eq: X}}]}`` is the
-    unassigned issues plus X's. An ``and`` at the level of ``IssueFilter`` is not this object:
-    ``{and: [{assignee: {null: true}}, {assignee: {name: {eq: X}}}]}`` is two relation filters
-    ANDed and answers none, as measured."""
+    (``{and: [{null: false}, {null: true}]}`` is the issues without it). An ``and`` at the level
+    of ``IssueFilter`` is not this object: ``{and: [{assignee: {null: true}}, {assignee: {name:
+    {eq: X}}}]}`` is two relation filters ANDed and answers none, as measured.
+
+    An ``or`` here is compiled as the union of its branches, each read by the rule above, so
+    ``{or: [{null: true}, {name: {eq: X}}]}`` is the unassigned issues plus X's. That union is
+    Backlot's, not the vendor's: #128 measured rows it loses -- ``{or: [{null: true}, {name: {eq:
+    X}}], name: {eq: Y}}`` is the issues without the relation where the union answers none, and
+    ``{or: []}`` is the issues with it where the union constrains nothing."""
     items = _relation_items(spec)
     own = [sub for key, sub, nested in items if key == "null" and not nested]
     inner = [sub for key, sub, nested in items if key == "null" and nested]
--- a/backlot/graphql/linear_filters.py
+++ b/backlot/graphql/linear_filters.py
@@ -907,7 +907,7 @@ def _sub_filter(conn, spec: dict, mapping: dict) -> tuple[str, list]:
     and: [{null: true}]}`` is the issues with the relation); between branches ``true`` wins
     (``{and: [{null: false}, {null: true}]}`` is the issues without it). An ``or`` is the union
     of its branches, each read by this rule, so ``{or: [{null: true}, {name: {eq: X}}]}`` is the
-    unassigned issues plus X's. An ``or`` at the level of ``IssueFilter`` is not this object:
+    unassigned issues plus X's. An ``and`` at the level of ``IssueFilter`` is not this object:
     ``{and: [{assignee: {null: true}}, {assignee: {name: {eq: X}}}]}`` is two relation filters
     ANDed and answers none, as measured."""
     items = _relation_items(spec)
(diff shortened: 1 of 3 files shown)


backlot/graphql/linear_filters.py
```diff
@@ -867,33 +867,76 @@ def add(frag, p):
...
+
+    ``and`` branches are read as more keys of the same object, so ``{and: [{null: true}], name:
+    {eq: X}}`` and ``{and: [{null: true}, {name: {eq: X}}]}`` are ``null: true`` too. A ``null``
+    written at the object's own level wins over one inside its ``and`` branches (``{null: false,
+    and: [{null: true}]}`` is the issues with the relation); between branches ``true`` wins
+    (``{and: [{null: false}, {null: true}]}`` is the issues without it). An ``or`` is the union
+    of its branches, each read by this rule, so ``{or: [{null: true}, {name: {eq: X}}]}`` is the
+    unassigned issues plus X's. An ``and`` at the level of ``IssueFilter`` is not this object:
+    ``{and: [{assignee: {null: true}}, {assignee: {name: {eq: X}}}]}`` is two relation filters
+    ANDed and answers none, as measured."""
```
khj809: This paragraph is measured vendor behaviour up to the `IssueFilter`-level `and`, and then says the `or` union is too. #128 measured that it is not, and the PR body agrees the union is being kept as an approximation — so the docstring should say whose rule it is and which rows it loses. I re-measured both losing rows today against the fixture: `{project: {or: [{null: true}], name: {eq: "runtime"}}}` answers `[]` here where Linear answers the issues without the project, and `{project: {or: []}}` answers all three where Linear answers the two with it.

```suggestion
    (``{and: [{null: false}, {null: true}]}`` is the issues without it). An ``and`` at the level
    of ``IssueFilter`` is not this object: ``{and: [{assignee: {null: true}}, {assignee: {name:
    {eq: X}}}]}`` is two relation filters ANDed and answers none, as measured.

    An ``or`` here is compiled as the union of its branches, each read by the rule above, so
    ``{or: [{null: true}, {name: {eq: X}}]}`` is the unassigned issues plus X's. That union is
    Backlot's, not the vendor's: #128 measured rows it loses -- ``{or: [{null: true}], name: {eq:
    X}}`` is the issues without the relation where the union answers none, and ``{or: []}`` is the
    issues with it where the union constrains nothing -- and no rule fitting every measured row is
    known yet."""
```

khj809 (approved):
Approving — both issues land, and I re-measured what I could without writing to the workspace.

**Verified independently**

- `IntegrationService` introspected on api.linear.app today: 45 members, `origin` between `gitlab` and `googleCalendarPersonal`. The SDL and `LINEAR_ENUMS` match it member for member and in order, so #127 is closed.
- 18 read-only probes of the relation rule against the workspace as it now stands (four issues, no assignee, creator or project) each answer what `_sub_filter` compiles, the two precedence rows included: `{assignee: {null: false, and: [{null: true}]}}` answers none and `{assignee: {and: [{null: false}, {null: true}]}}` all four, which pins "the object's own `null` wins, `true` wins between branches" without needing the projects. The deciding two-project rows need writes, so those I took from the PR.
- `{project: {null: true, id: {eq: "00000000-0000-0000-0000-000000000000"}}}` is still `Argument Validation Error` on Linear, and `_refuse_malformed_project_ids` runs before `_sub_filter`, so the early return does not swallow it. That is the one neighbouring behaviour the early return could have regressed and no cell covers it — worth one in `_RELATION_NULL_CELLS`.
- `tests/test_linear.py`: `438 passed`. Full suite green, `ruff` clean, `gen_docs.py --check` exit 0, `backlot diff --source linear`: `775 · 775 acknowledged · 0 new` with the baseline untouched. `linear_filters.py` back at `main` with the tests kept: exactly `18 failed`; `linear.graphql` back at `main`: exactly `1 failed`.

Three wording fixes before merge. Nothing in the compiler needs to change.

**1. The "other 9" enumeration lists ten items, and two of them name rows that fail at `main`**

Only one of the two `or` unions passes there — `{or: [{null: true, name: {eq: X}}]}` fails — and of the two two-relation rows only the `assignee: {name: {eq: "Bob Stone"}}` one passes. "The `null: false` rows" is ambiguous too: seven cells carry a `null: false` and five of them pass. The nine are exactly:

```
the other 9 are the five rows under `# `null: false` does AND`, `{or: [{null: true}, {name: {eq:
X}}]}`, the `IssueFilter`-level `and`, `{project: {null: true, name: {eq: "runtime"}}, assignee:
{name: {eq: "Bob Stone"}}}`, and `{null: true, name: {neq: X}}`, which the relation `neq` rule
(issues without the relation pass it) already answered right.
```

**2. The docstring states the `or` union as the vendor's rule**

"An `or` is the union of its branches, each read by this rule" sits inside a paragraph carrying "Measured 2026-09-04", so it reads as a source attribution — but #128 measured that api.linear.app does not read it that way, and this PR says so itself. Per AGENTS.md the docstring has to name the union as Backlot's and the rows it loses. Suggestion inline.

**3. The closing paragraph names a narrower gap than the one left to #128**

"when its branches carry more than one key, or an empty one" does not cover two rows that also diverge on this branch. Measured today, against the fixture: `{project: {or: [{null: true}], name: {eq: "runtime"}}}` — a single-key, non-empty branch — is the issues without the project on Linear and `[]` here, and `{project: {or: []}}` is the issues with the project on Linear and all three here. Suggested replacement:

```
What an `or` inside a relation filter does is not the union and not one other rule either -- an
empty `or`, an empty branch, a branch carrying more than one key, and an `or` sitting beside
another key of the object each diverge; those rows are #128 with the measurements, and this PR
leaves `_sub_filter`'s `or` as the union.
```

Same over-claim, smaller, in the served description the PR adds to `NullableUserFilter.null` and `NullableProjectFilter.null`: a client will read "reads nothing else in this filter, `and` branches included" as covering a `null: true` written inside an `or` branch, where #128 has `{or: [{null: true, name: {eq: "P1"}}, {name: {eq: "P2"}}]}` answering all four. Scoping it to the object's own keys would be honest, but this one I do not think blocks the merge.

Two cosmetics, take them or leave them: `col = next(m[1] for m in mapping.values())` is now evaluated on every call, `state` and `team` included, where no `null` key can appear and an empty mapping would leak a `StopIteration` out of a plain function — it belongs under `if null is not None:`. And the `null is True` return is the only exit that skips `_join`, so it hands back a bare `col IS NULL` where every other return is parenthesized; safe, since `IS` binds tighter than `AND`/`OR` and no caller negates the fragment, just inconsistent.

backlot/graphql/linear_filters.py
nuri-yoo: Taking the attribution point. The paragraph read as one measured rule from `null: true` through the `IssueFilter`-level `and`, and the union sat inside it saying nothing about whose rule it is; the PR body admitted it and the docstring did not. Both rows reproduce at `d6025bf` on the fixture: `{project: {or: [{null: true}], name: {eq: "runtime"}}}` answers `[]` and `{project: {or: []}}` answers `DES-1`, `ENG-1`, `ENG-2`.

Pushing your text with two edits.

The last clause, `and no rule fitting every measured row is known yet`, is the one I am not taking. #133 is open on top of this branch and compiles the rule: 354 filters against api.linear.app, 0 mismatches, the last 73 predicted before they were sent. It rewrites this docstring, so the clause would be false one commit later. What stays is the attribution and the lost rows, which are true whether #129 lands alone or with #133 behind it.

The first example I swapped for the row the table actually carries. `{or: [{null: true}], name: {eq: X}}` is not one of the 354 -- the nearest measured shapes are `{or: [{or: [{null: true}]}], name: {eq: "zz-pa"}}` and `{and: [{or: [{null: true}]}], name: {neq: "zz-pa"}}`, both the two without a project -- so the docstring now names row 15, `{or: [{null: true}, {name: {eq: X}}], name: {eq: Y}}`, which was sent as written. `{or: []}` is row 16, measured on the `assignee` mirror.

nuri-yoo:
All three wording fixes and both cosmetics are in: `959abe0` and `271d091`, on top of `e618d93` for the docstring from the inline thread.

**1. The "other 9".** Ran the 27 cells against `bd0e2c9`'s compiler: `18 failed, 9 passed`, and the nine are your list exactly -- the five rows in the `null: false` group, `{or: [{null: true}, {name: {eq: X}}]}`, the `IssueFilter`-level `and`, `{project: {null: true, name: {eq: "runtime"}}, assignee: {name: {eq: "Bob Stone"}}}`, and `{null: true, name: {neq: X}}`. `{or: [{null: true, name: {eq: X}}]}` and the `assignee: {null: true}` two-relation row fail there, as you said. Body says that now.

**3. The closing paragraph.** Took your four shapes. The one thing I did not take is `and not one other rule either`, for the reason in the inline thread: #133 is open on top of this branch with a rule that fits every measured row, so the clause would be false one commit later. Both rows you measured reproduce here at `d6025bf`: `{project: {or: [{null: true}], name: {eq: "runtime"}}}` is `[]` and `{project: {or: []}}` is `DES-1`, `ENG-1`, `ENG-2`.

**The served description.** Scoped in `271d091` to "reads nothing else written beside it in the same filter object, its `and` branches included". It claims nothing about an `or` branch now, where #128 has `{or: [{null: true, name: {eq: "P1"}}, {name: {eq: "P2"}}]}` answering every issue with the branch's `name` alive.

**Cosmetics.** Both taken in `959abe0`. `col` is looked up under `if null is not None:`, and the `null: true` return goes through `_join` like every other exit. With them the cell you asked for: `{project: {null: true, id: {eq: "00000000-0000-0000-0000-000000000000"}}}` is `Argument Validation Error` / `INVALID_INPUT`. It went into `test_a_malformed_project_id_is_refused_however_deep_the_filter_nests_it` rather than `_RELATION_NULL_CELLS`, since those cells assert identifiers and this one asserts an error; say the word if you would rather see it in the cell list.

One correction to my inline reply: #133 is 429 filters now, not the 354 I quoted -- it moved while I was writing.

Gates at `271d091`: `tests/test_linear.py` `438 passed`, full suite `2171 passed` with no skips, `ruff check .` and `ruff format --check .` clean, `scripts/gen_docs.py --check` exit 0. With `linear_filters.py` back at `bd0e2c9`, tests kept: `18 failed, 420 passed`; with `linear.graphql` back: `1 failed, 437 passed`.

PR #131, opened 2026-09-04
github: every listing pages the way real pages it, and a repo: qualifier that names nothing is refused

### LINKED ISSUE #126: github: eight list routes ignore per_page and send no Link, and page 99 of a one-item listing repeats it
**Source and commit** — `github`, at `3704070` (on the branch for #117).

Eight list routes on the GitHub router take no `page`/`per_page` and send no `Link`. #117 fixed two of these — `/branches` and `/tags` — on the premise that they were the only ones; counting the router's `@router.get` decorators says otherwise. The eight are what is left.

Seven of them are observable today, on the bundled corpus, with no new fixtures. The reason is that a listing does not need to span a page for a client to see the difference: real answers `[]` past the last page, and a route that ignores `page` serves its first page forever. `/repos/{owner}/{repo}/teams` holds at most one team and `/pulls/{n}/commits` exactly one commit, and both still diverge — asked for page 99 they hand back the same single item, which is a paginated walk with no termination condition.

| # | Route | Real pages it | Backlot at `?page=99` | Silent? |
|---|---|---|---|---|
| 1 | `GET /repos/{owner}/{repo}/issues/{n}/comments` | measured | serves the whole listing | yes |
| 2 | `GET /repos/{owner}/{repo}/pulls/{n}/comments` | measured | serves the whole listing | yes |
| 3 | `GET /repos/{owner}/{repo}/pulls/{n}/reviews` | measured | serves the whole listing | yes |
| 4 | `GET /repos/{owner}/{repo}/pulls/{n}/commits` | measured | serves its one commit | yes |
| 5 | `GET /repos/{owner}/{repo}/collaborators` | spec only | serves the whole listing | yes |
| 6 | `GET /orgs/{org}/teams` | spec only | serves the whole listing | yes |
| 7 | `GET /repos/{owner}/{repo}/teams` | spec only | serves its one team | yes |
| 8 | `GET /repos/{owner}/{repo}/statuses/{sha}` | spec only | `[]`, as real does | no |

All eight are already acknowledged in `backlot/fidelity/baseline/github.json` as `missing_param` gaps reading "the vendor accepts it; Backlot does not", so `backlot diff --source github` has been reporting them all along. Closing each one drops its acknowledgement, the way #117 dropped the four for the branch and tag listings.

**How you know** — measured against api.github.com on 2026-09-04.

- `psf/requests/issues/6730/comments?per_page=1` answers one comment with `rel="next"` and `rel="last"` page 5.
- `pydantic/pydantic/pulls/13731` (2 review comments, 2 reviews, 3 commits) answers `?per_page=1` on `/comments`, `/reviews` and `/commits` with one item each and a `Link` carrying `rel="next"` and `rel="last"`.
- `psf/requests/pulls/7616/commits` holds one commit: `?per_page=1&page=1` answers it, `?per_page=1&page=99` answers `[]`. This is the case that makes 4 and 7 observable.
- `psf/requests/statuses/main?page=2` answers `[]`, which is what it answers on page 1 — hence 8 being the one a client cannot tell apart.
- 5, 6 and 7 could not be called live: `/repos/psf/requests/collaborators` is a 403 ("Must have push access to view repository collaborators") and `/orgs/psf/teams` likewise. For those three, GitHub's published OpenAPI declaring `page`/`per_page` is all there is, which is the same source the baseline entries were read off. A fix should re-measure them against a repository and org the author administers rather than trust the spec alone.

```python
import backlot, requests

with backlot.serve() as m:
    n = requests.get(f"{api}/pulls", headers=h, params={"state": "all"}).json()[0]["number"]
    page99 = {"per_page": 1, "page": 99}
    for url in (
        f"{api}/pulls/{n}/commits",
        f"{api}/pulls/{n}/reviews",
        f"{api}/issues/{n}/comments",
        f"{api}/collaborators",
        f"{api}/teams",
        f"{m.base_url}/github/orgs/{org}/teams",
    ):
        r = requests.get(url, headers=h, params=page99)
        print(url.rsplit("/github/", 1)[1], len(r.json()), r.headers.get("Link"))
```

Page 99 of every one of them is non-empty and carries no `Link`: `1`, `2`, `1`, `10`, `1` and `7` items respectively. `/pulls/{n}/comments` is the seventh and is not in the loop because the bundled corpus gives this pull no review comments — a corpus that states one reproduces it the same way.

**What it costs a client** — the same cost #117 named, plus one these routes have that the ref listings did not. A client that pages by following `rel="next"` reads the first page and stops, never learning there was more. A client that composes its own urls instead — incrementing `page` until the answer is empty, which is how you page a listing that reports no total — never terminates: every page comes back full. `/pulls/{n}/commits` and `/repos/{owner}/{repo}/teams` are the worst of it, because their answer is a single item that repeats forever.

**Where it lands** — `issue_comments`, `pull_review_comments`, `pull_reviews`, `pull_commits`, `commit_statuses`, `list_collaborators`, `list_teams` and `list_repo_teams` in `backlot/routers/github.py`. `list_branches` and `list_tags`, as #117 left them, are the shape to follow: `clamp_page` against `default_page_size`/`max_page_size`, then `_paged`, with `_echo` deciding which of the caller's parameters the page urls carry back. None of these eight takes a filter, so `_echo` has nothing to carry and the `extra` dict is `{}` throughout.

**What a fix has to settle**

- Whether `/statuses/{sha}` is worth changing at all. It answers `[]` on every page on both sides, so accepting the parameters closes a baseline gap and changes no response a client can observe. The argument for doing it anyway is that the OpenAPI slice is what `backlot mcp` hands an agent as a tool, and a tool that declines a parameter real accepts is a divergence in the contract even when the body agrees.
- Whether these routes should cap at `default_page_size` when the caller sends nothing. They do not today — they serve everything — and switching them to a 100-item default changes an unpaginated caller's answer for any corpus with more than 100 comments on an issue. Real's own default is 30 across the API, which Backlot does not follow anywhere, so matching real here would make this router inconsistent with itself; matching `default_page_size` is the consistent choice and the one #117 made.
- Whether `pull_commits` and `list_repo_teams` deserve a test that walks to an empty page, given each can only ever hold one item. The measurement above says the empty page is exactly what real answers, and it is the only assertion that can fail for these two.

### LINKED ISSUE #130: github: a repo: search qualifier that resolves to no repository widens the issue search to the whole corpus, where real answers 422
**Source and commit** — `github`, at `4be7506` (on the branch for #120, PR #122).

`GET /search/issues?q=repo:<name>` narrows to one repository when `<name>` resolves to a container and to nothing at all when it does not: the loop in `search_issues` leaves `container` at `None` for an unresolved qualifier, and `None` means the whole corpus. So a caller asking for a repository that does not exist is answered with every repository's issues and pull requests. #122 closed the case-only version of this (`repo:DOCS-SITE` now resolves to `docs-site`), and its comment names the widening, but a name that resolves in no case still widens. The owner half of the qualifier is never read on this route — `v.split("/")[-1]` — so `repo:someone-else/docs-site` answers `docs-site`'s items too. `search/code` reads the same qualifier through `_code_repos`, which checks the owner and narrows an unresolved name to zero hits, so the two search routes disagree with each other before either is compared with real.

| Request (bundled corpus) | Real GitHub | Backlot at `4be7506` | Silent? |
|---|---|---|---|
| `search/issues?q=repo:ghost` | 422 `Validation Failed` | 200, 11 items — every repo's issues and pulls | yes |
| `search/issues?q=repo:acme/ghost` | 422 `Validation Failed` | 200, 11 items | yes |
| `search/issues?q=repo:someone-else/docs-site` | 422 `Validation Failed` | 200, 3 items — `docs-site`'s | yes |
| `search/code?q=repo:ghost` | 200, `total_count: 0`, `incomplete_results: true` | 200, `total_count: 0` | no |
| `search/code?q=repo:someone-else/docs-site` | 200, `total_count: 0`, `incomplete_results: true` | 200, `total_count: 0` | no |

Not in `backlot/fidelity/baseline/github.json`: the comparison there reads operations and parameters, not what a body answers for a given `q`.

**How you know** — measured against api.github.com on 2026-09-04.

- `search/issues?q=repo:psf/ghost-zz-9876` and `search/issues?q=repo:someone-else-zz/requests` both answer 422 with the body `{"message":"Validation Failed","errors":[{"message":"The listed users and repositories cannot be searched either because the resources do not exist or you do not have permission to view them.","resource":"Search","field":"q","code":"invalid"}],"documentation_url":"https://docs.github.com/v3/search/","status":"422"}`. The same message covers a repository the token cannot see, so the answer does not say which of the two it was.
- `search/code?q=repo:psf/ghost-zz-9876+def` and the wrong-owner spelling both answer 200 with `total_count: 0` and `incomplete_results: true`, no error.
- `search/issues?q=repo:PSF/Requests` and `repo:psf/requests` answer the same 7,234 items, which is the case #122 fixed. Its comments say 4,173; that is the `is:issue` count of the same repository, and the equality holds either way.

**Reproduction** — the bundled corpus, repos `docs-site` and `pipeline`:

```python
import backlot, requests

with backlot.serve() as m:
    api = f"{m.base_url}/github/search"
    for route in ("issues", "code"):
        for q in ("repo:docs-site", "repo:ghost", f"repo:{org}/ghost", "repo:someone-else/docs-site"):
            r = requests.get(f"{api}/{route}", headers=h, params={"q": q})
            print(route, repr(q), r.status_code, r.json()["total_count"])
```

prints `issues 'repo:docs-site' 200 3`, then `200 11` for `repo:ghost` and `repo:acme/ghost` and `200 3` for `repo:someone-else/docs-site`; the four `code` lines are `200 1`, `200 0`, `200 0`, `200 0`. The 11 is the corpus's whole issue-and-pull population across both repositories. A BYO corpus that states two repositories differing only in case (`Gateway` and `gateway`, which the schema's `^[A-Za-z0-9._-]+$` admits and a real organization cannot hold) reproduces the same widening for the third spelling: `store.container_spelling` reports `None` for `GATEWAY` because it has two candidates, and the search answers both repositories' items.

**What it costs a client** — a client that scopes a search with `repo:` and trusts the result to be that repository's gets other repositories' issues as hits, with nothing in the response to say the qualifier was dropped: every item's `repository_url` names a repository other than the one asked for, but only a client that re-checks would notice. A typo in a repository name, or an owner that is not the served org, is a search that succeeds against Backlot and fails with 422 against real, which is the ordering a mock is meant to prevent. For a scoped token the widening stops at the caller's own visibility (measured: a user who cannot see `vault` gets 0 for `repo:vault`), so nothing leaks; the answer is wrong, not over-permissive.

**Where it lands** — `search_issues` in `backlot/routers/github.py`: the `for v in quals.get("repo", [])` loop, where an unresolved `store.container_spelling` leaves `container = None`. `_code_repos` beside it is the shape the owner check should take, and `search_code`'s zero-hit answer for an unresolved name is what real does on that route, so the code path needs at most `incomplete_results`. The docstring of `store.container_spelling` says an ambiguous name "reports None and the caller 404s", which is true of `_canonical_path_repo` and not of the two search callers.

**What a fix has to settle**

- Whether `search/issues` answers real's 422 or narrows to zero hits. Real's answer is the 422 above, and it is the same 422 for a repository the caller cannot see, so answering it for an unresolved name and for a hidden one alike matches real and keeps the hidden repository's existence out of the answer. Zero hits would be quieter but is what `search/code` does and real does not, on this route.
- Whether the owner half is checked on `search/issues` the way `_code_repos` already checks it. Real rejects `repo:someone-else/requests` with the same 422, so reading only the name is a second way to widen, not a convenience.
- Whether `search/code` should say `incomplete_results: true` for an unresolved repository, as real does. It is the only field the two answers differ in, and no client is known to read it.
- Whether the importer should refuse two repositories that differ only in case. `container_spelling`'s `None` for that corpus is deliberate and documented; a 422 on the search route would then be the answer for a corpus real cannot hold, which is arguably right, but refusing the pair at import names the problem where it was made.

khj809:
Every list route on the GitHub router now pages, the `Link` header says what real's says on every page of a listing, and a `repo:` qualifier that names no searchable repository is refused rather than answered with the whole corpus.

Closes #126. Closes #130.

## The eight listings that took no page

`issue_comments`, `pull_review_comments`, `pull_reviews`, `pull_commits`, `commit_statuses`, `list_collaborators`, `list_teams` and `list_repo_teams` go through `clamp_page` and `_paged` like the rest of the router. None of them takes a filter, so the echoed query is `{}` throughout, and an unpaginated caller now gets `default_page_size` rows — the choice #117 made for the ref listings, kept here so the router is consistent with itself rather than with real's own default of 30.

A listing does not have to span a page for a client to see the difference. Real answers `[]` past the last page where a route that ignores `page` serves its first page forever, so a walk that increments `page` until the answer is empty never terminated: `/pulls/{n}/commits` and `/repos/{o}/{r}/teams` hold a single item and repeated it without end. `psf/requests/pulls/7616/commits` holds one commit and answers it at `?per_page=1&page=1` and nothing at `page=99`; the six collaborators of a repository the author administers answer the same way (measured on api.github.com on 2026-09-04).

`/statuses/{sha}` answers `[]` on every page on both sides, so its parameters change no body a client can read. They are here because real accepts them — `kubernetes/kubernetes` at `?per_page=1` answers one status with a `Link` — and because the OpenAPI slice is what `backlot mcp` hands an agent as a tool. GitHub's published spec declares this legacy route not at all, which is why no baseline entry covered it and the contract is the only place the divergence showed.

## What the Link header says

Three surfaces, three rule sets, none of them shared. Every rule below was measured against api.github.com on 2026-09-04, and each was previously wrong on at least one surface.

| | offset listings and `/search/issues` | `/search/code` | the cursor-paged `/issues` |
|---|---|---|---|
| rels | `prev, next, last, first`; no `first` on page 1, no `last` on the last page holding rows | `next, prev, first, last`; `first` and `last` on every page | `next, prev` only |
| a page past the last one | `prev` and `last` both name the last page holding rows | `prev` is the page asked for minus one, `last` still the end | no header at all |
| a page size the caller did not send | left out of the page urls | written as the size applied | written as the size applied |
| a page size the caller DID send | written in the caller's own spelling | written | written as the size applied |

`/search/code` having its own shape is the reviewer's measurement, reproduced here at five positions on `repo:psf/requests def` (45 hits, nine pages) — the two search routes are as different from each other as either is from a listing, so they build their headers separately.

The repository and organization in a page url are named by id on all three — `/repositories/1287077005/collaborators`, `/organizations/130592615/repos` — where Backlot named them by login.

The first costs a client that overshot: following `prev` handed it another empty page, and repeating that walked back one page at a time through however far it overshot, where real hands it the last real page in one hop. On the last page that HOLDS rows real sends no `last` at all, which Backlot already matched — the divergence was only past the end.

The page size rule is `_echo`'s reaching the one parameter `_echo` does not own: a size the handler defaulted is not one the caller asked for, and one the caller sent goes back in the caller's own spelling rather than as the size applied. `psf/requests/tags` with no query links `?page=2` and `?page=6` — six pages of 161 tags at real's own default of 30, no `per_page` anywhere — where `?per_page=1` links `per_page=1&page=2`, `?per_page=abc` links `per_page=abc` with `last` page 6 (it applied 30) and `?per_page=500` links `per_page=500` with `last` page 2 (it applied 100). `page` is not like this: every link recomputes it and so does real, answering `?page=abc` with `page=2`.

The id form has to resolve or the header hands a client a url it cannot follow, so `/github/repositories/{id}/…` and `/github/organizations/{id}/…` serve what the login-keyed paths serve, by path rewrite rather than by redirect: real answers 200 on `/repositories/{id}/collaborators` directly, so a 302 would be a shape a client only meets here. Resolution is by id alone and says nothing about visibility — the route it lands on applies the caller's ACL, so a repository this caller cannot read 404s exactly as it does by name. The id in the header is the corpus's spelling, not one minted from the spelling the request used: both segments resolve in any case, so hashing the caller's would name an id nothing holds and 404 the client that followed it. `github_user_id` is `1000 + digest % 9_000_000` and so not injective — `repo485` and `repo4107` both hash to 7755679 — and an id two repositories share names neither, the way `container_spelling` reports `None` for a spelling two of them answer to; walking a client onto the other repository's page with nothing in the response to say so is the failure that guard exists for. An id that names nothing is still rewritten on the prefix, with the id left where the name goes, so that these paths answer in the order the named ones do: `_validate_path_owner` puts credentials ahead of existence, and resolving first would let a caller with none tell an id the corpus holds from one it does not, and so confirm a guessed name. The resource urls in the same response keep the login form, because real's do: a comment fetched in the same breath still carries `url: …/repos/psf/requests/issues/comments/…`.

## The one listing real pages by cursor

`/repos/{owner}/{repo}/issues` is not offset-paged on real. Its header carries `next` and `prev` alone — never `last`, never `first` — each url pairing an opaque cursor with the caller's own page ± 1, and a window past the rows carries no header at all. Measured against a repository with 12 open issues at `per_page=5`: page 1 answers `next` alone, page 2 `next, prev` in that order, page 3 `prev` alone, pages 4 and 99 nothing, and those same 12 rows at `per_page=100` nothing. `/pulls` on the same server answers all four rels from `page=2`, so the asymmetry is real's, not a simplification.

The cursor is what selects the window: `?per_page=2&page=50` carrying page 1's `after` answers the rows `page=2` answers, where `page=50` alone answers a different window, and the page number is carried rather than computed — that request links `page=51`. Carried, and only where the caller claimed a page of 2 or more: `?after=C`, `?page=1&after=C` and `?page=0&after=C` all answer `next` and `prev` with no `page` in either url, so the alternative to a number is the absence of one rather than a `page=0` real never writes. `after`/`before` are read off the query rather than declared as parameters: GitHub's published OpenAPI declares neither for this operation, so declaring them would put an argument in Backlot's contract, and in the tool `backlot mcp` builds from it, that the vendor's spec does not have. This route is also the exception to the page-size rule above, and writes the effective size either way, as real does.

## A `repo:` qualifier that names nothing

`search/issues?q=repo:<unknown>` narrowed to one repository when the name resolved and to nothing at all when it did not, and nothing meant the whole corpus — so a client that scoped its search was answered with other repositories' issues, with nothing in the response to say the qualifier had been dropped. It is now real's 422, carrying real's own body: `code: "invalid"`, the message "The listed users and repositories cannot be searched either because the resources do not exist or you do not have permission to view them.", and a `documentation_url` that differs from the missing-parameter 422's by a trailing slash — real's, not a typo.

The refusal is for a qualifier that resolves to NOTHING. `repo:psf/requests repo:psf/ghost-zz-9876 timeout` is a 200 answering psf/requests' 846 items, so a name that resolves beside one that does not is not refused. Both search routes now read the qualifier through one helper, because a name that resolves for one and not the other is two answers to one question: the owner half is compared on `search/issues` the way `_code_repos` already compared it, and resolution is by spelling AND by visibility — a repository the caller cannot see counts as unresolved, which is real's own answer, its one message covering "does not exist" and "you cannot view it" so that neither answer confirms which.

`search/code` answers those same queries 200 with `total_count: 0` and `incomplete_results: true`, and `false` for the mix above. That flag was `false` for both, the one field its answer differed in.

## What it builds on

The search half needs the case-insensitive `container_spelling` that #122 added, now in `main`: without it, refusing a name that does not resolve would refuse `repo:GATEWAY`, which real resolves. The branch is twelve commits on current `main`.

## Decisions this leaves open

- The importer still accepts two repositories differing only in case, which is what makes `container_spelling` report `None` for either spelling. A corpus real cannot hold then gets the 422 on this route, which #130 calls arguably right; refusing the pair at import is a separate call about BYO load behaviour.
- `search/issues` still narrows to one repository where real unions every repository a `repo:` names — measured, 846 + 45 = 891 for two of them — because `store.search_documents` and `store.list_documents` take a single container. `search_code` already ORs correctly, so the two routes still disagree on this one point.

## Verification

`uv run pytest` with all extras: 2164 passed, 3 skipped. `ruff check` and `ruff format --check` clean, `scripts/gen_docs.py` reports both generated files current. `backlot diff --source github` drops fourteen acknowledgements, leaving 1235 and no stale one behind; its one new finding — `GET /repos/{}/{}/stargazers/history` — being spec drift that `main` reports too. Both issues' reproduction scripts answer what their "Real GitHub" column says, row for row.

### DIFF
--- a/backlot/main.py
+++ b/backlot/main.py
@@ -165,6 +165,28 @@ async def echo_github_api_version(request: Request, call_next):
 
 
 @app.middleware("http")
+async def resolve_github_id_paths(request: Request, call_next):
+    """Serve `/github/repositories/{id}/…` and `/github/organizations/{id}/…` as what the
+    login-keyed paths serve, because that is the form real's page urls take (see
+    ``_page_base_url``) and a page url a client cannot follow is worse than no header at all.
+
+    A rewrite rather than a redirect: real answers 200 on `/repositories/{id}/collaborators`
+    directly, so a 302 would be a shape a client only meets here. An id that names nothing is left
+    alone and 404s on the route it fails to match, the same answer a name that names nothing gets.
+    """
+    path = request.url.path
+    if path.startswith("/github/repositories/") or path.startswith("/github/organizations/"):
+        acl = getattr(request.app.state, "acl", None)
+        canonical = github.canonical_id_path(
+            request.app.state.conn, getattr(acl, "org_name", None) or get_settings().org_name, path
+        )
+        if canonical is not None:
+            request.scope["path"] = canonical
+            request.scope["raw_path"] = canonical.encode()
+    return await call_next(request)
+
+
+@app.middleware("http")
 async def parse_slack_form(request: Request, call_next):
     """Slack SDK POSTs urlencoded params; stash them for the router's param lookup."""
     if request.url.path.startswith("/slack/") and request.method == "POST":
--- a/backlot/main.py
+++ b/backlot/main.py
@@ -171,8 +171,11 @@ async def resolve_github_id_paths(request: Request, call_next):
     ``_page_base_url``) and a page url a client cannot follow is worse than no header at all.
 
     A rewrite rather than a redirect: real answers 200 on `/repositories/{id}/collaborators`
-    directly, so a 302 would be a shape a client only meets here. An id that names nothing is left
-    alone and 404s on the route it fails to match, the same answer a name that names nothing gets.
+    directly, so a 302 would be a shape a client only meets here.
+
+    The rewrite happens on the PREFIX, whatever the id turns out to name, so that these paths
+    answer in the order the named ones do — see ``canonical_id_path`` for why resolving first would
+    tell an unauthenticated caller which ids the corpus holds.
     """
     path = request.url.path
     if path.startswith("/github/repositories/") or path.startswith("/github/organizations/"):
--- a/backlot/pagination.py
+++ b/backlot/pagination.py
@@ -125,16 +125,21 @@ def github_link_header(
     per_page: int,
     total: int,
     *,
-    per_page_sent: bool = True,
+    per_page_param: str | None = None,
 ) -> str | None:
     """Build a Link header with rel=next/prev/first/last. ``params`` are extra query args.
 
-    ``per_page_sent`` says whether the CALLER named the page size; a size the handler defaulted is
-    left out of the urls, the rule a listing's filters already follow. Real omits it: `/tags` with
-    no query links `?page=2` and `?page=6` — six pages of 161 tags at its own default of 30 — where
-    `?per_page=1` links `per_page=1&page=2`, and search behaves the same (measured on
-    api.github.com on 2026-09-04). :func:`github_cursor_link_header` is the exception, and writes
-    the size either way.
+    ``per_page_param`` is the page size the CALLER sent, verbatim, or ``None`` when it sent none.
+    A size the handler defaulted is left out of the urls, the rule a listing's filters already
+    follow, and one the caller sent goes back in the caller's own spelling rather than as the size
+    the handler ended up applying. Both measured on api.github.com on 2026-09-04 against
+    `psf/requests/tags` and its 161 tags: no query links `?page=2` and `?page=6` with no `per_page`
+    anywhere, `?per_page=abc` links `per_page=abc` with `last` page 6 (it applied 30), and
+    `?per_page=500` links `per_page=500` with `last` page 2 (it applied 100). `page` is not like
+    this — every link recomputes it and real recomputes it too, answering `?page=abc` with `page=2`.
+
+    `/search/issues` pages by these rules too. `/search/code` and the cursor-paged issue listing do
+    not: see :func:`github_code_search_link_header` and :func:`github_cursor_link_header`.
 
     Returns ``None`` for a single page, which is what real sends there: no header at all, rather
     than one whose every rel points back at the page the caller is already holding. That holds
@@ -158,11 +163,12 @@ def github_link_header(
     if last_page <= 1:
         return None
 
-    size = {"per_page": per_page} if per_page_sent else {}
+    size = {} if per_page_param is None else {"per_page": per_page_param}
 
     def link(p: int) -> str:
         return _page_url(url_no_query, {**params, **size, "page": p})
 
+    # `last_page >= 2` here, so one of the two page tests always holds and `parts` is never empty
     parts = []
     if page > 1:
         parts.append(f'{link(min(page - 1, last_page))}; rel="prev"')
@@ -172,7 +178,7 @@ def github_link_header(
         parts.append(f'{link(last_page)}; rel="last"')
     if page > 1:
         parts.append(f'{link(1)}; rel="first"')
-    return ", ".join(parts) if parts else None
+    return ", ".join(parts)
 
 
 def github_cursor_offset(page: int, per_page: int, after: str | None, before: str | None) -> int:
@@ -190,8 +196,57 @@ def github_cursor_offset(page: int, per_page: int, after: str | None, before: st
     return (page - 1) * per_page
 
 
+def github_code_search_link_header(
+    url_no_query: str,
+    params: dict,
+    page: int,
+    per_page: int,
+    total: int,
+    *,
+    per_page_param: str | None = None,
+) -> str | None:
+    """Build a Link header for `/search/code`, which pages by none of the listings' rules.
+
+    Measured at five positions on api.github.com on 2026-09-04, `repo:psf/requests def` at
+    `per_page=5` (45 hits, nine pages), and again on `repo:kubernetes/kubernetes extension:md
+    kubelet`:
+
+    * `first` and `last` are on EVERY page — page 1 answers `next, first, last` and page 9, the last
+      that holds rows, answers `prev, first, last`, where a listing sends neither rel there.
+    * `prev` is the page before the one ASKED for even past the end: page 99 answers prev=98, where
+      a listing clamps it to the last page that holds rows.
+    * the rels come in the order next, prev, first, last.
+    * the size is written whether or not the caller named one — no `per_page` in the query still
+      links `per_page=30` — where a listing omits one the caller did not send.
+
+    `/search/issues` shares none of that and pages by :func:`github_link_header`. A result set that
+    fits one page carries no header, which is the one rule all three surfaces share.
+    """
+    last_page = max(1, (total + per_page - 1) // per_page)
+    if last_page <= 1:
+        return None
+    size = per_page if per_page_param is None else per_page_param
+
+    def link(p: int) -> str:
+        return _page_url(url_no_query, {**params, "per_page": size, "page": p})
+
+    parts = []
+    if page < last_page:
+        parts.append(f'{link(page + 1)}; rel="next"')
+    if page > 1:
+        parts.append(f'{link(page - 1)}; rel="prev"')
+    parts.append(f'{link(1)}; rel="first"')
+    parts.append(f'{link(last_page)}; rel="last"')
+    return ", ".join(parts)
+
+
 def github_cursor_link_header(
-    url_no_query: str, params: dict, page: int, per_page: int, total: int, offset: int
+    url_no_query: str,
+    params: dict,
+    page: int | None,
+    per_page: int,
+    total: int,
+    offset: int,
 ) -> str | None:
     """Build a Link header for a listing real pages by CURSOR rather than by offset.
 
@@ -201,21 +256,31 @@ def github_cursor_link_header(
     ever — page 1 answers `next` alone, page 2 `next, prev` in that order, page 3 `prev` alone, and
     pages 4 and 99 no header at all, as do those 12 rows at `per_page=100`.
 
-    Each url pairs an opaque cursor with the caller's own page ± 1. The page number is carried, not
-    computed: `?page=50` with page 1's cursor answers page 2's rows and links `page=51`. The cursor
-    is what names the window (see :func:`github_cursor_offset`).
+    ``page`` is the number the caller ASSERTED, or ``None`` when it asserted none — real writes a
+    page number only where the caller claimed a page of 2 or more, and drops `page` from both urls
+    entirely otherwise. Measured the same day on `?per_page=2`: `?after=C`, `?page=1&after=C` and
+    `?page=0&after=C` all answer `next` and `prev` with no `page` in either url, where
+    `?page=2&after=C` answers next=3/prev=1 and `?page=50&after=C` next=51/prev=49. So the number
+    is carried rather than computed, and the alternative to carrying one is writing none.
+
+    The size is written whether or not the caller sent one, and as the size APPLIED rather than as
+    the caller spelt it: `?per_page=abc` links `per_page=30` here, where `/tags` links `per_page=abc`
+    (both measured). The cursor is what names the window (see :func:`github_cursor_offset`).
     """
     if offset >= total:  # a window past the rows carries no header at all, not even `prev`
         return None
 
-    def link(p: int, **cursor) -> str:
-        return _page_url(url_no_query, {**params, "per_page": per_page, "page": p, **cursor})
+    def link(step: int, **cursor) -> str:
+        q = {**params, "per_page": per_page}
+        if page is not None:
+            q["page"] = page + step
+        return _page_url(url_no_query, {**q, **cursor})
 
     parts = []
     if offset + per_page < total:
-        parts.append(f'{link(page + 1, after=encode_cursor(offset + per_page))}; rel="next"')
+        parts.append(f'{link(1, after=encode_cursor(offset + per_page))}; rel="next"')
     if offset > 0:
-        parts.append(f'{link(page - 1, before=encode_cursor(offset))}; rel="prev"')
+        parts.append(f'{link(-1, before=encode_cursor(offset))}; rel="prev"')
     return ", ".join(parts) if parts else None
 
 
--- a/backlot/pagination.py
+++ b/backlot/pagination.py
@@ -118,7 +118,15 @@ def github_link_header(
     """Build a Link header with rel=next/prev/first/last. ``params`` are extra query args.
 
     Returns ``None`` for a single page, which is what real sends there: no header at all, rather
-    than one whose every rel points back at the page the caller is already holding.
+    than one whose every rel points back at the page the caller is already holding. That holds
+    however far past a one-page listing the caller asks.
+
+    A page PAST the last one is the caller's way out of a listing it overshot, so ``prev`` names the
+    last page that HOLDS rows rather than the page before the one asked for, and ``last`` comes back
+    to say where the rows end. Measured on api.github.com on 2026-09-04: pages 7 and 99 of a
+    six-page collaborator listing both answer prev=6, last=6, first=1, and a search whose 30 results
+    end on page 3 answers prev=3, last=3, first=1 for page 9. On the last page that holds rows there
+    is no ``last`` — a client already there needs no pointer to it.
 
     Values are percent-encoded because a param value is arbitrary text — a search `q` carries
     spaces and colons — and a URI cannot hold them raw. A client that follows the link has to
@@ -140,8 +148,10 @@ def github_link_header(
         parts.append(f'{link(page + 1)}; rel="next"')
         parts.append(f'{link(last_page)}; rel="last"')
     if page > 1:
-        parts.append(f'{link(page - 1)}; rel="prev"')
+        parts.append(f'{link(min(page - 1, last_page))}; rel="prev"')
         parts.append(f'{link(1)}; rel="first"')
+    if page > last_page:
+        parts.append(f'{link(last_page)}; rel="last"')
     return ", ".join(parts) if parts else None
 
 
--- a/backlot/pagination.py
+++ b/backlot/pagination.py
@@ -112,6 +112,12 @@ def clamp_page(
     return p, pp
 
 
+def _page_url(url_no_query: str, params: dict) -> str:
+    """One page url in a `Link` header, `<…>`-wrapped with its query percent-encoded."""
+    q = "&".join(f"{k}={quote(str(v), safe='')}" for k, v in params.items())
+    return f"<{url_no_query}?{q}>"
+
+
 def github_link_header(
     url_no_query: str, params: dict, page: int, per_page: int, total: int
 ) -> str | None:
@@ -140,11 +146,7 @@ def github_link_header(
         return None
 
     def link(p: int) -> str:
-        q = "&".join(
-            f"{k}={quote(str(v), safe='')}"
-            for k, v in {**params, "per_page": per_page, "page": p}.items()
-        )
-        return f"<{url_no_query}?{q}>"
+        return _page_url(url_no_query, {**params, "per_page": per_page, "page": p})
 
     parts = []
     if page > 1:
@@ -155,6 +157,50 @@ def github_link_header(
         parts.append(f'{link(last_page)}; rel="last"')
     if page > 1:
         parts.append(f'{link(1)}; rel="first"')
+    return ", ".join(parts) if parts else None
+
+
+def github_cursor_offset(page: int, per_page: int, after: str | None, before: str | None) -> int:
+    """Where a cursor-paged window starts: the cursor's offset, or the page's own when none came.
+
+    A cursor OVERRIDES ``page`` rather than adding to it — measured on api.github.com on 2026-09-04,
+    `/repos/psf/requests/issues?per_page=2&page=50` carrying page 1's `after` answers the rows
+    `page=2` answers, where `page=50` alone answers a different window. ``before`` names the window
+    that ENDS where the cursor points, which is how real's own `prev` walks back a page.
+    """
+    if after is not None:
+        return decode_cursor(after)
+    if before is not None:
+        return max(0, decode_cursor(before) - per_page)
+    return (page - 1) * per_page
+
+
+def github_cursor_link_header(
+    url_no_query: str, params: dict, page: int, per_page: int, total: int, offset: int
+) -> str | None:
+    """Build a Link header for a listing real pages by CURSOR rather than by offset.
+
+    `/repos/{owner}/{repo}/issues` is one; every other GitHub listing takes
+    :func:`github_link_header`. Measured on api.github.com on 2026-09-04 against a repository with
+    12 open issues at `per_page=5`: `next` and `prev` are the only rels — no `last`, no `first`,
+    ever — page 1 answers `next` alone, page 2 `next, prev` in that order, page 3 `prev` alone, and
+    pages 4 and 99 no header at all, as do those 12 rows at `per_page=100`.
+
+    Each url pairs an opaque cursor with the caller's own page ± 1. The page number is carried, not
+    computed: `?page=50` with page 1's cursor answers page 2's rows and links `page=51`. The cursor
+    is what names the window (see :func:`github_cursor_offset`).
+    """
+    if offset >= total:  # a window past the rows carries no header at all, not even `prev`
+        return None
+
+    def link(p: int, **cursor) -> str:
+        return _page_url(url_no_query, {**params, "per_page": per_page, "page": p, **cursor})
+
+    parts = []
+    if offset + per_page < total:
+        parts.append(f'{link(page + 1, after=encode_cursor(offset + per_page))}; rel="next"')
+    if offset > 0:
+        parts.append(f'{link(page - 1, before=encode_cursor(offset))}; rel="prev"')
     return ", ".join(parts) if parts else None
 
 
--- a/backlot/pagination.py
+++ b/backlot/pagination.py
@@ -128,6 +128,9 @@ def github_link_header(
     end on page 3 answers prev=3, last=3, first=1 for page 9. On the last page that holds rows there
     is no ``last`` — a client already there needs no pointer to it.
 
+    The rels come in one fixed order whichever of them a page has — prev, next, last, first — which
+    is the order real emits across all four of those pages.
+
     Values are percent-encoded because a param value is arbitrary text — a search `q` carries
     spaces and colons — and a URI cannot hold them raw. A client that follows the link has to
     arrive at the query it was paging, not at a truncated one.
@@ -144,14 +147,14 @@ def github_link_header(
         return f"<{url_no_query}?{q}>"
 
     parts = []
-    if page < last_page:
-        parts.append(f'{link(page + 1)}; rel="next"')
-        parts.append(f'{link(last_page)}; rel="last"')
     if page > 1:
         parts.append(f'{link(min(page - 1, last_page))}; rel="prev"')
-        parts.append(f'{link(1)}; rel="first"')
-    if page > last_page:
+    if page < last_page:
+        parts.append(f'{link(page + 1)}; rel="next"')
+    if page != last_page:  # short of the end, or past it
         parts.append(f'{link(last_page)}; rel="last"')
+    if page > 1:
+        parts.append(f'{link(1)}; rel="first"')
     return ", ".join(parts) if parts else None
 
 
--- a/backlot/pagination.py
+++ b/backlot/pagination.py
@@ -119,9 +119,22 @@ def _page_url(url_no_query: str, params: dict) -> str:
 
 
 def github_link_header(
-    url_no_query: str, params: dict, page: int, per_page: int, total: int
+    url_no_query: str,
+    params: dict,
+    page: int,
+    per_page: int,
+    total: int,
+    *,
+    per_page_sent: bool = True,
 ) -> str | None:
     """Build a Link header with rel=next/prev/first/last. ``params`` are extra query args.
+
+    ``per_page_sent`` says whether the CALLER named the page size; a size the handler defaulted is
+    left out of the urls, the rule a listing's filters already follow. Real omits it: `/tags` with
+    no query links `?page=2` and `?page=6` — six pages of 161 tags at its own default of 30 — where
+    `?per_page=1` links `per_page=1&page=2`, and search behaves the same (measured on
+    api.github.com on 2026-09-04). :func:`github_cursor_link_header` is the exception, and writes
+    the size either way.
 
     Returns ``None`` for a single page, which is what real sends there: no header at all, rather
     than one whose every rel points back at the page the caller is already holding. That holds
@@ -145,8 +158,10 @@ def github_link_header(
     if last_page <= 1:
         return None
 
+    size = {"per_page": per_page} if per_page_sent else {}
+
     def link(p: int) -> str:
-        return _page_url(url_no_query, {**params, "per_page": per_page, "page": p})
+        return _page_url(url_no_query, {**params, **size, "page": p})
 
     parts = []
     if page > 1:
--- a/backlot/pagination.py
+++ b/backlot/pagination.py
@@ -197,13 +197,7 @@ def github_cursor_offset(page: int, per_page: int, after: str | None, before: st
 
 
 def github_code_search_link_header(
-    url_no_query: str,
-    params: dict,
-    page: int,
-    per_page: int,
-    total: int,
-    *,
-    per_page_param: str | None = None,
+    url_no_query: str, params: dict, page: int, per_page: int, total: int
 ) -> str | None:
     """Build a Link header for `/search/code`, which pages by none of the listings' rules.
 
@@ -216,8 +210,10 @@ def github_code_search_link_header(
     * `prev` is the page before the one ASKED for even past the end: page 99 answers prev=98, where
       a listing clamps it to the last page that holds rows.
     * the rels come in the order next, prev, first, last.
-    * the size is written whether or not the caller named one — no `per_page` in the query still
-      links `per_page=30` — where a listing omits one the caller did not send.
+    * the size written is the one APPLIED, whether or not the caller named one and whatever they
+      spelt: no `per_page` in the query links `per_page=30`, `?per_page=500` on 1,808 hits links
+      `per_page=100` (its cap) and `?per_page=0` links `per_page=30`, where a listing omits an
+      unsent size and echoes a sent one verbatim. This cell it shares with the cursor listing.
 
     `/search/issues` shares none of that and pages by :func:`github_link_header`. A result set that
     fits one page carries no header, which is the one rule all three surfaces share.
@@ -225,10 +221,9 @@ def github_code_search_link_header(
     last_page = max(1, (total + per_page - 1) // per_page)
     if last_page <= 1:
         return None
-    size = per_page if per_page_param is None else per_page_param
 
     def link(p: int) -> str:
-        return _page_url(url_no_query, {**params, "per_page": size, "page": p})
+        return _page_url(url_no_query, {**params, "per_page": per_page, "page": p})
 
     parts = []
     if page < last_page:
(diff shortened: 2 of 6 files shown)


backlot/routers/github.py
```diff
@@ -373,7 +461,14 @@ def _search_paged(
...
     operation keeps the typed schema the MCP bridge reads.
     """
-    link = github_link_header(_base_url(request), {"q": q}, page, per_page, total)
+    link = github_link_header(
+        _page_base_url(request),
+        {"q": q},
+        page,
+        per_page,
+        total,
+        per_page_sent="per_page" in request.query_params,
```
nuri-yoo: Both search routes come through this helper, but the four rules it now applies were measured on `search/issues`. `search/code` answers differently on all four, and this PR moves two of them away from real.

Measured on api.github.com on 2026-09-04, `/search/code?q=repo:psf/requests+def&per_page=5` (45 hits, 9 pages):

```
page 1 : next=2, first=1, last=9            first is there on page 1
page 3 : next=4, prev=2, first=1, last=9    order is next, prev, first, last
page 99: prev=98, first=1, last=9           prev is page-1, not the last page holding rows
q with no per_page: <...&page=2&per_page=30>
```

`repo:kubernetes/kubernetes extension:md kubelet` (38 hits) answers the same four shapes. `/search/issues` answers all four the way this PR describes — I reproduced `prev=3, last=3, first=1` for page 9 of a 30-result query, and `?page=2` / `?page=29` with no `per_page` written anywhere — so the asymmetry is real's own.

Running this helper at both revisions:

```
search/code, per_page=5, total=45, page=99
  real (measured): prev=98, first=1, last=9
  e16e45e        : prev=98, first=1
  5a9777c        : prev=9,  last=9, first=1

per_page not sent
  e16e45e: <...?q=def&per_page=30&page=2>; rel="next", <...&per_page=30&page=2>; rel="last"
  5a9777c: <...?q=def&page=2>; rel="next", <...&page=2>; rel="last"
```

`last` is gained on this route and `prev` and `per_page` are lost. Backlot at head confirms it through the stack: `/github/search/code?q=line&per_page=1&page=99` links `prev=...&page=2` where real links `page=98`.

Two ways out that I can see. Measure `search/code` on its own and give it its own rels, or scope the four rules to the routes they were measured on and leave `search/code` where it was. If you'd rather keep this PR to what it measured, say so and I'll file the second as an issue.

backlot/pagination.py
```diff
@@ -128,23 +158,67 @@ def github_link_header(
...
+        return None
+
+    def link(p: int, **cursor) -> str:
+        return _page_url(url_no_query, {**params, "per_page": per_page, "page": p, **cursor})
+
+    parts = []
+    if offset + per_page < total:
+        parts.append(f'{link(page + 1, after=encode_cursor(offset + per_page))}; rel="next"')
+    if offset > 0:
+        parts.append(f'{link(page - 1, before=encode_cursor(offset))}; rel="prev"')
```
nuri-yoo: `page - 1` reaches 0 whenever a cursor puts the offset past the first window while `page` is 1. That is the shape a client gets by saving only the `after` cursor, which is the one durable thing this scheme hands it.

At head:

```
GET /github/repos/acme/diffable/issues?state=all&per_page=1&after=<cursor for offset 3>
  next: <...&per_page=1&page=2&after=bzo0>; rel="next"
  prev: <...&per_page=1&page=0&before=bzoz>; rel="prev"
```

Real emits no `page=0`. It also emits neither of those page numbers. The matrix on `/repos/psf/requests/issues?per_page=2`, measured 2026-09-04:

| request | page written into the page urls |
|---|---|
| `per_page=2`, no cursor | next `page=2` |
| `page=0` or `page=1`, no cursor | next `page=2` |
| `after=C` | none, on either rel |
| `page=1&after=C` | none |
| `page=0&after=C`, `page=abc&after=C` | none |
| `page=2&after=C` | next `page=3`, prev `page=1` |
| `page=50&after=C` | next `page=51`, prev `page=49` |

Real writes the number only where the caller asserted a page of 2 or more; with a cursor and no such assertion it drops `page` from both urls entirely. The docstring's "the caller's own page ± 1" holds for the `page=50` case it cites and for nothing below page 2.

The two cases the tests cover are `h(50, 5)` and `h(1, 0)` — page >= 2, and page 1 at offset 0. The combination that breaks is page 1 at offset > 0, and nothing reaches it: `github_cursor_link_header("http://x", {}, 1, 2, 12, 2)` returns a `page=0` url today and no test goes red.

backlot/routers/github.py
```diff
@@ -291,12 +297,94 @@ def _api_base(request: Request) -> str:
...
+    """
+    parts = path.strip("/").split("/", 3)
+    if len(parts) < 3 or parts[1] not in _ID_PATHS:
+        return None
+    tail = parts[3:]
+    if parts[1] == "organizations":
+        if str(synth.github_user_id(org)) != parts[2]:
+            return None
+        return "/".join(["/github/orgs", org, *tail])
+    for row in store.list_containers(conn, "github"):
```
nuri-yoo: `synth.github_user_id` is `1000 + digest % 9_000_000`, so it is not injective. Until this PR its output was a body field; here it becomes a routing key that has to round-trip, and this loop settles a collision by taking whichever name sorts first.

```
repo485 and repo4107 both hash to 7755679

  page url for 'repo485'  : /github/repositories/7755679/pulls
    canonical_id_path     -> /github/repos/acme/repo4107/pulls
  page url for 'repo4107' : /github/repositories/7755679/pulls
    canonical_id_path     -> /github/repos/acme/repo4107/pulls
```

A client walking `repo485` follows `next` and lands on `repo4107`'s second page, with nothing in the response to say so. Birthday bound over 9M values: 100 repos 0.05%, 500 repos 1.4%, 1,000 repos 5.4%, 5,000 repos 75%. The bundled corpus is far under that; a BYO one need not be.

`store.container_spelling` already settles the same question the other way — more than one candidate and it reports `None`, and the caller 404s. The same rule here turns a silent wrong answer into the 404 a name that names nothing already gets:

```python
    hits = [
        r["name"]
        for r in store.list_containers(conn, "github")
        if str(synth.github_user_id(r["name"])) == parts[2]
    ]
    # an id two repos share names neither, the way `container_spelling` reports None for a
    # spelling two of them answer to
    return "/".join(["/github/repos", org, hits[0], *tail]) if len(hits) == 1 else None
```

backlot/routers/github.py
```diff
@@ -291,12 +297,94 @@ def _api_base(request: Request) -> str:
...
-    link = github_link_header(_base_url(request), extra, page, per_page, rows_total)
-    headers = {"Link": link} if link else {}
-    return JSONResponse(body, headers=headers)
+    link = github_link_header(
+        _page_base_url(request),
+        extra,
+        page,
+        per_page,
+        rows_total,
+        per_page_sent="per_page" in request.query_params,
```
nuri-yoo: Not blocking, and narrower than the rule this PR lands: real does not write the size it applied, it writes back the one the caller sent. `psf/requests/tags`, 161 tags, measured 2026-09-04:

```
?per_page=abc -> <...?per_page=abc&page=2>, last=page=6   (it used 30)
?per_page=500 -> <...?per_page=500&page=2>, last=page=2   (it used 100)
?per_page=0   -> <...?per_page=0&page=2>,   last=page=6   (it used 30)
```

At head with `BACKLOT_DEFAULT_PAGE_SIZE=2`, all of `abc`, `0` and `-1` link `...&per_page=2&page=2`, and `per_page=500` would link `per_page=100`. `page` is not like this — it is recomputed per link and real recomputes it too (`?page=abc` still links `page=2`) — so `per_page` is the only one carrying the caller's own spelling.

`_absorb_page` exists so those values reach the handler, so this is reachable rather than theoretical, and the body already names the rule that covers it: `_echo` echoes what the caller sent. `request.query_params.get("per_page")` in place of the clamped int would do it here and in `_search_paged`.

backlot/main.py
```diff
@@ -164,6 +164,28 @@ async def echo_github_api_version(request: Request, call_next):
...
+    directly, so a 302 would be a shape a client only meets here. An id that names nothing is left
+    alone and 404s on the route it fails to match, the same answer a name that names nothing gets.
+    """
+    path = request.url.path
+    if path.startswith("/github/repositories/") or path.startswith("/github/organizations/"):
+        acl = getattr(request.app.state, "acl", None)
+        canonical = github.canonical_id_path(
+            request.app.state.conn, getattr(acl, "org_name", None) or get_settings().org_name, path
+        )
+        if canonical is not None:
```
nuri-yoo, line 186: Not blocking. The ACL comes through this path intact, so the docstring's claim holds:

```
admin  /github/repos/acme/vault/pulls      -> 200
admin  /github/repositories/3757977/pulls  -> 200
bob    /github/repos/acme/vault/pulls      -> 404
bob    /github/repositories/3757977/pulls  -> 404
```

The unauthenticated case splits, though:

```
noauth /github/repositories/373007/pulls   -> 401   (an id the corpus holds)
noauth /github/repositories/373008/pulls   -> 404   (one it does not)
noauth /github/repos/acme/diffable/pulls   -> 401
noauth /github/repos/acme/nosuchrepo/pulls -> 401
```

The name path puts credentials ahead of existence — `_validate_path_owner` calls `_require` first and says so in its docstring. This resolves before any of that, so a caller with no credentials can tell an id the corpus holds from one it does not, and since the id is `github_user_id(name)` a guessed name is confirmable that way. Real answers no 401 on these paths at all, so there is no shape to match here either.

Rewriting on the prefix whatever the id resolves to, leaving the id itself as the `{repo}` segment when it names nothing, would put both paths back in the same order.

nuri-yoo (commented):
Reviewed at 5a9777c in a worktree, with the four `Link` rules, the cursor listing and the 422 re-measured against api.github.com rather than read from the body.

Three are worth fixing before merge. Two of the three are one route each — `search/code` and the cursor listing's `page=0` — and the third is the id form's round-trip. Two smaller ones are marked inline as not blocking, and the prose below is the rest.

## Prose the approach change left behind

`_validate_path_owner`'s docstring (`backlot/routers/github.py:161-164`) still reads "The pagination `Link` header is the one url this does not reach: `_base_url` builds it from the path the request arrived on, deliberately, and real has no named url to match there — it paginates by numeric repository id instead". That last clause is what `_page_base_url` now does, so a reader of that paragraph learns the rule this PR reversed. It sits outside the diff, hence no inline.

`_base_url` (`backlot/routers/github.py:288`) has no callers left. `grep -rn '_base_url' backlot tests` at head returns the definition, that stale paragraph, `_page_base_url`'s own name, and an unrelated `self._base_url` in `integrations/fsspec.py`.

The body says "The branch is nine commits on current `main`"; `git rev-list --count e16e45e..HEAD` is 11. The body was edited at 05:40Z, after the last commit at 05:14Z, so this is the one line in it that did not keep up.

## Smaller things

`_ID_PATHS`'s values are never read — `canonical_id_path` writes `/github/orgs` and `/github/repos` itself. A set of the two keys would say the same thing.

`test_github_sub_resource_listings_page_and_answer_an_empty_page` has `from backlot import synth` in its body, and the same diff added `synth` to the module's imports.

`pull_review_comments` calls `_paged` twice with identical arguments. `_pr_files` is the only thing the early return saves, and `body` comes out `[]` from an empty `window` regardless.

`_page_base_url` calls `store.container_spelling` on every paged response, though `_canonical_path_repo` has already put the corpus spelling in `request.path_params["repo"]`. `canonical_id_path` hashes every container per id-path request, so a page walk pays that on each page.

`github_link_header`'s trailing `if parts else None` is unreachable: `last_page >= 2` means at least one of `page > 1` and `page < last_page` holds.

One `?page` gap survives in the baseline, `GET /repos/{}/{}/commits/{}?page`. Not a listing, so it does not contradict the body's claim.

## Verification

`uv run --all-extras pytest`: exit 0, 3 skipped (`mirage.core.google._client` twice, `llama_index.readers.hubspot`). `ruff check .` and `ruff format --check .` clean over 138 files. `scripts/gen_docs.py --check` exit 0. `backlot diff --source github`: 1236 divergences, 1235 acknowledged, 1 new — `GET /repos/{}/{}/stargazers/history`, which is #125. The 84 baseline lines removed are 14 entries, seven listings times two parameters, which matches.

## What I re-measured and could not break

The REST rel shapes at every position: page 1 `next, last`, a middle page `prev, next, last, first`, the last page holding rows `prev, first`, and past the end `prev` equal to `last`, then `last`, then `first`. `psf/requests/tags?per_page=1&page=9999` answers prev=162, last=162, first=1 in that order. A one-page listing stays header-less however far past it you ask (`pulls/7616/commits?page=99`).

Page urls in the id form with the resource urls in the same body keeping the login form, and `/user/repos` on its own path. The org id form also shows on `/orgs/kubernetes/repos` and `/orgs/kubernetes/members`.

`/repos/{o}/{r}/issues` carrying `next` and `prev` only, writing `per_page` even when the caller did not send it, and echoing `state`. An empty `after=` is offset 0 on both sides and real carries the page number along the same way Backlot does.

The 422 body matches byte for byte, `documentation_url` trailing slash included. A name that resolves beside one that does not is a 200 — 846 issues, 44 code hits, `incomplete_results: false` — and `search/code` alone on an unresolvable name is 200 with `total_count: 0` and the flag true.

`/statuses/{sha}` does page on real: kubernetes/kubernetes PR 141850's head carries two statuses and `?per_page=1` links `next, last`, in the id form.

`_qual_repos`'s visibility rule is `_visible_repos`'s, admin skip included, so the two agree on what the router calls visible.

backlot/routers/github.py
khj809: Went with (a). Re-measured `/search/code` at five positions (`repo:psf/requests def`, 45 hits at `per_page=5`) and it matches yours exactly: `first` and `last` on every page including the last one holding rows, `prev` = page asked for minus one even past the end, order next/prev/first/last, and the size written whether or not the caller sent one.

It has its own builder now, `github_code_search_link_header`, and `/search/issues` stays on the listings' rules — so no issue needed. The parametrized search test that asserted one shape for both routes is split by route.

nuri-yoo (commented):
Re-reviewed at e2a8301 in a fresh worktree. All five hold, and I checked each by reverting it rather than by reading the diff:

| reverted | suite |
|---|---|
| `code=True` on `search_code` | red — `test_github_search_pages_with_a_link_header[/github/search/code…]` |
| `named = hits[0] if len(hits) == 1 else parts[2]` | red — `test_github_an_id_two_repos_share_names_neither` |
| an unresolved id back to `return None` | red — `test_github_an_id_path_answers_in_the_order_the_named_path_does` |
| `per_page_param` back to the applied size | red — `test_github_link_header` |
| `link_page` back to the clamped `page` | green — inline |

Two things inline: the code-search builder writes the caller's spelling of `per_page` where real writes the size it applied, and the `page=0` fix has nothing guarding it. Neither is large.

## What I re-measured against api.github.com

`/search/code` at the position I had not taken: page 9, the last that holds rows, answers `prev=8, first=1, last=9` — `first` and `last` are there, where a listing sends neither. A result set that fits one page carries no header, so that rule does hold across all three surfaces. Head answers both shapes, and page 99 answers `prev=98, first=1, last=2` on the fixture.

The claim I expected to have to argue with is the one that held cleanest: the cursor listing really does write the size APPLIED where `/tags` writes the caller's spelling, on the same host on the same day.

```
/repos/psf/requests/issues?per_page=abc -> <...?per_page=30&after=…&page=2>
/repos/psf/requests/issues?per_page=500 -> <...?per_page=100&after=…&page=2>
/repos/psf/requests/issues?per_page=0   -> <...?per_page=30&after=…&page=2>
/repos/psf/requests/tags?per_page=abc   -> <...?per_page=abc&page=2>
```

The page-number rule answers the way the docstring says at every position I tried, including the two the old code got wrong: a cursor with no `page`, and a cursor with `page=1`, both answer no `page` in either url.

## Prose

Every item from the last round is gone — `_base_url` deleted, `_ID_PATHS` a set, the test-local `import synth` dropped, `pull_review_comments` down to one `_paged` call, `_page_base_url` reading `path_params`, the unreachable `if parts else None` replaced by the comment that explains why it was unreachable, and the body rewritten around three rule sets with the commit count now matching (`git rev-list --count e16e45e..e2a8301` is 12).

One line in the rewritten `_validate_path_owner` docstring: "`_page_base_url` reads the spelling this dependency settled rather than the one the request arrived with". For the org half that is this dependency; for the repository half `_page_base_url` reads `path_params["repo"]`, which `_canonical_path_repo` settles two functions down. Naming that one instead would keep the paragraph exact.

## Gates

`uv run --all-extras pytest`: exit 0, 3 skipped, the same three extras as before. `ruff check .` and `ruff format --check .` clean over 138 files. `scripts/gen_docs.py --check` exit 0. `backlot diff --source github` unchanged at 1236 divergences, 1235 acknowledged, 1 new, still `GET /repos/{}/{}/stargazers/history`.

backlot/pagination.py
```diff
@@ -128,20 +163,124 @@ def github_link_header(
...
+    * the size is written whether or not the caller named one — no `per_page` in the query still
+      links `per_page=30` — where a listing omits one the caller did not send.
+
+    `/search/issues` shares none of that and pages by :func:`github_link_header`. A result set that
+    fits one page carries no header, which is the one rule all three surfaces share.
+    """
+    last_page = max(1, (total + per_page - 1) // per_page)
+    if last_page <= 1:
+        return None
+    size = per_page if per_page_param is None else per_page_param
```
nuri-yoo: `/search/code` writes the size it APPLIED, not the caller's spelling — it is the one search surface that does not carry it. Measured on api.github.com on 2026-09-04:

```
/search/code?q=repo:kubernetes/kubernetes+kubelet&per_page=500   1808 hits, cap 100
  -> <...&page=2&per_page=100>; rel="next"
/search/code?q=repo:psf/requests+def&per_page=0                  45 hits
  -> <...&page=2&per_page=30>;  rel="next"

/search/issues?q=repo:psf/requests+timeout&per_page=500          same day
  -> <...&per_page=500&page=2>; rel="next"
/repos/psf/requests/tags?per_page=abc
  -> <...?per_page=abc&page=2>; rel="next"
```

So the third rule set splits from the first on this cell too, and it lands where the cursor listing already is — the size applied, both ways. At head:

```
per_page_param=None   -> per_page=100   (right)
per_page_param="500"  -> per_page=500   (real: 100)
per_page_param="0"    -> per_page=0     (real: 30)
```

5a9777c had this right by accident, since it wrote the clamped int for a caller who sent one. `size = per_page`, and the kwarg comes off this builder.

The test meant to hold it cannot: `assert "per_page=5" in h(1, per_page_param="5")` runs with `per_page=5`, so the applied size and the caller's spelling are the same string and both rules pass. Deleting the whole `per_page_param` branch from this function leaves `tests/test_github.py` and `tests/test_pagination.py` green. `h(1, per_page=30, per_page_param="500")` would tell them apart.

One more, not worth a rule of its own: `/search/code?per_page=abc` is a 400 on real, `Failed to deserialize query string: per_page: invalid digit found in string`. It is the only GitHub surface I found that refuses what `_absorb_page` absorbs — `/search/issues?per_page=abc` and `/tags?per_page=abc` are both 200. Whatever this builder does with an unparseable size describes a request real never answers.

backlot/routers/github.py
```diff
@@ -805,15 +978,27 @@ async def list_issues(
...
+    start = github_cursor_offset(page, per_page, q.get("after"), q.get("before"))
     rows = all_rows[start : start + per_page]
+    # A cursor and no page of 2 or more asserted beside it is the one case real writes no page
+    # number, so `None` says exactly that (see `github_cursor_link_header`). Without a cursor the
+    # page is the caller's own position and real writes it.
+    cursored = "after" in q or "before" in q
+    if not cursored:
+        link_page = page
+    else:
+        link_page = asserted if asserted and asserted >= 2 else None
```
nuri-yoo, line 1000: This is right, and nothing holds it. Reverting the one argument — `link_page` back to `page` at the `_paged_by_cursor` call — leaves `tests/test_github.py` and `tests/test_pagination.py` fully green, and puts `page=0` straight back:

```
GET /github/repos/acme/diffable/issues?state=all&per_page=1&after=<offset 3>
  e2a8301 : <...&per_page=1&after=bzo0>; rel="next", <...&per_page=1&before=bzoz>; rel="prev"
  reverted: <...&per_page=1&page=2&after=bzo0>; rel="next", <...&per_page=1&page=0&before=bzoz>; rel="prev"
```

`github_cursor_link_header` itself is covered — `h(None, 3)` in the unit test pins the helper. What is uncovered is this line: which requests get `None`. The existing route test never reaches it, because it follows `next` and from there a page of 2 or more is always in play.

Three requests would cover the branch, all against the same listing at `per_page=1`: `after=C` with no `page` and `page=1&after=C` both answering no `page` in either url, and `page=3&after=C` answering next=4/prev=2. I ran those three plus the no-cursor case against head and they answer exactly that.

backlot/routers/github.py
nuri-yoo: Reproduced (a) at head and it matches at every position I have measured, page 9 included — I had not taken that one, and `first` and `last` are indeed both there on the last page holding rows. Splitting the route test by shape is the right place for it.

One cell of the new builder is off, though, and it is the one your table left as just "written": `/search/code` writes the size it APPLIED, never the caller's spelling. New inline on `github_code_search_link_header` with the two measurements.

backlot/pagination.py
khj809: Fixed — the kwarg is off the builder and it writes the applied size, so this cell sits with the cursor listing.

You were right that the test could not tell the two apart. Replaced with a route-level one that overrides the cap so the applied size differs from the value sent: `per_page=500` links `per_page=2` on `/search/code` and `per_page=500` on `/search/issues`. It fails at e2a8301. Noted the `?per_page=abc` 400 as well — nothing here leans on that shape.

backlot/routers/github.py
khj809: Covered — added your three requests as a route test on that listing at `per_page=1`: `after=C` and `page=1&after=C` answer no `page` in either url, `page=3&after=C` answers next=4/prev=2, and the no-cursor case next=2.

Checked it the way you did: reverting the argument turns it red with `{'next': '2', 'prev': '0'}`.

nuri-yoo (approved):
LGTM at 313c90b. Everything below was checked at dd820d0; the merge on top of it is #133 landing in `main` and touches `graphql/linear.graphql`, `graphql/linear_filters.py` and `tests/test_linear.py` only, with `git diff dd820d0..313c90b` empty across the github router, `pagination.py`, `main.py`, `errors/github.py`, the baseline and both test files. Every gate re-runs the same at the head.

Both of your mutation results reproduce exactly. Running dd820d0's tests against e2a8301's `backlot/` turns one test red and only that one, `test_github_only_the_offset_surfaces_write_the_size_the_caller_spelt`; reverting the `link_page` argument gives `assert {'next': '2', 'prev': '0'} == {'next': None, 'prev': None}`, the same pair you quoted.

The three cells the new builder rests on re-measure as its docstring has them — `/search/code` at `per_page=500` on 1,808 hits links `per_page=100`, at `per_page=0` links `per_page=30`, and `/search/issues` at `per_page=500` links `per_page=500`.

At route level the page-number branch also answers real's way at the two positions the new test does not reach: `?page=0&after=C` and `?page=abc&after=C` both write no `page` in either url, which is what real answers for them. The id paths now answer 401 for an id the corpus holds and one it does not alike, so the order matches the named path.

`uv run --all-extras pytest` exit 0 with the same three extras skipped, `ruff check` and `ruff format --check` clean over 138 files, `scripts/gen_docs.py --check` exit 0, and `backlot diff --source github` still 1236 divergences, 1235 acknowledged, 1 new — `GET /repos/{}/{}/stargazers/history`, unchanged across the main merge.

Every prose item from both rounds is gone, the `_validate_path_owner` line included.

One thing for after this PR, not for it — filed as #137. `_absorb_page`'s "Real GitHub refuses no value here" is scoped to an issue listing in its own parenthesis, and `/search/code` is a 400 for `per_page=abc`, `-1` and `1.5` and for the same three `page` values, in `text/plain` rather than this router's envelope. `errors/github.py::validation_body` states the claim a second time and is the one that would decide the answer if the absorption came off, so the issue names both. It reproduces on `main` at 79ccdac, so none of it is this branch's.

PR #132, opened 2026-09-04
github: the star history real serves for a starless repo is an acknowledged gap

### LINKED ISSUE #125: Fidelity: github diverges from its vendor (first seen 2026-09-10)
`backlot diff --source github` reports divergences the baseline does not acknowledge. Each is either a bug in Backlot or a deliberate gap; a gap gets a note and an entry via `backlot diff --source github --update-baseline`. This body is the latest scheduled run's report, rewritten in place. Closing this issue keeps it as the account of what was triaged; a run that still diverges opens a new one.

```
🔍 github · https://raw.githubusercontent.com/github/rest-api-description/main/descriptions/api.github.com/api.github.com.json
   1240 divergence(s) · 1236 acknowledged · 4 new

⚠️  gap (4) — the vendor has surface Backlot does not
     missing_operation  GET /orgs/{}/code-scanning/ai-scan
       the vendor serves it; Backlot does not
     missing_operation  GET /repos/{}/{}/code-scanning/ai-scan
       the vendor serves it; Backlot does not
     missing_operation  PATCH /orgs/{}/code-scanning/ai-scan
       the vendor serves it; Backlot does not
     missing_operation  PATCH /repos/{}/{}/code-scanning/ai-scan
       the vendor serves it; Backlot does not

   4 new. Fix them, or run --update-baseline to acknowledge the gaps.
```

khj809:
`GET /repos/{owner}/{repo}/stargazers/history` is an acknowledged gap in `backlot/fidelity/baseline/github.json`. The reasoning that covers the whole stargazer family sits on `GET /repos/{owner}/{repo}/stargazers`, where it is above the entries it explains, and the history entry carries the measurement that makes it the route where a stand-in would break a client. No code change.

Closes #125.

## Why an acknowledgement rather than an empty list

No corpus states who starred a repository, which is why `/stargazers`, `/stargazers/count` and `/subscribers` are already gaps in this same file, and why the repo record omits `stargazers_count` rather than making one up. Answering this one route with 200 while its siblings 404 would contradict that.

Answering `[]` would not be a smaller version of the real answer, it would break a client that the absence does not. Real never answers an empty FIRST page for a repository that exists, because page 1 always carries at least the creation week. Measured against api.github.com on 2026-09-04 on repositories with zero stars: one row per calendar week from the creation week to now, every row `total: 0` with `days: [0, 0, 0, 0, 0, 0, 0]`. One created that same day answered a single row, whose week is the Sunday just past; one created 2026-05-02 answered 19, its oldest row the Sunday of that week. GitHub's published spec says the same in prose — weeks without stars carry zero counts, and pages move backward toward the repository's creation week — and caps `per_page` at 30. So `history[0]["week"]` is safe against real and unsafe against a Backlot that hands back `[]`, which is the breaking class by our own rule rather than a gap.

Real does answer `[]` past the last page — `per_page=10&page=3` of that 19-week repository, `page=2` of the one-week one — and 404s an unknown repository. The note says so, so that whoever serves this route later is not told by this file that `[]` is forbidden everywhere.

Serving it faithfully would need no schema change: the repo record carries `created_at`, so zero-count weeks could be emitted from that week to now, `per_page` capped at 30, paging backward, which is exactly what real answers for a starless repository. But that asserts the repository has zero stars, and then `/stargazers` → `[]`, `/stargazers/count` → `{"count": 0}` and `stargazers_count: 0` on the repo record have to follow. That is a decision about the whole stargazer family, and it stays open.

## What changed

One acknowledgement and two notes, plus `measured` moving to the day of the run. `backlot diff --source github` reports 1236 divergences, 1236 acknowledged, nothing new, and leaves no stale acknowledgement behind. Running `--update-baseline` again rewrites the file byte for byte, so both notes survive a later scheduled acknowledgement rather than being dropped by it.

## Verification

`uv run pytest` with all extras: 2339 passed, 3 skipped. `ruff check` and `ruff format --check` clean, and `scripts/gen_docs.py` reports both generated files current.

### DIFF
--- a/backlot/fidelity/baseline/github.json
+++ b/backlot/fidelity/baseline/github.json
@@ -4221,7 +4221,8 @@
       "kind": "missing_operation",
       "severity": "gap",
       "path": "GET /repos/{}/{}/stargazers",
-      "detail": "the vendor serves it; Backlot does not"
+      "detail": "the vendor serves it; Backlot does not",
+      "note": "No corpus states who starred a repository, so the whole stargazer family is a gap \u2014 this listing, `/stargazers/count`, `/stargazers/history` and `/subscribers` \u2014 and the repo record omits `stargazers_count` rather than inventing one. Serving any one of them commits Backlot to a star count: the history could be emitted from the record's own `created_at` with no schema change, but then this listing owes `[]`, the count owes `0` and the record owes `stargazers_count: 0`. That is a decision about the family, not about one route."
     },
     {
       "kind": "missing_operation",
@@ -4234,7 +4235,7 @@
       "severity": "gap",
       "path": "GET /repos/{}/{}/stargazers/history",
       "detail": "the vendor serves it; Backlot does not",
-      "note": "No corpus states who starred a repository, so `/stargazers`, `/stargazers/count` and `/subscribers` are gaps beside this one and the repo record omits `stargazers_count` rather than inventing it. Answering `[]` would be a shape real never has: measured against api.github.com on 2026-09-04, two repositories with zero stars each answered one row per calendar week from their creation week to now \u2014 2 rows for one created a week earlier, 19 for one created four months earlier \u2014 every row `total: 0` with `days: [0, 0, 0, 0, 0, 0, 0]`. GitHub's published spec says the same: weeks without stars carry zero counts, and pages move backward toward the creation week."
+      "note": "The family is a gap for the reason on `GET /repos/{}/{}/stargazers`; this is the route where a stand-in would break a client rather than merely be absent. Real never answers an empty FIRST page for a repository that exists, because page 1 always carries at least the creation week, so `history[0][\"week\"]` is safe against real and unsafe against a Backlot that answers `[]`. Measured against api.github.com on 2026-09-04 on repositories with zero stars: one row per calendar week from the creation week to now, every row `total: 0` with `days: [0, 0, 0, 0, 0, 0, 0]` \u2014 one created that same day answered a single row, one created 2026-05-02 answered 19, its oldest row the Sunday of that week. Past the last page real does answer `[]`, and an unknown repository 404s. GitHub's published spec says the same, capping `per_page` at 30 and paging backward toward the creation week."
     },
     {
       "kind": "missing_operation",
--- a/backlot/fidelity/baseline/github.json
+++ b/backlot/fidelity/baseline/github.json
@@ -1,7 +1,7 @@
 {
   "source": "github",
   "endpoint": "https://raw.githubusercontent.com/github/rest-api-description/main/descriptions/api.github.com/api.github.com.json",
-  "measured": "2026-09-02",
+  "measured": "2026-09-04",
   "acknowledged": [
     {
       "kind": "extra_operation",
@@ -4228,6 +4228,13 @@
       "severity": "gap",
       "path": "GET /repos/{}/{}/stargazers/count",
       "detail": "the vendor serves it; Backlot does not"
+    },
+    {
+      "kind": "missing_operation",
+      "severity": "gap",
+      "path": "GET /repos/{}/{}/stargazers/history",
+      "detail": "the vendor serves it; Backlot does not",
+      "note": "No corpus states who starred a repository, so `/stargazers`, `/stargazers/count` and `/subscribers` are gaps beside this one and the repo record omits `stargazers_count` rather than inventing it. Answering `[]` would be a shape real never has: measured against api.github.com on 2026-09-04, two repositories with zero stars each answered one row per calendar week from their creation week to now \u2014 2 rows for one created a week earlier, 19 for one created four months earlier \u2014 every row `total: 0` with `days: [0, 0, 0, 0, 0, 0, 0]`. GitHub's published spec says the same: weeks without stars carry zero counts, and pages move backward toward the creation week."
     },
     {
       "kind": "missing_operation",


backlot/fidelity/baseline/github.json
```diff
@@ -4229,6 +4229,13 @@
       "path": "GET /repos/{}/{}/stargazers/count",
       "detail": "the vendor serves it; Backlot does not"
     },
+    {
+      "kind": "missing_operation",
+      "severity": "gap",
+      "path": "GET /repos/{}/{}/stargazers/history",
+      "detail": "the vendor serves it; Backlot does not",
+      "note": "No corpus states who starred a repository, so `/stargazers`, `/stargazers/count` and `/subscribers` are gaps beside this one and the repo record omits `stargazers_count` rather than inventing it. Answering `[]` would be a shape real never has: measured against api.github.com on 2026-09-04, two repositories with zero stars each answered one row per calendar week from their creation week to now \u2014 2 rows for one created a week earlier, 19 for one created four months earlier \u2014 every row `total: 0` with `days: [0, 0, 0, 0, 0, 0, 0]`. GitHub's published spec says the same: weeks without stars carry zero counts, and pages move backward toward the creation week."
```
nuri-yoo: ``Answering `[]` would be a shape real never has`` is false, and it came out of my comment on #125 rather than anything you wrote, so treat this as my correction. Real answers `[]` for any page past the last one. Measured against api.github.com on 2026-09-04:

```
per_page=30&page=2   (the 2-row repo)    -> []
per_page=1&page=3    (the 2-row repo)    -> []
per_page=10          (the 19-row repo)   -> 10 rows, Link rel="last" is page=2
per_page=10&page=2   (the 19-row repo)   -> 9 rows
per_page=10&page=3   (the 19-row repo)   -> []
```

The argument is unaffected, but the claim carrying it has to be narrower. What real never answers is an empty FIRST page for a repository that exists, because page 1 always carries at least the creation week. That is what makes `history[0]["week"]` safe against real and unsafe against a Backlot that hands back `[]`.

The wording matters past precision, because this note is what someone reads before implementing the route. Told that `[]` is a shape real never has, whoever serves it faithfully later has to answer `[]` on page 3 of a repository whose history is 19 weeks long, and will find this file forbidding it.

One mechanical thing if you edit the string by hand: `Baseline.write` serialises with `json.dumps` defaults, so the em dashes live in the file as `\u2014`. A literal one is still valid JSON but no longer matches what the tool writes, and the next `--update-baseline` rewrites the escaping. I checked both directions: hand-editing with literal em dashes and then running `--update-baseline` once restores the escapes and the file goes back to byte-identical, so a single re-run after the edit is enough.

nuri-yoo (commented):
Reproduced the description against the head, 5b215bf.

`backlot diff --source github`: 1236 divergences, 1236 acknowledged, 0 new, exit 0, no stale section. On the base commit the same command exits 1 with exactly this one gap, so dropping the entry turns the run red. Re-running with `--update-baseline` leaves the file byte for byte identical, and `test_rewriting_a_baseline_keeps_its_notes_and_takes_the_new_identity` is what holds that, so the note does survive a later scheduled acknowledgement. `uv run pytest` with all extras: exit 0, 2342 collected, 3 skipped (`llama_index.readers.hubspot`, and `mirage.core.google._client` twice, none of them fidelity), which is the 2339 the description claims. `ruff check`, `ruff format --check` and `gen_docs.py --check` all exit 0.

The measurement holds. api.github.com on 2026-09-04 with a token: the repository created 2026-08-26 answers 2 rows and the one created 2026-05-02 answers 19, every row `total: 0` with `days: [0, 0, 0, 0, 0, 0, 0]`, newest week `1788048000` and oldest `1777161600`, which is the Sunday of the creation week. The spec transcription is accurate: the OpenAPI description of `activity/get-stargazer-history-for-repo` says "Pages move backward toward the repository's creation week" and "Weeks without stars contain zero counts", and caps `per_page` at 30. `/stargazers`, `/stargazers/count` and `/subscribers` are gaps in this same file, Backlot 404s all four in GitHub's error envelope, and `test_github.py` asserts the repo record omits `stargazers_count`. `created_at` is on the record, so the deferred design needs no schema change, as you say.

One sentence in the note is false, and it came from me rather than from you. Inline.

Three smaller things, none of them blocking.

This note is now the only place the stargazer-family reasoning lives, and it hangs off the newest member of the family. `/stargazers`, `/stargazers/count` and `/subscribers` carry no note at all, so someone asking why the list is a gap has to happen upon the history entry to find the answer. Putting the first sentence on `/stargazers` and leaving the history-specific measurement here would place it where it gets read, if you think entries this PR did not add are worth touching.

The two row counts age. `2` and `19` each grow by one every Sunday, and both are derivable from the invariant the same sentence already states: one row per week from the creation week to now, all zero. The counts are the only part of the note that stops being true.

The description names both repositories where the note deliberately does not. The argument needs only "a repository with no stars, created N weeks ago", and these are two people's personal repositories.

Last, a qualifier on "the note survives a later scheduled acknowledgement". True for re-acknowledgement, not for resolution: if someone serves this route later, including as `[]`, the finding disappears, `resolved()` reports the entry as stale, and `.github/workflows/fidelity.yml` acts only on exit 1. The run stays green, prints "drop from the baseline" into a log, and files nothing on #125. So the decision this note records is held by prose alone. That is the tool's shape rather than this PR's, and I can file it separately if you want it tracked.

nuri-yoo (approved):
Approving on 5b215bf. The entry itself is right and everything in the description reproduced, so the wording fix on the note is not worth holding the merge for.

The one inline finding stands as a follow-up: real does answer `[]` past the last page, so the sentence has to narrow to the first page of a repository that exists. Fix it here if you would rather not merge a note with a false sentence in it, or in whatever next touches this file. Either way `--update-baseline` keeps the note, so it will not be lost.

backlot/fidelity/baseline/github.json
khj809: Reproduced before changing the note. api.github.com on 2026-09-04: the one-week repository answers `[]` at `per_page=30&page=2` and at `per_page=1&page=3`, and the nineteen-week one answers 10, 9 and `[]` at `per_page=10` pages 1 to 3 — your five rows, row for row. The narrower claim holds in both directions too: two repositories created the day of the measurement each answer a single row on page 1, the Sunday just past, and an unknown repository 404s rather than answering `[]`.

The note now says real never answers an empty FIRST page for a repository that exists, and says outright that past the last page it does answer `[]`, so the file no longer forbids the page-3 answer to whoever serves this route later.

The three smaller ones are taken as well.

The family reasoning moved to `GET /repos/{}/{}/stargazers`, where it sits directly above `/stargazers/count` and `/stargazers/history` and names `/subscribers` too. The history note keeps only what is specific to it and points at that entry. The deferred design went with it — emit from the record's `created_at`, and then the listing owes `[]`, the count owes `0` and the record owes `stargazers_count: 0` — because that is the family's decision rather than this route's.

The counts no longer float free of what derives them. The note states the invariant, then one repository created the day of the measurement answering a single row and one created 2026-05-02 answering 19, its oldest row the Sunday of that week. Each is pinned by a creation date and a measurement date, so neither goes stale the way a bare `2` and `19` would.

The description no longer names either repository; both are identified by creation date.

Mechanically: `--update-baseline` ran once after the hand edit, and a further run rewrites the file byte for byte, so both notes store their em dashes escaped, the way the tool writes them.

The `resolved()` gap is worth tracking separately — please do file it. A note that records a decision is only load-bearing if dropping the entry turns the run red, and today it does not.

PR #133, opened 2026-09-04
linear: `or` on a relation filter reads a branch's keys as alternatives and an empty branch as the issues without the relation (#128)

### LINKED ISSUE #128: linear: or inside a relation filter ORs the keys of one branch, and an empty or null-only branch beside a comparator adds the issues without the relation

nuri-yoo:
Closes #128. Rebased onto `main` at `79ccdac`, the squash of #129: the `or` rule builds on the `null: true` and `and`-merge rules that PR put into `_sub_filter`.

**What Linear does.** An `or` inside a nullable relation filter (`project`, `assignee`, `creator`) is not the union of its branches, and the difference is not one rule but three, all measured against api.linear.app on 2026-09-04 over four issues: `BRE-1` in a throwaway project `zz-pa` and assigned to the viewer, `BRE-2` in `zz-pb`, `BRE-3` and `BRE-4` in no project. First, the keys of one `or` branch are alternatives: `{or: [{name: {eq: "zz-pa"}, id: {eq: <zz-pb>}}]}` is both issues with a project where `{name: {eq: "zz-pa"}, id: {eq: <zz-pb>}}` is none. This is Linear's `or` everywhere, not a relation quirk: `issues(filter: {or: [{number: {eq: 1}, title: {eq: <BRE-2's title>}}]})` answers both issues and `projects(filter: {or: [{name: {eq: "zz-pa"}, id: {eq: <zz-pb>}}]})` both projects. Second, a branch that says nothing about the related row beside one that does adds the issues without the relation: `{or: [{null: false}, {name: {eq: "nobody"}}]}`, `{or: [{}, {name: {eq: "nobody"}}]}` and `{or: [{null: true}, {name: {eq: "nobody"}}]}` are each the two without a project, while on their own `{or: [{null: false}, {null: false}]}` is the two with and `{or: [{}, {}]}` all four. Third, every branch carrying `null: true` is the issues without, comparators unread: `{or: [{null: true, name: {eq: "zz-pa"}}, {null: true}]}` is the two without, but `{or: [{null: true, name: {eq: "zz-pa"}}, {name: {eq: "nobody"}}]}` is those plus `zz-pa`'s.

**The mechanism.** Every one of the 458 measured rows is one compile procedure, and `_sub_filter` is now written as it:

- FLAGS. What the object says about the relation's existence is its own `null`; failing that, what its `and` branches say, merged with true winning over false, which is #124's rule for direct `null` keys applied to every branch: a branch says its own `null`, and a branch that is an `or` says true when every alternative carries `null: true` and false when it holds a bare `{null: false}` and no `null: true`. The object's own `or` contributes only a `null` common to all of its alternatives.
- ASSEMBLE. True is the issues without the relation, nothing else read. False drops every `null` key below the object and is the relation present AND the rest. Nothing decided is the rest, ORed with the issues without the relation when a `null: true` is anywhere below.
- LISTS. Each `or` is read against the related row: the keys of one alternative are alternatives, `null` keys are not read, an alternative with no operator-bearing comparator makes the list say nothing, a list whose every alternative says nothing about the row says nothing (an empty list is the relation present), and otherwise an alternative that says nothing about the row (`{}`, `{and: []}`, a bare `{null: *}`, or an `or` holding one) is the list's missing-relation alternative. A list does not need to know where it came from.

What looked like a position-dependent reading of a bare `{null: false}` (a missing-relation alternative in the object's own `or`, a requirement in an `or` lifted from an `and` branch) is FLAGS reading a disjunction that holds one as asserting `null: false`, and the object's own `null` dropping everything below it. The two readings that remain the vendor's rather than derivable are that and-merge reading and the keys of an alternative being alternatives; the rest is the null-merge #124 already measured, plus a row-side OR with knex-style elision of empty groups.

**How it was found.** 406 relation filters in eleven rounds (136 single branches and pairs, 93 nested shapes, then 53, 16, 4, 36, 31, 8, 13, 10 and 6), plus 11 on `issues` and `projects` for the branch-key rule; with the 24 rows of #128's first table and the 28 of #124's last round the relation rows are 458. Every round from the third was predicted from the rule before it was sent, and the misses fixed corners until the procedure fit every row. It is spelled out on `_sub_filter`; the short form is on the two nullable filters' `or` in the SDL.

**Where the mechanism was looked for.** Beyond the measurements: the field descriptions Linear serves for `or`, `and` and `null` on the nullable filters ("Compound filters, one of which need to be matched by the project", "Filter based on the existence of the relation"), which say nothing about the cases above; the validation errors the API returns for bad input, which are class-validator's shape and carry no query text; the filtering page of the developer docs and the linear/linear tracker, neither of which mentions relation `or`s; and the web client's bundle. That last one has the only public piece of Linear's filter code, the ModelMatcher the client runs over synced models: its `or` on a missing relation is "some branch says `null: true`" and its `and` "every branch does", the scan reading of the first two clauses. It ANDs the keys of a branch and reads `{and: [{null: true}, {name: {eq: X}}]}` as none where the API answers the issues without the relation, so it is a second implementation, not the API's compiler; `_sub_filter`'s docstring says so and keeps the clauses as measured.

**What changed.** `_sub_filter` is the procedure in that order: FLAGS from the object's own `null`, its `and` branches and its own `or`; ASSEMBLE, which returns `IS NULL`, or drops every `null` below and prefixes `IS NOT NULL AND`, or prefixes `IS NULL OR` when a `null: true` is below; and LISTS through `_relation_or`, which no longer needs to know whether the `or` is the object's own, with `_related_row_or` / `_related_row_and` rendering the alternatives against the related row with an alternative's keys ORed. `_issue_parts` reads an `or` branch with n keys as n branches, so `IssueFilter`'s `or` matches Linear too; the vacuous-branch rule it already had is unchanged, and no existing cell put two keys in one branch. `NullableUserFilter.or` and `NullableProjectFilter.or` carry a description that says what a bare `{null: false}` branch means in each position. `_labels_branches` and `compile_comment_filter` are not touched: the label collection has its own `some`/`every` semantics, and comment filters were not measured.

**Tests.** `_RELATION_OR_CELLS`, 143 measured answers on a fixture of the same shape as the workspace (names included), grouped by step; one test for the `IssueFilter`-level key rule, `priority` standing in for `number`, which `IssueFilter` here does not carry, as its docstring says; and two from review, pinning that `_RELATION_NEGATIVE_OPS` is not the labels rule's `_NEGATIVE_OPS` and that an unmapped nested field is refused with or without an operator. With `linear_filters.py` at `79ccdac` (main) and these tests kept: `97 failed, 49 passed`; at `35ef0ad`, the commit before the procedure: `6 failed, 137 passed` over the cells.

**Verification.**

```
2322 passed
```

**Not in this PR.** The fidelity baseline is schema-shaped and cannot acknowledge behaviour rows, so nothing there changes; `fidelity/graphql_diff.py` introspects with `descriptions=False`, so the two new `or` descriptions are invisible to it as well. `or` on `CommentFilter` keeps ANDing a branch's keys; that is now measured and filed as #136, with `_issue_parts`'s flattening as the shape to follow. `or` inside `IssueLabelCollectionFilter` stays as it is: its `some` / `every` polarity is tangled into the same code and nothing about its branch keys is measured.

### DIFF
--- a/backlot/graphql/linear.graphql
+++ b/backlot/graphql/linear.graphql
@@ -260,6 +260,10 @@ input NullableUserFilter {
   (measured 2026-09-04)."""
   null: Boolean
   and: [NullableUserFilter!]
+  """The keys of one branch are alternatives, as on every `or` here. A branch that says nothing about
+  the related row (`{}`, `{null: true}`, a bare `{null: false}`) beside one that does adds the issues
+  without the relation; every branch carrying `null: true` is the issues without it, comparators
+  unread (measured 2026-09-04)."""
   or: [NullableUserFilter!]
 }
 
@@ -279,6 +283,10 @@ input NullableProjectFilter {
   (measured 2026-09-04)."""
   null: Boolean
   and: [NullableProjectFilter!]
+  """The keys of one branch are alternatives, as on every `or` here. A branch that says nothing about
+  the related row (`{}`, `{null: true}`, a bare `{null: false}`) beside one that does adds the issues
+  without the relation; every branch carrying `null: true` is the issues without it, comparators
+  unread (measured 2026-09-04)."""
   or: [NullableProjectFilter!]
 }
 
--- a/backlot/graphql/linear.graphql
+++ b/backlot/graphql/linear.graphql
@@ -261,9 +261,10 @@ input NullableUserFilter {
   null: Boolean
   and: [NullableUserFilter!]
   """The keys of one branch are alternatives, as on every `or` here. A branch that says nothing about
-  the related row (`{}`, `{null: true}`, a bare `{null: false}`) beside one that does adds the issues
-  without the relation; every branch carrying `null: true` is the issues without it, comparators
-  unread (measured 2026-09-04)."""
+  the related row (`{}`, `{null: true}`) beside one that does adds the issues without the relation,
+  and every branch carrying `null: true` is the issues without it, comparators unread. A bare
+  `{null: false}` branch is that alternative in the object's own `or`, and the object's `null: false`
+  when the `or` sits in an `and` branch (measured 2026-09-04)."""
   or: [NullableUserFilter!]
 }
 
@@ -284,9 +285,10 @@ input NullableProjectFilter {
   null: Boolean
   and: [NullableProjectFilter!]
   """The keys of one branch are alternatives, as on every `or` here. A branch that says nothing about
-  the related row (`{}`, `{null: true}`, a bare `{null: false}`) beside one that does adds the issues
-  without the relation; every branch carrying `null: true` is the issues without it, comparators
-  unread (measured 2026-09-04)."""
+  the related row (`{}`, `{null: true}`) beside one that does adds the issues without the relation,
+  and every branch carrying `null: true` is the issues without it, comparators unread. A bare
+  `{null: false}` branch is that alternative in the object's own `or`, and the object's `null: false`
+  when the `or` sits in an `and` branch (measured 2026-09-04)."""
   or: [NullableProjectFilter!]
 }
 
--- a/backlot/graphql/linear.graphql
+++ b/backlot/graphql/linear.graphql
@@ -263,8 +263,9 @@ input NullableUserFilter {
   """The keys of one branch are alternatives, as on every `or` here. A branch that says nothing about
   the related row (`{}`, `{null: true}`) beside one that does adds the issues without the relation,
   and every branch carrying `null: true` is the issues without it, comparators unread. A bare
-  `{null: false}` branch is that alternative in the object's own `or`, and the object's `null: false`
-  when the `or` sits in an `and` branch (measured 2026-09-04)."""
+  `{null: false}` branch is that alternative in the object's own `or` unless every branch carries
+  `null: false`, and the object's `null: false` when the `or` sits in an `and` branch (measured
+  2026-09-04)."""
   or: [NullableUserFilter!]
 }
 
@@ -287,8 +288,9 @@ input NullableProjectFilter {
   """The keys of one branch are alternatives, as on every `or` here. A branch that says nothing about
   the related row (`{}`, `{null: true}`) beside one that does adds the issues without the relation,
   and every branch carrying `null: true` is the issues without it, comparators unread. A bare
-  `{null: false}` branch is that alternative in the object's own `or`, and the object's `null: false`
-  when the `or` sits in an `and` branch (measured 2026-09-04)."""
+  `{null: false}` branch is that alternative in the object's own `or` unless every branch carries
+  `null: false`, and the object's `null: false` when the `or` sits in an `and` branch (measured
+  2026-09-04)."""
   or: [NullableProjectFilter!]
 }
 
--- a/backlot/graphql/linear_filters.py
+++ b/backlot/graphql/linear_filters.py
@@ -1129,6 +1129,22 @@ def _sub_filter(conn, spec: dict, mapping: dict) -> tuple[str, list]:
       render nothing, and the object's own ``or: []`` is the issues with the relation, where
       ``{}`` and ``{and: []}`` are every issue.
 
+    Read as one procedure, which is how every row above was reproduced: MERGE -- ``and`` branches are
+    keys of the object, their ``or`` lists become the object's, ``null`` is the object's own or else
+    any-true over the branches, and a bare ``{null: false}`` branch of a list that came from an
+    ``and`` branch is that branch's ``null: false`` unless a ``null: true`` is anywhere in the list;
+    HOIST -- a list whose every branch carries ``null: true`` is the object saying ``null: true``, one
+    whose every branch carries ``null: false`` requires the relation; NULL -- own ``null: true`` is
+    the issues without the relation, nothing else read; LISTS -- each list against the related row,
+    the keys of one branch alternatives, ``null`` keys unread, a branch saying nothing skipped,
+    nothing left meaning every related row, a branch of operator-less comparators making the list
+    say nothing, and a branch that says nothing about the related row (``{}``, ``{and: []}``, an
+    ``or`` holding one, or in the object's own list a bare ``{null: false}``) being the list's
+    missing-relation alternative; ASSEMBLE -- relation required means present AND the rest, else a
+    ``null: true`` anywhere below means without OR the rest, else the rest. What the procedure
+    does not give is why the two positions read a bare ``{null: false}`` differently; that, and
+    the branch-keys-as-alternatives reading, is measured, not derived.
+
     This is the vendor's arithmetic as measured, not its code: the first two clauses read like a
     scan for ``null: true`` that decides whether rows without the relation are admitted, the
     rest like flags hoisted out of the branches before the related-row side compiles, but no
--- a/backlot/graphql/linear_filters.py
+++ b/backlot/graphql/linear_filters.py
@@ -751,7 +751,16 @@ def _issue_parts(conn, flt: dict, team_keys: dict | None) -> tuple[str, list, bo
             # `team_keys` rides down the recursion: without it a `team.key` nested under and/or
             # compiled against the name-derived key while the same filter at top level compiled
             # against the corpus's, so the two spellings of one query disagreed.
-            subs = [_issue_parts(conn, s, team_keys) for s in spec]
+            branches = spec
+            if key == "or":
+                # api.linear.app reads the keys of one `or` branch as alternatives, not as one
+                # conjunction: `{or: [{number: {eq: 1}, title: {eq: T2}}]}` answered the issue
+                # numbered 1 and the one titled T2, where `{and: [{number: {eq: 1}, title: {eq:
+                # T2}}]}` answered none, and `projects(filter: {or: [{name: {eq: A}, id: {eq:
+                # B}}]})` both projects (measured 2026-09-04). So a branch with n keys is n
+                # branches; one with none is still dropped, as measured above.
+                branches = [{k: v} for branch in spec for k, v in (branch or {}).items()]
+            subs = [_issue_parts(conn, s, team_keys) for s in branches]
             if key == "or" and any(v for _, _, v in subs):
                 vacuous = True
                 continue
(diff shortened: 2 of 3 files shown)


khj809 (commented):
The procedure holds up. I walked all sixteen rows of #128's table through `_sub_filter` by hand and every one lands where the issue says it does, and the cells are real gates: with `linear_filters.py` back at `79ccdac` and the tests kept, `96 failed, 48 passed` over the new parametrization. Full suite `2319 passed, 1 skipped` here (the skip is a missing `llama_index.readers.hubspot` in my venv, not this branch), `ruff check` / `ruff format --check` clean, `scripts/gen_docs.py --check` exit 0. The claim that the baseline cannot move is right for a second reason worth writing down: `fidelity/graphql_diff.py:259` introspects with `descriptions=False`, so the two new `or` descriptions are invisible to it by construction.

Three things to fix before merge.

**1. `_NEGATIVE_OPS` at `backlot/graphql/linear_filters.py:941` shadows the labels one.** The new tuple rebinds a name already bound at line 564 to the frozenset `_reads_as_negation` uses, so four operators silently leave the labels polarity rule:

```
this branch: _reads_as_negation({"name": {"notContains": "x"}}) -> False
79ccdac:     _reads_as_negation({"name": {"notContains": "x"}}) -> True
```

`_labels_predicate` keys what a label-less issue answers on that polarity, so `labels: {some: {name: {notContains: X}}}` flips from "every issue with no labels passes" to "none do" — the rule #116 measured. It is unreachable today only because `IssueLabelFilter.name` is a `StringComparator`, which does not declare `notContains` / `notContainsIgnoreCase` / `notStartsWith` / `notEndsWith`; `SourceTypeComparator` in this same SDL declares all four, so the day `StringComparator` catches up the labels rule breaks with nothing failing. ruff does not flag a rebound constant. Renaming to `_RELATION_NEGATIVE_OPS` fixes it, and the same three-op literal is inlined again at line 450 in `_derived_in`, so one shared relation-side constant would cover both.

**2. `elif sub == {}` at `backlot/graphql/linear_filters.py:1188` bypasses the unsupported-field raise.** It runs before `mapping.get(key)`, so a nested field that is not in the mapping is dropped instead of refused:

```
this branch: _issue_filter(None, {"project": {"bogus": {}}}) -> ('', [])
79ccdac:     GraphQLError("unsupported nested filter field 'bogus'")
```

`{"project": {"or": [{"bogus": {}}]}}` goes the same way, `_is_vacuous_branch` answering before `_related_row_term` can raise. Nothing is reachable now because every declared nested field is mapped, but the module docstring names that raise as the reason "the SDL and this module cannot drift apart unnoticed": the next field added to `NullableProjectFilter` / `UserFilter` / `TeamFilter` and forgotten here answers a full, wrong result set for `{field: {}}` rather than an error. Resolving `target` first and skipping the empty comparator after keeps both.

**3. The served `or` description overclaims the bare `{null: false}` case (`backlot/graphql/linear.graphql:265`, and the copy at `:289`).** It says a bare `{null: false}` branch is the missing-relation alternative "in the object's own `or`", but FLAGS runs before LISTS and decides `false` when every branch of that own `or` carries `null: false`. Two cells in this PR are the counterexample: `{or: [{null: false}, {null: false, name: {eq: "nobody"}}]}` is none and `{or: [{null: false}, {null: false, name: {eq: "zz-pa"}}]}` is `zz-pa`'s, where the description predicts the issues without a project in the first. Since `e12ab74` was specifically about saying what the bare branch means in each position, the third position belongs in it: unless every branch carries `null: false`.

One scope question rather than a defect. #128's "what a fix has to settle" answers the branch-key rule with "everywhere is the consistent choice", and its own measurement is that Linear ORs a branch's keys on `issues` and `projects` alike. This lands it on `_sub_filter` and `_issue_parts` and leaves `_labels_branches` and `compile_comment_filter` ANDing. Deferring the label collection reads fine — it has its own `some`/`every` polarity tangled into the same code. `CommentFilter` is the one I would not leave silent: nothing about it is measured, but the measurement already in hand says the AND is wrong, and the baseline cannot carry a behaviour row, so after this merges there is no artifact anywhere that remembers it. A follow-up issue referenced from "Not in this PR" would be enough.

Two numbers in the body to true up while you are in there. "swapped back to the previous commit's: `6 failed, 138 passed`" is `6 failed, 137 passed` now — 138 was the count before `e12ab74` dropped the duplicated cell, and read literally as `77c5e2d` the same sentence is `143 passed`. And the round breakdown, `138 + 93 + 53 + 16 + 4 + 37 + 31 + 8 + 13 + 10 + 6`, sums to 409 rather than the 458 in the sentence around it (#128 calls the same list 430); the commit messages run 354 → 429 → 458, and 354 does not match the first five either.

nuri-yoo:
All three fixed in `6349985`, each reproduced first.

**1. `_NEGATIVE_OPS`.** Confirmed: on `e12ab74` `_reads_as_negation({"name": {"notContains": "x"}})` is `False`, on `79ccdac` `True`. The relation-side tuple is `_RELATION_NEGATIVE_OPS`, defined next to the labels frozenset with a comment saying which is which, and `_derived_in` reads it instead of its inlined copy at the old line 450. `test_relation_negative_operators_do_not_shadow_the_labels_polarity_rule` pins `notContains` as a negation and that the relation tuple is a strict subset of the labels set.

**2. `elif sub == {}`.** Confirmed: `_issue_filter(None, {"project": {"bogus": {}}})` compiled to `('', [])` on `e12ab74`. The mapping lookup now comes before the empty-comparator skip, and `_relation_or` walks every field an `or` reaches (through `_through_and_or`) and refuses an unmapped one before `_is_vacuous_branch` gets to answer. `test_unmapped_nested_field_is_refused_even_with_no_operator` sends `{project: {bogus: {}}}`, the same inside an `or`, beside a real branch, and under `and`; all four raise `unsupported nested filter field 'bogus'`.

**3. The `or` description.** It now reads "a bare `{null: false}` branch is that alternative in the object's own `or` unless every branch carries `null: false`, and the object's `null: false` when the `or` sits in an `and` branch", on both types.

**Scope.** `CommentFilter` is measured now rather than left silent: two comments created on the workspace and deleted after, `comments(filter: {or: [{body: {eq: "zz-c1"}, id: {eq: <zz-c2>}}]})` answers both where `{body: {eq: "zz-c1"}, id: {eq: <zz-c2>}}` and the `and` form answer none, and an empty branch is dropped as `_comment_filter` already drops it. Filed as #136 with the bundled-corpus reproduction (`0, 2, 0` against Linear's `2, 2, 0`) and `_issue_parts`'s flattening as the shape; "Not in this PR" links it. Labels stay deferred for the reason you gave.

**Numbers.** The round list was the relation rows minus the 11 `issues` / `projects` probes and with one round miscounted; the true breakdown is 136 + 93 + 53 + 16 + 4 + 36 + 31 + 8 + 13 + 10 + 6 = 406 relation filters, plus 11 on `issues` and `projects` for the branch-key rule, and 406 + 24 (#128's first table) + 28 (#124's last round) = 458 is what the harness holds. The docstring, the body and #128 now say that. The fails-without sentence names its revision: `79ccdac` (main) with the tests kept is `97 failed, 49 passed`, `35ef0ad` (the commit before the procedure) is `6 failed, 137 passed`. Thank you for the `descriptions=False` reason; it is in "Not in this PR" now.

Full suite on `6349985`: `2322 passed`, ruff and format clean, `gen_docs.py --check` exit 0.

khj809 (approved):
All three are fixed at `6349985`, and I reproduced each: `_reads_as_negation({"name": {"notContains": "x"}})` is `True` again with `_RELATION_NEGATIVE_OPS` split out and `_derived_in` reading it, the unmapped-field refusal now also survives `{project: {or: [{or: [{bogus: {}}]}]}}` and the `and`-nested shapes, and the two `or` descriptions carry the every-branch case. Gates on my side: `2321 passed, 1 skipped` (the skip is a missing `llama_index.readers.hubspot` in my venv), ruff and format clean, `gen_docs.py --check` exit 0, `97 failed, 49 passed` with `linear_filters.py` at `79ccdac` — and thank you for measuring `CommentFilter` into #136 rather than leaving it noted.

PR #134, opened 2026-09-04
atlassian: an unresolved credential is anonymous on Jira, refused by Confluence, and a bearer refused as the real gateway does

### LINKED ISSUE #123: Atlassian: a failed Basic pair answers 401 where real Jira serves the request anonymously and Confluence answers 403
On the Jira and Confluence routes Backlot serves, a failed Basic pair gets a `401` from Backlot. The real services do not answer `401` there: Jira treats the request as anonymous and Confluence answers `403` in its own envelope. `401` is what real Jira gives only on `GET /rest/api/3/myself`, which Backlot does not serve (`fidelity/baseline/jira.json` lists it as a gap), and which is where #114's table was measured.

## Measured

`curl` with `nobody@example.com:wrongtoken`, and again with an empty password, against a public Atlassian Cloud site (`ecosystem.atlassian.net`) and a private one (`brekkylab.atlassian.net`). Both sites answered the same on every row.

| Route | Real | Backlot |
|---|---|---|
| `GET /rest/api/3/myself` | `401`, `text/html` "Client must be authenticated to access this resource.", `WWW-Authenticate: OAuth realm=…` | not served (`404`) |
| `GET /rest/api/3/search/jql?jql=` | `400` "Unbounded JQL", processed as anonymous | `401` |
| `GET /rest/api/3/project/search` | `200`, `{"values": []}`, processed as anonymous | `401` |
| `GET /rest/api/3/issue/ABC-1` | `404` "does not exist or you do not have permission", processed as anonymous | `401` |
| `GET /wiki/rest/api/space` | `403` `{"statusCode":403,"message":"…StacklessResponseStatusException: 403 FORBIDDEN \"Request rejected because caller cannot access Confluence\""}` | `401` |
| `GET /wiki/rest/api/content/1` | `403`, same envelope | `401` |

Every anonymous Jira row carries `X-Seraph-LoginReason: AUTHENTICATED_FAILED` and no `WWW-Authenticate`. Backlot's `401` on all of them is the hybrid envelope from `backlot/errors/atlassian.py`, which #44 shaped for client parsing rather than from a measurement of these routes.

## What a client sees

mcp-atlassian on Backlot logs `Authentication failed for Jira API (401). Token may be expired or invalid` and returns nothing (observed in `tests/test_mcp.py::test_mcp_atlassian_acl_enforced` with the pair corrupted). The raw answers real Jira gives on those routes are an empty project list or a `400`, with no authentication error to log. mcp-atlassian's own behaviour on real was not measured.

## Closing it

Needs an anonymous caller on these routes: `project/search` answers with the projects an anonymous user can see (none, in a corpus with no public project), `search/jql` with the anonymous view of the query, `issue/{key}` with `404`, and the Confluence routes with the `403` envelope, each with the `X-Seraph-LoginReason` header. It changes the answer for every failed credential on these routes, not only the pair, so it is its own change rather than part of #114.

## Not measured

- A valid token under another account's address on these routes (no second token on hand). `myself` gives `401` for it, per #114.
- What anonymous `search/jql` answers with a bounded JQL. The unbounded form was the only one tried.
- Whether real Jira's anonymous processing differs on a site with public projects.

### LINKED ISSUE #135: Atlassian: an unresolvable Bearer answers 401 where the real site answers 403 with a Connect-token parse error
Backlot answers `401` to an Atlassian read carrying a `Bearer` token it cannot resolve. Against a real `<site>.atlassian.net` an opaque bearer answers `403` on every route, with a body that is neither the Jira envelope nor the Confluence one.

#134 left this row alone: it fixed the Basic and no-credential rows on the same routes, where the measurement lines up with the mechanism Backlot models, and this one does not (see "The mechanism" below).

## Measured

`curl -H "Authorization: Bearer bogustoken123"` against `ecosystem.atlassian.net` and `brekkylab.atlassian.net` on 2026-09-04. Both sites answered the same on every row, and the answer did not vary between runs.

| Route | Real | Backlot |
|---|---|---|
| `GET /rest/api/3/project/search` | `403`, `application/json`, `{"error": "Failed to parse Connect Session Auth Token"}` | `401`, the Atlassian envelope |
| `GET /rest/api/3/issue/ABC-1` | `403`, same body | `401` |
| `GET /rest/api/3/search/jql?jql=` | `403`, same body | `401` |
| `GET /rest/api/3/myself` | `403`, same body | not served (`404`) |
| `GET /wiki/rest/api/space` | `403`, Confluence's `StacklessResponseStatusException` envelope | `403` since #134 |
| `GET /wiki/rest/api/content/1` | `403`, same envelope | `403` since #134 |

No `WWW-Authenticate` and no `X-Seraph-LoginReason` on any row. The Jira rows are the only place in this surface that answers with `{"error": ...}` — a single key, not `message`/`errorMessages`/`statusCode` — so a client reading the Atlassian envelope out of it finds nothing.

Note the contrast with a failed Basic pair on the same Jira routes, which is not refused at all: `project/search` is `200` and `issue/{key}` is `404`, processed anonymously (#123). A bearer is a harder refusal than a wrong password there.

## The mechanism

Against `<site>.atlassian.net` a `Bearer` header is read as a Connect session JWT, which is what the message says: an opaque string is not a JWT, so the parse fails before any identity is looked up. Atlassian's OAuth 2.0 3LO access tokens — the credential Backlot's bearer branch is documented as standing in for (`docs/auth.md`) — are sent to `api.atlassian.com/ex/jira/{cloudid}/...`, a base URL Backlot does not serve.

So the `403` above is measured for the wire shape Backlot serves, and a Backlot bearer really is an opaque non-JWT, which makes it the case this row covers. What is *not* measured is the answer a well-formed 3LO token that has expired or been revoked draws, on either base URL; that needs an OAuth app and a token to revoke.

## Closing it

Deciding which of the two the bearer branch is emulating comes first, because they answer differently and the choice is not observable from the routes:

- Treat a Backlot bearer as what it looks like on the wire — an opaque non-JWT to `<site>.atlassian.net` — and answer `403 {"error": "Failed to parse Connect Session Auth Token"}` on the Jira routes. Needs a third body shape in `backlot/errors/atlassian.py`, which today emits one hybrid envelope for both APIs.
- Or keep it as the 3LO stand-in and measure that path properly first, which means an OAuth app against a real site.

Either way the Confluence rows already match, since #134 refuses on the credential failing rather than on which scheme carried it.

khj809:
Backlot answered `401` to every Atlassian read whose credential it could not resolve. None of the three real answers is a `401` there. Jira drops a failed Basic pair to anonymous and serves the request; Confluence rejects it with a `403` of its own; and a bearer the site reads and cannot resolve is a `403` in a third body shape. All three now hold, and the spelling that counts as a bearer at all is the site's.

## What each API does

Measured against `ecosystem.atlassian.net` (public projects) and `brekkylab.atlassian.net` (none) on 2026-09-04, over a failed `email:api_token` pair, an empty password, a value that is not base64, an unknown scheme, an opaque bearer, a JWT-shaped bearer, and no `Authorization` header. Every row below answered alike on both sites, and repeats were stable.

| Route | Real, credential unresolved | Backlot now |
|---|---|---|
| `GET /rest/api/3/project/search` | `200`, the projects anonymous may see | `200`, `values: []` |
| `GET /rest/api/3/search/jql` (bounded) | `200`, that query's anonymous view | `200`, `issues: []` |
| `GET /rest/api/3/issue/{key}` | `404` "Issue does not exist or you do not have permission to see it." | the same, verbatim |
| `GET /rest/api/3/project/{key}/role[/{id}]` | `404` "No project could be found with key 'X'." for a key anonymous cannot see | the same, verbatim |
| `GET /rest/api/3/field`, `/issueLinkType` | `200` | `200` |
| `GET /wiki/rest/api/*` | `403`, Confluence's `StacklessResponseStatusException` envelope | the same message, in the Atlassian envelope this repo emits |
| any Jira route, unreadable `Bearer` | `403` `{"error": "Failed to parse Connect Session Auth Token"}` | the same, verbatim |

Jira reports a failed Basic username in `X-Seraph-LoginReason: AUTHENTICATED_FAILED`, on every answer including the `200`s, so that is middleware rather than something a refusal path carries. The header is keyed on the username alone: a value with a non-empty user before its first colon draws it, one with an empty user or no colon draws nothing, and a bearer failure draws none. Confluence sends it on nothing.

## Three credential shapes, three answers

Each API's answer turns on what it could read, not on which route was asked, so the shapes have to be told apart before any refusal is picked.

**Basic.** Confluence's `403` is for a pair it read and rejected, and a request carrying no credential gets the same `403`. A Basic value it could *not* read gets `401` with `WWW-Authenticate: OAuth realm="<site>/wiki"`. Exactly one colon with both halves non-empty is readable; an empty user, an empty password, a second colon, no colon, or a payload that is not base64 is not. So an empty password never reaches the `403` path — `auth.basic_credential_kind`. Jira reads all of them as anonymous.

**Bearer.** Read case-sensitively, separated from the token by exactly one space: `Bearer <t>` is a credential and `bearer <t>`, `BEARER <t>`, `token <t>`, `OAuth <t>`, `Bearer  <t>`, `Bearer<TAB><t>` and a bare `Bearer` are not — those are the anonymous request. Backlot's shared `bearer_token` accepts five of them — every one but `OAuth <t>`, `Bearer<t>` and the bare `Bearer` — so Atlassian gets its own parser, for the reason `slack_bearer_token` already has one: sharing it authenticates spellings that read nothing in production. A bearer it does read and cannot resolve is refused *before the route* — `serverInfo` and `field` answer the `403` too, though neither needs a credential — which is why that lives in middleware.

**Nothing at all.** Identical to a failed pair on every row, minus the Seraph header. So the fix covers the absent credential too, not only a failed one, and the four tests that pinned `401` for the unauthenticated case are rewritten: `test_atlassian.py::test_atlassian_401_keeps_the_atlassian_error_envelope`, `test_atlassian.py::test_atlassian_errors_use_atlassian_envelope`, `test_auth.py::test_require_basic_or_bearer_raises_401_with_no_credential`, and `test_cross_vendor.py::test_unauthenticated_is_rejected`.

## Anonymous in Backlot's terms

`acl.ANONYMOUS` is a caller with an empty principal set, where a scoped caller's is `{org} ∪ groups ∪ {their address}`. A corpus grants "public" to the org, and nobody outside the org is in it, so anonymous reaches no document — which is why each Jira listing comes back empty, the same answer `brekkylab.atlassian.net` gives. `store._acl_clause` already answered an empty set with `AND 0`.

A project listing had no ACL filter at all, so an anonymous caller would have been handed every project. A project is now listed when the caller can see an issue in it — Backlot grants per document, so there is nothing else to read "can browse this project" off. `store.has_visible_document` asks that as `LIMIT 1` rather than counting each project in full, which the listing never uses the number from.

That rule also applies to a scoped caller, and there it is **not** measured: real Jira lists a project by its browse permission, and a project can carry one while showing the caller no issue, so a real listing can hold a project this one drops. Measuring it needs two accounts and a per-project permission scheme on a live site. What the rule does rule out is the unfiltered listing, which handed every project to a caller who can open nothing in any of them.

## Measured and deliberately not emulated

A bearer shaped like a complete signed JWS — three segments, an `alg` in the header, a JSON payload, a non-empty signature — is *read* and then rejected with `401 text/html` "Client must be authenticated to access this resource.", not the `403`. Every incomplete shape falls back to the `403`: a missing signature, two segments, four segments, no `alg`, a header or payload that is not JSON (which answers `{"error": "Payload of JWS object is not a valid JSON object"}`, on Confluence as well). Backlot issues opaque `usr-<hex>` tokens, measured directly as the `403` shape, and reproducing Atlassian Connect's accept boundary would mean inventing the space between those probes for a mechanism Backlot does not implement — so a JWT-shaped bearer takes the `403` here.

`GET /rest/api/3/myself` is where `401` is the real answer, and Backlot does not serve it — `backlot/fidelity/baseline/jira.json` lists it as a gap. An unbounded `search/jql` answers `200` here where real answers `400`; that permissiveness is its own row, unchanged. What a valid token under another account's address does on these routes is still unmeasured, per #123.

Closes #123
Closes #135

### DIFF
--- a/backlot/auth.py
+++ b/backlot/auth.py
@@ -194,6 +194,50 @@ def require_bearer(request: Request, detail: str) -> Caller:
     return caller
 
 
+def atlassian_bearer_token(request: Request) -> str | None:
+    """Parse the ``Authorization`` header the way a ``<site>.atlassian.net`` gateway does, which
+    is stricter than :func:`bearer_token` and strict differently from :func:`slack_bearer_token`.
+
+    Measured against ecosystem.atlassian.net and brekkylab.atlassian.net on 2026-09-04 (a bogus
+    token is enough — a recognised credential is refused with a 403 and an unrecognised one is
+    served anonymously, so the answer says which the site read)::
+
+        Bearer <t>      read            bearer <t>      not read
+        ' Bearer <t>'   read            BEARER <t>      not read
+        Bearer <t>' '   read            token <t>       not read
+                                        OAuth <t>       not read
+                                        Bearer  <t>     not read
+                                        Bearer<TAB><t>  not read
+                                        Bearer<t>       not read
+                                        Bearer          not read
+
+    The scheme is case-sensitive and separated from the token by exactly one space. Sharing
+    :func:`bearer_token` would authenticate four spellings the real site serves anonymously — a
+    client sending ``Authorization: token <t>`` would pass every test here and read nothing in
+    production. Slack refuses a different set: it takes the double space this refuses.
+    """
+    hdr = (_authorization(request) or "").strip()
+    if not hdr.startswith("Bearer "):
+        return None
+    rest = hdr[len("Bearer ") :]
+    if not rest or rest[0].isspace():
+        return None
+    return rest.rstrip()
+
+
+def atlassian_bearer_unreadable(request: Request) -> bool:
+    """Whether the request carries a bearer the site would read and Backlot cannot resolve.
+
+    A Backlot token is an opaque string with no dots, and that is the shape the gateway reports as
+    unreadable — measured with ``usr-<hex>`` itself. A token shaped like a complete signed JWS is
+    read and then rejected with a 401 instead; Backlot issues none, and reproducing Atlassian
+    Connect's accept boundary would mean inventing the space between the shapes measured, so a
+    JWT-shaped bearer takes the 403 here too.
+    """
+    token = atlassian_bearer_token(request)
+    return bool(token) and acl(request).resolve(token) is None
+
+
 def atlassian_caller(request: Request) -> Caller:
     """The caller for an Atlassian read: Basic ``email:api_token`` or a bearer OAuth token, and
     :data:`backlot.acl.ANONYMOUS` when neither resolves.
@@ -203,7 +247,8 @@ def atlassian_caller(request: Request) -> Caller:
     Confluence refuses. Each router decides for itself, so this reports the identity and nothing
     else.
     """
-    return resolve_basic(request) or resolve_bearer(request) or ANONYMOUS
+    bearer = atlassian_bearer_token(request)
+    return resolve_basic(request) or acl(request).resolve(bearer) or ANONYMOUS
 
 
 def resolve_api_key(request: Request) -> Caller | None:
--- a/backlot/auth.py
+++ b/backlot/auth.py
@@ -154,9 +154,10 @@ def slack_bearer_token(request: Request) -> str | None:
     A tab is not a space to Slack, so the generic whitespace split in :func:`bearer_token` is
     wrong here, and so is its case-insensitive scheme match. That function stays permissive because
     GitHub really does accept ``token <t>`` and RFC 7235 really does make the scheme
-    case-insensitive; Slack implements neither. Sharing it would let Backlot authenticate six
-    spellings live Slack refuses outright, so a client sending ``Authorization: token <xoxb>``
-    would pass every test here and reach nothing in production.
+    case-insensitive; Slack implements neither. Of the five spellings above that Slack refuses,
+    sharing it would authenticate four — every one but ``Bearer<t>``, which it does not take
+    either — so a client sending ``Authorization: token <xoxb>`` would pass every test here and
+    reach nothing in production.
     """
     hdr = (_authorization(request) or "").strip()
     if not hdr.startswith("Bearer "):
@@ -212,9 +213,13 @@ def atlassian_bearer_token(request: Request) -> str | None:
                                         Bearer          not read
 
     The scheme is case-sensitive and separated from the token by exactly one space. Sharing
-    :func:`bearer_token` would authenticate four spellings the real site serves anonymously — a
-    client sending ``Authorization: token <t>`` would pass every test here and read nothing in
-    production. Slack refuses a different set: it takes the double space this refuses.
+    :func:`bearer_token` would authenticate five of those spellings — every "not read" row above
+    but ``OAuth <t>``, ``Bearer<t>`` and the bare ``Bearer`` — so a client sending
+    ``Authorization: token <t>`` would pass every test here and read nothing in production. Slack
+    refuses a different set: it takes the double space this refuses.
+
+    The leading ``strip`` is what serves the ``Bearer <t>' '`` row, so the token needs no second
+    one: nothing with trailing whitespace survives to reach it.
     """
     hdr = (_authorization(request) or "").strip()
     if not hdr.startswith("Bearer "):
@@ -222,7 +227,7 @@ def atlassian_bearer_token(request: Request) -> str | None:
     rest = hdr[len("Bearer ") :]
     if not rest or rest[0].isspace():
         return None
-    return rest.rstrip()
+    return rest
 
 
 def atlassian_bearer_unreadable(request: Request) -> bool:
@@ -239,8 +244,14 @@ def atlassian_bearer_unreadable(request: Request) -> bool:
 
 
 def atlassian_caller(request: Request) -> Caller:
-    """The caller for an Atlassian read: Basic ``email:api_token`` or a bearer OAuth token, and
-    :data:`backlot.acl.ANONYMOUS` when neither resolves.
+    """The caller for an Atlassian read: Basic ``email:api_token`` or
+    :func:`atlassian_bearer_token`, and :data:`backlot.acl.ANONYMOUS` when neither resolves.
+
+    The bearer is NOT an OAuth 3LO token standing in for Atlassian's own. A 3LO token goes to
+    ``api.atlassian.com/ex/jira/{cloudid}/…``, which Backlot does not serve; on the
+    ``<site>.atlassian.net`` surface it does serve, a bearer is read as a Connect session JWT, and
+    an opaque Backlot token is one the gateway cannot read at all — which is the ``403``
+    :func:`atlassian_bearer_unreadable` reports.
 
     No refusal here, unlike :func:`require_bearer`: the two Atlassian APIs disagree about what an
     unresolved credential means. Jira drops the caller to anonymous and answers the request; only
--- a/backlot/auth.py
+++ b/backlot/auth.py
@@ -17,7 +17,7 @@ from datetime import datetime, timezone
 from fastapi import HTTPException, Request
 
 from backlot import sigv4
-from backlot.acl import Acl, Caller
+from backlot.acl import ANONYMOUS, Acl, Caller
 
 
 def conn(request: Request) -> sqlite3.Connection:
@@ -62,20 +62,78 @@ def api_key_token(request: Request) -> str | None:
     return hdr
 
 
-def basic_password(request: Request) -> tuple[str | None, str | None]:
-    """Parse ``Authorization: Basic base64(user:pass)`` -> (user, pass)."""
+# How much of a Basic credential a request carries.
+BASIC_ABSENT = "absent"  # no header, another scheme, or `Basic` with nothing to decode
+BASIC_UNPARSEABLE = "unparseable"  # a value that is not one user and one password
+BASIC_PAIR = "pair"  # exactly one non-empty user and one non-empty password
+
+
+def _basic_value(request: Request) -> str | None:
+    """The raw base64 payload of an ``Authorization: Basic`` header, or None for any other."""
     hdr = _authorization(request)
     if not hdr:
-        return None, None
+        return None
     parts = hdr.split(None, 1)
     if len(parts) == 2 and parts[0].lower() == "basic":
-        try:
-            decoded = base64.b64decode(parts[1]).decode("utf-8", "replace")
-            user, _, pw = decoded.partition(":")
-            return user, pw
-        except (ValueError, UnicodeDecodeError):
-            return None, None
-    return None, None
+        return parts[1]
+    return None
+
+
+def _decoded_basic(request: Request) -> str | None:
+    """The decoded ``user:pass`` of a Basic header, ``None`` when there is none to decode and
+    ``""`` for a payload that is not base64 — a value that was there and could not be read."""
+    value = _basic_value(request)
+    if not value:
+        return None
+    try:
+        return base64.b64decode(value).decode("utf-8", "replace")
+    except (ValueError, UnicodeDecodeError):
+        return ""
+
+
+def basic_password(request: Request) -> tuple[str | None, str | None]:
+    """Parse ``Authorization: Basic base64(user:pass)`` -> (user, pass)."""
+    decoded = _decoded_basic(request)
+    if not decoded:
+        return None, None
+    user, _, pw = decoded.partition(":")
+    return user, pw
+
+
+def basic_credential_kind(request: Request) -> str:
+    """Which of :data:`BASIC_ABSENT` / :data:`BASIC_UNPARSEABLE` / :data:`BASIC_PAIR` the request
+    carries.
+
+    Confluence answers the three differently — a pair it read and rejected is its 403, a value it
+    could not read is a 401, and no credential at all is the 403 again — so which one a request
+    holds decides which refusal it draws. Measured against ecosystem.atlassian.net and
+    brekkylab.atlassian.net on 2026-09-04, which is where the split is observable at all: a single
+    colon with both halves non-empty is the pair; an empty user, an empty password, no colon, a
+    second colon, or a payload that is not base64 is unparseable; and a missing header, an unknown
+    scheme and a `Basic` with nothing after it are no credential.
+    """
+    if not _basic_value(request):
+        return BASIC_ABSENT
+    decoded = _decoded_basic(request)
+    if not decoded:  # a payload that is there and does not decode
+        return BASIC_UNPARSEABLE
+    user, _, pw = decoded.partition(":")
+    if user and pw and ":" not in pw:
+        return BASIC_PAIR
+    return BASIC_UNPARSEABLE
+
+
+def basic_names_a_user(request: Request) -> bool:
+    """Whether the request presented a Basic credential naming somebody.
+
+    Jira reports one it could not resolve in ``X-Seraph-LoginReason``, and the header is keyed on
+    the username alone: a value carrying a non-empty user before its first colon draws it whatever
+    follows, and one with an empty user, or with no colon at all, draws nothing. Measured on both
+    sites on 2026-09-04.
+    """
+    decoded = _decoded_basic(request) or ""
+    user, colon, _ = decoded.partition(":")
+    return bool(user and colon)
 
 
 def slack_bearer_token(request: Request) -> str | None:
@@ -136,13 +194,16 @@ def require_bearer(request: Request, detail: str) -> Caller:
     return caller
 
 
-def require_basic_or_bearer(request: Request, detail: str) -> Caller:
-    """Same, for Atlassian: it carries Basic ``email:api_token`` and also accepts a bearer OAuth
-    token, so both are tried before refusing."""
-    caller = resolve_basic(request) or resolve_bearer(request)
-    if caller is None:
-        raise HTTPException(status_code=401, detail=detail)
-    return caller
+def atlassian_caller(request: Request) -> Caller:
+    """The caller for an Atlassian read: Basic ``email:api_token`` or a bearer OAuth token, and
+    :data:`backlot.acl.ANONYMOUS` when neither resolves.
+
+    No refusal here, unlike :func:`require_bearer`: the two Atlassian APIs disagree about what an
+    unresolved credential means. Jira drops the caller to anonymous and answers the request; only
+    Confluence refuses. Each router decides for itself, so this reports the identity and nothing
+    else.
+    """
+    return resolve_basic(request) or resolve_bearer(request) or ANONYMOUS
 
 
 def resolve_api_key(request: Request) -> Caller | None:
--- a/backlot/errors/atlassian.py
+++ b/backlot/errors/atlassian.py
@@ -26,6 +26,17 @@ CONFLUENCE_FORBIDDEN = (
 )
 CONFLUENCE_UNAUTHORIZED = "Unauthorized"
 
+# What a `<site>.atlassian.net` gateway answers a `Bearer` it cannot read as a Connect session
+# JWT, on every Jira route — `serverInfo` and `field`, which need no credential, included. One
+# `error` key and nothing else: not this module's envelope, and not Confluence's either, so it is
+# returned as its own body rather than shaped. Measured on ecosystem.atlassian.net and
+# brekkylab.atlassian.net on 2026-09-04.
+CONNECT_TOKEN_UNREADABLE = "Failed to parse Connect Session Auth Token"
+
+
+def connect_token_body() -> dict:
+    return {"error": CONNECT_TOKEN_UNREADABLE}
+
 
 def owns(path: str) -> bool:
     return path.startswith(PREFIX)
--- a/backlot/errors/atlassian.py
+++ b/backlot/errors/atlassian.py
@@ -13,6 +13,19 @@ import http
 
 PREFIX = "/atlassian"
 
+# The two refusals Confluence gives a caller it will not serve, measured against
+# ecosystem.atlassian.net and brekkylab.atlassian.net on 2026-09-04. A credential it read and
+# rejected, and a request that carried none, both get the 403 below verbatim — the vendor reports
+# its own exception class in the message, and a client that logs the body logs that. A Basic value
+# it could not read gets a 401 instead, whose real body is Tomcat's HTML page titled "HTTP Status
+# 401 - Unauthorized"; that title is the message here, because the envelope this module exists to
+# emit is JSON and a client parsing `message` out of the page would find nothing.
+CONFLUENCE_FORBIDDEN = (
+    "com.atlassian.confluence.mvc.rest.common.exception.StacklessResponseStatusException: "
+    '403 FORBIDDEN "Request rejected because caller cannot access Confluence"'
+)
+CONFLUENCE_UNAUTHORIZED = "Unauthorized"
+
 
 def owns(path: str) -> bool:
     return path.startswith(PREFIX)
--- a/backlot/routers/atlassian.py
+++ b/backlot/routers/atlassian.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 
 import re
 from html import escape
+from urllib.parse import quote
 
 from fastapi import APIRouter, HTTPException, Request
 from pydantic import BaseModel, ConfigDict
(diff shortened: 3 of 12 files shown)


backlot/auth.py
```diff
@@ -136,13 +194,61 @@ def require_bearer(request: Request, detail: str) -> Caller:
...
+        ' Bearer <t>'   read            BEARER <t>      not read
+        Bearer <t>' '   read            token <t>       not read
+                                        OAuth <t>       not read
+                                        Bearer  <t>     not read
+                                        Bearer<TAB><t>  not read
+                                        Bearer<t>       not read
+                                        Bearer          not read
+
+    The scheme is case-sensitive and separated from the token by exactly one space. Sharing
+    :func:`bearer_token` would authenticate four spellings the real site serves anonymously — a
```
nuri-yoo: Five, not four. `bearer_token` takes `bearer <t>`, `BEARER <t>`, `token <t>`, `Bearer  <t>` and `Bearer<TAB><t>` — every one of them rejected here and served anonymously by the real site. I ran both parsers over the table above to count rather than reading it off the list. `tests/test_auth.py::test_atlassian_bearer_token_takes_only_the_spelling_the_real_site_reads` says five in its own docstring and the PR body says four, so the three disagree with each other.

backlot/auth.py
```diff
@@ -136,13 +194,61 @@ def require_bearer(request: Request, detail: str) -> Caller:
...
+    read and then rejected with a 401 instead; Backlot issues none, and reproducing Atlassian
+    Connect's accept boundary would mean inventing the space between the shapes measured, so a
+    JWT-shaped bearer takes the 403 here too.
+    """
+    token = atlassian_bearer_token(request)
+    return bool(token) and acl(request).resolve(token) is None
+
+
+def atlassian_caller(request: Request) -> Caller:
+    """The caller for an Atlassian read: Basic ``email:api_token`` or a bearer OAuth token, and
```
nuri-yoo: "a bearer OAuth token" is the framing #135 asked this PR to settle, and the PR settles it the other way. `atlassian_bearer_unreadable` reads a Backlot bearer as what it is on the wire to `<site>.atlassian.net` — an opaque non-JWT — and answers the Connect-token `403`, which is #135's first option; the 3LO stand-in is the one it contrasts that with, and 3LO tokens go to `api.atlassian.com/ex/jira/{cloudid}/...`, which Backlot does not serve. Leaving "OAuth token" on the helper that produces that `403` describes the branch the PR did not take. The same sentence survives in `tests/test_auth.py::test_atlassian_caller_accepts_either_scheme`'s docstring.

backlot/routers/atlassian.py
```diff
@@ -208,16 +263,38 @@ async def jira_project_search(request: Request):
...
+    ecosystem.atlassian.net, whose `AA` is public, and brekkylab.atlassian.net, which has no such
+    key, on 2026-09-04. An unreachable project is reported as an absent one, the way
+    `issue/{key}` reports an issue the caller cannot see.
+    """
+    ids = auth.visible_ids(request, _jira_caller(request))
+    container = _jira_container_for_key(conn, key, request)
+    # `ids is None` short-circuits for the same reason _reachable_projects returns every row for
+    # it: an admin sees the project in the listing, so a role read on it must not 404.
+    if container is not None and (
+        ids is None or store.has_visible_document(conn, "jira", container, ids)
```
nuri-yoo, line 284: `ids is None or` is a no-op today, and it is the one mutation of the thirteen I tried that leaves the suite green. For an admin `_acl_clause(visible_ids=None)` contributes no clause, so `has_visible_document` is already True for any container holding a row — and a container only exists because records put it there (`byo.write_containers` writes `self.containers`, and the only thing that fills it is `add(rec)`, once per record), so a project with no issues is not constructible. That makes the comment's case — "an admin sees the project in the listing, so a role read on it must not 404" — one nothing can reach, which is a reason a reader cannot check. Either it is defence against an importer that could write an empty container and could say so, or a project row with no issues is worth building in a test so the branch is pinned.

tests/test_atlassian.py
```diff
@@ -31,18 +33,225 @@ def test_admin_confluence_crawls_all(client, admin_h, ro_conn):
...
+    h = {"Authorization": f"Bearer {tokens_yaml['admin_token']}"}
+    assert client.get("/atlassian/rest/api/3/project/search", headers=h).json()["values"]
+    assert client.get("/atlassian/wiki/rest/api/space", headers=h).status_code == 200
+
+
+def test_confluence_refuses_an_unreadable_bearer_with_its_own_403_not_jiras(client):
+    """Confluence answers a bearer it cannot resolve with the same 403 envelope it gives a failed
+    pair — the refusal is keyed on the credential failing, not on which scheme carried it. Jira's
+    single-key Connect body does not appear on this side. Measured on both sites, for an opaque
+    token and for a JWT-shaped one."""
```
nuri-yoo, line 217: The docstring says a JWT-shaped bearer was measured, and the body says Backlot deliberately answers it with the `403` where real answers `401` — but nothing here pins that. `Bearer eyJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJ4In0.sig` answers `403` at head and `401 text/html` "Client must be authenticated to access this resource." on both real sites, so the divergence is real and chosen. A parametrize row carrying a JWS-shaped token on the Jira route would keep it from drifting silently if `atlassian_bearer_unreadable` ever grows a shape check.

nuri-yoo (approved):
Verified end to end: I re-ran the body's whole probe matrix against `ecosystem.atlassian.net` and spot-checked `brekkylab.atlassian.net` on 2026-09-04, then ran the same probes against Backlot at `2170295`. Every measured row reproduces, on both sites and on both sides. Four notes below, none blocking.

**Re-measured (real / Backlot at head, same probe both sides)**

| Probe | Real | Backlot |
|---|---|---|
| failed pair on `project/search`, bounded `search/jql`, `field`, `issueLinkType`, `serverInfo` | `200` + `X-Seraph-LoginReason: AUTHENTICATED_FAILED` | same |
| failed pair on `issue/ABC-1` | `404`, message verbatim | same |
| Seraph keying: empty user, no colon, not-base64, no header | no header | same |
| Seraph keying: empty password, second colon | header | same |
| `Bearer <t>`, `' Bearer <t>'`, `'Bearer <t> '` | `403 {"error": "Failed to parse Connect Session Auth Token"}` | same, `serverInfo` included |
| `bearer`, `BEARER`, `token`, `OAuth`, `Bearer  `, `Bearer<TAB>`, `Bearer`, `'Bearer '` | `200` anonymous | same |
| Confluence: readable pair, no credential, unreadable bearer | `403` Stackless envelope | same |
| Confluence: empty pass, empty user, `:`, no colon, two colons, not-base64 | `401` + `WWW-Authenticate: OAuth realm="https%3A%2F%2F<site>%2Fwiki"` | same, percent-encoding included |
| `project/AA/role[/10002]` (a project anonymous can see) | `401` "You cannot edit the configuration of this project." | n/a, anonymous sees no Backlot project |
| `project/NOSUCHKEY/role[/10002]` | `404`, message verbatim | same |
| JWS-shaped bearer | `401 text/html` | `403`, the disclosed divergence |
| Confluence sends Seraph on nothing, a bearer failure draws none | confirmed | same |

Suite at head: `2230 passed, 0 skipped, 0 failed` (junit counts, `-p no:randomly`). `ruff check`, `ruff format --check` and `scripts/gen_docs.py --check` each exit 0.

I also reverted each new guard one at a time and reran the four touched test files: 13 mutations, 12 go red (anonymous `visible_ids`, the project filter, the role 404 gate, the Confluence 401/403 split, the realm percent-encoding, the strict bearer parser, the bearer middleware, the Seraph colon rule, the Seraph middleware, the second-colon rule in `basic_credential_kind`, the empty-set handling in `has_visible_document`, the issue 404 message). The one that stays green has its own comment.

**`tests/test_mcp.py:436`** still names `auth.require_basic_or_bearer`, which this PR deletes — `atlassian_caller` is the name it wants now. The sentence around it is also narrower than it was: "no round-trip can tell the two apart" holds for a credential that resolves, but a failed bearer is now `403` where a failed pair is `200`.

**Body count.** "the two tests that pinned `401` for the unauthenticated case are rewritten" — four did: `test_atlassian.py::test_atlassian_401_keeps_the_atlassian_error_envelope`, `test_atlassian.py::test_atlassian_errors_use_atlassian_envelope`, `test_auth.py::test_require_basic_or_bearer_raises_401_with_no_credential`, and `test_cross_vendor.py::test_unauthenticated_is_rejected`.

**Out of scope, worth filing.** The reason `_reachable_projects` exists reads verbatim onto `/wiki/rest/api/space`, which is still unfiltered. On a two-space corpus where `ava` can read a page in one of them:

```
ava   spaces=['ENG82FE03', 'SEC20E8CD']   readable_pages=['Runbook']   projects=['payments']
```

The Jira project is dropped and the Confluence space is not, and `GET /wiki/rest/api/space/SEC20E8CD/permission` answers her `200` with `bob@acme.com`'s address in it. Not a regression from this PR — anonymous never gets there, `_confluence_caller` refuses first — and `store.has_visible_document` is now the helper that would close it. I can file it with the measurement if you want it separate.

**`docs/auth.md`.** "a credential check on Jira is a check of that header or of whether anything came back at all" holds for Basic only. A client that spells the scheme `bearer` gets `200`, an empty listing and no header, which is the same answer a working credential gives on a corpus it can read nothing in. The paragraph two above already says that spelling is not a credential, so it is a clause rather than a rewrite.

khj809:
Thanks — the independent re-measure is the part I most wanted checked, and the 13-mutation sweep caught the one guard I could not write a test for. All four notes addressed, in f51e1d3 on top of your merge of main.

**`tests/test_mcp.py`** — renamed to `auth.atlassian_caller` and the claim narrowed to what still holds: a credential that resolves reads the same under either scheme, and the schemes only became separable for one that fails, which says which credential failed rather than which scheme a working one arrived under.

**Body count** — corrected to four, named.

**`docs/auth.md`** — you are right that the sentence only held for Basic, and the "whether anything came back" half was wrong for Basic too: an empty listing is also the honest answer for a working credential over a corpus it can read nothing in. Now it says a Basic credential is checked by reading the header, that a misspelt bearer scheme has neither signal because it is not a credential to the site, and to check a bearer against Confluence or a Jira read the account is known to see.

**The space listing** — filed as #138 rather than folded in here, since it is not a regression and the scoped rule needs the same "unmeasured" treatment `_reachable_projects` got. I reproduced your case at 2170295 before filing; it is in the issue verbatim, including `space/SEC20E8CD/permission` answering `ava` with `bob@acme.com`. Two things worth flagging on it: `content?spaceKey=` is correctly empty, so it is exactly the three space-level routes (listing, single get, permission) that are unscoped; and the address disclosure needs no fidelity decision to fix, unlike the listing rule, which does. Happy to take it next if you want it before the Confluence work queued behind it.

One correction to my own earlier record, for the archive: my run reports `2227 passed, 3 skipped` where yours reports `2230 passed, 0 skipped`. The three are `llama_index.readers.hubspot` and two `mirage.core.google._client` imports — optional extras I do not have installed, not a difference in what ran.

backlot/auth.py
```diff
@@ -136,13 +194,61 @@ def require_bearer(request: Request, detail: str) -> Caller:
...
+    client sending ``Authorization: token <t>`` would pass every test here and read nothing in
+    production. Slack refuses a different set: it takes the double space this refuses.
+    """
+    hdr = (_authorization(request) or "").strip()
+    if not hdr.startswith("Bearer "):
+        return None
+    rest = hdr[len("Bearer ") :]
+    if not rest or rest[0].isspace():
+        return None
+    return rest.rstrip()
```
nuri-yoo: Unreachable, and the suite does not notice: `return rest` in place of this leaves `2230 passed, 0 failed` at `f51e1d3`. `hdr` is `.strip()`ed on the way in, so no trailing whitespace survives to reach this call — I put the docstring's own table plus `'Bearer t '`, `'Bearer t   '`, `'  Bearer t  '`, `'Bearer t\t'`, `'Bearer t\n'` and `'Bearer a b '` through the function with and without it, and no input separates the two. The `Bearer <t>' '   read` row of the table above is served by that `.strip()`, not by this line.

The shape looks carried over from `slack_bearer_token`, which ends `.strip() or None`. There both halves earn their place: the `.strip()` because Slack accepts extra leading spaces, which this parser refuses, and the `or None` for a bare `'Bearer '` — which `not rest` has already taken here.

backlot/errors/atlassian.py
```diff
@@ -13,6 +13,30 @@
...
+# rejected, and a request that carried none, both get the 403 below verbatim — the vendor reports
+# its own exception class in the message, and a client that logs the body logs that. A Basic value
+# it could not read gets a 401 instead, whose real body is Tomcat's HTML page titled "HTTP Status
+# 401 - Unauthorized"; that title is the message here, because the envelope this module exists to
+# emit is JSON and a client parsing `message` out of the page would find nothing.
+CONFLUENCE_FORBIDDEN = (
+    "com.atlassian.confluence.mvc.rest.common.exception.StacklessResponseStatusException: "
+    '403 FORBIDDEN "Request rejected because caller cannot access Confluence"'
+)
+CONFLUENCE_UNAUTHORIZED = "Unauthorized"
```
nuri-yoo, line 27: Nothing asserts this string. `CONFLUENCE_FORBIDDEN` is pinned four times in `tests/test_atlassian.py` (201, 211, 252, 253) and `CONNECT_TOKEN_UNREADABLE` once (162); this constant has no hit anywhere under `tests/`. Lowercasing it, and separately emptying it to `""`, each leave the full suite green at `f51e1d3`.

`test_confluence_answers_401_to_a_basic_credential_it_cannot_parse` asserts the status and the `WWW-Authenticate` realm and stops there, so `assert r.json()["message"] == errors_atlassian.CONFLUENCE_UNAUTHORIZED` in that loop covers it. Worth pinning because the comment above is the reason the string is this one rather than the vendor's page: a later reader who takes "Unauthorized" for the `require_basic_or_bearer` detail this PR deletes can change it and hear nothing back.

nuri-yoo (commented):
Re-ran everything at `f51e1d3`: `2230 passed, 0 skipped, 0 failed` (junit counts, `-p no:randomly`), and `ruff check`, `ruff format --check` and `scripts/gen_docs.py --check` each exit 0.

All four body notes are addressed — `test_mcp.py` names `atlassian_caller`, the body names the four tests, the auth guide's new clause is sharper than what I asked for ("an empty listing is also the honest answer for a working credential over a corpus it can read nothing in" is the half I could not phrase), and the Confluence one is #138. I left a correction on #138: the addresses are not the disclosure, since `POST /slack/api/users.list` already answers any scoped caller with every address on the same corpus (measured, `ok=true, n=2, ['ava@acme.com', 'bob@acme.com']`). The space-to-reader association is the part that is new, and the issue's title and its "no reading under which it should be served" sentence both claim more than that. Every other reason in it holds, and GitHub turns out to be a second precedent that states the rule in words — details there rather than here.

Two more below, both found by mutation against the full suite at `f51e1d3` rather than by reading, and both in code this PR adds. Neither blocks and the approval stands.

The body's line 25 still reads "Backlot's shared `bearer_token` accepts four of them", which is the same count as the `auth.py` thread above — five, not four.

backlot/routers/atlassian.py
nuri-yoo: Refining this — the idiom is the repo's rather than new here, and the difference is which source has a case for it.

`github._repo_visible:715` is the same shape: `ids is None or store.count_documents(conn, "github", repo, ids) > 0`. There it is live, because GitHub has a record that creates a container holding no document. `github.schema.json` on `subtype: repo`: "the repo stays visible to a scoped caller exactly when one of its documents is, and to the admin as soon as the record itself exists." Measured, on a corpus of one `subtype: repo` record for `pipeline` plus an issue in `docs-site`:

```
admin repos          ['docs-site', 'pipeline']
ava@acme.com repos   ['docs-site']
has_visible_document(db, 'github', 'pipeline', None)  ->  False
```

so dropping the short-circuit there would cost the admin the repo. Neither `jira.schema.json` nor `confluence.schema.json` has a container-only record — `project`/`space` is required on every record and no subtype omits the document — so on Jira there is nothing for the branch to catch, which is why removing it leaves the full suite green at `f51e1d3` (`2230 passed, 0 failed`).

Keeping it for symmetry with `_repo_visible` is a fine answer. It is the comment that then overstates: it names a listing case Jira cannot present, where the honest version is that the branch mirrors GitHub's.

While this is open — `_repo_visible:715` and `_visible_repos:728` still ask the question by counting, where `has_visible_document` now exists. The two agree on all three ACL states, checked for both repos of the bundled corpus over `None`, the empty set and a scoped set. The helper's own docstring names the Jira listing as its caller and GitHub's is the same question, so it is a rename each. Fine as its own change.

backlot/auth.py
nuri-yoo: Not this PR, but the same sentence next door is off in the same direction, which is worth knowing before this one is reworded. `slack_bearer_token` says sharing `bearer_token` "would let Backlot authenticate six spellings live Slack refuses outright". Its own table lists five that Slack refuses — `bearer <t>`, `BEARER <t>`, `token <t>`, `Bearer<TAB><t>` and `Bearer<t>` — and `bearer_token` takes four of them, since it does not take `Bearer<t>` either. Same counting method as above, run the same way. On `main` and outside this change, so I have left it alone.

khj809:
All seven addressed in b138300. I ran each claim rather than reading it — your counting method, and it changed one of my answers.

**Counts.** Five, confirmed by running both parsers over the table: `bearer_token` takes `bearer <t>`, `BEARER <t>`, `token <t>`, `Bearer  <t>` and `Bearer<TAB><t>`, and not `OAuth <t>`, `Bearer<t>` or the bare `Bearer`. The docstring now names the three it does not take instead of stating a bare number, so the next reader can check it without running anything. PR body and test docstring agree at five.

**`slack_bearer_token`'s "six".** Fixed in the same commit rather than left, since it is the identical error one function away and the method is now settled: refuses five, `bearer_token` takes four of them (all but `Bearer<t>`). Say the word if you would rather it moved out of this PR.

**`rest.rstrip()`.** Gone. No input separates it — the entry `strip` already serves the `Bearer <t>' '` row, and the docstring now says so, since that is the thing a reader would otherwise re-add.

**"a bearer OAuth token".** Removed from `atlassian_caller` and from the test docstring. It now says what the PR settled: a 3LO token goes to `api.atlassian.com/ex/jira/{cloudid}/…`, which Backlot does not serve, and on the surface it does serve a bearer is read as a Connect session JWT.

**`ids is None or`.** Kept, comment rewritten to your reading — it mirrors `github._repo_visible`, where the `subtype: repo` record makes it load-bearing, and no Jira record can create an empty container because `project` is required on every one. I verified the schema halves: github requires `['repo', 'source_type']`, jira and confluence both require the document fields. The comment no longer names a Jira listing case that cannot happen. `_repo_visible` / `_visible_repos` still counting is #139, with your `pipeline` measurement and the warning that the short-circuit must survive the swap there.

**The JWS row.** Added, and it earns its place: teaching `atlassian_bearer_unreadable` to treat a three-segment token as readable turns it red. It also pins Confluence's own `403` for the same token.

**`CONFLUENCE_UNAUTHORIZED`.** Your mutation found more than one gap. My first attempt was `assert r.json()["message"] == errors_atlassian.CONFLUENCE_UNAUTHORIZED`, which compares the body against the same constant the router emits — it passes whatever that constant says, and `-> ""` stayed green. The value is the measured part, so it is now asserted as the literal `"Unauthorized"`, and `-> ""` fails five parametrize rows.

Worth flagging: **`CONFLUENCE_FORBIDDEN` had the same hole**, despite the four sites. All four compared the constant against itself, so replacing its value with `"mutated"` left the suite green too. One of those four now asserts the vendor's wording literally, and that mutation fails. The other three keep the constant, which is the thing they are for — that the router reaches for the right one.

Suite `2394 passed, 3 skipped`; the three are the optional extras from my earlier note. `ruff check` and `ruff format --check` clean, jira and confluence baselines `0 new`.

Also, while integrating: main moved twice under this branch and `backlot/main.py` conflicted with #131's `resolve_github_id_paths` — both sides added a middleware in the same place. Resolved by keeping all three, each with its own `return await call_next(request)`; the prefixes are disjoint so ordering does not matter. Unrelated to any of the above, but it is in `da7872e` if you want to check the resolution.

PR #140, opened 2026-09-04
atlassian: a Confluence space is listed and readable exactly when the caller can read a page in it

### LINKED ISSUE #138: Confluence: the space listing and its reader roster are not ACL-scoped, so a caller is shown spaces they can read nothing in
The Confluence space listing carries no ACL filter, so a caller is shown every space in the corpus — including ones holding no document they may read. `GET /wiki/rest/api/space/{key}/permission` then answers that caller `200` with the space's readers, which associates a named person with a space they cannot see into.

The addresses themselves are not the new part, and an earlier draft of this issue claimed they were. `POST /slack/api/users.list` already answers any scoped caller with every address in the corpus — measured on the same two-record corpus below: `ok=true`, `n=2`, `['ava@acme.com', 'bob@acme.com']`. What these routes add is the **association**: which people read which space, for a space the caller reaches no document in. Whether the roster should be readable at all is the question, not whether the address is.

Raised in review of #134, which fixed the same rule on the Jira side. Not a regression from that PR: the listing is unfiltered on `main` today, and an anonymous caller never reaches it because `_confluence_caller` refuses first. What #134 changes is that the two APIs now disagree, and that the helper which closes this exists.

## Reproduced

Corpus: one page in `engineering` authored by `ava` and public, one in `secrets` authored by `bob` and private; the same split in Jira as `payments` / `secretproj`. Read as `ava`, at 2170295:

```
ava   spaces=['ENG8C85B6', 'SEC20E8CD']   readable_pages=['Runbook']   projects=['payments']
```

The Jira project she can open nothing in is dropped. The Confluence space is not, and neither space-scoped route refuses her:

```
GET /wiki/rest/api/space/SEC20E8CD             -> 200  {"id":1364036,"key":"SEC20E8CD","name":"secrets",...}
GET /wiki/rest/api/space/SEC20E8CD/permission  -> 200  {"results":[{"operation":{"operation":"read","targetType":"space"},
                                                        "subjects":{"user":{"results":[
                                                          {"accountId":"5b903d2879cd8995116bcd16","email":"bob@acme.com"}]}}}]}
```

`GET /wiki/rest/api/content?spaceKey=SEC20E8CD` is correctly empty — the document reads are all ACL-scoped. It is the three space-level routes that are not: the listing, the single-space get, and the permission read.

## What is measured, and what is not

Nothing here. The anonymous caller cannot be used to measure it, because Confluence answers `403` to an anonymous request before any space is resolved (measured on both sites for #134), and a scoped measurement needs two accounts plus a space permission scheme on a live site — the same gap #134 documents on `_reachable_projects` for Jira's browse permission. So closing this is a decision about Backlot's own ACL, informed by what Confluence's space permissions mean, not a row to be copied off a probe.

The association is the more tractable half: whatever a space listing should hold, a reader roster for a space the caller reaches no document in is the one answer no reading of Confluence's space permissions supports. GitHub is a second precedent that states the rule in words rather than leaving it to be inferred — `github.schema.json` on `subtype: repo` says a container "stays visible to a scoped caller exactly when one of its documents is", which is the rule `_reachable_projects` now applies to Jira.

## Closing it

`store.has_visible_document(conn, "confluence", container, ids)` — added in #134 for exactly this shape, `LIMIT 1` under the ACL clause — answers all three routes:

- `GET /wiki/rest/api/space` lists a space only when the caller can read a page in it, as `atlassian._reachable_projects` now does for Jira.
- `GET /wiki/rest/api/space/{key}` and `.../permission` refuse a space the caller reaches nothing in, the way `atlassian._require_project` refuses a project. Confluence's own not-found for a space is `404 {"message": "No space with the given key exists"}`, which the router already emits for an unknown key — an unreachable space reporting as an absent one matches how `issue/{key}` and `content/{id}` already answer here.

Whatever is chosen for the scoped case has to be written down as unmeasured, as #134 did, rather than attributed to the vendor.

khj809:
Closes #138.

## What changed

A Confluence space is visible to a caller exactly when one of its pages is — the rule #134 gave Jira and `github._repo_visible` already stated in words. `GET /wiki/rest/api/space` lists only the spaces the caller can read a page in (`_reachable_spaces`, the Confluence half of `_reachable_projects`), and `GET /wiki/rest/api/space/{key}` and `.../permission` answer a space the caller reaches no page in with Confluence's own `404 {"message": "No space with the given key exists"}` (`_require_space`, the Confluence half of `_require_project`), so an unreachable space reports as an absent one the way `issue/{key}` and `content/{id}` already do. The roster on `.../permission` is what makes the refusal matter: it names the people who hold `read` on a space, and that association is the one thing a caller cannot already read elsewhere — the addresses themselves are served to any scoped caller by `slack/api/users.list`.

Both space reads give a key that names nothing the same one answer, which is what makes the refusal hold: `.../permission` used to report an unresolvable key as `200 {"results": []}`, and leaving it there would have let a caller read a space's existence off the difference between that and the 404 an unreachable space now draws.

`_require_space` keeps `_require_project`'s `ids is None` short-circuit, so the two Atlassian readers are spelt the same way. The branch is load-bearing only on GitHub, where a `subtype: repo` record creates a container holding no document; `confluence.schema.json` requires `space` on every record and has no container-only subtype, so no space exists without a page. That schema now states the visibility rule in words, as `github.schema.json` does for a repo.

Not measured, and it cannot be from this side: Confluence answers an anonymous request `403` before any space is resolved (measured on brekkylab.atlassian.net and ecosystem.atlassian.net on 2026-09-04 for #134), and a scoped measurement needs two accounts plus a space permission scheme on a live site. Real Confluence lists a space by its space permissions, and a space can grant `read` while showing the caller no page, so a real listing can hold a space this one drops — the docstrings say so rather than attributing the rule to the vendor. What is settled is the other direction, which no reading of Confluence's space permissions supports: a caller who can open nothing in a space is not shown it, and is not handed its reader roster.

## Verification

- **Tests** — `pytest` on a `.[all]` install, rebased on `89f15ea`: `2625 passed, 3 skipped`. The two `tests/test_docs.py` prose checks (`test_no_prose_states_the_source_count`, `test_no_prose_states_the_bundled_corpus_size`) fail on this checkout for reasons outside the branch: every offender they name is untracked local material under `.superpowers/` and `/data/`, both gitignored.
- **Optional surfaces that ran rather than skipped** — `tests/test_sdk.py`, `tests/test_mcp.py`, `tests/test_s3.py`, `tests/test_llamaindex.py` and the reader/Google tests in `tests/test_integrations.py`. The three skips are the llama-index HubSpot reader and the two mirage Google-client cases, none of them installable here.
- **Lint** — `ruff check . && ruff format --check .`: `All checks passed!`, `138 files already formatted`.

`test_atlassian_lists_only_the_containers_the_caller_can_open` now carries the same public/private split in Jira and Confluence on one corpus, and fails on `main` with `assert ['engineering', 'secrets'] == ['engineering']`. Read as the author of the public page only:

```
ava   spaces=['engineering']   readable_pages=['Runbook']   projects=['payments']
GET /wiki/rest/api/space/SEC20E8CD             -> 404
GET /wiki/rest/api/space/SEC20E8CD/permission  -> 404
```

### DIFF
--- a/backlot/routers/atlassian.py
+++ b/backlot/routers/atlassian.py
@@ -837,12 +837,47 @@ def _space_container_for_key(conn, space_key: str) -> str | None:
     return None
 
 
+def _reachable_spaces(conn, ids) -> list:
+    """The spaces the caller can reach, as container rows — `_reachable_projects` for Confluence.
+
+    A space is listed when the caller can read a page in it: the ACL grants per document, so there
+    is nothing else to read "can view this space" off. NOT measured — Confluence 403s an anonymous
+    request before resolving a space, and a scoped measurement needs two accounts and a space
+    permission scheme on a live site. Real Confluence lists by space permission, which can grant
+    `read` while showing the caller no page, so a real listing can hold a space this one drops.
+    """
+    rows = store.list_containers(conn, "confluence")
+    if ids is None:
+        return rows
+    return [r for r in rows if store.has_visible_document(conn, "confluence", r["name"], ids)]
+
+
+def _require_space(request: Request, conn, key: str) -> str:
+    """The container behind a space key the caller can reach — `_require_project` for Confluence.
+
+    Both space reads answer an unreachable key with the SAME ``404 {"message": "No space with the
+    given key exists"}`` a key naming nothing gets, so neither confirms the space exists — the
+    roster on ``.../permission`` is what makes that worth withholding, since it names who reads a
+    space the caller cannot open. Unmeasured for a scoped caller, see :func:`_reachable_spaces`.
+    """
+    ids = auth.visible_ids(request, _confluence_caller(request))
+    container = _space_container_for_key(conn, key)
+    # The `ids is None` short-circuit is load-bearing only on GitHub (see `_require_project`'s
+    # note): `confluence.schema.json` requires `space` on every record, so no space exists
+    # without a page, and here it keeps the readers of the two APIs alike.
+    if container is not None and (
+        ids is None or store.has_visible_document(conn, "confluence", container, ids)
+    ):
+        return container
+    raise HTTPException(status_code=404, detail="No space with the given key exists")
+
+
 @router.get("/wiki/rest/api/space", response_model=ConfluenceResults)
 async def confluence_spaces(request: Request):
     conn = auth.conn(request)
-    _confluence_caller(request)
+    ids = auth.visible_ids(request, _confluence_caller(request))
     results = []
-    for r in store.list_containers(conn, "confluence"):
+    for r in _reachable_spaces(conn, ids):
         key = synth.confluence_space_key(r["name"])
         results.append(
             {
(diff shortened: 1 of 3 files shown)


nuri-yoo (approved):
Approving. The change does what the title says and I could not find a fault in it. What follows is one body cleanup and two nits, none blocking.

**Verification.** On a `.[all]` install: the new test run against `main` (89f15ea) fails with `assert ['engineering', 'secrets'] == ['engineering']`; the full suite on the PR head is 2577 passed, 1 skipped (Docker), plus one `test_docs` failure that is my worktree importing the editable-installed main checkout and passes with `PYTHONPATH=$PWD`; `ruff check`, `ruff format --check` and `gen_docs.py --check` pass. Beyond the suite, on a corpus with a public space, a private space, a space whose only page has `readers: []` and a `subtype: repo` record: the scoped caller is refused on all three space routes for the space she reaches no page in, the literal container name as a key is refused the same way, the two space reads return the same 404 body as a key naming nothing, `content?spaceKey=` and CQL `space=` answer an unreachable space and an unknown one with byte-identical bodies, the anonymous listing is still 403, and the admin reads the `readers: []` space.

**The github.py hunk is already on main.** #144 (89f15ea, merged today) made the same three `count_documents` -> `has_visible_document` changes, character for character; merging `main` into this branch leaves `git diff main -- backlot/routers/github.py` at zero lines, and the merged tree passes `tests/test_atlassian.py` and `tests/test_github.py` (238 passed). Nothing to do in code, git resolves it. But after the merge the fourth paragraph of "What changed" describes no change in this PR, and its "bundled corpus's 23 repos" does not reproduce from anything in the repo: `backlot/data/hello.jsonl` has 2 repos, `examples/bring-your-own-corpus/sample_corpus.jsonl` 1, the test SAMPLE 2. Deleting the paragraph settles both.

**nit: the two Atlassian readers choose oppositely on `ids is None`.** `_require_project` keeps `ids is None or ...` "only to keep the two readers alike"; `_require_space` drops it on the same reasoning. I checked that no page-less Confluence container can exist, so there is no behavior difference either way: containers are registered only from records (the `containers[(src, container)] = group` line in `byo.py`), the only early return I found in the per-record path before the document write is GitHub's `subtype: repo`, the only document-deleting path is gmail/slack thread trimming, and `--append` refuses re-homing a page (`content_id ... is already claimed by a previous import`). Purely style; one choice for both would read better.

**nit: the scoped-caller refusal is tested on the synthesized key only.** `_space_container_for_key` also resolves the literal container name, so adding `/wiki/rest/api/space/secrets` and its `/permission` to the loop in the `shut` block pins that path too. Both answer 404 for the scoped caller on my run.

PR #142, opened 2026-09-06
linear: the inbox notification root real serves is an acknowledged gap

### LINKED ISSUE #141: Fidelity: linear diverges from its vendor (first seen 2026-09-05)
`backlot diff --source linear` reports divergences the baseline does not acknowledge. Each is either a bug in Backlot or a deliberate gap; a gap gets a note and an entry via `backlot diff --source linear --update-baseline`. This body is the latest scheduled run's report, rewritten in place.

```
   775 divergence(s) · 774 acknowledged · 1 new

     missing_field  Query.agentSessionSshAddress
       vendor serves agentSessionSshAddress: String; Backlot does not

   1 new. Fix them, or run --update-baseline to acknowledge the gaps.
```

khj809:
`backlot diff --source linear` reported one new divergence: Linear now serves `Query.inboxNotifications: NotificationConnection!`, a root Backlot does not have. It joins the notification family — `notification`, `notifications`, `notificationsUnreadCount`, `notificationSubscription`, `notificationSubscriptions` — which the baseline already carried as gaps, so it is acknowledged rather than served, and the reasoning that was implicit for the other five is now written down on `Query.notifications` where a reviewer meets it.

No corpus states a notification. Records carry issues, comments, teams and users, and nothing that says who was told about what, when they read it, or whether it is snoozed. Measured against api.linear.app on 2026-09-06, `Notification` is an interface over 13 types whose 28 fields are almost all per-viewer event state — `readAt`, `emailedAt`, `snoozedUntilAt`, `unsnoozedAt`, `actor`, `groupingKey` — so serving any one root commits Backlot to an inbox it would have to invent. An empty connection is not a free stand-in: the workspace measured has no activity beyond its own creation and still answered a `WelcomeMessageNotification` from both roots, with `notificationsUnreadCount: 1`.

`inboxNotifications` gets its own note because a client cannot swap the two roots: it answers the same connection over the same nodes but takes only `first`, `after` and `unreadOnly`, and `filter`, `last`, `orderBy` and `includeArchived` — all accepted by `notifications` — each answer 400 `GRAPHQL_VALIDATION_FAILED`. Neither root is deprecated.

`backlot diff --source linear` now reports 776 acknowledged and nothing new. `pytest`: 2396 passed, 1 skipped.

Closes #141

### DIFF
--- a/backlot/fidelity/baseline/linear.json
+++ b/backlot/fidelity/baseline/linear.json
@@ -1,7 +1,7 @@
 {
   "source": "linear",
   "endpoint": "https://api.linear.app/graphql",
-  "measured": "2026-09-03",
+  "measured": "2026-09-06",
   "acknowledged": [
     {
       "kind": "missing_field",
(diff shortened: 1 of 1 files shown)


nuri-yoo (approved):
LGTM

PR #144, opened 2026-09-06
github: /search/code refuses a page value it cannot parse in text/plain, every search stops at the first 1000 results, and the repo visibility checks read existence instead of counting (#137, #139)

### LINKED ISSUE #137: github: /search/code refuses a page value the rest of the API absorbs, and refuses it in text/plain
**Source and commit** — `github`, at `79ccdac` (`main`).

`_absorb_page` takes every unparseable `page`/`per_page` on the GitHub router and hands the handler a default, on the premise that real refuses none of them. Real refuses them on one surface. `/search/code` answers `400` to a page value that is not a non-negative integer, and answers it in `text/plain` rather than the JSON envelope #104 built for this router. The premise is written down twice, and both docstrings state it without the route qualifier the measurement behind it carried.

|---|---|---|---|---|
| 1 | `GET /search/code?q=…&per_page=abc` | 400, `text/plain` | 200 with results | yes |
| 2 | `GET /search/code?q=…&per_page=-1` | 400, `text/plain` | 200 with results | yes |
| 3 | `GET /search/code?q=…&per_page=1.5` | 400, `text/plain` | 200 with results | yes |
| 4 | `GET /search/code?q=…&page=abc` | 400, `text/plain` | 200 with results | yes |
| 5 | `GET /search/code?q=…&page=-1` | 400, `text/plain` | 200 with results | yes |

Every other surface agrees with Backlot: `/repos/{o}/{r}/tags`, `/repos/{o}/{r}/issues`, `/user/repos` and `/search/issues` are each a 200 for all five values, which is what `_absorb_page` was measured against and is right for them. `/search/code` is the exception, and nothing in the router or the baseline records it — the divergence is in a runtime answer, so `backlot diff` cannot see it.

**How you know** — measured against api.github.com on 2026-09-04.

- `/search/code?q=repo:psf/requests+def&per_page=abc` answers `400` with `content-type: text/plain; charset=utf-8` and the body `Failed to deserialize query string: per_page: invalid digit found in string`. `per_page=-1` and `per_page=1.5` answer the same line; the `page` variants answer it with `page:` where it says `per_page:`.
- `per_page=0` and `page=0` are both `200` on `/search/code`. Zero parses, so what is refused is the parse and not the range.
- The same five values on `/repos/psf/requests/tags`, `/repos/psf/requests/issues`, `/user/repos` and `/search/issues?q=repo:psf/requests+timeout` are `200` each, all twenty.
- The body is not GitHub's error envelope. There is no `message`, no `documentation_url`, no `status`, and the content type is not JSON — so a client calling `.json()` on the response raises rather than reading an error object, which is a different failure from every other GitHub error it handles.

**Reproduction** — the bundled corpus, no fixtures:

```python
import backlot, requests

with backlot.serve() as m:
    for bad in ("per_page=abc", "per_page=-1", "per_page=1.5", "page=abc", "page=-1"):
        k, v = bad.split("=")
        r = requests.get(
            f"{m.base_url}/github/search/code", headers=h, params={"q": "extension:md", k: v}
        )
        print(f"{bad:14} {r.status_code} total_count={r.json().get('total_count')}")
```

All five print `200 total_count=2`, where real answers `400` to each.

**What it costs a client** — a page size is often computed rather than typed: a config value that arrives as a string, an offset arithmetic that yields `-1`, a division that yields `1.5`. Against Backlot that request is a normal search with results, so the client's error path for it is never reached; against real the same request is a 400 whose body will not parse as JSON. A client whose retry or fallback logic hangs off that error path is exercised nowhere in a Backlot-backed test suite, and `_absorb_page`'s own reasoning — that refusing "would hand a paginator computing an edge value a hard error where production absorbs it" — is exactly inverted here: on this route production is the hard error.

**Where it lands** — `_absorb_page` in `backlot/pagination.py` and `validation_body` in `backlot/errors/github.py`. The first says "Real GitHub refuses no value here" and then scopes its own evidence in the same paragraph, "(measured against a public repository's issue listing)". The second says "Real refuses no pagination value at all, which is why those now absorb … and never reach here", and it is the function that would decide the answer if `/search/code` stopped absorbing, so it is the one whose claim is load-bearing rather than only descriptive. `search_code` in `backlot/routers/github.py` is the single route that carries the divergence; `PageParam` is shared, so a fix cannot simply drop the validator without moving every other route with it.

**What a fix has to settle**

- Whether to refuse on this route or only to correct the two docstrings. Correcting them alone leaves a measured 400 unimplemented but costs nothing and can be done in a line; refusing makes the route match and is the project's usual direction, but it is the first place on this router where a client meets a non-JSON error.
- What shape the refusal takes if it is built. Real's is `text/plain` with no envelope, so it cannot come from `errors/github.py::validation_body`, which returns a dict that `_validation_exception_handler` JSON-encodes. Handing this route plain FastAPI validation instead does not reach real's answer either — that is a 422, not a 400, and it goes through the same JSON path.
- Whether the OpenAPI slice should change with it. `_absorb_page` is a `BeforeValidator` precisely so the parameter stays an integer in the schema, which is what real's spec declares and what `backlot mcp` hands an agent. Refusing at runtime on this one route does not change the declared type, so the slice would stay as it is either way — worth stating in whichever docstring survives, since the next reader will ask.
- Whether `page=0` and `per_page=0` staying `200` is worth a test of its own. They are the boundary that separates "the parse failed" from "the value is out of range", and they are the only two values on this route where Backlot and real already agree by accident rather than by rule.

### LINKED ISSUE #139: github: the repo visibility checks count every document where an existence read answers them
`github._repo_visible` and `github._visible_repos` ask "can this caller see anything in this repo?" by counting every document in it. `store.has_visible_document`, added in #134, is that question as `SELECT 1 … LIMIT 1`: no ORDER BY, and SQLite stops at the first row the ACL admits instead of totalling the container for a number neither caller uses.

Raised in review of #134, where the same swap was made on the Jira side.

## The two call sites

```
backlot/routers/github.py:876  def _repo_visible(conn, repo, ids)
    return ids is None or store.count_documents(conn, "github", repo, ids) > 0
backlot/routers/github.py:899  (inside _visible_repos)
    repos = [n for n in repos if store.count_documents(conn, "github", n, ids) > 0]
```

`_visible_repos` runs one count per repo, so a repo listing walks every visible document in every repo of the corpus.

## Measured

On the Jira listing this replaced, over 12 projects x 900 issues: **6.2 ms/req before, 1.8 ms/req after** for a scoped caller. Cost grows with documents-per-container, so the saving on a served corpus is larger than the fixture shows.

The two helpers agree on all three ACL states — checked across `None`, the empty set, and a scoped set. Worth keeping in mind while swapping: the `ids is None` short-circuit in `_repo_visible` is **load-bearing on GitHub** and must stay. A `subtype: repo` record creates a container holding no document (`github.schema.json` requires only `repo` and `source_type`), and `github.schema.json` states the intended rule: the repo "stays visible to a scoped caller exactly when one of its documents is, and to the admin as soon as the record itself exists". `has_visible_document(db, "github", "<container-only repo>", None)` is `False`, so dropping the short-circuit would cost the admin that repo. Measured on a corpus of one `subtype: repo` record for `pipeline` plus an issue in `docs-site`:

```
admin repos          ['docs-site', 'pipeline']
ava@acme.com repos   ['docs-site']
```

That case is why the branch reads as a no-op on Jira, where no record can create an empty container — #134's comment on `_require_project` now points here for it.

## Not a behaviour change

Both forms answer the same on every state, so this is a swap plus a test that the container-only repo still reaches the admin — which `github.schema.json` already asserts in prose and nothing pins.

nuri-yoo:
Closes #137. Closes #139. Sits on #152 (merged 2026-09-07), which set the GitHub page sizes to real's 30 and 100. Five commits carry the change: one per issue, a third for what the #137 measurement turned up on the same route and its sibling, the depth every search stops at and the version header code search does not read, and two from a second measurement on 2026-09-07 that found three answers the first commit got wrong (the last section). After them, merges of `main`, four commits that moved the two search routes onto #152's `_clamp` and the depth tests onto real's numbers, and the 2026-09-08 review's corrections (the section before Verification).

**What GitHub does on `/search/code` (#137).** Re-measured against api.github.com on 2026-09-06 with `q=repo:psf/requests+def` rather than taken from the issue, and the issue's five rows hold: `per_page=abc`, `per_page=-1`, `per_page=1.5`, `page=abc` and `page=-1` are each a 400 with `content-type: text/plain; charset=utf-8` and no envelope, the body a Rust deserializer's own line, `Failed to deserialize query string: per_page: invalid digit found in string` (or `page:`). The measurement went past the five to find the rule, which is Rust's `u32::from_str`: an optional single `+` and then one or more ASCII digits. Anything else is the invalid-digit line (`5abc`, `1e2`, ` 5`, which is how an unencoded `+5` arrives after decoding, a `+` on its own, `++5`, and the Arabic-Indic `١` and fullwidth `１` that Python's `int()` would read as 1), while an encoded `%2B5` is a 200 served at 5 and `page=%2B2` is page 2 (2026-09-07); an empty value is `cannot parse integer from empty string`; `4294967296` and anything larger is `number too large to fit in target type`, a value of 5000 digits included, while `4294967295` is a 200; `0`, `01` and 5000 zeros are a 200 (`01` served as 1), so what is refused is the parse and not the range or the length. The parameter given twice is `duplicate field `per_page``, reported at the second occurrence, so `per_page=5&per_page=abc` is the duplicate and `per_page=abc&per_page=5` the invalid digit, and `q` given twice is `duplicate field `q`` from the same deserializer, a blank `q=&q=` included (2026-09-07), where `sort` given twice is a 200. The first failure in query-string order is the one named (`page=abc&per_page=abc` names `page`, the reverse `per_page`, `q=a&per_page=abc&q=b` names `per_page`), it comes after the credential (no token with `per_page=abc` is the 401) and before the query (a blank `q` beside `per_page=abc` is this 400, not the 422), and `sort=abc`, `order=abc` or an unknown parameter beside a good value change nothing. The 400 carries the same headers as the route's 200 minus `Link`. On the other side, the same values are a 200 on `/search/issues`, `/search/repositories`, `/search/users`, `/search/commits`, `/search/labels`, `/search/topics`, `/repos/{o}/{r}/issues`, `/repos/{o}/{r}/tags` and `/user/repos`, so `_absorb_page` is right everywhere but here.

**What changed for #137.** `github_code_search_query_refusal(query_params)` in `backlot/pagination.py` returns the body real would answer or `None`, walking the query string in order and checking `q`, `page` and `per_page` for a repeat and the two page values for the u32 parse (a single leading `+` allowed, the size compared without leading zeros and by length before `int()` is asked, since Python refuses to convert more than 4300 digits and the route answered that with a 500), and `search_code` asks it before reading `q` and returns a `PlainTextResponse(…, status_code=400)` when it answers. `_absorb_page` drops leading zeros before its own parse for the same reason, so `0…05` is 5 here as on real. Nothing else moves: `PageParam`'s `BeforeValidator` still runs and still absorbs, so the route's OpenAPI slice still declares an integer, which the issue asked to keep and a test now asserts; `validation_body` is never reached by this route, as before. The two docstrings the issue named are corrected. `_absorb_page` now says which nine surfaces were measured to absorb, names `/search/code`, the tenth measured, as the one that refuses, and points at the refusal; `validation_body` no longer says "Real refuses no pagination value at all" but that the listings do not and the one measured route that does answers in text/plain from the route itself, after the validator has absorbed the value rather than before it. The existing test's first line said the same over-broad thing and is corrected too.

**What changed for #139.** `_repo_visible`, `_visible_repos` and the `repo:` qualifier's check in `_qual_repos` (the third site with the same shape, which the issue did not list) ask `store.has_visible_document` instead of `store.count_documents(…) > 0`. The `ids is None` short-circuit stays, and its docstring now says why it is load-bearing rather than a fast path: a `subtype: repo` record makes a container with no document, `has_visible_document` is False for it under every ACL including the admin's, and `github.schema.json` promises the admin that repo. Measured on 12 repos × 900 issues with a scoped caller who sees half: the listing's visibility pass goes from 2.52 ms to 0.08 ms (median of 50, store-level, the count and the existence read agreeing on every repo).

**What GitHub does past the first 1000 results, and to the version header on code search (third commit).** Every search stops at 1000 results whatever `total_count` says, and the two search backends draw the line differently (measured 2026-09-06 on an issue search of 2,813,432 results, a code search of 294,649,856, and the 846- and 44-result searches above). `/search/issues` refuses the page that STARTS past 1000, `(page - 1) * per_page >= 1000`: at 30 a page, page 34 (results 991 to 1020) is served in full and page 35 refused, at 7 a page, 143 (995 to 1001) is served and 144 refused, at 100 a page, 10 is served and 11 refused, and the 846-result search refuses page 11 at 100 a page just the same after serving page 10 empty. Its 422 is `{"message": "Only the first 1000 search results are available", "documentation_url": "https://docs.github.com/v3/search/", "status": "422"}`, no `errors` array, and `/search/repositories` answers the same. `/search/code` refuses the page that would REACH past 1000, `page * per_page > 1000`: page 33 at 30 is served and 34 refused, 142 at 7 served and 143 refused, 1000 at 1 served and 1001 refused, and `per_page=101` behaves as 100 because that is what is served. Its 422 is `{"message": "Cannot access beyond the first 1000 results", "documentation_url": "https://docs.github.com/rest/search/search#search-code", "status": "422"}`. The order differs too: on `/search/issues` the blank-`q` and unsearchable-`repo:` 422s come before the depth's, on `/search/code` the blank-`q` 422 comes before it and the `repo:` qualifier after (`repo:psf/ghost-zz-9876` at page 11 of 100 is the depth 422, where on page 1 it is the empty `incomplete_results: true` 200). Both routes' `Link` stops at the depth, `ceil(1000 / per_page)`: last=10 at 100, 34 at 30, 143 at 7, and for code 1000 at 1, with `prev, first` alone on that page for issues; for code that means page 33 at 30 links next=34 and last=34, a page the same route refuses, and this PR reproduces that rather than tidying it. Separately, code search does not read `X-GitHub-Api-Version` at all: a pinned `1999-01-01` or `garbage` is a 200 where `/search/issues` beside it answers the version 400, a pinned `2026-03-10` is a 200, and no response from the route, 200, 400 or 422, carries `X-GitHub-Api-Version-Selected`.

**What changed in the third commit.** `GITHUB_SEARCH_RESULT_CAP = 1000` in `pagination.py`, read by two functions beside it and nowhere else: `github_search_depth_refused(page, per_page, code=)` holds the two inequalities and `github_search_last_page(per_page)` the `ceil(1000 / per_page)` a `last` link stops at, so the router holds no copy of the number (the first version imported it by value, and the 422 and the `Link` read two copies). `_search_beyond_first_results(code=)` builds the two 422s; `search_issues` refuses after the `repo:` qualifier and before the search runs, `search_code` after the blank-`q` 422 and before the qualifier is read, each by its own inequality, in the size the route applies. `github_link_header` takes `max_page`, which `_search_paged` passes for the issue search and no listing passes; `github_code_search_link_header` caps its own `last`. `honours_api_version(request)` in the router is False for `/github/search/code` alone; `_validate_api_version` asks it before the version 400 and the echo middleware in `backlot/main.py` asks it before adding the header.

**What #137 asked a fix to settle.** Refuse rather than only correct the docstrings, and in the shape real has, which is a `PlainTextResponse` from the route and not anything `errors/github.py` can render. The OpenAPI slice stays an integer, stated in `_absorb_page`'s docstring and asserted. `page=0` and `per_page=0` staying 200 have their test, beside `01` and the u32 maximum.

**Tests.** `_CODE_SEARCH_PAGE_REFUSALS`, 28 measured refusals as one parametrized test asserting status, exact content type, exact body and that `.json()` raises; one test for what is not refused (`0`, `01`, `4294967295`, `%2B1`, `%2B100`, 5000 leading zeros, an unknown parameter, `sort` twice), `page=%2B2` serving page 2, `q=&q=` as the 400 and not the 422, `/search/issues` still absorbing a bad size and a repeated `q`, and the integer OpenAPI slice. For #139, one test on a corpus of a record-only `pipeline` beside a `docs-site` issue: `/user/repos` and `/orgs/{org}/repos` list both for the admin and `docs-site` alone for the scoped caller, `/repos/{org}/pipeline` is the admin's 200 and the scoped caller's 404, the `repo:pipeline` issue search is the admin's empty 200 and the scoped caller's 422 (the one a repo that does not exist gets, so the record's existence is not confirmed), and at the store level `has_visible_document` is False on `pipeline` under all three ACL states while agreeing with the count on `docs-site`. For the third commit: the code depth at 100, 30, 7 and 1 a page on both sides of the line, the issue depth the same, the applied-size cases (`per_page=0`, `per_page=101`, no size) at real's 30 and 100, the boundary pages written down as real's (33 served and 34 refused on code search, 34 and 35 on issue search), the order of each route's refusals, the version header on code search under four pinnings and on its 400 and 422s, the two Link builders against the measured `last` values at the real depth, and the routes' wiring of the depth into their `Link` and 422 with the depth lowered to 2, by the one name in `pagination.py` both read. With `routers/github.py`, `pagination.py`, `errors/github.py` and `main.py` at `bab49cb` and the tests kept (measured at the fifth commit, before the branch took #152): `20 failed, 1 passed` over the #137 and #139 tests, the one passing being the #139 test, since the swap is behaviour-preserving and the short-circuit it pins already existed; at `9c70341` (the second commit) `13 failed, 8 passed` over the third commit's tests, the 8 being the served side of each depth. Mutation-checked from a copy and restored: dropping the `ids is None` short-circuit or filtering the admin in `_visible_repos` each fails the #139 test; reverting the swap to the count fails nothing, as the issue says it should not; not asking the refusal, or raising it as a JSON envelope, fails 20; dropping the duplicate, empty, too-large or `per_page` lines fails 1, 2, 3 and 14; replacing the ASCII-digit check with `str.isdigit` survived until the three non-ASCII-digit cells were measured and added, and now fails 3; moving `GITHUB_CODE_SEARCH_PAGE_MAX` down by one fails 1; from the 2026-09-07 commit, dropping the leading `+`, the length guard, or the leading-zero strip in the refusal or in `_absorb_page` fails 1 each, dropping `q` from the duplicate check fails 3, swapping the code depth's inequality for the issue search's fails 2 and uncapping `github_search_last_page` fails 2; each depth inequality moved by one or swapped for the other route's fails 2 or 3, the two 422 bodies swapped fails 9, checking the code depth after the qualifier fails 6, restoring the version 400 or the echo on code search fails 1 each, and dropping the depth from either Link builder or from the issue route's call fails 1 (the last survived until the lowered-depth wiring test was added). One mutation survives by design: moving `_require` below the refusal in `search_code` changes nothing, because the router's dependencies answer the 401 before any handler runs. #137's reproduction prints five `400 text/plain` lines here where `bab49cb` prints five `200 total_count=2`.

**Second review, 2026-09-08.** Three suggestions applied as-is: the `search_issues` depth comment and the `search_code` docstring lose the clause that placed the unsent-size refusal at page 11 (it described the gap #152 closed; both now refuse where real does, 35 and 34), and `_absorb_page` says "the tenth measured surface" where "those ten" had nine antecedents. Three more by hand: `_version`'s docstring names `/search/code` as the route an unsupported version does reach, and says nothing there asks it; `http_body`'s docstring no longer claims every hand-shaped body carries `errors`, since the two depth 422s carry none and are hand-shaped for their `documentation_url` instead; the `X-GitHub-Api-Version` paragraph in `docs/supported-sources.md` now carries the code search exception. And the two depth tests write real's boundary pages down instead of deriving them from the router's constants, which is what #145 asked for.

**Verification.**

```
2629 passed, 1 skipped   (2026-09-08, at `9909112`; Docker down, so the one mcp test skipped)
```

**Not in this PR.** Two pre-existing gaps the measurement made visible, both wider than these routes, filed as #145 and #146. #145 was the page sizes: Backlot's were the server's settings, default 100 and cap 1000, where GitHub's are 30 and 100 on every route measured here, and since the depth is applied in the size the route serves, an unsent `per_page` would have been refused at page 11 here where real refuses at 35. It landed as #152 and this branch sits on it, so the two search routes page by `_clamp` at real's numbers, the boundary pages here are real's, and the four sites that described the mismatch while it was open are gone. Real's JSON responses carry `content-type: application/json; charset=utf-8` on every route but code search, where it is `application/json`; Backlot answers `application/json` everywhere, so code search is the one route where the two agree today (#146). The fidelity baseline is schema-shaped and carries none of this.

### DIFF
--- a/backlot/errors/github.py
+++ b/backlot/errors/github.py
@@ -178,12 +178,13 @@ def validation_body(path: str, errors) -> tuple[int, dict]:
     not it.
 
     A QUERY parameter is the residue. Real's listings refuse no pagination value, which is why
-    those absorb (see :func:`backlot.pagination._absorb_page`) and never reach here; the one route
-    that does refuse one, `/search/code`, answers in `text/plain` with no envelope and is refused by
-    the route itself before validation (see :func:`backlot.pagination.github_code_search_page_refusal`),
-    so it never reaches here either. What is left is a shape real has no measured answer for, so it
-    keeps the Validation Failed envelope real uses for a parameter it does refuse, with this route's
-    anchor rather than the bare root.
+    those absorb (see :func:`backlot.pagination._absorb_page`) and never reach here; the one
+    measured route that does refuse one, `/search/code`, answers in `text/plain` with no envelope,
+    and does so from the route itself, reading the raw query string after the validator has
+    absorbed the value (see :func:`backlot.pagination.github_code_search_query_refusal`), so
+    validation never fails there and this is never reached either. What is left is a shape real has
+    no measured answer for, so it keeps the Validation Failed envelope real uses for a parameter it
+    does refuse, with this route's anchor rather than the bare root.
     """
     if any(tuple(e.get("loc", ()))[:1] == ("path",) for e in errors):
         return 404, {
--- a/backlot/errors/github.py
+++ b/backlot/errors/github.py
@@ -140,10 +140,13 @@ def http_body(path: str, exc) -> dict | None:
     """Render an exception into the envelope.
 
     An error the router shaped by hand carries its own body as ``github_body`` — the version 400
-    and the search 422, both of which have an ``errors`` member no generic rendering could invent —
-    and that wins. Everything else is the three-member envelope over the exception's own detail,
-    whose wording is already real's ("Not Found", "Bad credentials") because the routers were
-    written against measured responses; only the shape around it was FastAPI's.
+    and the two Validation Failed search 422s, which have an ``errors`` member no generic rendering
+    could invent, and the two depth 422s, whose ``documentation_url`` is not the route's anchor in
+    :data:`ROUTE_DOCS` (the issue search's is the bare ``/v3/search/`` with its trailing slash, code
+    search's the ``#search-code`` anchor) — and that wins. Everything else is the three-member
+    envelope over the exception's own detail, whose wording is already real's ("Not Found", "Bad
+    credentials") because the routers were written against measured responses; only the shape
+    around it was FastAPI's.
 
     A 405 is the exception, and keeps FastAPI's ``detail``. Every route here declares GET, so a
     wrong method is refused by Starlette with ``http.HTTPStatus(405).phrase`` — a string no
--- a/backlot/errors/github.py
+++ b/backlot/errors/github.py
@@ -177,10 +177,13 @@ def validation_body(path: str, errors) -> tuple[int, dict]:
     query parameter at once has one answer, and which of the two FastAPI happens to list first is
     not it.
 
-    A QUERY parameter is the residue. Real refuses no pagination value at all, which is why those
-    now absorb (see :func:`backlot.pagination._absorb_page`) and never reach here; what is left is
-    a shape real has no measured answer for, so it keeps the Validation Failed envelope real uses
-    for a parameter it does refuse, with this route's anchor rather than the bare root.
+    A QUERY parameter is the residue. Real's listings refuse no pagination value, which is why
+    those absorb (see :func:`backlot.pagination._absorb_page`) and never reach here; the one route
+    that does refuse one, `/search/code`, answers in `text/plain` with no envelope and is refused by
+    the route itself before validation (see :func:`backlot.pagination.github_code_search_page_refusal`),
+    so it never reaches here either. What is left is a shape real has no measured answer for, so it
+    keeps the Validation Failed envelope real uses for a parameter it does refuse, with this route's
+    anchor rather than the bare root.
     """
     if any(tuple(e.get("loc", ()))[:1] == ("path",) for e in errors):
         return 404, {
--- a/backlot/main.py
+++ b/backlot/main.py
@@ -154,10 +154,12 @@ async def echo_github_api_version(request: Request, call_next):
     the header describes would ship without it.
 
     A rejected version gets no echo, matching real: it selected nothing. That is `None` from
-    ``selected_api_version``, the same call the router's 400 is raised from.
+    ``selected_api_version``, the same call the router's 400 is raised from. Code search gets no
+    echo either, whatever it pinned: real's code search backend does not read the header (see
+    ``github.honours_api_version``).
     """
     response = await call_next(request)
-    if request.url.path.startswith("/github"):
+    if request.url.path.startswith("/github") and github.honours_api_version(request):
         version = github.selected_api_version(request)
         if version is not None:
             response.headers[github.SELECTED_VERSION_HEADER] = version
--- a/backlot/pagination.py
+++ b/backlot/pagination.py
@@ -80,10 +80,15 @@ def clamp_limit(limit: int | None, default: int, maximum: int) -> int:
 def _absorb_page(v):
     """A page parameter's value, or ``None`` for one that will not parse.
 
-    Real GitHub refuses no value here: `per_page=0`, `per_page=abc`, `page=0`, `page=-1` and
-    `page=abc` are each a 200 with the defaults applied, and a `per_page` over the cap is a 200 at
-    the cap (measured against a public repository's issue listing). Refusing them would hand a
-    paginator computing an edge value a hard error where production absorbs it.
+    Real GitHub refuses no value on its listings and on every search but code search: `per_page=0`,
+    `per_page=abc`, `page=0`, `page=-1` and `page=abc` are each a 200 with the defaults applied, and
+    a `per_page` over the cap is a 200 at the cap (measured against a public repository's issue
+    listing; the same five values are a 200 on `/repos/{o}/{r}/tags`, `/user/repos`,
+    `/search/issues`, `/search/repositories`, `/search/users`, `/search/commits`, `/search/labels`
+    and `/search/topics`, 2026-09-06). Refusing them would hand a paginator computing an edge value a
+    hard error where production absorbs it. `/search/code` is the one surface that refuses, and
+    refuses before this validator's answer matters: see :func:`github_code_search_page_refusal`,
+    which that route asks first.
 
     A `BeforeValidator` rather than a `str` annotation: the parameter stays an integer in the
     OpenAPI schema, which is what real's own spec declares, so the tolerance is in the runtime
@@ -101,6 +106,59 @@ def _absorb_page(v):
 #: Only the GitHub surface uses it — the other vendors' answers to an unparseable page value are
 #: not measured, and a helper that spread this tolerance to them would be asserting they share it.
 PageParam = Annotated[int | None, BeforeValidator(_absorb_page), Query()]
+
+#: The largest `page` / `per_page` value real's code search parses: it deserializes both into an
+#: unsigned 32-bit integer, so 4294967295 is a 200 and 4294967296 the "number too large" refusal.
+GITHUB_CODE_SEARCH_PAGE_MAX = 4_294_967_295
+
+
+def github_code_search_page_refusal(query_params) -> str | None:
+    """The body of the 400 real's `/search/code` answers for a `page` / `per_page` it cannot parse,
+    or ``None`` when it would parse both.
+
+    Code search is the one GitHub surface that refuses a page value, and it refuses in a shape no
+    other GitHub error has: status 400, `content-type: text/plain; charset=utf-8`, no JSON envelope,
+    and a body that is a Rust deserializer's own message (measured against api.github.com on
+    2026-09-06 with `q=repo:psf/requests+def`):
+
+    * a value with anything but ASCII digits in it, `abc`, `-1`, `1.5`, `5abc`, `1e2`, ` 5` (which is
+      how `+5` arrives) and the Arabic-Indic `١` or fullwidth `１` that Python's ``int`` would read
+      as 1, is `Failed to deserialize query string: per_page: invalid digit found in string`;
+    * an empty value is `… per_page: cannot parse integer from empty string`;
+    * a value over 4294967295 is `… per_page: number too large to fit in target type`, and
+      4294967295 itself is a 200;
+    * the parameter given twice is `… duplicate field `per_page``, reported at the second
+      occurrence, so `per_page=5&per_page=abc` is the duplicate and `per_page=abc&per_page=5` the
+      invalid digit;
+    * `page` answers the same lines with `page:` in place of `per_page:`.
+
+    The first failure in query-string order is the one reported (`page=abc&per_page=abc` names
+    `page`, the reverse names `per_page`), it comes before `q` is looked at (a blank `q` beside
+    `per_page=abc` is this 400, not the 422), and it is only the parse: `0` and `01` are a 200, as is
+    `per_page=101` (served at the cap), and `sort=abc`, `order=abc` or an unknown parameter beside a
+    good page value change nothing. Every other GitHub route absorbs these values, which is what
+    :func:`_absorb_page` is for; that validator still runs on this route and its answer is unused
+    when this function refuses, so the parameter stays an integer in the OpenAPI slice, as real's
+    spec declares it, on this route as on the others.
+    """
+    seen: set[str] = set()
+    for key, value in query_params.multi_items():
+        if key not in ("page", "per_page"):
+            continue
+        if key in seen:
+            return f"Failed to deserialize query string: duplicate field `{key}`"
+        seen.add(key)
+        if value == "":
+            return (
+                f"Failed to deserialize query string: {key}: cannot parse integer from empty string"
+            )
+        if not all(c in "0123456789" for c in value):
+            return f"Failed to deserialize query string: {key}: invalid digit found in string"
+        if int(value) > GITHUB_CODE_SEARCH_PAGE_MAX:
+            return (
+                f"Failed to deserialize query string: {key}: number too large to fit in target type"
+            )
+    return None
 
 
 def clamp_page(
--- a/backlot/pagination.py
+++ b/backlot/pagination.py
@@ -107,6 +107,12 @@ def _absorb_page(v):
 #: not measured, and a helper that spread this tolerance to them would be asserting they share it.
 PageParam = Annotated[int | None, BeforeValidator(_absorb_page), Query()]
 
+#: How deep a GitHub search pages: real serves the first 1000 results of a search and refuses a
+#: page past them with a 422 (measured 2026-09-06 on `/search/issues`, `/search/code` and
+#: `/search/repositories`), whatever the total. Where exactly the refusal falls differs by route
+#: and is each route's own rule; this is the number both share, and what caps a `last` link.
+GITHUB_SEARCH_RESULT_CAP = 1000
+
 #: The largest `page` / `per_page` value real's code search parses: it deserializes both into an
 #: unsigned 32-bit integer, so 4294967295 is a 200 and 4294967296 the "number too large" refusal.
 GITHUB_CODE_SEARCH_PAGE_MAX = 4_294_967_295
@@ -184,8 +190,16 @@ def github_link_header(
     total: int,
     *,
     per_page_param: str | None = None,
+    max_page: int | None = None,
 ) -> str | None:
     """Build a Link header with rel=next/prev/first/last. ``params`` are extra query args.
+
+    ``max_page`` is the deepest page the route will serve, for a search: `last` never names a page
+    past it, and the page it names is the end of the listing for `next` too. Measured on
+    api.github.com on 2026-09-06 on an issue search of 2.8 million results: `per_page=100` links
+    last=10, `per_page=30` last=34 and `per_page=7` last=143, which is ``ceil(1000 / per_page)``
+    each time, and on that page the header carries `prev, first` alone, as on any last page. A
+    listing has no such depth and passes nothing.
 
     ``per_page_param`` is the page size the CALLER sent, verbatim, or ``None`` when it sent none.
     A size the handler defaulted is left out of the urls, the rule a listing's filters already
@@ -218,6 +232,8 @@ def github_link_header(
     arrive at the query it was paging, not at a truncated one.
     """
     last_page = max(1, (total + per_page - 1) // per_page)
+    if max_page is not None:
+        last_page = min(last_page, max_page)
     if last_page <= 1:
         return None
 
@@ -275,8 +291,16 @@ def github_code_search_link_header(
 
     `/search/issues` shares none of that and pages by :func:`github_link_header`. A result set that
     fits one page carries no header, which is the one rule all three surfaces share.
+
+    `last` stops at the search depth, ``ceil(1000 / per_page)``: on 294 million hits `per_page=100`
+    links last=10, `per_page=30` last=34, `per_page=7` last=143 and `per_page=1` last=1000
+    (measured 2026-09-06). That is one page further than the route serves at `per_page=30` and
+    `per_page=7`, where page 34 and page 143 are the 422 (see ``search_code`` for the route's own
+    boundary, ``page * per_page > 1000``): real's `last`, and so this one, names a page real refuses.
+    Page 33 at `per_page=30` links next=34 for the same reason.
     """
     last_page = max(1, (total + per_page - 1) // per_page)
+    last_page = min(last_page, -(-GITHUB_SEARCH_RESULT_CAP // per_page))
     if last_page <= 1:
         return None
 
--- a/backlot/pagination.py
+++ b/backlot/pagination.py
@@ -86,9 +86,9 @@ def _absorb_page(v):
     listing; the same five values are a 200 on `/repos/{o}/{r}/tags`, `/user/repos`,
     `/search/issues`, `/search/repositories`, `/search/users`, `/search/commits`, `/search/labels`
     and `/search/topics`, 2026-09-06). Refusing them would hand a paginator computing an edge value a
-    hard error where production absorbs it. `/search/code` is the one surface that refuses, and
-    refuses before this validator's answer matters: see :func:`github_code_search_page_refusal`,
-    which that route asks first.
+    hard error where production absorbs it. Of those ten surfaces `/search/code` is the one that
+    refuses, and it refuses before this validator's answer matters: see
+    :func:`github_code_search_query_refusal`, which that route asks first.
 
     A `BeforeValidator` rather than a `str` annotation: the parameter stays an integer in the
     OpenAPI schema, which is what real's own spec declares, so the tolerance is in the runtime
@@ -96,6 +96,10 @@ def _absorb_page(v):
     """
     if v is None or isinstance(v, int):
         return v
+    if isinstance(v, str):
+        # Leading zeros are dropped before the parse, not by it: Python refuses to convert a
+        # string of more than 4300 digits at all, and real reads `0000…05` as 5 however many zeros.
+        v = v.lstrip("0") or "0"
     try:
         return int(v)
     except (TypeError, ValueError):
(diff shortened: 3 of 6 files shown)


khj809:
@nuri-yoo 

While reviewing this branch, I found that landing #145 first would resolve part of it for free, measured rather than guessed: I ran this branch with `BACKLOT_DEFAULT_PAGE_SIZE=30 BACKLOT_MAX_PAGE_SIZE=100`, GitHub's real two numbers.

One defect goes away outright — code search's `page * per_page > 1000` makes `floor(1000 / per_page)` zero once the applied size exceeds 1000, so with `BACKLOT_MAX_PAGE_SIZE=2000` and `per_page=1001` page 1 itself is the depth 422 and no page of a code search is reachable, which is unreachable under a cap of 100. And four sites here exist only to describe the size mismatch (`search_issues`'s depth comment, `search_code`'s docstring, and the computed boundary cells in the two depth tests), so landing this first means writing them and then deleting them.

The tests need no edit either way: `tests/test_github.py` is 187 passed under 30/100, and the boundaries land exactly on the measured cells (`/search/issues` serves page 34 and refuses 35, `/search/code` serves 33 and refuses 34), so afterwards those cells can be written down as real's numbers instead of derived from configuration — which is what #145's own "what a fix has to settle" asks for.

So I'd suggest submitting PR for #145 first and then continuing here. How does that sound?

nuri-yoo:
> @nuri-yoo
> 
> While reviewing this branch, I found that landing #145 first would resolve part of it for free, measured rather than guessed: I ran this branch with `BACKLOT_DEFAULT_PAGE_SIZE=30 BACKLOT_MAX_PAGE_SIZE=100`, GitHub's real two numbers.
> 
> One defect goes away outright — code search's `page * per_page > 1000` makes `floor(1000 / per_page)` zero once the applied size exceeds 1000, so with `BACKLOT_MAX_PAGE_SIZE=2000` and `per_page=1001` page 1 itself is the depth 422 and no page of a code search is reachable, which is unreachable under a cap of 100. And four sites here exist only to describe the size mismatch (`search_issues`'s depth comment, `search_code`'s docstring, and the computed boundary cells in the two depth tests), so landing this first means writing them and then deleting them.
> 
> The tests need no edit either way: `tests/test_github.py` is 187 passed under 30/100, and the boundaries land exactly on the measured cells (`/search/issues` serves page 34 and refuses 35, `/search/code` serves 33 and refuses 34), so afterwards those cells can be written down as real's numbers instead of derived from configuration — which is what #145's own "what a fix has to settle" asks for.
> 
> So I'd suggest submitting PR for #145 first and then continuing here. How does that sound?

Good point. would proceed as suggested.

nuri-yoo:
#145 is now #152, one commit off `main`: `PER_PAGE_DEFAULT = 30` / `PER_PAGE_MAX = 100` in the router (the `TREE_MAX_ENTRIES` shape, so a test lowers them with `monkeypatch.setattr`), read through `_clamp` at the sixteen sites; the settings stay for Slack, Gmail, Drive and the Jira search and now say which reads which. Once it lands, this PR rebases onto it, the two depth tests write `34`/`35` and `33`/`34` as real's numbers instead of deriving them, and the four sites that only describe the size mismatch go.

backlot/routers/github.py
```diff
@@ -554,13 +573,23 @@ async def search_issues(
...
+    page, per_page = _clamp(page, per_page)
+    start = (page - 1) * per_page
+    # A page that STARTS past the first 1000 results is refused, whatever the total: at real's
+    # default 30 a page, page 34 (results 991 to 1020) is served in full and page 35 refused, at
+    # 7 a page, page 143 (995 to 1001) is served and 144 refused, at 100 a page, 10 is served and
+    # 11 refused, and an 846-result search refuses page 11 at 100 a page just the same, after
+    # serving page 10 empty (measured 2026-09-06; the rule is `github_search_depth_refused`'s).
+    # After the blank-`q` and `repo:` 422s, which real answers first on this route; `/search/code`
+    # draws its line elsewhere, see there. Measured in the size THIS server applies, so with an
+    # unsent size the refusal falls at page 11 here (default 30) where real's falls at 35.
```
khj809: This says the refusal falls at page 11 here, but since #152 it falls at 35 — real's own page (verified on this head: `page=34` is a 200, `page=35` the 422, `page=11` a 200). The clause only existed to describe the gap #145 closed, so it wants deleting rather than having its number swapped: the sentence above it already carries the measurement and is now true of Backlot too.

```suggestion
    # draws its line elsewhere, see there.
```

backlot/routers/github.py
```diff
@@ -815,10 +862,39 @@ async def search_code(
...
+    page 34 (991 to 1020) refused, at 7 a page, 142 is served and 143 refused, at 1 a page, 1000 is
+    served and 1001 refused, at 100 a page, 10 is served and 11 refused, and `per_page=101` is
+    refused at 11 too because it is served at 100. The total does not enter: a 44-result search
+    refuses page 34 at 30 a page after serving pages 3 to 33 empty. That is not `/search/issues`'s
+    line, which serves page 34 in full and refuses the page that STARTS past 1000 (both rules are
+    :func:`backlot.pagination.github_search_depth_refused`'s). The refusal comes after the blank-`q`
+    422 and before the `repo:` qualifier is read: `repo:psf/ghost-zz-9876` at page 11 of 100 is
+    this 422, not the qualifier's answer (all measured 2026-09-06). It is measured in the size THIS
+    server applies: with an unsent size the refusal falls at page 11 here (default 100) where
+    real's, at its default 30, falls at 34.
```
khj809: Same clause here, and this one still reads `default 100`: since #152 code search refuses at page 34, not 11 (`page=33` is a 200, `page=34` the 422). The "size APPLIED" point it was making is already carried above by `per_page=101` being refused at 11 because it is served at 100.

```suggestion
    this 422, not the qualifier's answer (all measured 2026-09-06).
```

backlot/pagination.py
```diff
@@ -80,17 +80,26 @@ def clamp_limit(limit: int | None, default: int, maximum: int) -> int:
...
-    the cap (measured against a public repository's issue listing). Refusing them would hand a
-    paginator computing an edge value a hard error where production absorbs it.
+    Real GitHub refuses no value on its listings and on every search but code search: `per_page=0`,
+    `per_page=abc`, `page=0`, `page=-1` and `page=abc` are each a 200 with the defaults applied, and
+    a `per_page` over the cap is a 200 at the cap (measured against a public repository's issue
+    listing; the same five values are a 200 on `/repos/{o}/{r}/tags`, `/user/repos`,
+    `/search/issues`, `/search/repositories`, `/search/users`, `/search/commits`, `/search/labels`
+    and `/search/topics`, 2026-09-06). Refusing them would hand a paginator computing an edge value a
+    hard error where production absorbs it. Of those ten surfaces `/search/code` is the one that
+    refuses, and it refuses before this validator's answer matters: see
```
khj809: Nine surfaces are enumerated just above (the issue listing plus eight), so "those ten" has no antecedent — the tenth is `/search/code`, which is not one of them.

```suggestion
    hard error where production absorbs it. The tenth measured surface, `/search/code`, is the
    one that refuses, and it refuses before this validator's answer matters: see
```

backlot/routers/github.py
```diff
@@ -147,7 +165,7 @@ async def _validate_api_version(request: Request) -> None:
     is what is wrong. Declared before ``_validate_path_owner`` in the router's dependency list, which
     is what puts it first.
     """
-    if selected_api_version(request) is None:
+    if honours_api_version(request) and selected_api_version(request) is None:
```
khj809, line 172: `_version`'s docstring just above still says "``_validate_api_version`` is a router-wide dependency, so an unsupported one never reaches a handler", which no longer holds for `/search/code`. No bug today — `_version` is not reachable from `search_code` — but that sentence is the stated reason its `or DEFAULT_API_VERSION` fallback is safe, so it needs the exception.

backlot/errors/github.py
```diff
@@ -177,10 +177,14 @@ def validation_body(path: str, errors) -> tuple[int, dict]:
     query parameter at once has one answer, and which of the two FastAPI happens to list first is
     not it.
 
-    A QUERY parameter is the residue. Real refuses no pagination value at all, which is why those
-    now absorb (see :func:`backlot.pagination._absorb_page`) and never reach here; what is left is
-    a shape real has no measured answer for, so it keeps the Validation Failed envelope real uses
-    for a parameter it does refuse, with this route's anchor rather than the bare root.
+    A QUERY parameter is the residue. Real's listings refuse no pagination value, which is why
```
khj809, line 183: While in this file: `http_body`'s docstring above describes the hand-shaped bodies as "the version 400 and the search 422, both of which have an ``errors`` member no generic rendering could invent". This PR adds two search 422s that carry no `errors` member, so the class it names no longer holds.

backlot/main.py
```diff
@@ -171,10 +171,12 @@ async def echo_github_api_version(request: Request, call_next):
...
 
     A rejected version gets no echo, matching real: it selected nothing. That is `None` from
-    ``selected_api_version``, the same call the router's 400 is raised from.
+    ``selected_api_version``, the same call the router's 400 is raised from. Code search gets no
+    echo either, whatever it pinned: real's code search backend does not read the header (see
+    ``github.honours_api_version``).
     """
     response = await call_next(request)
-    if request.url.path.startswith("/github"):
+    if request.url.path.startswith("/github") and github.honours_api_version(request):
```
khj809, line 179: `docs/supported-sources.md`'s `X-GitHub-Api-Version` paragraph still says "anything else is the real API's 400, and every response reports its choice in `X-GitHub-Api-Version-Selected`". Both clauses are now false for `/search/code`. Noting it here since that file is not in the diff.

khj809 (commented):
PR body is stale on #145: it still describes the page sizes as the server's 100/1000 and the depth as "refused at page 11 here", and Verification still reads `2448 passed` at the fifth commit (2628 at `817b0de`). Inline notes below — the first two are comments that now contradict the code they annotate.

backlot/routers/github.py
nuri-yoo: Right, and that sentence was the stated reason the fallback is safe; it stopped being true of `/search/code` at the commit that added `honours_api_version`. Rewritten in 0b4f1f5: `_validate_api_version` is router-wide, so on every route that honours the header an unsupported version never reaches a handler; `/search/code` does let one through, and nothing on that route asks `_version`, so the `or DEFAULT_API_VERSION` keeps the return a `str` and is not a case any route reaches.

backlot/errors/github.py
nuri-yoo: The class it named no longer held: the two depth 422s carry no `errors`. 0b4f1f5 splits it: the version 400 and the two Validation Failed search 422s are hand-shaped for their `errors` member, the two depth 422s for their `documentation_url`, which is not the route's anchor in `ROUTE_DOCS` (the issue search's is the bare `/v3/search/` with its trailing slash, code search's the `#search-code` anchor).

backlot/main.py
nuri-yoo: Both clauses were false for `/search/code`. 0b4f1f5 adds the exception to that paragraph: real's code search backend does not read the header, so a pinned version there is served whatever it says and no response from it carries the `Selected` header (measured 2026-09-06). The paragraph sits outside the generated block, so `gen_docs.py --check` is unaffected.

nuri-yoo:
Body rewritten against `9909112`: the opening says the branch sits on #152 and what the commits after the fifth are, the #137 paragraph counts nine absorbing surfaces and `/search/code` as the tenth, the tests paragraph says the applied-size cells are real's pages written down (33/34 code, 34/35 issues) rather than computed, the `bab49cb` revert measurement is dated to the fifth commit, "Not in this PR" reads #145 as landed, a new paragraph lists this review's six changes, and Verification is this head's run: `2629 passed, 1 skipped` (Docker down), ruff, format and `gen_docs.py --check` clean.

The three suggestions went in as-is (`dda1c10`, `78abc7e`, `6a73564`). The three inline notes without one are `0b4f1f5`, prose only; `9909112` is the test change, the two depth tests writing real's boundary pages instead of deriving them from `PER_PAGE_DEFAULT`/`PER_PAGE_MAX`, which the derived cells could not catch: setting the router's default to 100 now fails both.

khj809 (approved):
Approving — all six notes and the derived-boundary nit are addressed, and I re-verified on `830e820`: 2629 passed, ruff clean, `gen_docs --check` exit 0, `backlot diff --source github` 0 new, and setting `PER_PAGE_DEFAULT` to 100 now fails the two depth tests where the derived cells passed.

One non-blocking note on `http_body`'s new docstring: it enumerates four of the six `github_body` sites — the ref 404 (`No commit found for the ref`) and `_no_commit_for_sha`'s 422 are missing, and both fit the second reason it gives, a `documentation_url` that is not the route's `ROUTE_DOCS` anchor. Pre-existing omission, just more visible now that the sentence reads as a taxonomy; fine as a follow-up.

nuri-yoo:
On the `http_body` note: the count is right, six sites and four named, and it goes in #147, which already edits that docstring, as `90f24b0` rather than another commit here. One correction to the reason, checked on `830e820`: the ref 404 does fit the `documentation_url` reason (`/v3/repos/contents/` against `ROUTE_DOCS`'s `repos/contents#get-repository-content`), but `_no_commit_for_sha`'s 422 does not, its `documentation_url` is `commits/commits#get-a-commit`, the same string `ROUTE_DOCS` holds for `/commits/{sha}`. What makes that body hand-shaped is a third reason the paragraph never named: a `message` the exception's `detail` does not carry, the ref echoed back, which the ref 404 shares. The paragraph now lists all six sites under the three reasons.

PR #147, opened 2026-09-06
github: a JSON response carries the charset real's carry, on every route but code search (#146)

### LINKED ISSUE #146: github: JSON responses carry no charset where real's carry charset=utf-8 on every route but code search
**Source and commit** — `github`, at `bab49cb` (`main`; #144 does not touch this).

Every JSON response the GitHub router sends is `content-type: application/json`, FastAPI's default. Real sends `application/json; charset=utf-8` on every GitHub route measured, 200 and error alike, with one exception: `/search/code`, served by a different backend, sends `application/json`, which is the one route where the two agree. The gap is in a header, so `backlot diff --source github` cannot see it, and it is the same on every status the router answers.

| # | Response | Real | Backlot at `bab49cb` | Silent? |
|---|---|---|---|---|
| 1 | `GET /repos/{o}/{r}/issues` 200 | `application/json; charset=utf-8` | `application/json` | yes |
| 2 | `GET /repos/{o}/{r}` 200, `/orgs/{org}` 200, `/user/repos` 200 | `application/json; charset=utf-8` | `application/json` | yes |
| 3 | `GET /repos/{o}/nope` 404 | `application/json; charset=utf-8` | `application/json` | yes |
| 4 | `GET /search/issues?q=` 422 | `application/json; charset=utf-8` | `application/json` | yes |
| 5 | `X-GitHub-Api-Version: 1999-01-01` 400 | `application/json; charset=utf-8` | `application/json` | yes |
| 6 | no credential, 401 | `application/json; charset=utf-8` | `application/json` | yes |
| 7 | `GET /search/code?q=…` 200 | `application/json` | `application/json` | no, agrees |

**How you know** — measured against api.github.com on 2026-09-06.

- `application/json; charset=utf-8` on: `/repos/psf/requests/issues?per_page=1`, `/repos/psf/requests`, `/repos/psf/requests/branches?per_page=1`, `/repos/psf/requests/commits?per_page=1`, `/repos/psf/requests/contents/README.md`, `/orgs/psf`, `/user/repos?per_page=1`, `/search/issues?q=repo:psf/requests+timeout&per_page=1`, the 404 for `/repos/psf/ghost-zz-9876`, the 422 for `/search/issues?q=`, the version 400 for `1999-01-01`, and the 401 for `/user` with no credential. Thirteen responses, one content type.
- `application/json` on `/search/code`, 200 and its 422s alike (the depth 422 and the blank-`q` 422), measured for #144, which also found that route sends no `X-GitHub-Api-Version-Selected`; it is that backend's shape throughout.
- Backlot answers `application/json` on all of the above: the listing, the repo, the 404, the search 200 and 422, the version 400, the 401 and code search. Measured on the bundled corpus at `bab49cb`.

**Reproduction** — the bundled corpus, no fixtures:

```python
import backlot, requests

with backlot.serve() as m:
    org = requests.get(f"{m.base_url}/_meta/users", headers=h).json()["org"]
    b = f"{m.base_url}/github"
    for label, url, params, extra in (
        ("listing 200", f"{b}/repos/{org}/pipeline/issues", {"per_page": 1}, {}),
        ("404", f"{b}/repos/{org}/ghost-zz-9876", {}, {}),
        ("search 422", f"{b}/search/issues", {"q": ""}, {}),
        ("version 400", f"{b}/repos/{org}/pipeline", {}, {"X-GitHub-Api-Version": "1999-01-01"}),
        ("401", f"{b}/user/repos", {}, {"Authorization": "Bearer nope"}),
        ("code search 200", f"{b}/search/code", {"q": "extension:md"}, {}),
    ):
        r = requests.get(url, headers={**h, **extra}, params=params)
        print(f"{label:16s} {r.status_code} {r.headers['content-type']}")
```

Every line prints `application/json`; real prints `application/json; charset=utf-8` for the first five and `application/json` for the last.

**What it costs a client** — little, and that is worth saying plainly. JSON is UTF-8 by definition and every mainstream HTTP client decodes it without the parameter, so the response body is read the same either way. What differs is the header string: a client or a recorded-fixture test that matches `content-type` exactly, or a proxy rule keyed on it, passes against one side and fails against the other, and the direction is the usual one, a fixture recorded against Backlot fails against production. The other cost is the one #144 stopped short of: the code-search exception is only visible as an exception once the rest of the router carries the charset.

**Where it lands** — `router = APIRouter(prefix="/github", …)` in `backlot/routers/github.py`, which takes a `default_response_class` and today takes FastAPI's; `_http_exception_handler` and `_validation_exception_handler` in `backlot/main.py`, which build a `JSONResponse` for every vendor's error envelope and would need to know the path is GitHub's; and `search_code`, which must keep `application/json`. The raw, diff and patch media types on this router already carry `charset=utf-8` in their own content types and are not this.

**What a fix has to settle**

- Where the header is set. A `default_response_class` on the router covers every handler that returns a dict and nothing else, so the two error handlers in `backlot/main.py` need their own rule for `/github` paths, the way `errors/github.py` already owns the envelope; a middleware that rewrites the header on every `/github` response covers both in one place but then has to exempt `/search/code` and every non-JSON response the router returns. Either way the exemption for code search is a measured fact and belongs in the same commit.
- Not the other vendors. Slack's, Jira's, Confluence's and Drive's content types are not measured here, and a global response class would assert they share GitHub's. Each is its own measurement and its own change.
- Whether the version 400 and the 401, which are raised before the router's handler runs, get the header from the same place as the 404 and 422 do. On real all four carry it; in Backlot all four are rendered by the same handler in `backlot/main.py`, so one rule covers them if it is there.
- Whether a test asserts the exact header on one response of each kind, 200, 404, 422, 400, 401 and the code-search 200, or only on one. Six responses, six measured values; a table is cheap and the code-search row is the one a later change is most likely to break.

nuri-yoo:
Closes #146. On `main` at `89f15ea`, which is #144 merged; this was stacked on #144 until it landed and the exemption reuses its `CODE_SEARCH_PATH`. Seven commits: `bb4b14d` the change, `55ac839` from a self-review on 2026-09-07 that measured the code search 401, `d7454c8` and `499b760` from a second self-review, prose only, `b30484e` and `82b81bf` the two suggestions of the 2026-09-08 review applied as-is, and `142107e` its third note, the middleware's registration order (the two sections before Verification).

**What GitHub does.** Every JSON response measured on api.github.com on 2026-09-06 carries `content-type: application/json; charset=utf-8`: the 200s of `/repos/{o}/{r}/issues`, `/repos/{o}/{r}`, `/repos/{o}/{r}/branches`, `/repos/{o}/{r}/commits`, `/repos/{o}/{r}/contents/{path}`, `/orgs/{org}`, `/user/repos` and `/search/issues`, the 404 for `/repos/psf/ghost-zz-9876`, the 422 for `/search/issues?q=`, the version 400 for `X-GitHub-Api-Version: 1999-01-01`, the 401 for `/user` with no credential, and an unauthenticated `/repos/psf/requests` 200. Thirteen responses, one header. `/search/code` is the exception and answers `application/json` with no parameter on its 200 and on its 422s alike, the blank-`q` one and the depth one, which is the same backend #144 found reading no version header. The exception is the backend's, not the path's: the 401 on `/search/code` for no credential or a bad one is the gateway's answer and carries the charset like every other 401, and so does the 404 real answers an authenticated `POST` to `/search/code` or `/search/issues` with (both measured 2026-09-07). Backlot answered FastAPI's bare `application/json` on all of these (measured on the bundled corpus at `bab49cb`: the listing, the repo, the 404, the search 200 and 422, the version 400, the 401 and code search), so code search was the one route where the two agreed, and by accident.

**What changed.** `backlot/errors/github.py` gains `json_media_type(path, status_code)`, the `content-type` real puts on a JSON body answered at that path with that status: `application/json; charset=utf-8` everywhere under `/github` except `CODE_SEARCH_PATH` on the statuses the code search backend answers itself, 200 and 422, where it is `application/json`; the 401 and the 405 on that path are the gateway's kind of answer and keep the charset (the first version exempted the path whole, and a self-review on 2026-09-07 measured the 401); the router import is lazy, as `_routes()`'s is, so the package's dependency on the router stays one-way. `backlot/errors/__init__.py` dispatches it like `http_body` and `validation_body`, as an optional member of an envelope module, so a vendor without one keeps FastAPI's default and no claim is made about what it sends. `backlot/main.py` applies it in a middleware, `vendor_json_media_type`, that rewrites a response's `content-type` only when it is exactly `application/json`, which covers the handlers and the two error handlers in one place and leaves the raw, diff, patch and text/plain answers alone.

**Why a middleware and not `default_response_class`.** #146 left that open. A response class whose `media_type` carries the charset is written by FastAPI into the OpenAPI document as the key of the 200's `content` (checked on FastAPI in this environment: a router with such a class documents `content: {"application/json; charset=utf-8": …}`), where real's own spec keys it `application/json`. That would hand `backlot diff --source github` a new gap on every route and `backlot mcp` a content key real does not have. The charset is a fact about the wire and not about the contract, so it is set on the wire.

**Tests.** One test, twenty cells: the 200s of a listing, a repo, an org, `/user/repos`, `/search/issues` and the id-keyed `/repositories/{id}` (which the path-rewriting middleware serves), the 404, the search 422, the version 400, the 401 with no credential and with a bad one, all `application/json; charset=utf-8`; code search's 200, blank-`q` 422 and depth 422, all `application/json`, and its 401 with no credential and with a bad one, both `application/json; charset=utf-8`; its parse 400 still `text/plain; charset=utf-8`; a `POST` to code search and to a repo, Starlette's 405 with the charset on both; a pull asked for as `application/vnd.github.diff` still `application/vnd.github.diff; charset=utf-8`. Then that the OpenAPI document still keys the JSON body `application/json`, and that `/openapi.json` itself, outside `/github`, is untouched. With the three source files at `main` (`89f15ea`) and the test kept: `1 failed`. Mutation-checked from a copy and restored: dropping the code-search exemption fails the test on its code-search cells; exempting the path whole, 401 and 405 included, fails it on the 401 cells; a middleware that rewrites nothing fails it; a middleware that rewrites every content-type fails 25 (the text/plain 400 cells of #144 among them); a dispatch that answers for every vendor fails the `/openapi.json` cell.

**Second self-review, 2026-09-08.** Read the way #144's second review read it, for prose the code had outgrown. `json_media_type`'s docstring said thirteen responses and enumerated twelve; the thirteenth, the 200 `/repos/{o}/{r}` answers with no credential, is now in the list, and all thirteen were re-measured on 2026-09-08 with the same header, as were code search's bare 200, its charset 401 and the 404 real answers an authenticated `POST` with. The comment on `_CODE_SEARCH_BACKEND_STATUSES` claimed "everything else on the path" for the gateway, more than was measured; it names the 401 and the 405 this server answers. The middleware's docstring placed the exception handlers below it; they are above. The test docstring counts nine 200s. And `docs/supported-sources.md`'s media-type paragraph now says a JSON body carries the charset except on `search/code`, since that page is where the other honoured headers are recorded. No cell of the test depends on the page sizes #152 changed (the depth cell is `per_page=1&page=1001`), and the rebase onto #144's current head applied without conflict; `http_body`'s docstring was edited by both branches in different paragraphs. Then `d7454c8`, from khj809's non-blocking note on #144's approval: the first paragraph of that docstring read as a taxonomy but named four of the six `github_body` sites and two reasons; it now names all six under three, the `errors` member, a `documentation_url` that is not the route's `ROUTE_DOCS` anchor, and a `message` the exception's `detail` does not carry (the ref 404 and `_no_commit_for_sha`'s 422 echo the ref; the latter's `documentation_url` is the route's own, so it fits only that third reason).

**Review, 2026-09-08.** Three notes. `vendor_json_media_type` was registered first, which under `@app.middleware("http")` makes it the innermost user middleware, so a JSON body another middleware returns itself (`refuse_a_bearer_jira_cannot_read`'s 403) went around it; harmless while Atlassian has no `json_media_type`, but it would have made `errors/__init__.py`'s "a module plus one entry" promise untrue for the first vendor that gets one. It is registered last now, the outermost, and the docstring says why; the exception handlers run inside `ExceptionMiddleware` and are covered from either position. The test gains an Atlassian `serverInfo` cell, so the dispatch is exercised walking past an envelope with no `json_media_type` and not only past the app's own `/openapi.json`. And the docs sentence carries the 401 distinction that lived only in the code: on `search/code` the 401 is the gateway's answer and keeps the charset.

**Verification.**

```
2630 passed, 1 skipped   (2026-09-08, at `142107e`, on `main`; Docker down, so the one mcp test skipped)
```

**Not in this PR.** The other vendors' content types are not measured and do not move; the dispatch is built so each can carry its own once it is. The 405 a wrong method earns is Starlette's: real has no 405 there, an authenticated `POST` to a search route is its 404 Not Found and an unauthenticated one to a repo its 401 (see `errors/github.py::http_body`, which now records both), so the status is a gap on its own and not this PR's; what this PR does to it is carry the charset, as real's 404 does. `HEAD` is a wider gap of the same kind, a 405 on every GitHub route here where real answers it as the GET with no body, filed as #151.

### DIFF
--- a/backlot/main.py
+++ b/backlot/main.py
@@ -172,9 +172,9 @@ async def vendor_json_media_type(request: Request, call_next):
     real's own spec says `application/json` there: the charset is a fact about the wire, not about
     the contract, and `backlot diff` compares the contract. Only a response that is exactly
     `application/json` is touched, so the raw, diff and text/plain answers keep their own types, and
-    the error handlers below are covered along with the handlers. The status goes along with the
-    path because the answer can depend on it: code search's own 200 and 422 carry no charset where
-    the gateway's 401 on the same path does.
+    the two exception handlers above are covered along with the route handlers. The status goes
+    along with the path because the answer can depend on it: code search's own 200 and 422 carry no
+    charset where the gateway's 401 on the same path does.
     """
     response = await call_next(request)
     if response.headers.get("content-type") == "application/json":
--- a/backlot/main.py
+++ b/backlot/main.py
@@ -162,29 +162,6 @@ async def _validation_exception_handler(request: Request, exc: RequestValidation
 
 
 @app.middleware("http")
-async def vendor_json_media_type(request: Request, call_next):
-    """Put the vendor's own `content-type` on a JSON body, where one is measured.
-
-    Real GitHub answers `application/json; charset=utf-8` on every JSON response but code search's
-    (see ``backlot.errors.github.json_media_type``); FastAPI's ``JSONResponse`` answers
-    `application/json`. A middleware rather than a ``default_response_class`` on the router, because
-    FastAPI writes a response class's media type into the OpenAPI document as the content key, and
-    real's own spec says `application/json` there: the charset is a fact about the wire, not about
-    the contract, and `backlot diff` compares the contract. Only a response that is exactly
-    `application/json` is touched, so the raw, diff and text/plain answers keep their own types, and
-    the two exception handlers above are covered along with the route handlers. The status goes
-    along with the path because the answer can depend on it: code search's own 200 and 422 carry no
-    charset where the gateway's 401 on the same path does.
-    """
-    response = await call_next(request)
-    if response.headers.get("content-type") == "application/json":
-        media_type = errors.json_media_type(request.url.path, response.status_code)
-        if media_type is not None:
-            response.headers["content-type"] = media_type
-    return response
-
-
-@app.middleware("http")
 async def echo_github_api_version(request: Request, call_next):
     """Report which API version served the response, as real GitHub does on every github request.
 
@@ -275,6 +252,32 @@ async def parse_slack_form(request: Request, call_next):
         if "application/x-www-form-urlencoded" in ctype:
             request.state._form = dict(await request.form())
     return await call_next(request)
+
+
+@app.middleware("http")
+async def vendor_json_media_type(request: Request, call_next):
+    """Put the vendor's own `content-type` on a JSON body, where one is measured.
+
+    Real GitHub answers `application/json; charset=utf-8` on every JSON response but code search's
+    (see ``backlot.errors.github.json_media_type``); FastAPI's ``JSONResponse`` answers
+    `application/json`. A middleware rather than a ``default_response_class`` on the router, because
+    FastAPI writes a response class's media type into the OpenAPI document as the content key, and
+    real's own spec says `application/json` there: the charset is a fact about the wire, not about
+    the contract, and `backlot diff` compares the contract. Only a response that is exactly
+    `application/json` is touched, so the raw, diff and text/plain answers keep their own types, and
+    the two exception handlers above are covered along with the route handlers. Registered last of
+    the middlewares, which makes it the outermost, so a JSON body another middleware returns
+    (``refuse_a_bearer_jira_cannot_read``'s 403) passes through it too rather than around it; the
+    exception handlers run inside ``ExceptionMiddleware`` and are covered from either position. The status goes
+    along with the path because the answer can depend on it: code search's own 200 and 422 carry no
+    charset where the gateway's 401 on the same path does.
+    """
+    response = await call_next(request)
+    if response.headers.get("content-type") == "application/json":
+        media_type = errors.json_media_type(request.url.path, response.status_code)
+        if media_type is not None:
+            response.headers["content-type"] = media_type
+    return response
 
 
 @app.get("/health")
--- a/backlot/main.py
+++ b/backlot/main.py
@@ -172,11 +172,13 @@ async def vendor_json_media_type(request: Request, call_next):
     real's own spec says `application/json` there: the charset is a fact about the wire, not about
     the contract, and `backlot diff` compares the contract. Only a response that is exactly
     `application/json` is touched, so the raw, diff and text/plain answers keep their own types, and
-    the error handlers below are covered along with the handlers.
+    the error handlers below are covered along with the handlers. The status goes along with the
+    path because the answer can depend on it: code search's own 200 and 422 carry no charset where
+    the gateway's 401 on the same path does.
     """
     response = await call_next(request)
     if response.headers.get("content-type") == "application/json":
-        media_type = errors.json_media_type(request.url.path)
+        media_type = errors.json_media_type(request.url.path, response.status_code)
         if media_type is not None:
             response.headers["content-type"] = media_type
     return response
--- a/backlot/main.py
+++ b/backlot/main.py
@@ -162,6 +162,27 @@ async def _validation_exception_handler(request: Request, exc: RequestValidation
 
 
 @app.middleware("http")
+async def vendor_json_media_type(request: Request, call_next):
+    """Put the vendor's own `content-type` on a JSON body, where one is measured.
+
+    Real GitHub answers `application/json; charset=utf-8` on every JSON response but code search's
+    (see ``backlot.errors.github.json_media_type``); FastAPI's ``JSONResponse`` answers
+    `application/json`. A middleware rather than a ``default_response_class`` on the router, because
+    FastAPI writes a response class's media type into the OpenAPI document as the content key, and
+    real's own spec says `application/json` there: the charset is a fact about the wire, not about
+    the contract, and `backlot diff` compares the contract. Only a response that is exactly
+    `application/json` is touched, so the raw, diff and text/plain answers keep their own types, and
+    the error handlers below are covered along with the handlers.
+    """
+    response = await call_next(request)
+    if response.headers.get("content-type") == "application/json":
+        media_type = errors.json_media_type(request.url.path)
+        if media_type is not None:
+            response.headers["content-type"] = media_type
+    return response
+
+
+@app.middleware("http")
 async def echo_github_api_version(request: Request, call_next):
     """Report which API version served the response, as real GitHub does on every github request.
 
--- a/backlot/main.py
+++ b/backlot/main.py
@@ -264,12 +264,12 @@ async def vendor_json_media_type(request: Request, call_next):
     FastAPI writes a response class's media type into the OpenAPI document as the content key, and
     real's own spec says `application/json` there: the charset is a fact about the wire, not about
     the contract, and `backlot diff` compares the contract. Only a response that is exactly
-    `application/json` is touched, so the raw, diff and text/plain answers keep their own types, and
-    the two exception handlers above are covered along with the route handlers. Registered last of
-    the middlewares, which makes it the outermost, so a JSON body another middleware returns
-    (``refuse_a_bearer_jira_cannot_read``'s 403) passes through it too rather than around it; the
-    exception handlers run inside ``ExceptionMiddleware`` and are covered from either position. The status goes
-    along with the path because the answer can depend on it: code search's own 200 and 422 carry no
+    `application/json` is touched, so the raw, diff and text/plain answers keep their own types.
+    Registered last of the middlewares, which makes it the outermost, so a JSON body another
+    middleware returns (``refuse_a_bearer_jira_cannot_read``'s 403) passes through it too rather
+    than around it; the route handlers and the two exception handlers above are covered from either
+    position, those handlers because they run inside ``ExceptionMiddleware``. The status goes along
+    with the path because the answer can depend on it: code search's own 200 and 422 carry no
     charset where the gateway's 401 on the same path does.
     """
     response = await call_next(request)
--- a/docs/supported-sources.md
+++ b/docs/supported-sources.md
@@ -105,7 +105,8 @@ A `repo:` search qualifier resolves the same way.
 returns the file's bytes; `…diff`/`…patch` on a pull returns a real unified diff / `git am` mbox; and
 `…text-match+json` on `search/code` adds each hit's `text_matches` fragment. A JSON body is
 `application/json; charset=utf-8`, as real's is on every route measured, except on `search/code`,
-whose own 200 and 422 are the bare `application/json` real's code search backend sends.
+whose own 200 and 422 are the bare `application/json` real's code search backend sends; the 401
+there is the gateway's answer rather than that backend's, and carries the charset.
 
 **`X-GitHub-Api-Version` is honoured too**, in both values real currently supports: `2026-03-10`
 drops `assignee` (issues and pulls) and `merge_commit_sha` (pulls), `2022-11-28` keeps them, an
--- a/docs/supported-sources.md
+++ b/docs/supported-sources.md
@@ -103,7 +103,9 @@ A `repo:` search qualifier resolves the same way.
 
 **Media types are honoured.** `Accept: application/vnd.github.raw` on `contents`/`readme`/`git/blobs`
 returns the file's bytes; `…diff`/`…patch` on a pull returns a real unified diff / `git am` mbox; and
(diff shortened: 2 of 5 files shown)


backlot/main.py
```diff
@@ -161,6 +161,29 @@ async def _validation_exception_handler(request: Request, exc: RequestValidation
     return JSONResponse(status_code=status_code, content=body)
 
 
+@app.middleware("http")
+async def vendor_json_media_type(request: Request, call_next):
```
khj809: Registered here this is the innermost user middleware, so a JSON response returned by an outer one never reaches it — `refuse_a_bearer_jira_cannot_read`'s 403 is exactly that shape today. Harmless while Atlassian has no `json_media_type`, but it makes `errors/__init__.py`'s "Adding a vendor is a module plus one entry below — not an edit to the handler" untrue for the first vendor that gets one.

Moving the function below `parse_slack_form` makes it outermost and covers those responses too; the exception handlers stay covered either way, since they run inside `ExceptionMiddleware`. Measured with it moved last: `tests/test_github.py` and `tests/test_atlassian.py` are 239 passed, the `/github/repositories/{id}` cell (whose path an inner middleware rewrites) included.

tests/test_github.py
```diff
@@ -3196,6 +3196,81 @@ def test_github_api_version_selects_the_payload(
...
+    for path in ("/github/search/code?q=extension:md", f"/github/repos/{gh_org}/gateway"):
+        r = c.post(path, headers=gh_admin_h)
+        assert (r.status_code, r.headers["content-type"]) == (405, utf8), path
+    # the OpenAPI document still keys the JSON body as `application/json`, as real's spec does: the
+    # charset is on the wire, not in the contract `backlot diff` compares
+    op = c.get("/openapi.json").json()["paths"]["/github/repos/{owner}/{repo}/issues"]["get"]
+    assert list(op["responses"]["200"]["content"]) == ["application/json"]
+    # ...and a JSON response outside `/github` is untouched: not a claim about what another vendor
+    # sends, only that this rule is GitHub's alone
+    assert c.get("/openapi.json").headers["content-type"] == bare
```
khj809: This is the only cell holding the rule to GitHub, and `/openapi.json` is not a vendor route — nothing else in the suite pins another vendor's JSON content type. An Atlassian cell exercises the dispatch walking past an envelope that has no `json_media_type`, rather than only the prefix check.

```suggestion
    # ...and a JSON response outside `/github` is untouched, the app's own routes and another
    # vendor's alike: not a claim about what Atlassian sends, only that this rule is GitHub's alone
    assert c.get("/openapi.json").headers["content-type"] == bare
    server_info = c.get("/atlassian/rest/api/2/serverInfo", headers=gh_admin_h)
    assert (server_info.status_code, server_info.headers["content-type"]) == (200, bare)
```

Verified: 199 passed with `tests/test_docs.py`, `ruff format --check` and `ruff check` clean.

docs/supported-sources.md
```diff
@@ -103,7 +103,9 @@ A `repo:` search qualifier resolves the same way.
 
 **Media types are honoured.** `Accept: application/vnd.github.raw` on `contents`/`readme`/`git/blobs`
 returns the file's bytes; `…diff`/`…patch` on a pull returns a real unified diff / `git am` mbox; and
-`…text-match+json` on `search/code` adds each hit's `text_matches` fragment.
+`…text-match+json` on `search/code` adds each hit's `text_matches` fragment. A JSON body is
+`application/json; charset=utf-8`, as real's is on every route measured, except on `search/code`,
+whose own 200 and 422 are the bare `application/json` real's code search backend sends.
```
khj809: The sentence scopes the exception to code search's own 200 and 422 correctly, but the distinction the second commit measured — the 401 on that path is the gateway's and keeps the charset — lives only in the code. Read alone this invites "`search/code` is bare".

```suggestion
whose own 200 and 422 are the bare `application/json` real's code search backend sends; the 401
there is the gateway's answer rather than that backend's, and carries the charset.
```

backlot/main.py
nuri-yoo: Right: `@app.middleware("http")` puts each new middleware outside the ones before it, so registered first this was the innermost and the Jira 403 went around it. Moved after `parse_slack_form` in 142107e, which makes it the outermost (`app.user_middleware` lists it first), and the docstring says why; the exception handlers run inside `ExceptionMiddleware` and are covered from either position. `tests/test_github.py`, `test_atlassian.py` and `test_docs.py`: 248 passed with the two suggestions in. The branch is rebased onto `main` in the same push, since #144 landed.

tests/test_github.py
nuri-yoo: Applied as-is (`bdfdf6c`, `b30484e` once rebased onto `main`). With the middleware now outermost this cell also exercises a response that passes through it rather than around it, and it is the 200 bare.

docs/supported-sources.md
nuri-yoo: Applied as-is (`53bd681`, `82b81bf` once rebased onto `main`).

backlot/main.py
```diff
@@ -254,6 +254,32 @@ async def parse_slack_form(request: Request, call_next):
...
+    FastAPI writes a response class's media type into the OpenAPI document as the content key, and
+    real's own spec says `application/json` there: the charset is a fact about the wire, not about
+    the contract, and `backlot diff` compares the contract. Only a response that is exactly
+    `application/json` is touched, so the raw, diff and text/plain answers keep their own types, and
+    the two exception handlers above are covered along with the route handlers. Registered last of
+    the middlewares, which makes it the outermost, so a JSON body another middleware returns
+    (``refuse_a_bearer_jira_cannot_read``'s 403) passes through it too rather than around it; the
+    exception handlers run inside ``ExceptionMiddleware`` and are covered from either position. The status goes
+    along with the path because the answer can depend on it: code search's own 200 and 422 carry no
+    charset where the gateway's 401 on the same path does.
```
khj809: The paragraph now states that the exception handlers are covered twice — here, and again in the `ExceptionMiddleware` clause at the end. The insertion also left line 271 at 111 characters where the rest of the docstring wraps at 100. Merging the two clauses settles both. Non-blocking.

```suggestion
    `application/json` is touched, so the raw, diff and text/plain answers keep their own types.
    Registered last of the middlewares, which makes it the outermost, so a JSON body another
    middleware returns (``refuse_a_bearer_jira_cannot_read``'s 403) passes through it too rather
    than around it; the route handlers and the two exception handlers above are covered from either
    position, those handlers because they run inside ``ExceptionMiddleware``. The status goes along
    with the path because the answer can depend on it: code search's own 200 and 422 carry no
    charset where the gateway's 401 on the same path does.
```

khj809 (approved):
Approved. All three notes are in, and I re-ran the branch at `142107e`: full suite 2628 passed / 3 skipped (optional deps missing in my environment), `ruff check`, `ruff format --check` and `gen_docs.py --check` clean, and a sweep over all 32 GET routes still finds `/search/code`'s 200 the only bare one.

Checked the move behaviourally rather than by position alone: the 403 `refuse_a_bearer_jira_cannot_read` short-circuits now reaches the dispatch — `json_media_type` is called with `('/atlassian/rest/api/2/serverInfo', 403)` — and stays bare because Atlassian has no hook, which is the right answer. `/github/repositories/{id}` still carries the charset with the path rewrite now running inside. I also re-measured the charset table against api.github.com on 2026-09-08 before the first pass: all thirteen, plus code search's bare 200/422s, its 401 with the charset, the `text/plain` parse 400 and the 404 an authenticated POST earns — every value matches what the docstrings claim.

One prose nit inline; nothing blocking.

PR #149, opened 2026-09-07
CI scans the repository with HOL's plugin-scanner

khj809:
Backlot is being listed in [awesome-ai-plugins](https://github.com/hashgraph-online/awesome-ai-plugins), whose catalog scores every listed repository with [HOL's plugin-scanner](https://github.com/hashgraph-online/hol-guard) whether or not the repository runs it. A repository that runs the scanner itself keeps the full trust score; one that does not takes a 10% reduction. Running it here also means a finding arrives on the pull request that introduces it, rather than in someone else's scan.

The scanner installs a hash-pinned wheel from PyPI and reads the checkout; it needs no secrets and no network beyond the install, so the job runs with `contents: read`. The action is pinned by commit SHA rather than by the `v1` tag it publishes several times a week.

`min_score: 80` is HOL's own passing bar, so a score regression is a red build here rather than a quieter listing later. `fail_on_severity` is `critical`, where [their guide](https://github.com/hashgraph-online/awesome-ai-plugins/blob/main/SCANNER_GUIDE.md) pairs that bar with `high`, and the deviation is the one thing worth reviewing in this change.

Everything the scanner reports as `high` is `HARDCODED_SECRET` on Backlot's own fake tokens, in seven files: `admin-service-token` in `README.md`, `README.pypi.md`, `backlot/server.py`, `examples/using-official-sdk/linear/index.ts` and `tests/test_importer_erb.py`, `xoxb-not-a-real-token` in `tests/test_slack.py`, and `AKIABOGUS0000000BOGU` / `AKIADOESNOTEXIST0000` in `tests/test_s3.py`. The detector's false-positive escape hatch only recognises `your-`, `example-`, `-here` and `xxxx` shapes, so clearing them means renaming the token in the README's first example.

So the trade is explicit: a corpus of plausible credentials is the thing being shipped, gating on `high` would gate on it, and the cost of not gating is that HOL's passing criteria — score ≥ 80 **and** no critical or high findings — reads this repository as advisory rather than passing on their centralized scan. The 10% trust-score reduction that running the scanner at all avoids is unaffected. What is worth fixing is the detector's placeholder heuristics, upstream rather than here.

`SECURITY.md` was the eighth such file until this branch stopped it naming a token in a form the detector matches; it points at the file that holds it instead, which is what a reporter needs anyway.

#150 is merged into this branch and accounts for the rest of the diff: it acts on what the scanner found, taking the score from 57/100 (F) to 90/100 (A). Its own description covers that work.

### DIFF
--- a/.agents/plugins/marketplace.json
+++ b/.agents/plugins/marketplace.json
@@ -7,11 +7,12 @@
     {
       "name": "backlot",
       "source": {
-        "source": "url",
-        "url": "./"
+        "source": "local",
+        "path": "./"
       },
       "policy": {
-        "installation": "AVAILABLE"
+        "installation": "AVAILABLE",
+        "authentication": "ON_INSTALL"
       },
       "category": "Developer Tools"
     }
--- a/.codex-plugin/plugin.json
+++ b/.codex-plugin/plugin.json
@@ -9,12 +9,35 @@
   "homepage": "https://github.com/brekkylab/backlot",
   "repository": "https://github.com/brekkylab/backlot",
   "license": "MIT",
-  "keywords": ["emulator", "rag", "connector", "acl", "slack", "gmail", "github", "jira", "s3"],
+  "keywords": [
+    "emulator",
+    "rag",
+    "connector",
+    "acl",
+    "slack",
+    "gmail",
+    "github",
+    "jira",
+    "s3"
+  ],
   "skills": "./skills/",
   "interface": {
     "displayName": "Backlot",
     "shortDescription": "Emulate an enterprise SaaS source over your own corpus",
     "developerName": "brekkylab",
-    "category": "Developer Tools"
+    "category": "Developer Tools",
+    "longDescription": "Backlot serves Slack, Gmail, Google Drive, GitHub, Jira, Confluence, Notion, Linear, HubSpot, Fireflies and S3 from one local process, over a corpus you supply. It reproduces the response shapes, pagination, authentication, errors and per-document access controls those APIs have, so an integration, an agent or a RAG pipeline can be built and tested against the official vendor SDKs with no vendor account, no OAuth approval, no secrets in CI and no network. Every user in the corpus gets a token, so what one caller may read and another may not is testable too.",
+    "capabilities": [
+      "Serve every supported enterprise SaaS API locally from one corpus",
+      "Answer the official vendor SDKs unchanged but for a base URL",
+      "Enforce per-document access controls per corpus user",
+      "Expose every source as MCP tools over stdio",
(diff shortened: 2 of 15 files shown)


nuri-yoo (approved):
Approving. Checked at eb653be: every `uses:` SHA resolves to the tag in its comment, the pinned action verifies the wheel's SHA-256 and PyPI provenance before installing, the touched tests, `ruff` and `uv lock --check` pass locally, the old `eval` reader and the new `literal_eval` one return the same 38 files and 91 records, and plugin-scanner 3.0.113 run locally reproduces 57/F on the first commit and 90/A on head.

One wording point. The body frames the remaining `high` findings as seven points, but HOL's bar is "score ≥ 80 with no critical or high findings", and the catalog's centralized scan gates on `high`, so the listing check will read "scan findings (advisory)" rather than "scan passed" while the tokens stay. Keeping them is right; the header and body should name that as the trade being made. The scanner also flags two files the body does not list, one of them SECURITY.md itself.

Optional: `min_score: 80` next to `fail_on_severity: critical`, so a score regression fails CI.

khj809:
Both points hold, and I checked them rather than taking them.

**The two unlisted files.** `SECURITY.md` was one — line 37 named `xoxb-not-a-real-token` in full, and the detector's Slack prefix pattern does not care that the file's whole point is that the string is not a secret. It now points at `tests/test_slack.py` as the file that holds it, which is what a reporter needs anyway, and the file is off the list. The other was `backlot.egg-info/PKG-INFO`, which only exists in a working tree that has been installed; the scan job is checkout plus scanner with no install, so CI never saw it. CI now reports `high: 14` against the 16 it reported before, which is the seven remaining files across both ecosystems.

**The passing-criteria wording.** `SCANNER_GUIDE.md:61` is explicit — score ≥ 80 *with no critical or high findings* — so you are right that the centralized scan reads advisory rather than passing while the tokens stay, and the body was glossing that as merely a points question. Both the workflow header and the body now name it as the trade being made, alongside the fact that the 10% trust-score reduction is what running the scanner at all avoids and that is unaffected.

**`min_score: 80`** added. Verified it gates rather than assuming: `--min-score 80` exits 0 at the current 90, `--min-score 95` exits 1, and `action_runner.py:643` is `if result.score < min_score: return finish(1)`, checked before the severity gate. 80 rather than a tighter number because it is HOL's own bar, so the value documents itself.

`fail_on_severity` stays `critical`, and the header now says it deviates from your guide's `high` on purpose rather than leaving a reader to notice.

CI green on 5674ac6: scan 90/100 (A), critical 0, high 14, medium 0, and 2573 passed / 4 skipped locally against current main.

PR #150, opened 2026-09-07
Apply fixes to increase plugin-scanner score

khj809:
Stacked on #149, which added the scanner. That run reported **57/100 (F)**; this one reports **90/100 (A)**, trust 50.18 → 61.67, with `policy_pass` and `verify_pass` both true and no critical findings. The scanner scores a repository as two ecosystems and averages them — `claude:.` out of 64 points and `codex:.` out of 112 — so a fix that lands in both is worth roughly twice one that lands in codex alone.

Nothing here is written for the scanner's benefit alone; each item is something the repository was missing, and the score is how it was found.

**A security policy.** `SECURITY.md` says where to report privately, what we treat as a vulnerability, and — because this project's whole business is handing out plausible credentials to a fake corpus — what we do not. A document reaching a caller whose token should not see it is a vulnerability. `admin-service-token` appearing in the README is not.

**Every action pinned to a commit SHA**, across all four workflows, with the release tag in a trailing comment. `.github/dependabot.yml` is what keeps those pins from rotting: weekly groups for github-actions, for uv, and for the Linear example's npm dependencies.

**`uv.lock` is now committed.** Worth knowing what this changes: CI installs with `pip install -e ".[all]"` and does not read the lock, but `uv run` and `uv sync` locally now do, so a contributor's resolution is pinned where CI's still floats. The `uv` Dependabot group is what moves it.

**The two manifests the catalogs read are complete.** `.codex-plugin/plugin.json` gained the `interface` fields a catalog profile renders and this project actually has: `longDescription`, `capabilities`, `websiteURL`, and `screenshots` — the demo GIF and the architecture diagram, as the repository-relative paths the field is specified to take.

`.agents/plugins/marketplace.json` declared `"source": {"source": "url", "url": "./"}`. The two manifests `tests/test_skills.py` was written against disagree here — [obra/superpowers](https://github.com/obra/superpowers/blob/main/.agents/plugins/marketplace.json) ships that shape, [hashicorp/agent-skills](https://github.com/hashicorp/agent-skills/blob/main/.agents/plugins/marketplace.json) ships `{"source": "local", "path": "./plugins/<name>"}` — and only the second passes the ecosystem's validator, which requires `source.source == "local"` and a `"./"`-prefixed `path`. This repository's plugin is its root, so the path is `"./"`; the test's locator and the comment recording the disagreement moved with it. `policy.authentication` is new alongside the `policy.installation` that was already there, and its value is copied rather than invented: both reference manifests write `ON_INSTALL`, and nothing reachable — the validator included, which only checks the field is non-empty — documents the vocabulary.

`.claude-plugin/marketplace.json` is untouched. Claude Code reads that one, and it was never what the scanner was reading.

**`.codexignore`** lists what an agent client should not pull into context when it loads the plugin: corpora, build output, `.venv`, local scratch.

**A flake8-bandit subset in ruff.** Not `S` wholesale — that is 4269 hits, and the ones under `backlot/` are the project being what it is: 92 `S608` from the FTS and listing queries it composes, 13 `S324` from etag hashes, `S104` from `--host 0.0.0.0`, 4009 `S101` from test asserts. The rules selected — exec, pickle, marshal, eval, and the shell-spawning subprocess calls — pass, so this is a gate rather than a backlog.

**No `eval` anywhere.** `tests/test_schema.py` read each example's `CORPUS` expression by evaluating it against a restricted scope, because an example cannot simply be imported — most of them call `serve_or_connect(CORPUS)` in the module body, so importing one starts a server and drives it. It now flattens the expression instead and hands it to `ast.literal_eval`: a new `_Resolve` transformer substitutes each `Name` for the module constant it already looked up, and folds `"...".format(k=v)` on a string literal, which is the one other thing a corpus is written with — `examples/using-fsspec/s3.py` shares a CSV template between two quarters that way. Anything it does not recognise stays a `Call`, `literal_eval` refuses it, and that corpus is skipped exactly as it was before. The fold runs before the `literal_eval` that is allowed to fail, so it must not raise on any template at all: a parametrized test walks every way a format field can miss — a name not passed, a positional index, a bad format code, an attribute, a subscript — and each is either folded or left for `literal_eval`.

The replacement was checked against the old reader rather than trusted: both were imported side by side and their output compared, and it is equal — the same 38 files and 91 records, byte for byte. `# noqa: S307` is gone with the call it annotated.

**And the guard on stale source counts covers more than the README.** `tests/test_docs.py` already refused a written-out count of the sources — the number goes stale the next time one is added, and `docs/supported-sources.md` carries the generated list — but it only ever read `README.md`. It now sweeps `.md`, `.py`, `.json` and `.toml`, which is where the description a catalog profile renders actually lives. That caught three: the `interface.capabilities` entry and the ruff comment this branch had just written, and one in `scripts/gen_docs.py` that predates it. The regex's own comment no longer spells out an example, for the same reason the bundled-corpus guard below it does not: this file is scanned like any other.

### What is still missing, and why it stays missing

- **`No hardcoded secrets` (7 points in each ecosystem, the largest single item).** The matches are `admin-service-token` in the two READMEs, `backlot/server.py` and the Linear example, `xoxb-not-a-real-token` in `tests/test_slack.py`, and `AKIABOGUS0000000BOGU` / `AKIADOESNOTEXIST0000` in `tests/test_s3.py`. The scanner's false-positive escape hatch only recognises `your-`, `example-`, `-here` and `xxxx` shapes, so clearing it means renaming the token in the README's first example. The documentation is worth more than the seven points.
- **`Interface links and assets valid if declared` (3, codex only).** It wants six fields and the check is all-or-nothing, so the three points stay unclaimed on two counts. `composerIcon` and `logo` must name an image that exists here; the repository has an architecture diagram and a demo GIF but no logo, pointing `logo` at the diagram would be false, and inventing a mark is not a decision a scanner gets to force. `privacyPolicyURL` and `termsOfServiceURL` are deliberately absent: Backlot is a package you run, not a service you sign up for, and publishing terms of service and a privacy policy for it would describe a relationship that does not exist.

### DIFF
--- a/.agents/plugins/marketplace.json
+++ b/.agents/plugins/marketplace.json
@@ -7,11 +7,12 @@
     {
       "name": "backlot",
       "source": {
-        "source": "url",
-        "url": "./"
+        "source": "local",
+        "path": "./"
       },
       "policy": {
-        "installation": "AVAILABLE"
+        "installation": "AVAILABLE",
+        "authentication": "ON_INSTALL"
       },
       "category": "Developer Tools"
     }
--- a/.codex-plugin/plugin.json
+++ b/.codex-plugin/plugin.json
@@ -9,12 +9,30 @@
   "homepage": "https://github.com/brekkylab/backlot",
   "repository": "https://github.com/brekkylab/backlot",
   "license": "MIT",
-  "keywords": ["emulator", "rag", "connector", "acl", "slack", "gmail", "github", "jira", "s3"],
+  "keywords": [
+    "emulator",
+    "rag",
+    "connector",
+    "acl",
+    "slack",
+    "gmail",
+    "github",
+    "jira",
+    "s3"
+  ],
   "skills": "./skills/",
   "interface": {
     "displayName": "Backlot",
     "shortDescription": "Emulate an enterprise SaaS source over your own corpus",
     "developerName": "brekkylab",
-    "category": "Developer Tools"
+    "category": "Developer Tools",
+    "longDescription": "Backlot serves Slack, Gmail, Google Drive, GitHub, Jira, Confluence, Notion, Linear, HubSpot, Fireflies and S3 from one local process, over a corpus you supply. It reproduces the response shapes, pagination, authentication, errors and per-document access controls those APIs have, so an integration, an agent or a RAG pipeline can be built and tested against the official vendor SDKs with no vendor account, no OAuth approval, no secrets in CI and no network. Every user in the corpus gets a token, so what one caller may read and another may not is testable too.",
+    "capabilities": [
+      "Serve eleven enterprise SaaS APIs locally from one corpus",
+      "Answer the official vendor SDKs unchanged but for a base URL",
+      "Enforce per-document access controls per corpus user",
+      "Expose every source as MCP tools over stdio",
+      "Import a corpus from JSONL, or ship the bundled one"
+    ]
   }
 }
--- a/.codex-plugin/plugin.json
+++ b/.codex-plugin/plugin.json
@@ -28,11 +28,16 @@
     "category": "Developer Tools",
     "longDescription": "Backlot serves Slack, Gmail, Google Drive, GitHub, Jira, Confluence, Notion, Linear, HubSpot, Fireflies and S3 from one local process, over a corpus you supply. It reproduces the response shapes, pagination, authentication, errors and per-document access controls those APIs have, so an integration, an agent or a RAG pipeline can be built and tested against the official vendor SDKs with no vendor account, no OAuth approval, no secrets in CI and no network. Every user in the corpus gets a token, so what one caller may read and another may not is testable too.",
     "capabilities": [
-      "Serve eleven enterprise SaaS APIs locally from one corpus",
+      "Serve every supported enterprise SaaS API locally from one corpus",
       "Answer the official vendor SDKs unchanged but for a base URL",
       "Enforce per-document access controls per corpus user",
       "Expose every source as MCP tools over stdio",
       "Import a corpus from JSONL, or ship the bundled one"
+    ],
+    "websiteURL": "https://github.com/brekkylab/backlot",
+    "screenshots": [
+      "./assets/demo.gif",
+      "./assets/figures/architecture.png"
     ]
   }
 }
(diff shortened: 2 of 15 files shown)


khj809:
@nuri-yoo Note that this is a stacked PR on #149. Let's review and merge this first, and I'll make #149 ready to review.

nuri-yoo (approved):
Checked at a3a92a9. Every claim in the body I could re-run holds.

- All seven `uses:` pins resolve to the commented tag (`git/ref/tags/<tag>` dereferenced through the annotated tag object where there is one).
- `ruff check .` and `ruff format --check .` pass on this head with the pinned 0.15.22, so the new `S*` selection is a gate with zero backlog as described.
- `tests/test_schema.py`, `tests/test_docs.py`, `tests/test_skills.py`: 160 passed.
- Old `eval` reader and the new `_Resolve` + `literal_eval` reader imported side by side over `examples/`: 38 files, 91 records, identical after `json.dumps(sort_keys=True)`.
- `uv lock --check` passes against `pyproject.toml`.
- `./assets/demo.gif` and `./assets/figures/architecture.png` exist; both reference manifests write `policy.authentication: "ON_INSTALL"` and the shapes are as the body says (superpowers `url`, hashicorp `local`+`path`); Codex's own plugin-creator spec uses the `local`+`path` shape.
- Scanner run 34092619498 on this head: Final Score 90/100 (A), trust 61.67, critical 0.

One nit, not blocking: `_Resolve.visit_Call` catches `IndexError, KeyError, ValueError` from `str.format`, but a field like `{x.y}` raises `AttributeError` and `{x[0]}` on a non-subscriptable constant raises `TypeError`, and both escape the tree walk before the `try` around `literal_eval`. No example writes either today, so this only matters for the docstring's "same skip as any other unreadable corpus" promise. Adding `AttributeError, TypeError` to the tuple would make it true.

khj809:
Thanks — the nit is right, and it was more than the two you named. I checked every way a format field can miss and they land on five distinct exceptions, not three: a name the call does not pass and a nested `{q:>{w}}` spec raise `KeyError`, a positional index `IndexError`, a malformed template or format code `ValueError`, `{q.y}` `AttributeError`, and `{q[0]}` on a non-subscriptable constant `TypeError`. The tuple is now all five, so the docstring's promise holds for any template, not just the ones examples happen to write.

Added a parametrized test over those five plus the fold that actually fires, asserting the walk either folds to a `Constant` or leaves a `Call` that `literal_eval` refuses. Narrowing the tuple back to the original three fails it on exactly the `{q.y}` and `{q[0]}` rows, so it is guarding the fix rather than restating it.

`ruff check .` / `ruff format --check .` clean, 2400 passed / 4 skipped, and the scanner still reports 90/100 (A).

PR #143, opened 2026-09-06
linear: `or` reads the keys of one branch as alternatives on `CommentFilter` and `IssueLabelFilter`, and by precedence on the label collection, as Linear does (#136)

nuri-yoo:
Closes #136. On `main` at `bab49cb`. Six commits: the `CommentFilter` fix #136 asks for; the `or` compiler #136 and #133 both left as "not measured", `_labels_branches`, now measured, which turned out to be two different answers for the two label filter types; the corners the second commit had listed as unmeasured, measured; from a mutation check of the result, the one rule line no cell protected and the one polarity corner no cell pinned, both measured and pinned; and, from review, the polarity of an empty `and` / `or` beside a negative key and the vacuous `or: [{}]` twin, measured and pinned, with a cell that was pinned twice pinned once; and, from review, the second evaluator of `IssueLabelFilter` (and of `UserFilter`, `TeamFilter`, `AttachmentFilter`, `ReleaseFilter`), `_match_fields` in `linear_resolvers.py`, brought to the same reading, measured on each route it serves.

[...]

PR #152, opened 2026-09-07
github: an unsent per_page is served at 30 and a sent one capped at 100, real's two numbers

### LINKED ISSUE #145: github: an unsent per_page is served at 100 and a sent one up to 1000, where real serves 30 and caps at 100
**Source and commit** — `github`, at `bab49cb` (`main`; #144 does not touch this, and its depth tests compute their boundary pages from it).

Every paged GitHub route sizes its page from the server's settings, `default_page_size` 100 and `max_page_size` 1000, and those are the same two numbers every other vendor's router reads. Real GitHub serves 30 when `per_page` is not sent and caps a sent value at 100, on every listing and search measured. The page size is the one paging rule #131 left alone: that PR made the `Link` header say what real says (an unsent size omitted, a sent size in the caller's spelling, `last` computed from the size applied), so the header is right while the page under it is the wrong length.

| # | Request | Real | Backlot at `bab49cb` | Silent? |
|---|---|---|---|---|
| 1 | `GET /repos/{o}/{r}/issues?state=all` | 30 items | 100 items | yes |
| 2 | `GET /repos/{o}/{r}/issues?state=all&per_page=500` | 100 items | up to 500 | yes |
| 3 | `GET /repos/{o}/{r}/issues?state=all&per_page=101` | 100 items | 101 | yes |
| 4 | `GET /search/issues?q=…` | 30 items | 100 items | yes |
| 5 | `GET /search/issues?q=…&per_page=500` | 100 items | up to 500 | yes |
| 6 | `GET /search/code?q=…&per_page=101&page=11` | 422, the depth (101 served as 100) | 422 | no, agrees by accident |

Row 6 is #144's depth check reading the size the route applies: it agrees today because 101 × 11 exceeds 1000 either way, and disagrees as soon as the size is unsent (real refuses page 34 at 30 a page, Backlot page 11 at 100 a page). #144's tests compute that boundary from the settings for this reason, and say so.

**How you know** — measured against api.github.com on 2026-09-06.

- `/repos/psf/requests/issues?state=all`, `/repos/psf/requests/tags`, `/repos/psf/requests/pulls?state=all` and `/search/issues?q=repo:psf/requests+timeout` each answer 30 items with no `per_page`, and 100 items for `per_page=100`, `per_page=101` and `per_page=500` alike. `/user/repos` answers 30 unsent and its full 94 for each of the three, which is under the cap.
- The cap is the one #131 measured on `psf/requests/tags` (`?per_page=500` links `per_page=500` with `last` page 2 over 161 tags, so 100 applied) and the one #144 measured on `/search/code` (`per_page=101` refused at page 11 and served at page 10, so 100 applied). GitHub's own OpenAPI description declares the shared `per-page` parameter as "The number of results per page (max 100)." with `schema: {type: integer, default: 30}` (github/rest-api-description, `descriptions/api.github.com/api.github.com.yaml`, read 2026-09-06), which is the two numbers measured.
- Backlot's two numbers are `Settings.default_page_size = 100` and `Settings.max_page_size = 1000` in `backlot/config.py`, read through `get_settings()` at each `clamp_page` call.

**Reproduction** — an in-code corpus, since the bundled one has no repository with more than 30 documents:

```python
import backlot, requests

records = [
    {
        "source_type": "github",
        "doc_id": f"gh-{i}",
        "repo": "wide",
        "group": "engineering",
        "title": f"Issue {i}",
        "content": "body",
        "author_groups": ["engineering"],
        "visibility": "public",
        "state": "open",
        "created": "2026-01-01T00:00:00Z",
    }
    for i in range(120)
]
with backlot.serve(records) as m:
    org = requests.get(f"{m.base_url}/_meta/users", headers=h).json()["org"]
    for path in (f"/github/repos/{org}/wide/issues", f"/github/search/issues?q=repo:{org}/wide"):
        for params in ({}, {"per_page": 500}, {"per_page": 101}):
            body = requests.get(f"{m.base_url}{path}", headers=h, params=params).json()
            print(path, params, len(body["items"]) if isinstance(body, dict) else len(body))
```

It prints `100`, `120`, `101` for the listing and the same three for the search; real prints `30`, `100`, `100` for each.

**What it costs a client** — a client that pages without naming a size walks a 120-issue repository in two pages here and five on real, so a loop bounded by "page until fewer than 30 come back" ends after the first page against real and never ends early here; one that computes offsets from `per_page=30` reads items 31 to 60 on page 2 against real and items 101 to 200 here. A client that asks for `per_page=500` expecting the cap gets five times real's page, so a buffer or a rate budget sized to 100 is exceeded only in production. `backlot mcp` hands the same `page`/`per_page` arguments to an agent with the same wrong defaults behind them.

**Where it lands** — every `clamp_page(page, per_page, get_settings().default_page_size, get_settings().max_page_size)` call in `backlot/routers/github.py` (the listings through `_paged`, the two searches through their handlers), and `Settings` in `backlot/config.py`, whose two fields are what the other routers read too. `github_link_header` and `github_code_search_link_header` in `backlot/pagination.py` take the applied size as an argument and need no change; `backlot/pagination.py::_absorb_page`'s docstring says "a `per_page` over the cap is a 200 at the cap", which is true of both sides and says nothing about which cap.

**What a fix has to settle**

- Whether the GitHub router carries its own two constants (30 and 100) and leaves `Settings` for the vendors not yet measured, or whether the settings become per-vendor. The first is one file and is what the measurement supports; the second is a configuration change every router feels. Slack's, Jira's and Drive's defaults are not measured here and should not move on GitHub's numbers.
- What the two settings then mean for GitHub. `test_github_a_page_url_omits_a_page_size_the_caller_did_not_send` overrides `BACKLOT_DEFAULT_PAGE_SIZE` to 1 to make a small fixture span pages; if GitHub stops reading the setting, that test needs another way to a short default page, and the setting's documentation says which vendors it still governs.
- Which existing assertions encode 100: `test_github_tolerates_the_pagination_values_real_tolerates` asserts a 200 for `per_page=100_000` "at the cap" without saying which, and the #131 Link tests were written with sizes the fixture spans. Each should keep passing with the real numbers or be corrected to them.
- #144's depth tests, which compute the boundary page from the settings. Once the numbers are real's, those cells should read `page=34` served and `page=35` refused on `/search/issues`, `page=33` served and `page=34` refused on `/search/code`, as the measurements in #144 have them.
- #126's open question, whether the routes it paged should cap at the default when nothing is sent, is this question again: on real the default is 30 on those routes too.

nuri-yoo:
Closes #145. On `main` at `bab49cb`, three commits: the change, a prose correction from a self-review, and the spec's declared defaults (the last two sections); a fourth merges `main` at `7f1df12` (#143) into the branch and changes nothing here; three more apply the review's suggestions as written, and `7ef30e4` takes its other two notes (the last section). Opened first at the suggestion on #144, which rebases onto this once it lands and writes its depth cells as real's numbers.

**What GitHub does.** Real serves 30 items when `per_page` is not sent and 100 for any sent size at or above it, on every listing and search measured against api.github.com on 2026-09-06: `psf/requests/issues?state=all`, `psf/requests/tags`, `psf/requests/pulls?state=all` and `/search/issues?q=repo:psf/requests+timeout` each answer 30 with no `per_page` and 100 for `per_page=100`, `101` and `500` alike; `/user/repos` answers 30 unsent and its full 94 for each of the three, which is under the cap. The cap is the one #131 measured on `psf/requests/tags` (`?per_page=500` links `per_page=500` with `last` page 2 over 161 tags) and the one #144 measured on `/search/code` (`per_page=101` refused at page 11 and served at page 10). GitHub's OpenAPI description declares the shared `per-page` parameter as "The number of results per page (max 100)." with `schema: {type: integer, default: 30}` (github/rest-api-description, `descriptions/api.github.com/api.github.com.yaml`, read 2026-09-06). Backlot sized every GitHub page from `Settings.default_page_size` (100) and `Settings.max_page_size` (1000), the two numbers Slack, Gmail, Drive and the Jira search read too, so the `Link` header #131 made right (`last` computed from the size applied) sat over a page of the wrong length: a client walking a 120-issue repository without naming a size took two pages here and five on real, and one asking for 500 got five times real's page.

**What changed.** `PER_PAGE_DEFAULT = 30` and `PER_PAGE_MAX = 100` at module level in `backlot/routers/github.py`, beside `TREE_MAX_ENTRIES` and for the same reason (a test can lower them; no repository in the bundled corpus spans a page of 30), read through one helper, `_clamp(page, per_page)`, at the sixteen sites that read the two settings: the offset listings through `_paged`, the cursor-paged issue listing through `_paged_by_cursor`, and the two searches. The three `Link` builders already take the applied size and do not move. `Settings.default_page_size` and `max_page_size` stay, and `backlot/config.py` and the Paging table in `docs/configuration.md` now say which vendor reads which: Slack's `limit`, Gmail's and Drive's `maxResults` / `pageSize` and the Jira search's `maxResults` default to the first, Slack and Google cap at the second (the Jira search caps nowhere), and GitHub (30/100), Linear (50/250), Notion (100), Fireflies (50), HubSpot (100, 500 for associations), S3 (1000 either way) and Confluence (25) carry their own; the cap row names what Slack reads through it, `limit`, `count` and `page`. The router still imports `get_settings`, for `org_name`.

**What #145 asked a fix to settle.**

- Router constants rather than per-vendor settings: the measurement supports GitHub's two numbers and nothing else's, and the shape already exists twice in this repository, `TREE_MAX_ENTRIES` in this router and `_PAGE_MAX` in Notion's. Slack's, Jira's and Drive's defaults do not move.
- The two tests that overrode the settings through the environment (`BACKLOT_DEFAULT_PAGE_SIZE=1` to make a short default page, `BACKLOT_MAX_PAGE_SIZE=2` to make the applied size differ from the sent one) lower `gh.PER_PAGE_DEFAULT` and `gh.PER_PAGE_MAX` with `monkeypatch.setattr`, the way the tree tests lower `TREE_MAX_ENTRIES`; the `get_settings.cache_clear()` bracketing goes with the env var. Their docstrings' "the 100 rows Backlot defaults to" reads "the 30 rows a GitHub page defaults to".
- The assertions that encoded 100: `test_github_tolerates_the_pagination_values_real_tolerates` still asserts `per_page=100_000` is a 200, now with the comment saying the cap it lands at is real's 100; the #131 Link tests pass unchanged, since the sizes they send are ones the fixture spans under either cap.
- #144's depth cells stay computed from the settings on that branch until it rebases onto this, when they become `page=34` served and `35` refused on `/search/issues`, `33` and `34` on `/search/code`, as its measurements have them. That is the follow-up on #144, not here.
- #126's open question, whether the routes it paged should cap at the default when nothing is sent: they go through `_clamp` like every other listing on this router, so they default to 30 as real does, and matching real there no longer makes the router inconsistent with itself.

**Tests.** One new test for the page, `test_github_pages_at_reals_thirty_and_caps_at_its_hundred`, on an in-code corpus of 120 issues and 120 files in one repository: the issue listing, `/search/issues` and `/search/code` each serve 30 with no `per_page`, 30 at `per_page=30`, 100 at `100`, 100 at `101` and 100 at `500`, the cursor-paged listing linking `next` to page 2 and the searches linking `last` to the page the applied size reaches (4 at 30, 2 at 100); and the two constants are 30 and 100 while the settings' declared defaults stay 100 and 1000 (read off `Settings.model_fields`, not `get_settings()`, so an exported `BACKLOT_DEFAULT_PAGE_SIZE` does not fail the test). Mutation-checked from a copy and restored: both constants set to the settings' numbers fails the unsent cell (`assert 100 == 30`), the cap alone set to 1000 fails the `per_page=101` cell (`assert 101 == 100`), the default alone set to 100 fails the unsent cell; `test_github_tolerates_the_pagination_values_real_tolerates` passes under each, as it asserts the status and not the length. The issue's reproduction prints `30, 100, 100` for the listing and for the search here, which is what it says real prints, where `bab49cb` prints `100, 120, 101`.

**Verification.**

```
python -m pytest -rs -p no:cacheprovider
2395 passed, 3 skipped   (2026-09-07; the skips are importorskip on llama_index.readers.hubspot and mirage.core.google._client, neither touched here)
python -m pytest tests/test_github.py      137 passed
ruff check .                                All checks passed!
ruff format --check .                       138 files already formatted
python scripts/gen_docs.py --check          exit 0
```

**Self-review, second commit.** Prose only. The settings comment and the Paging table named Jira/Confluence as readers of the two settings; Confluence pages at a fixed 25 and only the Jira search's `maxResults` reads the default, and the cap is read by Slack and Google alone, so both now say that. The new test's docstring said the `Link` header at `bab49cb` named four pages over a page of 100; it named two, since #131 already computed `last` from the size applied, and the sentence now says the header described the wrong length faithfully. Re-run at the second commit: `tests/test_github.py` 137 passed, `ruff check` and `ruff format --check` clean, `gen_docs.py --check` exit 0.

**The spec's defaults, third commit.** GitHub's OpenAPI description declares the shared `per-page` parameter as `{type: integer, default: 30}` and `page` as `{type: integer, default: 1}` (`components/parameters`, read 2026-09-07), and sixteen of the seventeen routes here that page reference those two; the seventeenth, `GET /repos/{owner}/{repo}/statuses/{sha}`, is the legacy alias the description names only in the prose of `/commits/{ref}/statuses`, which references them (the four routes whose inline `per_page` defaults differ, `/notifications` at 50, `/orgs/{org}/copilot/billing/seats` at 50, `/organizations/{org}/settings/billing/budgets` at 10 and `/zen`, are not served). Backlot's slice declared neither default. The handlers declare both as `PageParam = None` and have to, since the `None` is how a handler tells an unsent size from a sent one (the `Link` header omits a size the caller did not send), and FastAPI writes no `default` for a query parameter whose runtime default is `None` whatever the annotation says: `Query(json_schema_extra={"default": 30})` keeps every other extra key and drops that one, and `WithJsonSchema({"type": "integer", "default": 30})` keeps the type and drops the default (both measured on FastAPI 0.141.1, pydantic 2.13.5). So `openapi.github_page_defaults(spec, defaults)` writes the two onto the served document after FastAPI builds it, on GitHub's operations alone, and `app.openapi` in `backlot/main.py` is FastAPI's builder followed by that edit, made in place on the document FastAPI caches so it is the same edit each time. The numbers are the caller's: `main.py` passes `{"per_page": github.PER_PAGE_DEFAULT, "page": 1}`, so 30 is stated once, in the router, and the document cannot declare one size while the route applies another; `openapi` imports no router, since six routers import it for `qp`. The `/_meta/openapi/github` slice `backlot mcp` hands an agent is built from the same document and carries them, which is where the default matters most: an agent reading the tool has no vendor documentation behind it. One test, `test_github_the_spec_declares_reals_page_defaults`: all 34 `page`/`per_page` parameters across the seventeen GitHub routes carry real's default and are still declared integers, Slack's `search.messages` `page` carries none, and the MCP slice's `/search/code` `per_page` reads 30. Mutation-checked from a copy and restored: leaving `app.openapi` unassigned fails it on the first parameter (`KeyError: 'default'`), and dropping the `/github` path filter fails it on Slack's `page`. Re-run at the third commit: full suite `2396 passed, 3 skipped` (the same three skips), `ruff check`, `ruff format --check` and `gen_docs.py --check` clean. Re-run at the merge commit `2942372`, with every extra installed so nothing skips: `2572 passed`, `ruff check` and `ruff format --check` clean, `gen_docs.py --check` exit 0; CI green on 3.11 through 3.14.

**Review, `7ef30e4`.** Five notes, all prose or test hygiene, each reproduced before it was taken. Three are applied as suggested: the settings comment names Linear (50/250), S3 (1000, default and cap) and HubSpot's associations cap (500) beside the vendors it already listed; the `BACKLOT_MAX_PAGE_SIZE` row says Slack reads `limit`, `count` and `page` through it (`slack.py:788` and the `_int` helper); and the docstring of `test_github_only_the_offset_surfaces_write_the_size_the_caller_spelt` gives the reason the cap is lowered that still holds, that no listing in the corpus spans a page of 100 so neither surface would link a `next` (setting it to 100 fails with `KeyError: 'Link'`). Two are taken by hand. The page test asserted the settings through `get_settings()`, which reads the environment, so `BACKLOT_DEFAULT_PAGE_SIZE=200` failed it on a documented knob; it reads `Settings.model_fields[...].default` now, with the import to match. And `openapi.GITHUB_PAGE_DEFAULTS` stated 30 a second time; the function takes the mapping from `main.py`, which passes the router's constant, as the third section now says (mutation: `PER_PAGE_DEFAULT = 25` fails `test_github_the_spec_declares_reals_page_defaults` with `assert 25 == 30`). The reviewer's count of sixteen operations referencing the shared parameters is the right one, and the third section, the spec test's docstring and this body say so. Re-run at `7ef30e4`, every extra installed: `2572 passed`, `ruff check` and `ruff format --check` clean, `gen_docs.py --check` exit 0.

### DIFF
--- a/backlot/config.py
+++ b/backlot/config.py
@@ -59,8 +59,9 @@ class Settings(BaseSettings):
     # `maxResults` / `pageSize` and the Jira search's `maxResults` default to the first, and Slack
     # and Google cap a sent value at the second (the Jira search caps nowhere). A vendor whose
     # numbers are measured or documented carries them in its own router instead: GitHub pages at 30
-    # and caps at 100 (backlot.routers.github.PER_PAGE_DEFAULT / PER_PAGE_MAX), Notion caps at 100,
-    # Fireflies at 50, HubSpot at 100, Confluence pages at 25.
+    # and caps at 100 (backlot.routers.github.PER_PAGE_DEFAULT / PER_PAGE_MAX), Linear at 50 and
+    # caps at 250, Notion caps at 100, Fireflies at 50, HubSpot at 100 (500 for associations), S3
+    # at 1000 either way, Confluence pages at 25.
     default_page_size: int = 100
     max_page_size: int = 1000
 
--- a/backlot/config.py
+++ b/backlot/config.py
@@ -55,10 +55,12 @@ class Settings(BaseSettings):
     admin_token: str = "admin-service-token"
 
     # --- pagination defaults ---
-    # For the vendors whose real page sizes are not measured (Slack, Google, Jira/Confluence). A
-    # vendor whose numbers are measured carries them in its own router instead: GitHub pages at 30
+    # For the vendors whose real page sizes are not measured: Slack's `limit`, Gmail's and Drive's
+    # `maxResults` / `pageSize` and the Jira search's `maxResults` default to the first, and Slack
+    # and Google cap a sent value at the second (the Jira search caps nowhere). A vendor whose
+    # numbers are measured or documented carries them in its own router instead: GitHub pages at 30
     # and caps at 100 (backlot.routers.github.PER_PAGE_DEFAULT / PER_PAGE_MAX), Notion caps at 100,
-    # Fireflies at 50, HubSpot at 100.
+    # Fireflies at 50, HubSpot at 100, Confluence pages at 25.
     default_page_size: int = 100
     max_page_size: int = 1000
 
--- a/backlot/config.py
+++ b/backlot/config.py
@@ -55,6 +55,10 @@ class Settings(BaseSettings):
     admin_token: str = "admin-service-token"
 
     # --- pagination defaults ---
+    # For the vendors whose real page sizes are not measured (Slack, Google, Jira/Confluence). A
+    # vendor whose numbers are measured carries them in its own router instead: GitHub pages at 30
+    # and caps at 100 (backlot.routers.github.PER_PAGE_DEFAULT / PER_PAGE_MAX), Notion caps at 100,
+    # Fireflies at 50, HubSpot at 100.
     default_page_size: int = 100
     max_page_size: int = 1000
 
--- a/backlot/openapi.py
+++ b/backlot/openapi.py
@@ -112,6 +112,37 @@ def qp(
     return p
 
 
+# What GitHub's own OpenAPI description declares for the two paging parameters every one of its
+# listings and searches shares: `per-page` is `{type: integer, default: 30}` and `page` is
+# `{type: integer, default: 1}` (github/rest-api-description, `components/parameters`, read
+# 2026-09-07; every route Backlot serves references those two or repeats the same numbers inline).
+GITHUB_PAGE_DEFAULTS = {"per_page": 30, "page": 1}
+
+
+def github_page_defaults(spec: dict) -> dict:
+    """``spec`` with real's declared defaults written onto every GitHub operation's `page` and
+    `per_page`, in place.
+
+    The GitHub handlers declare both as ``PageParam = None`` and have to: the handler tells an unsent
+    size from a sent one by that ``None`` (a `Link` header omits a size the caller did not send), so
+    the runtime default cannot be 30. FastAPI writes no ``default`` for a query parameter whose
+    default is ``None``, and neither ``json_schema_extra`` nor ``WithJsonSchema`` gets one past it
+    (both measured on 0.141.1 / pydantic 2.13). So the number the route applies, which is the number
+    real declares, is written here, after FastAPI has built the document. Only GitHub: the other
+    vendors' declared defaults are not measured."""
+    for path, item in spec.get("paths", {}).items():
+        if path != "/github" and not path.startswith("/github/"):
+            continue
+        for method, op in item.items():
+            if method not in _METHODS:
+                continue
+            for param in op.get("parameters", []):
+                default = GITHUB_PAGE_DEFAULTS.get(param.get("name"))
+                if default is not None and param.get("in") == "query":
+                    param["schema"]["default"] = default
+    return spec
+
+
 def slice_spec(spec: dict, prefixes: list[str]) -> dict:
     """Copy ``spec`` keeping only paths under one of ``prefixes``."""
     paths = {
--- a/backlot/openapi.py
+++ b/backlot/openapi.py
@@ -112,24 +112,22 @@ def qp(
     return p
 
 
-# What GitHub's own OpenAPI description declares for the two paging parameters every one of its
-# listings and searches shares: `per-page` is `{type: integer, default: 30}` and `page` is
-# `{type: integer, default: 1}` (github/rest-api-description, `components/parameters`, read
-# 2026-09-07; every route Backlot serves references those two or repeats the same numbers inline).
-GITHUB_PAGE_DEFAULTS = {"per_page": 30, "page": 1}
+def github_page_defaults(spec: dict, defaults: dict[str, int]) -> dict:
+    """``spec`` with ``defaults``, parameter name to the number the route applies when that
+    parameter is not sent, written onto every GitHub operation's query parameter of that name, in
+    place.
 
-
-def github_page_defaults(spec: dict) -> dict:
-    """``spec`` with real's declared defaults written onto every GitHub operation's `page` and
-    `per_page`, in place.
-
-    The GitHub handlers declare both as ``PageParam = None`` and have to: the handler tells an unsent
-    size from a sent one by that ``None`` (a `Link` header omits a size the caller did not send), so
-    the runtime default cannot be 30. FastAPI writes no ``default`` for a query parameter whose
-    default is ``None``, and neither ``json_schema_extra`` nor ``WithJsonSchema`` gets one past it
-    (both measured on 0.141.1 / pydantic 2.13). So the number the route applies, which is the number
-    real declares, is written here, after FastAPI has built the document. Only GitHub: the other
-    vendors' declared defaults are not measured."""
+    The GitHub handlers declare `page` and `per_page` as ``PageParam = None`` and have to: the
+    handler tells an unsent size from a sent one by that ``None`` (a `Link` header omits a size the
+    caller did not send), so the runtime default cannot be 30. FastAPI writes no ``default`` for a
+    query parameter whose default is ``None``, and neither ``json_schema_extra`` nor
+    ``WithJsonSchema`` gets one past it (both measured on 0.141.1 / pydantic 2.13). So the numbers
+    are written here, after FastAPI has built the document, and they come from the caller rather
+    than from a constant of this module's own: the router that applies them holds them
+    (``backlot.routers.github.PER_PAGE_DEFAULT``, whose comment cites GitHub's description
+    declaring the same 30), so the document cannot declare one size while the route applies
+    another. Six routers import this module for :func:`qp`, which is why it does not import that
+    one. Only GitHub: the other vendors' declared defaults are not measured."""
     for path, item in spec.get("paths", {}).items():
         if path != "/github" and not path.startswith("/github/"):
             continue
@@ -137,7 +135,7 @@ def github_page_defaults(spec: dict) -> dict:
             if method not in _METHODS:
                 continue
             for param in op.get("parameters", []):
-                default = GITHUB_PAGE_DEFAULTS.get(param.get("name"))
+                default = defaults.get(param.get("name"))
                 if default is not None and param.get("in") == "query":
                     param["schema"]["default"] = default
     return spec
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -19,8 +19,8 @@ the var is unset. There are nine, and this page is all of them.
 
 | Env var | Default | What it does |
 |---|---|---|
-| `BACKLOT_DEFAULT_PAGE_SIZE` | `100` | Page size when a request names none, on Slack, Google and Jira/Confluence |
-| `BACKLOT_MAX_PAGE_SIZE` | `1000` | Ceiling a request may ask for, on the same three |
+| `BACKLOT_DEFAULT_PAGE_SIZE` | `100` | Page size when a request names none: Slack's `limit`, Gmail's and Drive's `maxResults` / `pageSize`, the Jira search's `maxResults` |
+| `BACKLOT_MAX_PAGE_SIZE` | `1000` | Ceiling a sent size is cut to, on Slack and Google |
 
 **A vendor's own numbers still win.** Where the real API documents or measures a page size, Backlot
 serves that one instead: GitHub pages at 30 when `per_page` is not sent and caps it at 100, on every
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -20,7 +20,7 @@ the var is unset. There are nine, and this page is all of them.
 | Env var | Default | What it does |
 |---|---|---|
 | `BACKLOT_DEFAULT_PAGE_SIZE` | `100` | Page size when a request names none: Slack's `limit`, Gmail's and Drive's `maxResults` / `pageSize`, the Jira search's `maxResults` |
-| `BACKLOT_MAX_PAGE_SIZE` | `1000` | Ceiling a sent size is cut to, on Slack and Google |
+| `BACKLOT_MAX_PAGE_SIZE` | `1000` | Ceiling a sent value is cut to, on Slack (`limit`, `count`, `page`) and Google |
 
 **A vendor's own numbers still win.** Where the real API documents or measures a page size, Backlot
 serves that one instead: GitHub pages at 30 when `per_page` is not sent and caps it at 100, on every
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -19,13 +19,16 @@ the var is unset. There are nine, and this page is all of them.
 
 | Env var | Default | What it does |
 |---|---|---|
-| `BACKLOT_DEFAULT_PAGE_SIZE` | `100` | Page size when a request names none |
-| `BACKLOT_MAX_PAGE_SIZE` | `1000` | Ceiling a request may ask for |
+| `BACKLOT_DEFAULT_PAGE_SIZE` | `100` | Page size when a request names none, on Slack, Google and Jira/Confluence |
+| `BACKLOT_MAX_PAGE_SIZE` | `1000` | Ceiling a request may ask for, on the same three |
 
-**A vendor's own cap still wins.** Where the real API documents a maximum, Backlot enforces that
-one instead: Fireflies clamps `limit` to 50 rather than erroring, and HubSpot to 100 — the value its
-official client pages at. Raising `BACKLOT_MAX_PAGE_SIZE` does not lift either, because a client
-that gets 1000 rows from a call the real API caps at 50 is a client that breaks in production.
+**A vendor's own numbers still win.** Where the real API documents or measures a page size, Backlot
+serves that one instead: GitHub pages at 30 when `per_page` is not sent and caps it at 100, on every
+listing and search (its OpenAPI declares `per-page` as "The number of results per page (max 100)."
+with default 30, and api.github.com answers exactly that); Fireflies clamps `limit` to 50 rather
+than erroring, and HubSpot to 100 — the value its official client pages at. Neither variable moves
+any of these, because a client that gets 1000 rows from a call the real API caps at 100 is a client
+that breaks in production.
 
 ## SQLite
 
--- a/tests/test_github.py
+++ b/tests/test_github.py
@@ -1656,9 +1656,9 @@ def test_github_pages_at_reals_thirty_and_caps_at_its_hundred(tmp_path):
     Backlot sized every GitHub page from the server's `default_page_size` (100) and `max_page_size`
     (1000), the two numbers every other router reads: a client walking a 120-issue repository without
     naming a size took two pages here and five on real, and one asking for 500 got five times real's
-    page. The `Link` header was already right (#131 computes `last` from the size applied), so the
-    header said 4 pages over a page that held 100. The corpus is built here because no repository in
-    the bundled one holds more than 30 documents.
+    page. The `Link` header already paged the way real's does (#131: an unsent size omitted, `last`
+    computed from the size applied), so it described the wrong page length faithfully. The corpus is
+    built here because no repository in the bundled one holds more than 30 documents.
     """
     from backlot.routers import github as gh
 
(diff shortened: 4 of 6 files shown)


backlot/config.py
```diff
@@ -55,6 +55,12 @@ def _resolve_path_defaults(cls, values):
     admin_token: str = "admin-service-token"
 
     # --- pagination defaults ---
+    # For the vendors whose real page sizes are not measured: Slack's `limit`, Gmail's and Drive's
+    # `maxResults` / `pageSize` and the Jira search's `maxResults` default to the first, and Slack
+    # and Google cap a sent value at the second (the Jira search caps nowhere). A vendor whose
+    # numbers are measured or documented carries them in its own router instead: GitHub pages at 30
+    # and caps at 100 (backlot.routers.github.PER_PAGE_DEFAULT / PER_PAGE_MAX), Notion caps at 100,
+    # Fireflies at 50, HubSpot at 100, Confluence pages at 25.
```
khj809: The list reads exhaustive, but two routers carrying their own numbers are missing: Linear (`PAGE_DEFAULT = 50` / `PAGE_MAX = 250`, `backlot/graphql/linear_resolvers.py:32`) and S3 (`_MAX_KEYS = 1000`, `backlot/routers/s3.py:40`, which is also its default). HubSpot carries a second number too, `_ASSOC_PAGE_MAX = 500` on the associations read. Someone adding a paged surface reads this and concludes Linear and S3 page from the settings.

```suggestion
    # For the vendors whose real page sizes are not measured: Slack's `limit`, Gmail's and Drive's
    # `maxResults` / `pageSize` and the Jira search's `maxResults` default to the first, and Slack
    # and Google cap a sent value at the second (the Jira search caps nowhere). A vendor whose
    # numbers are measured or documented carries them in its own router instead: GitHub pages at 30
    # and caps at 100 (backlot.routers.github.PER_PAGE_DEFAULT / PER_PAGE_MAX), Linear at 50 and
    # caps at 250, Notion caps at 100, Fireflies at 50, HubSpot at 100 (500 for associations), S3
    # at 1000 either way, Confluence pages at 25.
```

docs/configuration.md
```diff
@@ -19,13 +19,16 @@ the var is unset. There are nine, and this page is all of them.
...
 |---|---|---|
-| `BACKLOT_DEFAULT_PAGE_SIZE` | `100` | Page size when a request names none |
-| `BACKLOT_MAX_PAGE_SIZE` | `1000` | Ceiling a request may ask for |
-
-**A vendor's own cap still wins.** Where the real API documents a maximum, Backlot enforces that
-one instead: Fireflies clamps `limit` to 50 rather than erroring, and HubSpot to 100 — the value its
-official client pages at. Raising `BACKLOT_MAX_PAGE_SIZE` does not lift either, because a client
-that gets 1000 rows from a call the real API caps at 50 is a client that breaks in production.
+| `BACKLOT_DEFAULT_PAGE_SIZE` | `100` | Page size when a request names none: Slack's `limit`, Gmail's and Drive's `maxResults` / `pageSize`, the Jira search's `maxResults` |
+| `BACKLOT_MAX_PAGE_SIZE` | `1000` | Ceiling a sent size is cut to, on Slack and Google |
```
khj809: `max_page_size` is not only a size ceiling on Slack: the same `_int` helper reads `page` through it (`backlot/routers/slack.py:788`, helper at `:1142`), so `search.messages?page=5000` is cut to 1000. `count` goes through it as well, though it defaults to 20 rather than to the other setting.

```suggestion
| `BACKLOT_MAX_PAGE_SIZE` | `1000` | Ceiling a sent value is cut to, on Slack (`limit`, `count`, `page`) and Google |
```

tests/test_github.py
```diff
@@ -1888,21 +2001,18 @@ def test_github_only_the_offset_surfaces_write_the_size_the_caller_spelt(
     `per_page=abc`. The cap is overridden here so Backlot's own applied size differs from the value
     sent, which is what tells the two rules apart at all.
```
khj809: This reason no longer holds: with `PER_PAGE_MAX` at 100 and 500 sent, the applied size already differs from the value sent. Checked what the override buys now — setting it to 100 instead of 2 fails the test with `KeyError: 'Link'`, because no listing in the corpus spans a page of 100 and neither surface emits a `next` at all. So it is still load-bearing, for the other reason.

```suggestion
    `per_page=abc`. The cap is lowered here because no listing in the corpus spans a page of 100,
    so without it neither surface links a `next` to read the size off at all.
```

tests/test_github.py
```diff
@@ -1644,6 +1644,119 @@ def test_github_tolerates_the_pagination_values_real_tolerates(gh_client, gh_adm
...
+            ):
+                params = dict(base) if sent is None else {**base, "per_page": sent}
+                r = c.get(path, headers=h, params=params)
+                assert r.status_code == 200, (path, params)
+                assert len(items(r.json())) == served, (path, params)
+                expected = 2 if rel == "next" else -(-120 // served)
+                assert f"page={expected}" in _link_rels(r.headers["Link"])[rel], (path, params)
+        # ...and the two numbers are real's, not the server's settings
+        assert (gh.PER_PAGE_DEFAULT, gh.PER_PAGE_MAX) == (30, 100)
+        assert (get_settings().default_page_size, get_settings().max_page_size) == (100, 1000)
```
khj809: `get_settings()` reads the environment (`env_prefix="BACKLOT_"`, `env_file=".env"`), so this line fails on a documented knob: with `BACKLOT_DEFAULT_PAGE_SIZE=200` exported the test fails `assert (200, 1000) == (100, 1000)` while every cell above it still passes. The field defaults say the same thing without reading the environment, and `tests/test_docs.py:196` already reaches for `Settings.model_fields`. Takes `Settings` in place of `get_settings` on the `backlot.config` import at the top, which this line is now the last user of.

```suggestion
        assert Settings.model_fields["default_page_size"].default == 100
        assert Settings.model_fields["max_page_size"].default == 1000
```

backlot/openapi.py
```diff
@@ -112,6 +112,37 @@ def qp(
     return p
 
 
+# What GitHub's own OpenAPI description declares for the two paging parameters every one of its
+# listings and searches shares: `per-page` is `{type: integer, default: 30}` and `page` is
+# `{type: integer, default: 1}` (github/rest-api-description, `components/parameters`, read
+# 2026-09-07; every route Backlot serves references those two or repeats the same numbers inline).
+GITHUB_PAGE_DEFAULTS = {"per_page": 30, "page": 1}
```
khj809: Not part of my original pass — this arrived in b6046b7, and it is the one thing in it that reintroduces what the rest of the PR removed: `30` is now stated in two places, here and `github.PER_PAGE_DEFAULT`. Move that constant and the served spec keeps declaring 30 while the router applies something else, which is the divergence #145 opened about.

A module-level `from backlot.routers import github` would invert the layering (six routers import this module for `qp`), so either read it in the function or pin the invariant in the test that commit added:

```python
    from backlot.routers import github as gh

    assert openapi.GITHUB_PAGE_DEFAULTS["per_page"] == gh.PER_PAGE_DEFAULT
```

The `page: 1` half has the same shape against `clamp_page`'s hardcoded 1, which matters less.

khj809 (commented):
Reviewed the change against api.github.com and against GitHub's own OpenAPI description; the 30/100 part holds everywhere I could measure it, including two cells the body does not claim. Five notes below, all prose or test hygiene — nothing that changes what is served.

Measured on api.github.com on 2026-09-07, on endpoints this PR does not name: `/orgs/github/repos` answers 30 unsent with no `per_page` in the `Link`, and 100 for `per_page=100`, `101` and `500` alike, each echoing the caller's spelling with `last` computed from the applied 100; `/search/issues?q=repo:psf/requests+timeout` answers 30 unsent with `last` page 29 over 846 results, and 100 for `101` and `500`. All sixteen served operations `$ref` the shared `per-page` / `page` parameters in `descriptions/api.github.com` — no operation declares its own — so the two numbers really are the whole surface's.

Two cells the cap at 100 fixes that the body does not mention: the cursor-paged issue listing writes the size it APPLIED, and real writes `per_page=100` for `?per_page=500` there (measured), where this branch's parent wrote `per_page=500`; and an unsent request on that listing now writes `per_page=30`, which is what real writes. `/search/code` unsent likewise moves from `per_page=100` to `per_page=30`.

Full suite at 2942372: 2570 passed, 2 skipped (Docker and `llama_index.readers.hubspot`, neither touched here); `ruff check`, `ruff format --check` and `gen_docs.py --check` clean.

tests/test_github.py
nuri-yoo: Reproduced: `BACKLOT_DEFAULT_PAGE_SIZE=200` fails that line with `assert (200, 1000) == (100, 1000)` and nothing above it. Taken as written, plus the import it needs: `from backlot.config import Settings` replaces `get_settings` at the top, which that line was the last user of. `BACKLOT_DEFAULT_PAGE_SIZE=200` now passes the test.

backlot/openapi.py
nuri-yoo: Agreed, and taken the third way: `github_page_defaults(spec, defaults)` takes the mapping from its caller, and `backlot/main.py`, which already imports both the router and `openapi`, passes `{"per_page": github.PER_PAGE_DEFAULT, "page": 1}`. The constant in `openapi` is gone, so 30 is stated once, in the router, whose comment now says the served spec reads it; `openapi` still imports no router. Mutation: `PER_PAGE_DEFAULT = 25` fails `test_github_the_spec_declares_reals_page_defaults` with `assert 25 == 30`. The `page: 1` half stays a literal beside the call, with the comment naming `clamp_page` as where it comes from.

Your count is the right one, too: sixteen operations `$ref` the shared parameters. The seventeenth route here, `GET /repos/{owner}/{repo}/statuses/{sha}`, is the legacy alias the description names only in the prose of `/commits/{ref}/statuses`, which does reference them. The spec test's docstring and the PR body said seventeen; both now say sixteen plus the alias.

khj809 (approved):
All five notes taken, and the last one better than it was suggested: `github_page_defaults(spec, defaults)` removes the second statement of `30` instead of pinning it, and it holds — moving `PER_PAGE_DEFAULT` to 7 moves the served spec's declared default to 7, so the document cannot describe a page the route does not serve. The page test survives an exported `BACKLOT_DEFAULT_PAGE_SIZE=200` now. Verified at `7ef30e4`: 2570 passed, 2 skipped here (Docker, and `llama_index.readers.hubspot` not installed), `ruff check`, `ruff format --check` and `gen_docs.py --check` clean. The sixteen-plus-alias count checks out against the description too — the legacy statuses path declares `post` alone, and `/commits/{ref}/statuses`, which does reference the shared parameters, names the read in its prose.

One prose slip in the third section, not worth a commit of its own: the parenthetical calls it four routes whose inline `per_page` default differs and puts `/zen` among them. There are three — `/notifications` at 50, `/orgs/{org}/copilot/billing/seats` at 50 and `/organizations/{org}/settings/billing/budgets` at 10, and all three are indeed unserved — while `/zen` declares no parameters at all, so it is not in that set. Worth folding in whenever the body is next edited.

The other half of the spec line these defaults come from, `per_page`'s description and the ceiling only it states, is #153.

PR #157, opened 2026-09-07
linear: `Team.initiativesEnabled` and `Project.resourceCount` are served, so `@linear/sdk`'s own Team, Teams and Project documents validate

khj809:
`@linear/sdk` ships one pre-built document per operation and graphql-core validates before it executes, so a single field a fragment selects and Backlot does not declare rejects the whole query. #154 moves the Linear example from `@linear/sdk` 88.3.0 to 93.0.1, whose Team and Project fragments select two such fields: `client.teams()` came back as `UnknownLinearError: Cannot query field 'initiativesEnabled' on type 'Team'` and the example step failed on all four Python versions. This PR stacks on #154 so its own CI runs the example against the new SDK.

`Team.initiativesEnabled` (`Boolean!`) and `Project.resourceCount` (`Int!`) are now declared and resolved. Both were acknowledged gaps in `backlot/fidelity/baseline/linear.json` — surface the vendor has and Backlot does not — which holds right up to the point where the SDK's own documents select them. The two entries are gone from the baseline, and `backlot diff --source linear` reports nothing new.

Which fields, measured rather than guessed: each version's bundle ships 663 documents and 656 of the names are in both, and validating those against the served SDL, exactly three query documents were valid at 88.3.0 and invalid at 93.0.1 — `TeamDocument` and `TeamsDocument` on `Team.initiativesEnabled`, and `ProjectDocument` on `Project.resourceCount`, which the SDK reaches lazily through `await issue.project`. Both fields exist on api.linear.app at the types declared here, and a real team answers `initiativesEnabled: false` (measured 2026-09-07). Each value states what Backlot has rather than a vendor default: initiatives are out of scope here, so no team has them enabled, and Backlot serves no project resources, so the count is 0.

`tests/test_linear.py` holds the declarations to the vendor's types alongside #101's findings and pins that the resolvers produce values, since a declaration with nothing behind it voids a non-null result just as surely as a missing declaration rejects the query. The example runs to completion against 93.0.1 and `npx tsc --noEmit` passes under TypeScript 7.0.2.

### DIFF
--- a/backlot/graphql/linear.graphql
+++ b/backlot/graphql/linear.graphql
@@ -757,6 +757,7 @@ type Project {
   scope: Float!
   priorityLabel: String!
   priority: Int!
+  resourceCount: Int!
   color: String!
   content: String
   slugId: String!
@@ -921,6 +922,7 @@ type Team {
   slackNewIssue: Boolean!
   slackIssueStatuses: Boolean!
   triageEnabled: Boolean!
+  initiativesEnabled: Boolean!
   inviteHash: String!
   issueOrderingNoPriorityFirst: Boolean!
   issueSortOrderDefaultToBottom: Boolean!
--- a/backlot/graphql/linear_resolvers.py
+++ b/backlot/graphql/linear_resolvers.py
@@ -422,7 +422,7 @@ def _state(name: str, team: str, info) -> dict:
 
 
 def _project(name: str | None, info) -> dict | None:
-    """A ``Project``. The corpus knows a project only by name, so the 26 non-null fields the SDK's
+    """A ``Project``. The corpus knows a project only by name, so the 27 non-null fields the SDK's
     fragment demands take neutral values — empty history arrays, zero progress/scope — rather than
     invented burndown data. `state` is Linear's project state string; "started" is the only claim
     Backlot can make about a project it sees issues in."""
@@ -445,6 +445,7 @@ def _project(name: str | None, info) -> dict | None:
         # `Float!` the SDK's TypeScript `number` once suggested.
         "priority": 0,
         "priorityLabel": "No priority",
+        "resourceCount": 0,
         "progress": 0.0,
         "scope": 0.0,
         "sortOrder": 0.0,
@@ -660,7 +661,7 @@ def resolve_team_issue_count(team, info) -> int:
 
 
 def _team(container: str, info) -> dict:
-    """A ``Team``. 42 of its fields are non-null in the SDK's fragment; the ones Backlot cannot
+    """A ``Team``. 43 of its fields are non-null in the SDK's fragment; the ones Backlot cannot
     know take Linear's own product defaults (cycles off, 2-week duration, estimate scale
     ``notUsed``) rather than zero values that would read as configured."""
     key = _team_key(container, info)
@@ -702,6 +703,7 @@ def _team(container: str, info) -> dict:
         "issueOrderingNoPriorityFirst": False,
         "requirePriorityToLeaveTriage": False,
         "triageEnabled": False,
+        "initiativesEnabled": False,
         "groupIssueHistory": True,
         "ledInitiativeCount": 0,  # `Int!` in Linear, like `issueCount`
         "aiDiscussionSummariesEnabled": False,
--- a/backlot/fidelity/baseline/linear.json
+++ b/backlot/fidelity/baseline/linear.json
@@ -1,7 +1,7 @@
 {
   "source": "linear",
   "endpoint": "https://api.linear.app/graphql",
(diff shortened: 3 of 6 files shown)


backlot/graphql/linear_resolvers.py
```diff
@@ -660,7 +661,7 @@ def resolve_team_issue_count(team, info) -> int:
 
 
 def _team(container: str, info) -> dict:
-    """A ``Team``. 42 of its fields are non-null in the SDK's fragment; the ones Backlot cannot
+    """A ``Team``. 43 of its fields are non-null in the SDK's fragment; the ones Backlot cannot
```
nuri-yoo, line 664: Nit, inherited: the fragment does not select `issues`, so the count of non-null fields the SDK's fragment demands is 42 here (43 is the served type's non-null count, `issues: IssueConnection!` included). The 42 on `main` had the same off-by-one against 41. Either 42, or "43 of its fields are non-null in the served type". The 27 in `_project` is exact either way: the Project fragment selects every non-null field.

backlot/graphql/linear.graphql
```diff
@@ -921,6 +922,7 @@ type Team {
   slackNewIssue: Boolean!
   slackIssueStatuses: Boolean!
   triageEnabled: Boolean!
+  initiativesEnabled: Boolean!
```
nuri-yoo, line 925: Observation, not for this PR: this line is guarded by `TeamsDocument` in the example, but `TeamDocument` (the `team(id:)` root, which also selects `initiativesEnabled`) is one of 19 SDK documents that validate against this branch and are never sent by the example, so a future bump that touches only a by-id fragment would pass CI. The DRIFT paragraph at the top says `backlot diff` notices the next drift, which holds for the vendor's schema and not for the SDK's documents, as this PR shows.

nuri-yoo (approved):
Approving. Every claim in the description reproduced, so the notes below are prose and process, not code.

**What I measured, all at 436d134.** Introspecting api.linear.app answers `Team.initiativesEnabled` as `Boolean!` and `Project.resourceCount` as `Int!` (description: "The number of resources associated with the project, including documents, external links, and attachments"); a real team answers `initiativesEnabled: false`. Validating the SDK bundle with graphql-core against the served SDL: 88.3.0 and 93.0.1 both ship 663 documents, 656 of them under the same name in both versions, and over those 656 exactly `ProjectDocument`, `TeamDocument` and `TeamsDocument` flip from valid to invalid; against this branch the three validate and nothing that validated before stops. The example on `main` with 93.0.1 fails with `UnknownLinearError: Cannot query field 'initiativesEnabled' on type 'Team'`; on this branch it runs to completion and `npx tsc --noEmit` passes. `backlot diff --source linear` live: 774 acknowledged, 0 new. `tests/test_linear.py`: 766 passed. Reverting the two resolver values alone turns the new test and the counts test red with `Cannot return null for non-nullable field`; reverting the two SDL lines alone turns them and the `LINEAR_TYPES` rows red with `Cannot query field`. The 27 in `_project` and "Eight of the nine `Int!` counts" both count out. The four bumped packages' lockfile integrity hashes match the registry.

**Baseline `measured`.** The file still says `2026-09-06`. Running `--update-baseline` on this branch produces a file that differs from yours in that one line only (entries and order identical), and the description dates the measurement to 09-07. Of the three earlier commits that changed entries, #101 and #142 bumped the date and #116 did not, so this is a consistency nit rather than a rule.

**"all 663 documents".** Both versions ship 663, but seven were removed and seven added between them, so the "valid at 88.3.0 and invalid at 93.0.1" comparison is over the 656 common names. The finding of exactly three is right; the sentence just implies a larger denominator than it has.

**What caught this, and what would not have.** Both fields were acknowledged gaps, so `backlot diff` reports nothing new on `main` either (measured: 776 acknowledged, 0 new, exit 0, on a checkout where the example fails). The only tripwire for SDK-document drift is the example step, and instrumenting the shim shows it sends nine operations: `viewer`, `teams`, `issues`, `workflowState`, `user`, `project`, `issue_labels`, `team_issues`, `issue_comments`. Of the 28 SDK documents that validate against this branch, the other 19 are unguarded, `TeamDocument` among them (the example calls `teams`, never `team(id:)`). Not for this PR; a line in the DRIFT paragraph of `linear.graphql` or a follow-up issue would record it.

**Stacking.** The dependabot commit here is byte-identical to #154's head as a patch. Squashing #154 first and then merging this branch gives a tree identical to this head; squashing this first leaves #154 as an empty change on `examples/`. GitHub's docs say Dependabot closes such PRs once the update lands on the default branch, with the caveat that it may occasionally stay open, in which case `@dependabot recreate` forces the check.

PR #160, opened 2026-09-08
ruff's lint selection includes isort, and every import block in the repository is sorted

khj809:
## What changed

`[tool.ruff.lint] select` carries `"I"`, so import order is a gate rather than a matter of taste, and the 66 `I001` findings that turned up are fixed across 49 files. Every edit is ruff's own safe autofix: imports reordered inside the block they already sat in, plus the blank line isort wants between a stdlib, third-party and first-party group. No import moved to a different place in the file, and nothing else in the repository changed.

The selection needs no isort configuration. The default `src = ["."]` classifies `backlot` as first-party correctly, and the examples' sibling helpers (`_agent`, `_helpers`, `_common.google_creds`) sort as third-party — cosmetic, and not worth a `known-local-folder` entry that would then have to be kept in step with the examples tree.

A comment beside `select` records why `I` does not disturb the `E402` sites the ignore under it exists for: isort sorts within a run of consecutive imports and never lifts one over a statement, so the examples' `sys.path` shadowing guards, and the tests' imports below a module-level `importorskip` or a mid-file section break, all stay where they are. `examples/using-official-sdk/github.py`, which drops its own directory from `sys.path` because the file shadows PyGithub, is untouched.

No vendor surface is involved: no route, schema, ACL or response shape changes, and the 11 lines E402 fires on are the same 11 before and after.

## Verification

The claim that no import crossed a statement is checked rather than asserted. For each of the 49 changed Python files, the module was parsed before and after and reduced to a sequence of statements in which each run of consecutive imports collapses to the set of imports it contains, recursively through every function and class body; the two sequences are identical for all 49. An import that had crossed a `sys.path` guard or an `importorskip` would have changed that sequence.

- **Tests** — `pytest -rs` on a `uv sync --all-extras` install: 2631 passed, 3 skipped, 0 failed.
- **Optional surfaces that ran rather than skipped** — the S3/SigV4 tests, the SDK tests, the MCP tests, the LlamaIndex readers and the mirage and Google integration tests all ran. The 3 skips are the two absent optional imports that skip identically on `main`: `llama_index.readers.hubspot` (its stale `hubspot-api-client<9` pin, documented in `pyproject.toml`) and `mirage.core.google._client`.
- **Lint** — `ruff check . && ruff format --check .`: `All checks passed!` and `138 files already formatted`, under both 0.16.6 (`uv.lock`, and what the CI lint job installs) and 0.15.22 (the floor of the `dev` extra's range). The formatter and isort do not disagree: the formatted-file count is unchanged.

- [ ] A test fails without this change
- [ ] A new endpoint serving corpus content is ACL-scoped, proved for both an admin and a scoped token

The first box stays unchecked deliberately: the gate for this change is `ruff check .`, which fails on `main` with 66 errors the moment `"I"` is selected and passes here. The second does not apply — no route, schema or ACL is touched.

### DIFF
--- a/backlot/fidelity/__init__.py
+++ b/backlot/fidelity/__init__.py
@@ -17,9 +17,9 @@ schema loader, a spec fetcher, a path walker — would invite a caller to assemb
 hand and get a different answer than the command does.
 """
 
+from backlot.fidelity.comparisons import COMPARISONS, baseline_path, divergences
 from backlot.fidelity.errors import CredentialsMissing, FidelityError
 from backlot.fidelity.findings import BREAKING, GAP, Baseline, Finding
-from backlot.fidelity.comparisons import COMPARISONS, baseline_path, divergences
 
 __all__ = [
     "BREAKING",
--- a/backlot/fidelity/comparisons.py
+++ b/backlot/fidelity/comparisons.py
@@ -16,9 +16,8 @@ from __future__ import annotations
 
 import os
 from dataclasses import dataclass
-from typing import Any, Callable, Mapping
 from pathlib import Path
-
+from typing import Any, Callable, Mapping
 
 from backlot.fidelity import (
     google_discovery_diff,
--- a/backlot/fidelity/graphql_diff.py
+++ b/backlot/fidelity/graphql_diff.py
@@ -37,10 +37,9 @@ from graphql import (
     is_input_object_type,
     is_interface_type,
     is_object_type,
-    is_union_type,
     is_specified_scalar_type,
+    is_union_type,
 )
-
 
 import backlot.graphql
 from backlot.fidelity.errors import FidelityError
--- a/backlot/fidelity/openapi_diff.py
+++ b/backlot/fidelity/openapi_diff.py
@@ -17,8 +17,8 @@ from backlot.fidelity.errors import FidelityError
 from backlot.fidelity.fetch import fetch_json
 from backlot.fidelity.findings import Finding
 from backlot.fidelity.operations import (
-    Operation,
     _METHODS,
+    Operation,
     _query_params,
     _resolve,
     canonical,
--- a/backlot/fidelity/s3_probe.py
+++ b/backlot/fidelity/s3_probe.py
@@ -35,9 +35,9 @@ from urllib.parse import urlsplit
 import httpx
 
 from backlot import sigv4
-from backlot.fidelity.findings import BREAKING, GAP, Finding
 from backlot.fidelity.errors import FidelityError
 from backlot.fidelity.fetch import fetch_json
+from backlot.fidelity.findings import BREAKING, GAP, Finding
 
 _ROOT = re.compile(r"<\??[a-zA-Z]*[^>]*>\s*<([A-Za-z][\w.-]*)")
 _EMPTY_SHA = hashlib.sha256(b"").hexdigest()
--- a/backlot/pagination.py
+++ b/backlot/pagination.py
@@ -14,7 +14,6 @@ from urllib.parse import quote
 from fastapi import Query
 from pydantic import BeforeValidator
 
-
 # --- opaque offset cursor (Slack next_cursor, Gmail/Drive pageToken, Jira token) ---
 
 
--- a/backlot/routers/atlassian.py
+++ b/backlot/routers/atlassian.py
@@ -15,10 +15,10 @@ from fastapi import APIRouter, HTTPException, Request
 from pydantic import BaseModel, ConfigDict
 
 from backlot import auth, store, synth
-from backlot.openapi import qp
 from backlot.acl import Caller
 from backlot.config import get_settings
 from backlot.errors import atlassian as errors_atlassian
+from backlot.openapi import qp
 from backlot.pagination import confluence_next_link, decode_cursor, next_page_token
 
 router = APIRouter(prefix="/atlassian", tags=["atlassian"])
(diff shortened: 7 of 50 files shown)


nuri-yoo (approved):
Approving. Everything the body claims, I reproduced; the notes below are about the wording of the new comment, and two things outside this PR that it makes more pressing.

**Reproduced**

- From `main` at `ea3dfc7` with only this PR's `pyproject.toml` applied, `ruff check --fix .` produces a tree byte-identical to `b2f31b7`, under both ruff 0.15.22 (the range floor) and 0.16.6 (uv.lock, and what the CI lint job installed per its log). 66 I001 across 49 files, as stated.
- An independent AST pass (each run of consecutive imports collapsed to a set, recursing into function, class, try and if bodies) gives the same statement sequence and the same import-binding set for all 49 files. The only name bound twice within one run is `backlot` in `tests/test_server.py` (`import backlot` + `import backlot.server`), the same object either way.
- The one reordered module with import-time code is `backlot/integrations/mirage.py` (the certifi load into aiohttp's cached SSL context); it is written to be order-independent and its position relative to `mirage` did not change. `_agent`, `_helpers`, `_common.google_creds`, `tests/_helpers.py` and `tests/conftest.py` have no top-level calls beyond one `re.compile`.
- `ruff check --show-files .` lists 138 `.py` files, the same 138 `find` sees; no `.pyi` or notebooks. With `required-imports` unset, `"I"` turns on I001 only.
- `pytest -rs` on an all-extras venv: 2633 passed, 1 skipped (no Docker), 0 failed.

**What the E402 ignore actually covers (non-blocking, wording)**

The new comment and the body say the ignore exists for "`sys.path` inserts, `importorskip` calls and shadowing guards". Measured with `ruff check --isolated --select E402 --ignore-noqa .` on this head under 0.16.6, E402 fires at 11 lines in 7 files, and the breakdown is different.

- The ten `sys.path.insert` calls under `examples/` are not E402 sites. The rule text says E402 "makes an exception for `sys.path` modifications", and a three-line probe (`import sys`, `sys.path.insert(0, "x")`, `import os`) passes with no ignore at all. The ignore does nothing for them.
- `tests/` has no `sys.path` manipulation (the `tests/test_mcp.py` docstring forbids it), so "the examples' and tests' `sys.path` inserts" attributes to tests something only the examples do.
- What the ignore does cover: the five shadowing guards (the inline `sys.path[:] =` in `examples/using-official-sdk/github.py` and `hubspot.py`, and `drop_self_from_syspath(__file__)` in the three `using-llamaindex-readers` examples), the module-level `importorskip` at `tests/test_s3.py:654`, and one kind the list misses, the mid-file section import at `tests/test_pagination.py:230-232`.

The claim the comment exists to make, that isort sorts within a run and never lifts an import over a statement, holds; the AST pass above is exactly that check. Narrowing the last clause to something like "so the examples' shadowing guards and the tests' module-level `importorskip` calls still precede the imports they protect" would match what E402 fires on.

**Outside this PR**

- Of the 11 E402 sites, 9 also carry `# noqa: E402`, so the global ignore and the markers overlap; the 3 markers at `tests/test_mcp.py:50-52` are stale (nothing there triggers E402 any more); only `examples/using-llamaindex-readers/github.py:22` and `examples/using-official-sdk/github.py:24` depend on the global ignore alone. One mechanism or the other could go.
- `pyproject.toml:59` says the ruff requirement is "Pinned, not a lower bound ... so local and CI cannot disagree". That sentence dates from #44, when the specifier was `~=`; since #156 it is `>=0.15.22,<0.17.0`, CI resolves the newest release in range through pip, and `uv sync` installs the locked 0.16.6. They coincide today. With isort in the gate, the first 0.16.x that changes a sort decision makes them disagree in exactly the way the comment says cannot happen. CI installing from `uv.lock`, or an `==` in the dev extra that dependabot moves, would make the sentence true again.

One small thing on the body: the template's "Why this is what the real API does" section is absent. #158 kept it with a one-line "no vendor surface is involved", which reads better than dropping the heading.

khj809:
Comment narrowed in 2c78568, and the body says the same thing now (the vendor heading is back, with a one-liner). The clause reads:

> so the examples' sys.path shadowing guards, and the tests' imports below a module-level `importorskip` or a mid-file section break, all stay where they are

Reproduced your E402 breakdown before rewording: `--isolated --select E402 --ignore-noqa` gives the same 11 sites in 7 files, a three-line probe confirms the rule already excepts a plain `sys.path.insert`, and `tests/` has no `sys.path` manipulation at all — the two hits are prose in `test_docs.py:160` and `test_mcp.py:21`. Also checked the range floor: 0.15.22 reports `All checks passed!` and `138 files already formatted` on this head, same as 0.16.6.

One correction, which makes your follow-up stronger rather than weaker: all 11 sites carry `# noqa: E402`, not 9. `examples/using-llamaindex-readers/github.py:22` and `examples/using-official-sdk/github.py:24` both have the marker inline. `ruff check --isolated --select E402 .` (noqa honored, so the pyproject ignore is out of the picture) is clean, while `--ignore-noqa` reports all 11. So `ignore = ["E402"]` is redundant today — dropping it leaves the lint green, and the three stale markers at `tests/test_mcp.py:50-52` come out in the same sweep. Left for its own PR; nothing in this one depends on which mechanism survives.

On `pyproject.toml:59`: the sentence has been wrong since #156 regardless of this PR, and the risk it describes is latent rather than live — both ends of the range produce this exact tree, so local and CI agree today. Worth its own change, and it is a repo-policy call rather than a lint one: CI installing from `uv.lock` pins the gate to the lockfile dependabot already bumps, while `==` in the dev extra keeps pip installing it but makes every ruff bump a reviewed diff. Say which you prefer and I will send it.

Issue #159, opened 2026-09-08
sheets: a spreadsheet is one sheet of single-column lines, so a real grid — tabs, columns, typed cells — has no representation


khj809:
Backlot models a spreadsheet as **one sheet whose every cell is a whole line of the stored text**. A real spreadsheet is a workbook of named sheets over a 2D grid of typed cells. Nothing in the corpus format, the store, or the served API can express a table today, so a client that needs one — reading a real grid now, writing one later — cannot get it.

This is a tracking issue: the work below is a stack, not a single PR.

## What is served today

| Fact | Where |
|---|---|
| Every spreadsheet is one sheet titled `Sheet1`, `sheetId` 0, `index` 0 | `SHEETS_SHEET_TITLE`, `sheets_get` in `backlot/routers/google.py` |
| A cell is a whole line: `content.split("\n")` -> `[[line], [line], ...]`, one column | `_sheets_grid` |
| Every cell is served as a string — `effectiveValue` is always `{"stringValue": ...}` | `_sheets_grid_data` |
| The grid is declared 1000x26, but only column A can ever hold anything | `SHEETS_GRID_ROWS` / `SHEETS_GRID_COLS` |
| The store holds one `content TEXT` blob per file; there is no grid column and no sheet table | `gdrive_files` in `backlot/store.py` |
| The BYO corpus record has `content` and `subtype` only; no field can carry rows and columns | `backlot/schemas/google_drive.schema.json` |
| `files.export?mimeType=text/csv` serves `content` verbatim, independent of the grid the Sheets API builds from it | `drive_files_export` |

`_sheets_grid_data` already states two of these as known divergences in its own docstring. This issue is about the one underneath them: the grid itself is not a grid.

## Why the current shape is right, and why it is still not enough

This is not an oversight. #35 measured 1,875 spreadsheet records in the bench corpus and found none of them delimiter-uniform CSV: 82.6% prose, 17.4% prose wrapped around a pipe-delimited table. Splitting those on a comma manufactures columns out of sentence punctuation, so `_sheets_grid` deliberately splits on nothing. That reasoning holds and must survive this work.

What it establishes is that **prose cannot be parsed into a grid**, not that a grid has no place. A corpus owner who has an actual table has no way to say so. The fix is to let the corpus state its columns rather than to let the mock guess them.

## Why this matters now

Two independent reasons, either sufficient on its own.

[...]

Issue #35, opened 2026-08-04
Sheets/Drive disagree on CSV cell splitting for 94 bench spreadsheets


khj809:
`spreadsheets.get` (and now `values.get`/`values:batchGet`, #34) split a spreadsheet's stored text with `line.split(",")`, while `files.export?mimeType=text/csv` serves the raw text — so a CSV-aware consumer parses it properly. For most of the corpus those agree. For some they do not.

Measured over the bench corpus (1,875 spreadsheets, 104,781 non-empty lines):

| | |
|---|---|
| spreadsheets where `split(",")` != `csv.reader` | **94** (5.0%) |
| non-empty lines that differ | **512** (0.49%) |

The cause is not a quoted comma inside a field — it is a **whole line wrapped in double quotes**:

```
"1,EU fallback,Lena Rodriguez (SRE),H,2026-09-06,in_progress,limited spare GPUs,spin 2 nodes + reroute,1-2"
```

- `split(",")` -> 9 cells, the first being `"1`
- `csv.reader`  -> **1** cell holding the whole line

So for those 94 documents the mock already gives two different answers depending on which API you ask:

| call | result |
|---|---|
| `spreadsheets.get` / `values.get` | 9 columns, first cell carries a stray `"` |
| `files.export` (text/csv), parsed | 1 column |

Real Sheets has one grid; a client cannot get two shapes for one document.

## What needs deciding

Which reading is correct is genuinely ambiguous, and that is why this is an issue rather than a patch:

1. **The quoting is spurious** (the generator wrapped rows it should not have) -> intended shape is 9 columns, and the fix is to strip a fully-wrapping quote pair before splitting.
2. **The quoting is meaningful CSV** -> intended shape is 1 column, and the fix is to use `csv.reader` everywhere.

Option 2 is the correct reading of CSV as a format, but it changes what `spreadsheets.get` has served since it was added, and the stray `"` that option 1 removes is itself evidence the quoting was not intended to be read as CSV.

Worth checking what the raw ERB source records look like for one of the 94 before choosing — the answer is in the source data, not in the mock.

## Scope

- `_sheets_rows` / `sheets_get` in `app/routers/google.py` (one splitter, shared)
- whichever choice is made, `files.export` and the Sheets APIs must end up agreeing
- a test over the affected shape, since the SAMPLE corpus has no quoted lines and cannot catch this

[...]

PR #161, opened 2026-09-08
Improve fidelity on Google Sheets APIs

khj809:
A `google_drive` record can state a spreadsheet's actual grid — named sheets, 2D rows, typed cells — and that grid carries through import, storage, the Sheets v4 read surface and Drive's CSV/TSV export. Work list B, C, D, F, G and H of #159, plus the Sheets read surface's fidelity gaps that the grid work put within reach.

Before this, a spreadsheet was one sheet titled `Sheet1` whose every cell was a whole line of the stored text. That shape stays, and stays the default: #35 measured 1,875 real spreadsheet records and found none of them delimiter-uniform CSV, so splitting prose on a comma manufactures columns out of sentence punctuation. What changes is that a corpus owner who *has* a table can now say so, instead of the mock guessing.

## The corpus

```json
{"source_type": "google_drive", "subtype": "spreadsheet", "title": "Q3 pipeline", "folder": "sales",
 "sheets": [{"title": "Summary", "grid": [["Region", "Deals", "Won"], ["EMEA", 12, true]]},
            {"grid": [["id", "amount"], ["a-1", 1250.5]]}]}
```

A cell is a JSON scalar — string, number, boolean, or `null` for an empty cell — so the type a corpus states is the type the API serves. `sheets` and `content` are mutually exclusive: `content` is derived from the first sheet at import, so Drive's CSV export and the Sheets API cannot describe one document two ways. A record with no `sheets` imports as before, and serves as before but for the grid height noted under review.

Dates, formulas and error cells have no representation. That is a stated limitation rather than a divergence — a corpus says what a cell holds, and JSON has no date.

## What is served

- `spreadsheets.get` returns one entry per sheet with a real `sheetId`, `index`, `title` and per-sheet `gridProperties`, beside the `autoRecalc`, `timeZone`, `defaultFormat` and `spreadsheetTheme` real always carries. `ranges` filters the `sheets` array itself, and a sheet several ranges touch gets one `data` block per range, each with its own `rowMetadata` and `columnMetadata`.
- An A1 range names any sheet in the workbook — qualified, quoted, unquoted, case-insensitively — and the echoed `range` normalises the title and requotes it. An unknown sheet is 400 `INVALID_ARGUMENT`.
- A cell carries `userEnteredValue`, `effectiveValue`, `formattedValue` and `effectiveFormat`, so `valueRenderOption` finally distinguishes: `UNFORMATTED_VALUE` answers `12`, `FORMATTED_VALUE` answers `"12"`.
- `files.export` serialises the first sheet — RFC 4180 for CSV, and the lossy space-substitution real TSV does.
- `spreadsheets:getByDataFilter` and `values:batchGetByDataFilter` serve the same reads addressed by `a1Range` or `gridRange`. Reads issued over POST, because the filters do not fit in a query string.
- Every Sheets read honours `fields` (a nested partial-response mask), `prettyPrint`, `alt`, `callback`, `quotaUser` and `upload_protocol`, and refuses a value the real API refuses.

## Measurements

Every fidelity claim here was measured against the live Sheets v4 and Drive v3 APIs over a purpose-built multi-sheet workbook; work list A of #159 records them. The non-obvious ones the implementation turns on:

- `sheetId` is not the index. Only the sheet created with the spreadsheet is `0`; every later one carries a large pseudo-random integer.
- The sheet/range separator is the **last** `!`, and a bare `A1` is a cell on the first sheet even when a sheet named `A1` exists.
- `FORMULA` and `UNFORMATTED_VALUE` are identical for any cell that is not a formula.
- Trailing-empty trimming is per row, so `values.get` answers ragged — while CSV export pads rows rectangular to the last used column.
- All three read enums match case-insensitively, and an *empty* value is not an absent one: it is refused rather than defaulted.
- `includeGridData` is a protobuf boolean, so `1`, `t`, `y` and `yes` all mean true — and `on` does not.
- `values:batchGetByDataFilter` answers in the order the ranges *start*, column before row, not the order the filters were sent.
- `prettyPrint` defaults to true, and unlike the other booleans an unparseable value is treated as true rather than refused.
- A cell's `effectiveFormat` varies across the types a corpus can state in exactly one field — a string sits `LEFT`, a number `RIGHT`, a boolean `CENTER` — so it is derived rather than stored, the way `formattedValue` is.
- A spreadsheet's `defaultFormat` and `spreadsheetTheme` are *not* derivations. Across six real workbooks they came in three variants, differing where someone had applied a theme or the document had been imported from .xlsx. What is served is the variant a freshly created spreadsheet has.

## Fidelity

`backlot diff --source google_drive` goes from 224 acknowledged divergences to 213. Twenty-one close because the behaviour is now served — nineteen parameters and two operations — and ten open, because the two endpoints that previously did not exist bring their own standard-parameter surface.

**Every read method is served.** The Sheets v4 discovery document declares 17 methods: seven reads and ten writes. Five of the seven are served — `spreadsheets.get`, `values.get`, `values:batchGet`, `values:batchGetByDataFilter` and `spreadsheets:getByDataFilter` — and the two that are not are `developerMetadata.get` and `developerMetadata.search`, which #159 names in its non-goals. The ten writes are work list E.

What stays open is 37 acknowledged gaps: those twelve operations, and five standard parameters on each of the five read endpoints. Each is left for a reason rather than for want of effort. `key`, `access_token` and `oauth_token` are credential transports — honouring one means a second path through `backlot.auth`, which every source shares, not a parameter this module can read. `uploadType` the real API *refuses* on a read, so the baseline's "the vendor accepts it" does not describe the vendor. `$.xgafv` is implementable and measured — `1` adds a legacy `errors[]` array to the envelope, `2` is what is already served, anything else is a 400 naming the system parameter — but it is declared on every Google endpoint rather than these five, so closing it here would leave the rest of Drive, Docs and Slides inconsistent with Sheets. It belongs to a change over that whole family.

No gap was closed by declaring a parameter without implementing it. That was the shape of one of the bugs this PR fixes: `dateTimeRenderOption` was declared on the values routes and never validated, so the discovery diff counted it as served while a client's typo passed silently.

## Notes for review

- A prose-backed spreadsheet and a gridded one agree with Drive export *differently*, and the tests assert each as itself. A prose document's cells are its lines, so its export round-trips byte for byte; running a CSV parser over it would find columns the Sheets API never claimed.
- `_workbook()` makes the prose representation one more grid, so both paths are the same code — which is why the existing Sheets tests pass unmodified.
- A sheet is addressed by the sheet it is, never by re-parsing its own title: a title reading as a cell or column reference (`A1`, `AB`) would otherwise resolve against the first sheet.
- The ERB importer deliberately still produces prose spreadsheets. Measured over the bench's 1,875: 25 state a header whose rows parse to exactly that width, and none of those 25 is the table alone — each carries 1–13 further content fields. `sheets` forbids `content`, so converting even the cleanest would drop the document's own prose from search and from export.
- A `fields` mask is validated against the fields Backlot can emit rather than the whole Sheets schema, so one naming a real-but-unmodelled field is refused here and answered there. Stated in the code; the alternative is a typo becoming a silently empty response.
- One behaviour on the prose path does move, and it is not a no-op: a spreadsheet with more than 1000 lines now reports `rowCount` equal to its line count rather than 1000, and a range inside it answers rather than 400ing. That follows the same measurement the gridded path uses — a grid is never smaller than the data it holds — but it is a change, so `test_a_prose_spreadsheet_is_still_one_synthesized_sheet` pins the shape that did not move.
- R1C1 ranges are still refused, and `values.get` documents them beside A1 — #174. Present at the base too; the completeness claim above is about methods, not notations.
- No migration path: a corpus is re-imported, not upgraded in place. A DB built before this change has no `gdrive_sheets`, is opened read-only, and would otherwise answer every Sheets read with a bare `OperationalError`; the server now names the missing table at startup and says to re-import.

**Every field these reads return is now served — a key-by-key diff against the live API over `spreadsheets.get` (with and without grid data), `values.get` and `values:batchGet` finds nothing missing in either direction, and nothing invented.**

Getting there meant revisiting #159's formatting non-goal, whose stated reason is "nothing in the corpus can carry them". Measured, that splits in two. A cell's `effectiveFormat` *is* carried, in the only way that matters: the sole field that varies across the types a corpus can state is decided by the cell's type, so the object is derived exactly as `formattedValue` is. A spreadsheet's `defaultFormat` and `spreadsheetTheme` are genuinely settings a person chooses — six workbooks gave three variants — so they are served as the values a freshly created spreadsheet has, on the same footing as `locale`, `autoRecalc`, `timeZone`, `sheetId` and the 1000x26 grid this module already synthesizes. `userEnteredFormat` is the one field left unserved, and measured it is the right call: it carries only what was explicitly set on a cell, so of 17 typed cells the only ones that had one were the percent, date, datetime, time and scientific cells, each holding a lone `numberFormat` Sheets inferred from what was typed. A corpus states no formatting and none of those value types, so no cell it can describe has anything to put there.

Still open on #159: the write surface (work list E). Developer metadata, merges, named ranges, charts and `spreadsheets.create`/`batchUpdate` remain non-goals.


nuri-yoo (commented):
Reviewed at 12505d4. Read the whole diff, ran the suite in a worktree at this head (exit 0; the seven touched test files: 1047 passed), `ruff check` / `ruff format --check` clean, `scripts/gen_docs.py --check` clean, and reproduced every inline finding against this head with the TestClient. Inline: two High (the doc_id collision, and an existing DB answering 500 on every Sheets read), the rest Medium or prose.

Not anchorable in the diff:

- PR body arithmetic. Sheets entries in the baseline go 48 to 37 with 10 new ones, so 21 close (19 parameters + 2 operations), not nineteen; 224 - 19 + 10 is 215, not 213. The 37 that stay are 12 operations plus *five* standard parameters on each of the five read endpoints (`key`, `access_token`, `oauth_token`, `uploadType`, `$.xgafv`), not one. "the four Sheets read endpoints" sits a few paragraphs from "Five of the seven are served".
- `examples/bring-your-own-corpus/README.md:47` still shows the `content` CSV record, and line 35 says every record states `content`; `sample_corpus.jsonl` moved to `sheets`.
- A behaviour change on the prose path that the body does not state: at 364e071 a 1500-line prose spreadsheet is `rowCount 1000` and `A1500` is refused; here it is `rowCount 1500` and `A1500` answers `line1499`. Consistent with A.6 in #159, but the body says a record without `sheets` imports exactly as before, and `test_a_prose_spreadsheet_is_still_one_synthesized_sheet` says its served shape must not move.

Follow-up issue rather than this PR: the discovery document describes `range` as "The A1 notation or R1C1 notation", and the installed llama-index `GoogleSheetsReader` sends `R1C1:R{rows}C{cols}` and `{title}!R1C1:R{rows}C{cols}`. Both are 400 `Unable to parse range` here and at the base too. Worth filing given the "read surface is complete" claim.

Low, no action needed from me:

- `GridData` keys come out `rowMetadata, columnMetadata, rowData`; the proto numbering (row_data 3, row_metadata 4, column_metadata 5) puts `rowData` first. Your measured body will settle it; the module orders other objects by measurement.
- `sheets_grid.formatted` on a non-integral float is `str()`: 0.1+0.2 gives "0.30000000000000004", 1e-7 gives "1e-07". Measured cases are 3.0 and 3.5 only. Stating the rest as unmeasured is enough.
- `grid.items.minItems: 1` refuses `[]` as a row ("[] should be non-empty") while the schema says short rows are padded. `normalise_grid` handles `[]` fine, and the commit that added the constraint (392ffb6) gives no reason.
- `land()` in `_gmask_parse` has a `sub` parameter no call site passes.
- `callback` is reflected into the `text/javascript` body unvalidated. Bearer auth keeps it out of a `<script src>`, but restricting it to identifier characters costs nothing.
- `_workbook` loads every sheet's grid for a single-range read: 5.4 ms vs 0.9 ms per `values.get A1` on a 5 x 1000 x 26 workbook. Fine at this scale; `_a1_sheet` only needs titles and ids, so the grids could load lazily if it ever matters.

backlot/importer/byo.py
```diff
@@ -1711,6 +1795,19 @@ def add(self, rec: dict, where: str = "record") -> None:
                     s["speaker_id"] = ordinals[s["speaker_name"] or ""]
             rec = {**rec, "content": synth.fireflies_transcript_text(sentences)}
 
+        # google_drive: a stated grid IS the document, so `content` is DERIVED from the first sheet
+        # rather than supplied. `files.export?mimeType=text/csv` serves `content` verbatim, so this
+        # is what stops Drive and the Sheets API describing one document two ways -- and it leaves
+        # FTS text to index for a spreadsheet that states no prose. Like fireflies above, it runs
+        # before `_doc_id`, which hashes the content, so the id covers the grid.
```
nuri-yoo: `_doc_id` hashes `source_type + title + content` (line 145), and for a gridded record `content` is the first sheet's CSV, so the id covers the first sheet only, not the grid. Two records without `doc_id` that share a title and a first sheet but differ on a later sheet collide.

Reproduced on this head: two `Twin` records, first sheet `[["a"]]`, second sheets `[["first"]]` and `[["second"]]`. `files.list` serves one `Twin`; `values.get Sheet2` on it answers `second`; the import prints nothing. The comment above `seen` justifies later-wins for "a direct import of the same documents"; these are different documents, and a Summary sheet shared by several workbooks is an ordinary corpus.

Hashing `_j(rec["sheets"])` next to `content` when `sheets` is present would make the id cover what this comment says it covers.

backlot/importer/byo.py
```diff
@@ -3433,6 +3544,8 @@ def run(
                 problems.append((lineno, f"invalid JSON: {e}"))
                 continue
             errs = record_errors(rec)
+            if isinstance(rec, dict) and rec.get("source_type") == "google_drive":
+                errs += _sheets_pairing_errors(rec)
```
nuri-yoo, line 3573: When `sheets` is not a list, `record_errors` has already reported the type error and this line iterates the value anyway. `{"sheets": 5}` under `--dry-run`:

```
TypeError: 'int' object is not iterable
```

`"sheets": true` does the same. The load path is safe (the schema check at 1721 precedes 1730); only the dry run, which exists to report instead of failing, ends in a traceback. `if not isinstance(sheets, list): return []` at the top of `_sheets_pairing_errors` is enough, since the schema already names the problem.

backlot/routers/google.py
```diff
@@ -2032,9 +2641,150 @@ async def sheets_values_batch_get(spreadsheet_id: str, request: Request):
...
+    Half-open indices, and an omitted bound means that edge of the grid."""
+    if isinstance(f.get("a1Range"), str):
+        return f["a1Range"]
+    grid = f.get("gridRange")
+    if not isinstance(grid, dict):
+        raise gerr.invalid_argument("dataFilter.filter must be specified.")
+    sheet = next((s for s in sheets if s.sheet_id == grid.get("sheetId", 0)), None)
+    if sheet is None:
+        raise gerr.invalid_argument(f"No sheet with id: {grid.get('sheetId', 0)}")
+    r0 = int(grid.get("startRowIndex") or 0)
```
nuri-yoo: The `gridRange` members are used without a typed check, unlike the enums and booleans on the same reads. Measured on this head, `values:batchGetByDataFilter`:

| gridRange | answer |
|---|---|
| `{"startRowIndex": "abc"}` | `ValueError` escapes, 500 |
| `{"startRowIndex": 0, "endRowIndex": 0}` | 400 `Range (Data!A0:Z1) exceeds grid limits` |
| `{"startRowIndex": -1}` | 400 `Range (Data!A0:Z1000) ...` |
| `{"startRowIndex": 2, "endRowIndex": 1}` | 200, range `Data!A1:Z3`, three rows served |
| `{"startRowIndex": 1.7}` | 200, silently 1 |
| `{"sheetId": null}` | 400 `No sheet with id: None` |
| `{"sheetId": "0"}` | 400 `No sheet with id: 0` |

`A0` is not a cell: `_a1_name` builds `Data!A1:Z0` for `r1 <= r0` and `_a1_range` re-parses it into a start row of -1. The reversed filter answers a range it did not ask for. `None` is a Python repr. The discovery document declares these `int32`, and the proto3 JSON mapping accepts a number or a string for int32, so `"0"` should resolve. I did not measure real's wording for any of these, so no message proposals; but the 500 needs to be a 400, and a check of the TYPE_INT32 kind seems due before the values reach `_a1_name`.

backlot/routers/google.py
```diff
@@ -2032,9 +2641,150 @@ async def sheets_values_batch_get(spreadsheet_id: str, request: Request):
...
+    "/sheets/v4/spreadsheets/{spreadsheet_id}:getByDataFilter",
+    openapi_extra={"parameters": _P_SHEETS_STD},
+)
+async def sheets_get_by_data_filter(spreadsheet_id: str, request: Request):
+    """``spreadsheets.get`` addressed by DataFilter. Same response, and the filters scope the
+    ``sheets`` array exactly as ``ranges`` does — measured, including that NO filter means every
+    sheet rather than the refusal its values-level sibling gives."""
+    row, sheets = _workbook(request, spreadsheet_id)
+    body, specs = await _sheets_filters(request, sheets, required=False, indexed=False)
+    grid = bool(body.get("includeGridData"))
```
nuri-yoo: `bool()` on the body value: `"false"`, `"no"` and `"0"` all come back with `data`, and only a JSON `0` or `false` does not. The GET path parses the same parameter through `_sheets_bool`. 41e4e96c gives the enums one rule for query string and body; the boolean did not get it. The discovery document declares `GetSpreadsheetByDataFilterRequest.includeGridData` as `boolean`, so a string is arguably a refusal, but under no reading is `"false"` true. Either run a string through `_sheets_bool`'s spellings or accept only a JSON boolean.

backlot/routers/google.py
```diff
@@ -1735,26 +1974,66 @@ async def sheets_get(spreadsheet_id: str, request: Request):
...
-    sheet = {
-        "properties": {
-            "sheetId": 0,
-            "title": SHEETS_SHEET_TITLE,
-            "index": 0,
-            "sheetType": "GRID",
-            # the grid, not the data extent — see SHEETS_GRID_ROWS
-            "gridProperties": {"rowCount": SHEETS_GRID_ROWS, "columnCount": SHEETS_GRID_COLS},
+    row, sheets = _workbook(request, spreadsheet_id)
+    grid = _sheets_bool(request, "includeGridData", "include_grid_data")
```
nuri-yoo, line 1994: Measured on this head: `GET ...?fields=sheets.data.rowData.values.formattedValue` with no `includeGridData` answers `{"sheets":[{}]}`, and `:getByDataFilter` with the same mask does too. The discovery document says of `includeGridData`, on both methods:

> True if grid data should be returned. This parameter is ignored if a field mask was set in the request.

So with a mask present the mask decides, and one that reaches `sheets.data` (or `sheets`, or `*`) should build the grid and project it. Every test that sends `fields` also sends `includeGridData=true` (test_google.py:3459), which is why this is uncovered.

backlot/schemas/google_drive.schema.json
```diff
@@ -120,16 +120,63 @@
     "trashed": {
       "type": "boolean",
       "description": "Whether the file is trashed (files.list excludes trashed by default)."
+    },
+    "sheets": {
+      "type": "array",
+      "minItems": 1,
+      "description": "A real grid, for a `spreadsheet`. Authoritative when present, and then `content` must be ABSENT: content is derived from the first sheet at import, so a record stating both would let Drive search find words no cell holds. Omit `sheets` and `content` keeps its meaning unchanged -- one sheet whose every cell is a whole line of it.",
```
nuri-yoo: This states one direction (words no cell holds). The other is unstated: `content` is the first sheet, and the FTS index is title+content (store.py:2083), so a word held only on a later sheet is not found by `fullText contains`. Measured: `zebrawombat` on sheet 2 answers `files: []`; `EMEA` on sheet 1 is a hit.

Drive's search reference defines `fullText` as matching "text in the file's content or metadata of the file", without scoping it to a sheet; I have not measured a multi-sheet workbook against it. Either way a client searching a gridded corpus needs to know. State it here and in docs/corpus.md, or feed FTS every sheet.

backlot/store.py
```diff
@@ -367,6 +367,31 @@ def id_seed_scope(source_type: str) -> str | None:
...
+-- `sheet_id` is what the Sheets API emits, assigned at import (see backlot.importer.byo) rather
+-- than derived from the index at serve time. Measured on a real workbook: the sheet created with
+-- the spreadsheet is 0 and every sheet added afterwards carries a large pseudo-random integer
+-- (562149769, 1609058389, ...), so `sheet_id == sheet_index` is never a safe assumption. UNIQUE
+-- per file because it is what a client addresses a sheet by -- two sharing one would leave the
+-- loser unreachable -- and a collision here is a loud import failure instead of a silent shadow.
+--
+-- A cell in `grid` is a JSON scalar, so the type a corpus stated survives the round trip and
+-- `valueRenderOption` has something to distinguish.
+CREATE TABLE IF NOT EXISTS gdrive_sheets (
```
nuri-yoo, line 388: A DB imported before this change has no `gdrive_sheets`, and the server opens it read-only (`connect_ro`, main.py:53), so this DDL never runs for it. Built a corpus with 364e071 and served it with this head:

```
sheets.get   sqlite3.OperationalError: no such table: gdrive_sheets
values.get   sqlite3.OperationalError: no such table: gdrive_sheets
export tsv   sqlite3.OperationalError: no such table: gdrive_sheets
export csv   200
```

Prose spreadsheets included: `_workbook` queries the table before it knows the document states no grid. The PR body says a corpus is re-imported rather than upgraded, but nothing tells the operator: `meta` carries no schema version and the lifespan reads only `source_documents`. A startup check that names the missing table and says to re-import would turn a 500 per Sheets read into one message.

backlot/sheets_grid.py
```diff
@@ -0,0 +1,98 @@
...
+the importer is the other caller and the two layers do not import each other -- an importer that
+reached into a router would pull the whole web layer into ``backlot import``. A vendor-specific
+pure helper that more than one layer needs lives at this level: ``sigv4`` is the S3 router's, and
+``synth`` is every source's.
+"""
+
+Cell = str | int | float | bool | None
+
+
+def _scalar(v):
```
nuri-yoo, line 17: `json.loads` accepts `1e400`, `Infinity` and `NaN`, and jsonschema's `number` does not refuse them, so a corpus cell `1e400` passes `--dry-run` (`OK: 1 records valid`) and this function keeps it: `values.get` answers `"values":[[Infinity]]` and CSV export `inf`. RFC 8259 has no such literals; `JSON.parse` fails on the body. `math.isfinite` here, or in `_sheets_pairing_errors` so the dry run names it, would refuse at import.

Related, worth a sentence in the schema description: `numberValue` is a `double` per the discovery document, so `2**53+1` served as the exact integer `9007199254740993` is a value the real API cannot hold.

backlot/routers/google.py
```diff
@@ -1723,7 +1737,232 @@ def _sheets_grid(content: str | None) -> list[list[str]]:
...
+        raise gerr.bad_field_mask(mask)
+    land(cur, token)
+    return tree
+
+
+def _gmask_check(tree: dict, allowed: dict, path: str = "") -> None:
+    """Refuse a name the response has no field for, naming the full path as the real API does.
+
+    Validated against the fields Backlot CAN emit rather than against the whole Sheets schema:
+    a mask naming a real field this module does not model -- a cell's `effectiveFormat`, say --
```
nuri-yoo: `effectiveFormat` is in `_F_CELL` now, and `fields=sheets.data.rowData.values.effectiveFormat.horizontalAlignment` answers 200 on this head. The example needs a field this module really does not model, `userEnteredFormat` or `note` say.

backlot/routers/google.py
```diff
@@ -1907,99 +2374,241 @@ def _rstrip_empty(cells: list[str]) -> list[str]:
...
+    ``rowMetadata``/``columnMetadata`` cover the RANGE, one entry per row and column of it —
+    measured, 2 and 2 for ``Data!A1:B2`` against the same sheet whose unscoped block carries 1000
+    and 26. Every entry is identical (``pixelSize`` 21 for a row, 100 for a column), those being
+    the default track sizes; a corpus states no track size, so there is nothing to vary.
 
-    Two divergences, stated rather than hidden: real Sheets pads ``rowData`` to the WHOLE 1000-row
-    grid and this stops at the last row holding data; and real cells carry format objects plus
-    ``rowMetadata``/``columnMetadata``, none of which Backlot models."""
-    r0, c0, _r1, c1, block = _sheets_block(rows, spec)
+    Two divergences, stated rather than hidden: real Sheets pads ``rowData`` out to the WHOLE
```
nuri-yoo: `effectiveFormat` is emitted at line 2457, so the second divergence no longer exists and "Two divergences" is one.

backlot/importer/byo.py
khj809: Confirmed and fixed at e67b04c. Reproduced your case independently: two `Twin` records collapsed to one row and `values.get S2` answered `second`, silently.

`_doc_id` now folds `_j(rec['sheets'])` into the seed when the record states one. The comment above the derivation claimed the id covered the grid, which it did not — that wording was the bug in miniature, so it now says why the first sheet alone is not enough.

backlot/importer/byo.py
khj809: Fixed at e67b04c with the guard you proposed, moved to the top of `_sheets_pairing_errors` so every rule below it is safe rather than just the loop. Parametrised over `5`, `true`, `"abc"` and `{}`.

backlot/routers/google.py
khj809: Fixed at e67b04c. Added `_sheets_int32`, applied to every member including `sheetId`, before anything reaches `_a1_name`:

- a non-numeric string, a fractional float and a negative index are 400s, not a 500 and not a silent truncation
- a decimal string resolves, per the proto3 JSON mapping you cite — `"0"` now finds sheet 0
- an end at or before its start is refused outright, which is what stops `_a1_name` from building the `Data!A1:Z0` you found and the parser reading it back as a start row of -1

Messages are Backlot's own and say so in the docstring — I did not measure real's wording for these either.

backlot/routers/google.py
khj809: Fixed at e67b04c. `_sheets_bool` now delegates to `_sheets_bool_value`, which the body path calls too — the same split 41e4e96 gave the enums. A JSON boolean is taken as itself; anything else goes through the measured spellings, so `"false"`, `"no"` and `"0"` are false and a value that is neither is a 400 rather than truthy.

backlot/routers/google.py
khj809: Fixed at e67b04c, and measuring it narrowed the rule. The vendor sentence reads as "a mask present ⇒ grid", but live:

| `fields` | grid built |
|---|---|
| `sheets.data.rowData.values.formattedValue` | yes |
| `sheets` | yes |
| `*` | **no** |
| `sheets.properties.title` | no |

So `*` selects everything and still does not build it. `_gmask_wants_grid` implements "the mask reaches `sheets.data`" rather than "a mask is set", on both methods, and the parametrised test carries the `*` case since that is the one a reading of the sentence gets wrong.

backlot/schemas/google_drive.schema.json
khj809: Confirmed — and it is the sharper half of the pair, since the first direction is a corpus author's mistake while this one is invisible. Stated at e67b04c in the schema description and in docs/corpus.md, both naming the first sheet and telling an author to put what should be findable there.

Not feeding FTS every sheet: `content` is what `files.export` serves verbatim, so making it every sheet would either change what export returns or split the two apart again, which is the disagreement the derivation exists to prevent. Indexing beyond `content` would mean a second FTS source column, which is a store-wide change rather than a Sheets one.

backlot/store.py
khj809: Confirmed — reproduced by dropping the table from a built DB and serving it: 500 on `sheets.get`, `values.get` and TSV export, prose spreadsheets included, because `_workbook` asks the table before it can know the document states no grid.

Fixed at e67b04c the way you suggest, and generally rather than for this one table: `store.missing_tables` diffs `SCHEMA`'s `CREATE TABLE` names against `sqlite_master`, and the lifespan raises naming what is missing and saying to re-import — beside the existing "DB not found" refusal, which is the same shape. No migration: a corpus is re-imported, not upgraded, so naming the gap is the remedy.

backlot/sheets_grid.py
khj809: Confirmed and fixed at e67b04c. `_sheets_pairing_errors` now refuses a non-finite cell by position, so `--dry-run` names it rather than the value reaching a served body — the point you make about `JSON.parse` is what settles it: a body real Sheets could never send and a browser cannot read.

Put in the pairing check rather than `sheets_grid` so the dry run reports it, which is what you asked for. The `2**53+1` note is a different claim and I have not measured it; leaving it rather than writing an unmeasured sentence into the schema.

backlot/routers/google.py
khj809: Fixed at e67b04c — the example is `userEnteredFormat` now, which measurement says is the one a corpus genuinely cannot produce: of 17 typed cells only the percent, date, datetime, time and scientific ones had one, each a lone `numberFormat` Sheets inferred from what was typed.

backlot/routers/google.py
khj809: Fixed at e67b04c — one divergence now, and it is the `rowData` padding.

khj809:
Every inline finding is fixed at e67b04c and cf6495f, each reproduced first — replies are on the threads. Both Highs reproduced exactly as you described them.

On the three not anchorable in the diff:

**The arithmetic was wrong and your count is right.** Computed from the two baselines: 21 sheets entries close (19 parameters + 2 operations) and 10 open, and 224 − 21 + 10 = 213. The body said "nineteen", which is where the 215 came from. Also corrected: five standard parameters on each of the five read endpoints, not one; and the key-by-key diff sentence now names the four *calls* it covered (`spreadsheets.get` with and without grid data, `values.get`, `values:batchGet`) rather than saying "endpoints", which read as though it covered five.

**The prose-path change is real and the body was wrong to say otherwise.** Reproduced: a 1500-line prose spreadsheet is `rowCount 1500` here and `A1500` answers `line1499`, where the base gives 1000 and a 400. It follows the same A.6 measurement the gridded path uses — a grid is never smaller than the data it holds — so I am keeping it, but it is a change and the body now says so beside the test that pins what did *not* move.

**`examples/bring-your-own-corpus/README.md`** — the drive line now states `sheets`, matching the sample corpus, and the "every record states `content`" sentence carries the exception with a link to the spreadsheet section of docs/corpus.md.

**R1C1 is #174**, which already had it — mine was a duplicate and is deleted; the measurements are now a comment there. Measured before filing, and it is more clearly a gap than the discovery sentence alone suggests: real takes `R1C2`, `R1C1:R2C2`, `Data!R1C1:R2C2` and the quoted form, and *echoes the A1 equivalent* — so a client diffing two backends sees the same `range` string it would from us. Backlot 400s all of them, at the base too. The completeness claim in the body is now scoped to methods, with the notation gap called out and linked.

The dead `sub` parameter in `land()` is gone. On two of the low items I am not acting, with reasons rather than silence:

- **`grid.items.minItems: 1`.** Deliberate, though 392ffb6 does not say so — `grid: []` is an empty sheet and `grid: [[]]` is a row with no cells, which is a typo rather than a shape a corpus means. `normalise_grid` tolerating it is not an argument for accepting it; the schema refuses several things the loader would survive. I have added the reason to the description so the next reader does not have to ask.
- **`callback` reflection.** Restricting it to identifier characters would be a refusal the real API does not have, which is the divergence this PR spent its time removing. I did not measure real's handling of a hostile callback name, so I would be inventing a rule. Worth its own measurement if the reflected value ever leaves a bearer-authenticated response.

The rest of the low list I agree with and have left alone: the `GridData` key order (my captured body has `rowMetadata, columnMetadata, rowData`, so it stands on measurement), the float formatting beyond the two measured cases, and `_workbook`'s eager grid load at this scale.

backlot/importer/byo.py
```diff
@@ -1711,6 +1818,19 @@ def add(self, rec: dict, where: str = "record") -> None:
                     s["speaker_id"] = ordinals[s["speaker_name"] or ""]
             rec = {**rec, "content": synth.fireflies_transcript_text(sentences)}
 
+        # google_drive: a stated grid IS the document, so `content` is DERIVED from the first sheet
+        # rather than supplied. `files.export?mimeType=text/csv` serves `content` verbatim, so this
+        # is what stops Drive and the Sheets API describing one document two ways -- and it leaves
+        # FTS text to index for a spreadsheet that states no prose. Like fireflies above, it runs
+        # before `_doc_id`, which hashes the content, so the id covers the grid.
```
nuri-yoo: Nit, non-blocking: this sentence is true now, but for a different reason than it gives. The id covers the grid because `_doc_id` folds `_j(rec["sheets"])` into its seed, not because the derived `content` is hashed; `content` still carries the first sheet only. Something like "before `_doc_id`, which hashes the content and, for a gridded record, the sheets" would attribute it to the right place.

nuri-yoo (approved):
Approving at cf6495f. Re-ran every inline finding against this head in a fresh worktree: the two workbooks sharing a title and a first sheet are two documents; `--dry-run` reports `5`, `true`, `"abc"` and `{}` instead of raising; the gridRange members are 400s for `"abc"`, `-1`, `1.7` and an end at or before its start, and `"0"` resolves; `"false"`, `"no"` and `"0"` in the body are false and `"maybe"` is a 400; a mask reaching `sheets.data` or `sheets` builds the grid and `*` does not; `1e400` and `NaN` are refused on both the dry run and the load; and a DB built at 364e071 fails at startup naming `gdrive_sheets` and saying to re-import. Suite exit 0, ruff and `gen_docs.py --check` clean, CI green. Body arithmetic, the prose-path statement and the README example all check out, and the two low items you declined have their reasons where the next reader will find them. On GridData key order your captured body settles it and my proto-number guess was wrong. One non-blocking nit inline.

backlot/importer/byo.py
khj809: Fixed at e58e699. You are right that it was true for the wrong reason — and it is the same misattribution that produced the bug in the first place, so the wording now says the derived content is what gets hashed there, that the rest of the workbook reaches the id through `_doc_id` folding `sheets` in, and that content alone would cover the first sheet and nothing else.

PR #162, opened 2026-09-09
Add missing spec comparisons (gdocs/gsheets/gslides/jira/hubspot)

khj809:
Seventeen vendor routes were reached by no document, and the test whose name promised to prevent that counted the wrong thing. This closes the gap and fixes the test.

`test_the_comparisons_are_exactly_the_sources_backlot_serves` asserts `set(COMPARISONS) == set(store.SOURCE_TABLE) == set(cli.FIDELITY_SOURCES)`. That reads as a coverage guarantee and is not one: it counts **source types**, the corpus dimension. `google_drive` is one source type served through five vendor APIs, so Docs, Sheets and Slides sat compared against nothing while the suite stayed green — a Sheets response shape could be rewritten and `backlot diff --source google_drive` would still answer `0 new`. Its neighbour asserts only that a mount matches *something*, which `/drive/v3` satisfies on its own.

## A comparison holds one document per API

The blocker was structural: a spec-backed comparison was one `spec_url` plus one `mount` plus one `strip`, so a source served through several vendor APIs could not be expressed. Widening `mount` does not help — the extra paths get diffed against the wrong document and report as invented.

`OpenAPIComparison` and `GoogleDiscoveryComparison` now hold `specs: tuple[Spec, ...]`, and `divergences()` runs each and merges. `strip` belongs on the `Spec` rather than the comparison because how much of the mount a vendor's own document repeats differs per document: `/sheets/v4/...` pairs against the Sheets discovery document only with `strip="/sheets"`, while Gmail's document already spells `gmail/v1/...` and strips nothing. `Baseline.endpoint` becomes `endpoints`, since one string can no longer name what a source was measured against — and a `Spec` whose document is chosen out of an index names the entry, not the index, so HubSpot's two do not both report the same URL.

Merging rather than concatenating, because two documents of one source can declare the same path. Atlassian's do: 11 paths sit outside `/rest/api/{2,3}` — `/rest/atlassian-connect/1/*`, `/rest/forge/1/*` and one internal worklog route — so both Jira documents carry them and each reported as `missing_operation` twice. That is a different case from the mirrored `/rest/api/2/...` and `/rest/api/3/...` entries, which are distinct paths a client can call and both belong in the baseline. A baseline is keyed on `kind:path` and cannot hold the second copy, so leaving it in made the file claim more than it could load. The merged list is also sorted, because `diff_operations` orders breaking-first-then-path per document and that does not survive joining two — without it the next real baseline change would arrive buried in a reordering of the whole file.

The registry is still keyed by `source_type` — the vocabulary a BYO record and `backlot import` use — rather than gaining an entry per vendor API, which would make `--source sheets` name something no corpus can contain.

## What is now compared

| Source | Was | Now |
|---|---|---|
| `google_drive` | Drive v3 | Drive v3 + Docs v1 + Sheets v4 + Slides v1 |
| `jira` | v3 document | v3 document + **v2 document** |
| `hubspot` | CRM v3 | CRM v3 + Associations v4 |

Thirteen documents where there were eight, covering 11 paths and 12 operations that no document reached before: Jira v2's five paths and six operations, HubSpot v4's one, Docs' one, Sheets' three, Slides' one. Zero `extra_operation` across all five additions — Backlot serves nothing these documents do not declare.

**Jira v2 is not legacy.** Measured: `atlassian-python-api` hardcodes `api_version = "2"` in its Jira constructor and the `jira` PyPI client defaults `rest_api_version` to `"2"`, probing `/rest/api/2/serverInfo` on connect — so at default settings both SDKs call v2 exclusively and nothing calls v3. The code carried a note saying "Atlassian publishes a v3 document"; it publishes one per version, named `swagger[-<apiVersion>].<oasVersion>.json`, so the suffix-less `swagger.v3.json` is the v2 API with 403 `/rest/api/2` paths. Each version is now compared against its own document. `docs/fidelity.md` separately claimed the v2 paths "are compared through their v3 twins" — that mechanism never existed, and the line is corrected.

## Every declared path is accounted for

Replacing the mount half of the old test: every path in `app.openapi()` must be under some document's mount, probe-compared, or listed in `UNCOMPARED` with the reason no document covers it — Backlot's own `/health`, `/oauth2/token` and `/_meta`, and Google's `/batch`, which is one endpoint carrying requests for every Google API and so appears in no single discovery document. Two further tests carry three checks between them: a mount that selects nothing fails, and — in one test — an `UNCOMPARED` prefix whose route is gone fails, as does one that sits under a comparison's mount, so the list cannot claim a surface that is in fact compared.

Declared, not served. Six live routes are `include_in_schema=False` and invisible to that check: FastAPI's own four, and the GraphQL POSTs at `/fireflies/graphql` and `/linear/graphql`, which their own comparison covers by introspection and so have no mount to say so. Walking `app.routes` instead was ruled out for `gen_docs.py` and the same reasoning holds here, so `docs/fidelity.md` says *declares* rather than *serves*.

The vocabulary assertion stays, renamed to `test_the_registry_names_exactly_the_source_types_backlot_serves` — a real invariant, just not the coverage one.

## Baselines

The five new documents surface findings that were real and previously invisible: `google_drive` 146 → 224, `hubspot` 14 → 23, `jira` 640 → 1243. All gaps, acknowledged in one pass, and each file now round-trips through `--update-baseline` byte for byte.

Three are not gaps and were classified by hand. `POST /rest/api/2/search/jql` accepts `jql`, `maxResults` and `nextPageToken` as query parameters where the v2 document declares them in a `SearchAndReconcileRequestBean` body. The v3 comparison already carries the identical three with a note, and the two routes are one handler under two decorators — one divergence surfaced at a second version, not a new one. Each new entry says so and records the v2 measurement rather than copying the v3 note.

The jira growth includes 584 `missing_operation` entries mirroring v3's own. Accepted rather than de-duplicated: `/rest/api/2/attachment/{id}` and `/rest/api/3/attachment/{id}` are different paths a client can call, and a reader asking what of Jira v2 Backlot serves should get the answer from the v2 entries rather than by inferring it from v3's.

## Verification

`uv run pytest` with all extras: 2645 passed, 3 skipped. `ruff check . && ruff format --check .` clean.

Ten of the eleven sources report `0 new`, both GraphQL ones run with credentials. `linear` reports one — `missing_field Query.agentSessionSshAddress`, at `gap`. It reproduces identically on `main`, so it is Linear schema drift that predates this branch and belongs to the scheduled job; it is not touched here.

Each of the three new guards was checked by reverting the thing it guards: a deliberately uncompared route fails only the served-path test and names the path, a dead mount fails only the mount test, and rolling `Spec.endpoint` back to a bare `spec_url` fails the distinctness assertion with HubSpot's index reported twice.

## Not in scope

Confluence is still not fully covered — its reads have moved to a v2 document whose paths are shaped differently from the v1 ones Backlot serves, so the eight reads Atlassian removed from the v1 document remain acknowledged rather than compared.

`SOURCE_PREFIXES` in `backlot/openapi.py` has the same blind spot this fixes in the fidelity registry — the `gdrive` MCP slice is Drive only, so the Docs, Sheets and Slides routes have no tool. Filed as #170.

### DIFF
--- a/backlot/fidelity/comparisons.py
+++ b/backlot/fidelity/comparisons.py
@@ -71,34 +71,29 @@ class Spec:
 
 @dataclass(frozen=True)
 class OpenAPIComparison:
-    """A source compared against the OpenAPI document its vendor publishes.
+    """A source compared against the OpenAPI document(s) its vendor publishes.
 
     Public, so this kind needs no credential, no quota and no account: the scheduled job carries no
     secret for it, and anyone can reproduce a finding locally.
 
-    Most vendors publish the document at a stable URL. One does not, and ``resolve_url`` is for
-    that: ``spec_url`` addresses an index, and the hook picks the entry out of it.
+    SEVERAL documents where a vendor publishes several: Atlassian publishes Jira's v2 and v3 REST
+    APIs separately and Backlot serves both, and HubSpot's associations are their own API at their
+    own version. Each is a :class:`Spec` with its own mount and strip, and the findings
+    concatenate.
     """
 
     name: str
-    spec_url: str
-    mount: tuple[str, ...]
-    strip: str = ""
-    # Set when ``spec_url`` addresses an INDEX rather than the document itself: given what was
-    # fetched, it returns the URL to fetch instead. A vendor that publishes one document per
-    # release needs this — see `hubspot_catalog` for why pinning the resolved URL is
-    # not the simplification it appears to be.
-    resolve_url: Callable[[Mapping[str, Any]], str] | None = None
+    specs: tuple[Spec, ...]
     credentials: tuple[Credential, ...] = ()
 
     @property
-    def endpoint(self) -> str:
-        return self.spec_url
+    def endpoints(self) -> tuple[str, ...]:
+        return tuple(s.spec_url for s in self.specs)
 
     def divergences(
         self, credentials: Mapping[str, str] | None = None, *, timeout: float = 120.0
     ) -> list[Finding]:
-        return openapi_diff.divergences(self, timeout=timeout)
+        return [f for s in self.specs for f in openapi_diff.divergences(s, timeout=timeout)]
 
 
 @dataclass(frozen=True)
@@ -109,22 +104,25 @@ class GoogleDiscoveryComparison:
     nest under resources, paths join to a ``servicePath``, and the standard query parameters are
     declared once for the whole document. A shared class would have carried a ``kind`` field that
     picked a parser, which is a type doing a type's job in a string.
+
+    SEVERAL documents where a source is served through several APIs: a Drive file is also read
+    through Docs, Sheets and Slides, each of which publishes its own discovery document.
     """
 
     name: str
-    spec_url: str
-    mount: tuple[str, ...]
-    strip: str = ""
+    specs: tuple[Spec, ...]
     credentials: tuple[Credential, ...] = ()
 
     @property
-    def endpoint(self) -> str:
-        return self.spec_url
+    def endpoints(self) -> tuple[str, ...]:
+        return tuple(s.spec_url for s in self.specs)
 
     def divergences(
         self, credentials: Mapping[str, str] | None = None, *, timeout: float = 120.0
     ) -> list[Finding]:
-        return google_discovery_diff.divergences(self, timeout=timeout)
+        return [
+            f for s in self.specs for f in google_discovery_diff.divergences(s, timeout=timeout)
+        ]
 
 
 @dataclass(frozen=True)
@@ -143,6 +141,12 @@ class GraphQLComparison:
     # Fireflies takes `Bearer <key>`, and Linear's personal API keys go in BARE — sending Linear a
     # `Bearer` prefix is answered 400, not 401.
     auth_scheme: str = "Bearer"
+
+    @property
+    def endpoints(self) -> tuple[str, ...]:
+        """One, always: introspection is a single live schema, not a set of documents. Plural so
+        every kind answers the same question and no caller needs to know which it is holding."""
+        return (self.endpoint,)
 
     def authorization(self, resolved: Mapping[str, str]) -> str:
         key = resolved["api_key"]
@@ -177,8 +181,8 @@ class ProbeComparison:
     run: Callable[[str, Mapping[str, Any], float], list[Finding]]
 
     @property
-    def endpoint(self) -> str:
-        return self.spec_url
+    def endpoints(self) -> tuple[str, ...]:
+        return (self.spec_url,)
 
     # None: a probe asks Backlot, not the vendor, and signs with the corpus's own keys. A probe
     # against a live vendor would declare what it needs here. Note that `run` is the only part a
@@ -195,46 +199,71 @@ class ProbeComparison:
 OPENAPI = {
     "slack": OpenAPIComparison(
         name="slack",
-        spec_url="https://raw.githubusercontent.com/slackapi/slack-api-specs/master/web-api/slack_web_openapi_v2.json",
-        mount=("/slack/api",),
-        strip="/slack/api",
+        specs=(
+            Spec(
+                spec_url="https://raw.githubusercontent.com/slackapi/slack-api-specs/master/web-api/slack_web_openapi_v2.json",
+                mount=("/slack/api",),
+                strip="/slack/api",
+            ),
+        ),
     ),
     "github": OpenAPIComparison(
         name="github",
-        spec_url="https://raw.githubusercontent.com/github/rest-api-description/main/descriptions/api.github.com/api.github.com.json",
-        mount=("/github",),
-        strip="/github",
+        specs=(
+            Spec(
+                spec_url="https://raw.githubusercontent.com/github/rest-api-description/main/descriptions/api.github.com/api.github.com.json",
+                mount=("/github",),
+                strip="/github",
+            ),
+        ),
     ),
     "jira": OpenAPIComparison(
         name="jira",
-        spec_url="https://developer.atlassian.com/cloud/jira/platform/swagger-v3.v3.json",
-        # v3 only. Backlot serves the `/rest/api/2` aliases too, because real Jira does, but
-        # Atlassian publishes a v3 document — comparing a v2 path against it would report Backlot
-        # inventing every one of them, which is a statement about the document, not about Jira.
-        mount=("/atlassian/rest/api/3",),
-        strip="/atlassian",
+        specs=(
+            Spec(
+                spec_url="https://developer.atlassian.com/cloud/jira/platform/swagger-v3.v3.json",
+                mount=("/atlassian/rest/api/3",),
+                strip="/atlassian",
+            ),
+            # v3 only. Backlot serves the `/rest/api/2` aliases too, because real Jira does, but
+            # Atlassian publishes a v3 document — comparing a v2 path against it would report
+            # Backlot inventing every one of them, which is a statement about the document, not
+            # about Jira.
+        ),
     ),
     "confluence": OpenAPIComparison(
         name="confluence",
-        spec_url="https://developer.atlassian.com/cloud/confluence/swagger.v3.json",
-        mount=("/atlassian/wiki",),
-        strip="/atlassian",
+        specs=(
+            Spec(
+                spec_url="https://developer.atlassian.com/cloud/confluence/swagger.v3.json",
+                mount=("/atlassian/wiki",),
+                strip="/atlassian",
+            ),
+        ),
     ),
     "notion": OpenAPIComparison(
         name="notion",
-        spec_url="https://developers.notion.com/openapi.json",
-        mount=("/notion/v1",),
-        strip="/notion",
+        specs=(
+            Spec(
+                spec_url="https://developers.notion.com/openapi.json",
+                mount=("/notion/v1",),
+                strip="/notion",
+            ),
+        ),
     ),
     "hubspot": OpenAPIComparison(
         name="hubspot",
-        spec_url="https://api.hubspot.com/public/api/spec/v1/specs",
-        # crm/v3 only: the v4 associations surface Backlot also serves is a SEPARATE API in
-        # HubSpot's catalog with its own document, so comparing it against this one would report
-        # it as invented.
-        mount=("/hubspot/crm/v3",),
-        strip="/hubspot",
-        resolve_url=hubspot_catalog.entry("Custom Objects", "3"),
+        specs=(
+            Spec(
+                spec_url="https://api.hubspot.com/public/api/spec/v1/specs",
+                mount=("/hubspot/crm/v3",),
+                strip="/hubspot",
+                resolve_url=hubspot_catalog.entry("Custom Objects", "3"),
+            ),
+            # crm/v3 only: the v4 associations surface Backlot also serves is a SEPARATE API in
+            # HubSpot's catalog with its own document, so comparing it against this one would
+            # report it as invented.
+        ),
     ),
     # S3 is not an entry here: its contract is not a path map at all, so it is compared by asking
     # a running server instead. See backlot.fidelity.s3_probe.
@@ -244,15 +273,23 @@ OPENAPI = {
 GOOGLE_DISCOVERY = {
     "gmail": GoogleDiscoveryComparison(
         name="gmail",
-        spec_url="https://gmail.googleapis.com/$discovery/rest?version=v1",
-        # Nothing stripped: Google's own document already spells `gmail/v1/...`, which is exactly
-        # where Backlot mounts it.
-        mount=("/gmail",),
+        specs=(
+            Spec(
+                spec_url="https://gmail.googleapis.com/$discovery/rest?version=v1",
+                # Nothing stripped: Google's own document already spells `gmail/v1/...`, which is
+                # exactly where Backlot mounts it.
+                mount=("/gmail",),
+            ),
+        ),
     ),
     "google_drive": GoogleDiscoveryComparison(
         name="google_drive",
-        spec_url="https://www.googleapis.com/discovery/v1/apis/drive/v3/rest",
-        mount=("/drive/v3",),
+        specs=(
+            Spec(
+                spec_url="https://www.googleapis.com/discovery/v1/apis/drive/v3/rest",
+                mount=("/drive/v3",),
+            ),
+        ),
     ),
 }
 
--- a/backlot/fidelity/comparisons.py
+++ b/backlot/fidelity/comparisons.py
@@ -104,7 +104,12 @@ class OpenAPIComparison:
     def divergences(
         self, credentials: Mapping[str, str] | None = None, *, timeout: float = 120.0
     ) -> list[Finding]:
-        return [f for s in self.specs for f in openapi_diff.divergences(s, timeout=timeout)]
+        # One memo across this run's specs: HubSpot reaches every document through a single index,
+        # so without it a source comparing two of its APIs reads that index once per spec.
+        seen: dict[str, dict] = {}
+        return [
+            f for s in self.specs for f in openapi_diff.divergences(s, timeout=timeout, seen=seen)
+        ]
 
 
 @dataclass(frozen=True)
--- a/backlot/fidelity/comparisons.py
+++ b/backlot/fidelity/comparisons.py
@@ -184,6 +184,11 @@ class ProbeComparison:
     def endpoints(self) -> tuple[str, ...]:
         return (self.spec_url,)
 
+    # Which served paths this comparison speaks for. Every kind answers that question; only a
+    # probe was never asked it, because it needs no mount to select paths out of a document — it
+    # asks a running server instead. The prober does not read this; what does is the check that
+    # every served path is compared by something.
+    mount: tuple[str, ...] = ()
     # None: a probe asks Backlot, not the vendor, and signs with the corpus's own keys. A probe
     # against a live vendor would declare what it needs here. Note that `run` is the only part a
     # probe supplies: fetching the model and starting the server are `s3_probe.divergences`, and
@@ -348,7 +353,33 @@ PROBE = {
     "s3": ProbeComparison(
         name="s3",
         spec_url="https://raw.githubusercontent.com/boto/botocore/develop/botocore/data/s3/2006-03-01/service-2.json",
+        mount=("/s3",),
         run=s3_probe.run,
+    ),
+}
+
+
+# Served paths no published document covers, and why. Matched by prefix, so `/batch` covers
+# `/batch/{api}/{version}`.
+#
+# The escape hatch the coverage check is built around, and deliberately narrow: an entry here is a
+# reason a reviewer reads, not a silence. Two kinds qualify — Backlot's own surface, which no
+# vendor publishes because it is not a vendor's, and a vendor surface that genuinely has no
+# document to compare against.
+UNCOMPARED = {
+    "/health": "Backlot's own liveness endpoint, not a vendor's surface",
+    "/oauth2/token": (
+        "Backlot's own token exchange. The vendors' own auth surfaces are compared under their "
+        "mounts; this one is how a caller gets a token for Backlot itself."
+    ),
+    "/_meta": (
+        "Backlot's own introspection — the corpus's principals and the per-source OpenAPI. No "
+        "vendor has it, by construction."
+    ),
+    "/batch": (
+        "Google's batch protocol is one endpoint carrying requests for every Google API, so no "
+        "single discovery document declares it: Drive's, Gmail's and the editor APIs' each "
+        "describe only their own operations."
     ),
 }
 
--- a/backlot/fidelity/comparisons.py
+++ b/backlot/fidelity/comparisons.py
@@ -3,7 +3,8 @@
 Four kinds, and they differ in what each side of the comparison is. For the two GraphQL sources,
 Backlot's side is the SDL the server builds its engine from and the vendor's is a live introspection
 response, which needs a credential. For the eight document sources, Backlot's side is the app's own
-``app.openapi()`` and the vendor's is a document it publishes, which needs none. For S3, whose
+``app.openapi()`` and the vendor's is one or more documents it publishes, which needs none. For
+S3, whose
 operations are selected by query string rather than by path, both sides are answers from a running
 server that ``backlot.serve()`` starts.
 
@@ -27,7 +28,32 @@ from backlot.fidelity import (
     s3_probe,
 )
 from backlot.fidelity.errors import CredentialsMissing, FidelityError
-from backlot.fidelity.findings import Finding
+from backlot.fidelity.findings import BREAKING, Finding
+
+
+def _one_report(per_spec: list[list[Finding]]) -> list[Finding]:
+    """Several documents' findings as one source's report: de-duplicated by key, then ordered the
+    way a single document's are.
+
+    Two documents of one source can declare the SAME path. Atlassian's do: 11 paths sit outside
+    ``/rest/api/{2,3}`` — ``/rest/atlassian-connect/1/*``, ``/rest/forge/1/*`` and one internal
+    worklog route — so both Jira documents carry them and each reports as ``missing_operation``
+    twice. That is a different case from the mirrored ``/rest/api/2/...`` and ``/rest/api/3/...``
+    entries, which are distinct paths a client can call and both belong in the baseline. A
+    :class:`~backlot.fidelity.findings.Baseline` is keyed on ``kind:path`` and cannot hold the
+    second copy, so leaving it in makes the file claim more than it can load and prints a genuinely
+    new finding on such a path twice.
+
+    Ordered here rather than left concatenated because ``diff_operations`` sorts breaking-first-
+    then-path per document, and that ordering does not survive joining two. The baseline is written
+    in the order it is given, so without this the next real change lands buried in a reordering of
+    the whole file.
+    """
+    seen: dict[str, Finding] = {}
+    for findings in per_spec:
+        for f in findings:
+            seen.setdefault(f.key, f)
+    return sorted(seen.values(), key=lambda f: (f.severity != BREAKING, f.path, f.kind))
 
 
 @dataclass(frozen=True)
@@ -110,10 +136,10 @@ class OpenAPIComparison:
     ) -> list[Finding]:
         # One memo across this run's specs: HubSpot reaches every document through a single index,
         # so without it a source comparing two of its APIs reads that index once per spec.
-        seen: dict[str, dict] = {}
-        return [
-            f for s in self.specs for f in openapi_diff.divergences(s, timeout=timeout, seen=seen)
-        ]
+        fetched: dict[str, dict] = {}
+        return _one_report(
+            [openapi_diff.divergences(s, timeout=timeout, seen=fetched) for s in self.specs]
+        )
 
 
 @dataclass(frozen=True)
@@ -140,9 +166,9 @@ class GoogleDiscoveryComparison:
     def divergences(
         self, credentials: Mapping[str, str] | None = None, *, timeout: float = 120.0
     ) -> list[Finding]:
-        return [
-            f for s in self.specs for f in google_discovery_diff.divergences(s, timeout=timeout)
-        ]
+        return _one_report(
+            [google_discovery_diff.divergences(s, timeout=timeout) for s in self.specs]
+        )
 
 
 @dataclass(frozen=True)
@@ -335,19 +361,23 @@ GOOGLE_DISCOVERY = {
             # empty `servicePath` and spell the version themselves (`v4/spreadsheets/...`), where
             # Drive's document repeats `drive/v3/`. Measured — with nothing stripped, every served
             # operation reports as one Backlot invented.
+            #
+            # The version is in the mount, matching `gen_docs.SOURCES`. A bare `/docs` would also
+            # select FastAPI's own `/docs` and `/docs/oauth2-redirect`, which escape this only
+            # because `include_in_schema=False` keeps them out of `app.openapi()`.
             Spec(
                 spec_url="https://www.googleapis.com/discovery/v1/apis/docs/v1/rest",
-                mount=("/docs",),
+                mount=("/docs/v1",),
                 strip="/docs",
             ),
             Spec(
                 spec_url="https://www.googleapis.com/discovery/v1/apis/sheets/v4/rest",
-                mount=("/sheets",),
+                mount=("/sheets/v4",),
                 strip="/sheets",
             ),
             Spec(
                 spec_url="https://www.googleapis.com/discovery/v1/apis/slides/v1/rest",
-                mount=("/slides",),
+                mount=("/slides/v1",),
                 strip="/slides",
             ),
         ),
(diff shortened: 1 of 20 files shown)


backlot/fidelity/comparisons.py
```diff
@@ -46,35 +46,74 @@ class Credential:
...
 
     def divergences(
         self, credentials: Mapping[str, str] | None = None, *, timeout: float = 120.0
     ) -> list[Finding]:
-        return openapi_diff.divergences(self, timeout=timeout)
+        # One memo across this run's specs: HubSpot reaches every document through a single index,
+        # so without it a source comparing two of its APIs reads that index once per spec.
+        seen: dict[str, dict] = {}
+        return [
+            f for s in self.specs for f in openapi_diff.divergences(s, timeout=timeout, seen=seen)
```
nuri-yoo: The two Jira documents declare 11 paths in common, so this concatenation emits 18 findings twice.

Measured at this head, per spec:

```
v3 spec -> 640 findings
v2 spec -> 621 findings
total 1261, unique keys 1243, duplicates 18
```

The shared paths are `/rest/atlassian-connect/1/*` (eight operations), `/rest/forge/1/*` (two) and `/rest/internal/api/latest/worklog/bulk` (one) — I fetched both documents and intersected their `paths`, which gives those 11. They surface as `missing_operation`, which no mount filters: `diff_operations` reports every vendor operation Backlot does not serve, whichever mount selected the served side.

Four consequences:

- `jira.json` ships 18 redundant entries. `Baseline.load` keys by `kind:path`, so `file entries 1261 -> loaded 1243`.
- `backlot diff --source jira` reports `total 1261`, and `--update-baseline` prints "now acknowledges 1261 divergence(s)" for a file that can only hold 1243.
- A genuinely new finding on one of those paths is printed twice and counted twice. Deleting one entry from a copy of the baseline and re-running gives `new: 2`, the same `missing_operation PUT /rest/forge/1/app/properties/{}` listed twice.
- Two identical entries carry nothing to say which document each came from, and these 11 are precisely the paths where the path itself does not say.

"Accepted rather than deduplicated" in the body covers the 584 mirrored `missing_operation` entries, where `/rest/api/2/attachment/{id}` and `/rest/api/3/attachment/{id}` really are different paths a client can call. These 18 lack that property. 584 is also 602 - 18, so that figure was already computed on the de-duplicated set.

De-duplicating by `Finding.key` here is safe: both copies carry the same `detail`, and `Baseline` cannot hold two entries under one key in any case. Sorting the concatenated list at the same time also fixes the baseline ordering described in the review body — `diff_operations` sorts breaking-first-then-path per spec, and the concatenation loses that across specs.

backlot/fidelity/comparisons.py
```diff
@@ -85,22 +124,25 @@ class GoogleDiscoveryComparison:
...
-        return self.spec_url
+    def endpoints(self) -> tuple[str, ...]:
+        return tuple(s.endpoint for s in self.specs)
 
     def divergences(
         self, credentials: Mapping[str, str] | None = None, *, timeout: float = 120.0
     ) -> list[Finding]:
-        return google_discovery_diff.divergences(self, timeout=timeout)
+        return [
+            f for s in self.specs for f in google_discovery_diff.divergences(s, timeout=timeout)
```
nuri-yoo: Same shape as the OpenAPI concatenation above, and nothing de-duplicates across specs here either. `google_drive` has no collisions today — 224 findings, 224 unique keys, because Drive, Docs, Sheets and Slides have disjoint operation namespaces — so this is not a live bug. But whatever fixes the Jira duplication belongs here too, rather than only on the one kind that happens to expose it.

backlot/fidelity/comparisons.py
```diff
@@ -220,15 +314,43 @@ def divergences(
...
+                mount=("/drive/v3",),
+            ),
+            # A Drive file is also read through the editor APIs, and each publishes its own
+            # discovery document. Each strips its own mount, unlike Drive: these three declare an
+            # empty `servicePath` and spell the version themselves (`v4/spreadsheets/...`), where
+            # Drive's document repeats `drive/v3/`. Measured — with nothing stripped, every served
+            # operation reports as one Backlot invented.
+            Spec(
+                spec_url="https://www.googleapis.com/discovery/v1/apis/docs/v1/rest",
+                mount=("/docs",),
```
nuri-yoo: These three mounts carry no version segment, and `/docs` collides with FastAPI's own Swagger UI: `GET /docs` and `GET /docs/oauth2-redirect` are live routes on this app, both answering 200. They are invisible to the comparison only because `include_in_schema=False` keeps them out of `app.openapi()`, so `from_backlot` never selects them. That, rather than the mount, is what stops them being diffed against the Docs discovery document and reported as invented surface.

`scripts/gen_docs.py:55` already spells the same four prefixes with their versions:

```
"google_drive": (..., ("/drive/v3", "/docs/v1", "/sheets/v4", "/slides/v1")),
```

Narrowing to `/docs/v1`, `/sheets/v4` and `/slides/v1` while keeping `strip` as it is selects exactly the same operations. Measured:

```
/docs/v1   strip=/docs   -> v1/documents/{}
/sheets/v4 strip=/sheets -> v4/spreadsheets/{}, v4/spreadsheets/{}/values/{}, v4/spreadsheets/{}/values:batchGet
/slides/v1 strip=/slides -> v1/presentations/{}
```

No served path falls outside the narrower prefix, so nothing is lost. It also lines this table up with the other one, and with Drive, Notion and HubSpot in this same registry, which all carry their version already.

backlot/fidelity/comparisons.py
```diff
@@ -251,10 +373,36 @@ def divergences(
...
+        mount=("/s3",),
         run=s3_probe.run,
     ),
 }
 
+
+# Served paths no published document covers, and why. Matched by prefix, so `/batch` covers
+# `/batch/{api}/{version}`.
+#
+# The escape hatch the coverage check is built around, and deliberately narrow: an entry here is a
```
nuri-yoo, line 414: Nothing holds this to "deliberately narrow". I added `"/notion/v1": "..."` to `UNCOMPARED` — a surface that is in fact fully compared — and the suite stayed green. So an entry here can claim a surface has no document when it has one, and a reader cannot take the list at face value.

The coverage guarantee itself survives: with that entry in place I also broke notion's mount, and `test_every_mount_selects_something_backlot_serves` caught the combination. So this is about the list being trustworthy to read, not about a hole in the check. One assertion in `test_no_uncompared_declaration_outlives_its_route` closes it — that no `UNCOMPARED` prefix sits under any comparison mount, and none does today, so it passes as written.

docs/fidelity.md
```diff
@@ -199,19 +199,33 @@ each of those is compared, and a test fails if the two sets ever drift apart.
...
+also read through Docs, Sheets and Slides, and HubSpot's associations are their own API at their own
+version.
+
+**Jira's `/rest/api/2` paths** are served because the clients call them — `atlassian-python-api`
+hardcodes `api_version = "2"` in its Jira constructor, and the `jira` PyPI client defaults
+`rest_api_version` to `"2"` and probes `/rest/api/2/serverInfo` on connect. They are compared
+against Atlassian's own v2 document, which it publishes beside the v3 one: the naming is
+`swagger[-<apiVersion>].<oasVersion>.json`, so the suffix-less `swagger.v3.json` is the v2 API.
+
+Every path Backlot serves is under one of those documents' mounts, probed, or declared in
```
nuri-yoo: "Every path Backlot serves" is more than the test measures. It counts `app.openapi()["paths"]`, which is what Backlot *declares*. Six live routes are outside it:

```
GET,HEAD  /docs                  include_in_schema=False
GET,HEAD  /docs/oauth2-redirect  include_in_schema=False
POST      /fireflies/graphql     include_in_schema=False
POST      /linear/graphql        include_in_schema=False
GET,HEAD  /openapi.json          include_in_schema=False
GET,HEAD  /redoc                 include_in_schema=False
```

121 route paths, 115 declared. Two of the six are vendor surface: the GraphQL POSTs at `routers/fireflies.py:32` and `routers/linear.py:34`. They *are* compared, by their GraphQL comparison — they simply have no mount to say so, and that asymmetry is where the gap sits. This PR gave `ProbeComparison` a `mount` purely so the coverage check could see it, and left `GraphQLComparison` with `endpoints` only, so `_comparison_mounts` reads an empty tuple for both.

The repo already carries this sentence in the two places that ran into it. `tests/test_docs.py:185` calls the GraphQL sources "the one part of the table `/openapi.json` cannot vouch for", and `scripts/gen_docs.py:20-28` records that walking `app.routes` was ruled out and that REST prefixes are checked against `/openapi.json` instead. So route introspection is not the fix. Either narrow this line to every path Backlot declares in its own OpenAPI, or give `GraphQLComparison` a `mount` of `("/fireflies/graphql",)` and `("/linear/graphql",)` so all four kinds answer the same question and only FastAPI's four are left to declare.

`/docs` and `/docs/oauth2-redirect` are a separate matter: they are currently swallowed by the new `/docs` mount rather than genuinely covered. See the comment on `comparisons.py`.

tests/test_fidelity.py
```diff
@@ -357,7 +544,11 @@ def test_the_mount_comes_off_only_where_the_vendor_does_not_repeat_it():
 
     def mounted(name):
         c = OPENAPI.get(name) or GOOGLE_DISCOVERY[name]
-        return {p for _, p in operations.from_backlot(served, c.mount, c.strip)}
+        return {
+            p
+            for spec in c.specs
```
nuri-yoo, line 660: The loop was widened over `specs` but the fixture was not, so this asserts exactly what it did before: `served` above holds no `/docs`, `/sheets` or `/slides` path, so the three new stripping specs select nothing and contribute nothing to `mounted("google_drive")`.

That leaves the strip decision for the editor APIs with no offline guard at all. I deleted `strip` from all three new Google specs and ran the whole fidelity suite: 69 tests, zero failures. The only thing that catches it is the nightly network run, which would then report the five served operations as surface Backlot invented.

Two lines fix it. Adding `"/sheets/v4/spreadsheets/{spreadsheet_id}"` to `served` and

```python
assert mounted("google_drive") == {"drive/v3/files", "v4/spreadsheets/{}"}
```

passes as written and fails once `strip` is removed from the sheets spec. I checked both directions. (`{}` because `operations.canonical` flattens placeholders.)

The docstring above also still reads "Google's own document already spells drive/v3, so nothing does", which is now true of one of `google_drive`'s four documents.

nuri-yoo (changes requested):
The design holds and I could not fault a measurement. I fetched both Atlassian documents (`swagger.v3.json` is 414 paths, 403 of them `/rest/api/2` and zero v3; `swagger-v3.v3.json` is the mirror; both carry the same `info.version` and `info.title`, so the paths are what identify them, not the file), read the two SDK defaults off the installed packages (`atlassian/jira.py:30-31` sets `kwargs["api_version"] = "2"`, and `jira/client.py:435` defaults `rest_api_version` to `"2"` with `get_server_info=True`, reaching `_get_json("serverInfo")` at line 661), confirmed `search/jql` splits GET-query from POST-`SearchAndReconcileRequestBean` in both documents down to the schema name, and confirmed the three editor discovery documents declare an empty `servicePath` so their `strip` is required. Reverting each of the three new guards fails exactly the intended test and nothing else: a deliberately uncompared route fails only `test_every_served_path_is_compared_or_says_why_not` and names the path, a dead mount fails only `test_every_mount_selects_something_backlot_serves`, and rolling `Spec.endpoint` back to a bare `spec_url` fails the distinctness assertion with `hubspot: (index, index)`. Every commit in the stack is green on `tests/test_fidelity.py` individually.

I audited the 4,992 lines of baseline JSON entry by entry rather than by count, comparing `(kind, severity, path, detail, note)` multisets against `main`: no source lost an entry, no existing key had its detail or note rewritten, and the growth is exactly the 708 you describe. Re-running `--update-baseline` into a copy of the directory reproduces `google_drive.json` and `hubspot.json` byte for byte.

One blocker, in the inline comment on `comparisons.py`: the two Jira documents share 11 paths, so 18 findings are emitted and stored twice. The rest is smaller. Four items have no line in the diff to hang on, so they are here.

**`docs/fidelity.md:53` still names the JSON key `endpoint`.** `_emit_json` now emits `endpoints` as a list, and the payload keys are `endpoints, new, resolved, source, total`. The line sits directly under the `jq` example that documents `--json` for piping, so a reader following it reads null. Nothing in `tests/` asserts on the payload shape, so there is no guard against this drifting again.

**`docs/fidelity.md:21-22` still asserts one document per source.** "an **OpenAPI** document for GitHub, Slack, Jira, Confluence, Notion and HubSpot, and a **Google API Discovery** document for Gmail and Drive" is singular, and it names Drive alone where the table 180 lines below now reads "Drive, Docs, Sheets and Slides". `comparisons.py:5-6` has the same singular ("the vendor's is a document it publishes"); the count of eight sources there is still right.

**`.github/workflows/fidelity.yml:117-118` carries a measurement this PR invalidates.** The comment justifies the 55000-character cap with "github's report is 135k characters and jira's 69k". Measured against an empty baseline at this head, jira is 136,865 characters and github is 133,621, so jira is now the larger of the two. The cap still holds, which is what a cap is for, but neither the number nor the ordering does.

**`tests/test_fidelity.py:135` is a call site the rename missed.** `Baseline.empty("fireflies", "https://api.fireflies.ai/graphql")` passes a bare string where the parameter is now `endpoints: tuple[str, ...]`, and `write` does `list(self.endpoints)`, so the fixture writes a 32-element list of single characters and loads it back the same way. The test immediately below was updated to `("https://new.invalid",)`; this one wants `("https://api.fireflies.ai/graphql",)`. It stays green only because `_baseline()`'s callers never assert on `endpoints`. The registry is guarded against exactly this in both directions — a string `mount` fails `test_every_mount_selects_something_backlot_serves` and a string `endpoints` fails the `startswith("http")` assertion in `test_every_comparison_names_the_documents_it_was_measured_against` — so this fixture is the one path around the guards.

**The shipped `jira.json` is not what the tool writes.** Round-tripping it through `--update-baseline` gives the same multiset with the same notes, the same `endpoints` and the same `measured`, in a different order: the six breaking entries sit at indices 0-5 in the file and at 0-2 and 640-642 in the tool's output, and `diff` between the two is 5,629 lines. `diff_operations` sorts breaking-first-then-path, and concatenating per-spec results means that ordering now holds per document rather than per source. The cost is that the next legitimate baseline update produces a 5,629-line reordering with the real change buried in it. Sorting the concatenation alongside the de-duplication in the inline comment makes the file canonical and the round-trip stable. The human report is unaffected, since `_echo_findings` regroups by severity.

Two things in the body:

**"Seventeen vendor routes were compared against nothing ... This compares all of them."** The 17 reproduces exactly: served paths under no pre-PR document mount, less Backlot's own five, is 17. But six of them are not routes this PR compares. `/s3`, `/s3/`, `/s3/{bucket}` and `/s3/{bucket}/{key}` were already compared, by the probe — the same document calls S3 "probed" — and `/batch` with `/batch/{api}/{version}` this PR declares in `UNCOMPARED` rather than compares. What it newly compares is 11 paths and 12 operations: jira v2's five paths and six operations, hubspot v4's one, docs' one, sheets' three, slides' one.

**"All eleven sources report `0 new`, including the two GraphQL ones run with credentials."** I ran all eleven with both GraphQL credentials. Nine report 0 new; linear reports 1, `missing_field Query.agentSessionSshAddress` at `gap`. It reproduces identically on `main`, so it is Linear drift that predates this branch and belongs to the scheduled job rather than to anything here — but the sentence is not true at this head.

Nit: `comparisons.py:406` reads "The three kinds differ" where the union below it has four members and the module docstring says "Four kinds". Pre-existing, and this PR adds text immediately above it.

Adjacent and out of scope, so I have not filed it yet: the same blind spot this PR fixes is still in `backlot/openapi.py:35`. `SOURCE_PREFIXES["gdrive"]` is `["/drive"]`, so the gdrive MCP slice is six Drive tools and the five Docs, Sheets and Slides routes have none — `/_meta/openapi/gdrive` returns only `/drive/v3/*`. `SOURCE_PREFIXES["atlassian"]` also carries `/wiki`, which has matched no served path since #44. `gen_docs.SOURCES` is checked for dead prefixes in `validate()`; that table is not, and the new dead-mount test does not reach it. Say the word and I will open it.

nuri-yoo:
Filed the adjacent one from the end of my review as #170, against `main` rather than this branch — `SOURCE_PREFIXES["gdrive"]` is `["/drive"]`, so `/_meta/openapi/gdrive` and `backlot mcp --source gdrive` carry six Drive operations and none of the five Docs, Sheets and Slides routes. Nothing for this PR to do; it is here because it is the same blind spot in the other prefix table.

khj809:
Thanks — the blocker reproduced exactly, and every other item was real. All addressed in 2c4134d; the PR body is rewritten rather than appended to.

**Blocker.** `_one_report()` now merges each source's per-document findings by `Finding.key` and re-sorts breaking-first-then-path, on both the OpenAPI and the Discovery kind. Confirmed before the fix: 1261 entries, 1243 unique keys, 18 duplicates, on exactly the `/rest/atlassian-connect/1/*`, `/rest/forge/1/*` and worklog paths you named. `jira.json` is now 1243 with none, and all three rewritten baselines round-trip through `--update-baseline` byte for byte. You were right that the mirrored v2/v3 argument does not stretch to these — that reasoning is now in `_one_report`'s docstring, next to the case it does not cover.

**The fixture.** Fixed, and guarded rather than only corrected: `Baseline.__post_init__` refuses a `str` for `endpoints`, which covers `empty`, `load` and `identified_as` together. Your point that this was the one path around the registry's guards is what made a guard rather than a fix seem right.

**Mounts.** Narrowed to `/docs/v1`, `/sheets/v4` and `/slides/v1`, matching `gen_docs.SOURCES`. The comment now records that a bare `/docs` also selects FastAPI's own and escapes only via `include_in_schema=False`.

**`UNCOMPARED`.** `test_no_uncompared_declaration_outlives_its_route` now also asserts no prefix sits under a comparison mount. Checked with your `/notion/v1` case: it fails and names the overlap.

**Docs.** `endpoint` → `endpoints` (a list) at :53, the singular document claim at :21-22 and in `comparisons.py:5-6`, and :224 now says *declares* rather than *serves*, with the six `include_in_schema=False` routes spelled out and why walking `app.routes` is not the answer here. Also the "three kinds" nit.

**The workflow comment.** Re-measured here rather than taking the numbers: jira 134,754, github 133,637. Slightly under yours — mine are post-de-duplication — but the ordering you flagged is the point and it holds, so the comment now reads jira 135k / github 134k.

**Body.** Both corrected: what this newly compares is 11 paths and 12 operations, not 17 routes; and ten of eleven sources report `0 new`, with linear's `Query.agentSessionSshAddress` named as pre-existing drift that reproduces on `main`.

You already opened the `SOURCE_PREFIXES` blind spot as #170 — thank you. Fixing it is a follow-up PR rather than part of this one: it is a different table serving the MCP slice, and the bidirectional check it wants ("every served path is under some source's prefixes") belongs with that fix rather than bolted onto this registry's tests. Left in this PR's "Not in scope" with the issue link.

tests/test_fidelity.py
nuri-yoo: Still open — this one is not in `2c4134d` and I could not find it in the reply either, so flagging rather than assuming it was declined.

Re-verified at `2c4134d`: deleting `strip` from all three Google editor specs leaves `tests/test_fidelity.py` at zero failures. The narrowed mounts do not change that, because the fixture's `served` still holds no `/docs/v1`, `/sheets/v4` or `/slides/v1` path, so the three specs select nothing here and the widened loop asserts what it did before.

The two-line change still works unaltered against this head. I applied it and ran both directions: adding `"/sheets/v4/spreadsheets/{spreadsheet_id}"` to `served` and

```python
assert mounted("google_drive") == {"drive/v3/files", "v4/spreadsheets/{}"}
```

passes as written, and fails once `strip` goes from the sheets spec.

The docstring above is also still on the old rule — "Google's own document already spells drive/v3, so nothing does" is now true of one of `google_drive`'s four documents, and the other three are the reason `strip` moved onto `Spec` in the first place.

nuri-yoo (commented):
Re-ran everything against `2c4134d` rather than reading the commit. The blocker is genuinely fixed and the merge is the right shape — `_one_report` on both kinds, keyed on `Finding.key`, re-sorted with `diff_operations`' own key.

```
baseline            file  unique  loaded  dups
confluence           147     147     147     0
fireflies             14      14      14     0
github              1236    1236    1236     0
gmail                165     165     165     0
google_drive         224     224     224     0
hubspot               23      23      23     0
jira                1243    1243    1243     0
linear               774     774     774     0
notion                53      53      53     0
s3                    43      43      43     0
slack                200     200     200     0
```

`file == unique == loaded` on all eleven, so no file claims more than it can load any more. `--update-baseline` into a copy reproduces `jira.json`, `google_drive.json` and `hubspot.json` byte for byte, which is the ordering half as well. Reverting `_one_report` to the old concatenation fails both new tests and nothing else. The `/notion/v1` case fails the new overlap assertion and names it: `/notion/v1 is declared uncompared but sits under ['/notion/v1']`. Mounts narrowed: 115 declared paths, zero unclassified, zero dead mounts, zero `UNCOMPARED`/mount overlaps, and no served path falls outside the version-qualified prefixes. Gates here: `2644 passed, 4 skipped` (the extra skip is Docker, so your 2645/3 matches), `ruff check` and `ruff format --check` clean, `gen_docs.py --check` clean. Adding a deliberately uncompared route still fails only `test_every_served_path_is_compared_or_says_why_not` and names the path.

I re-measured the workflow numbers rather than taking them: jira 134,738 and github 133,621, sixteen characters off yours in each case, which is the vendor documents drifting between fetches. Both round the way the comment now reads, and jira is the larger. `584` is still exact — 640 + 603 = 1243, of which 584 are `missing_operation`. The body's 11 paths and 12 operations breakdown adds up.

Three things still open. None is a correctness blocker; I will approve once the first two land, both one to two lines.

**The guard misses the one path its own comment names.** `__post_init__` tests `isinstance(self.endpoints, str)`, but `load` does `tuple(raw.get("endpoints", ()))` before constructing, so a string in the file is already a tuple of characters by the time the check runs:

```
Baseline.empty("s", "https://one.invalid")                     -> TypeError: endpoints is a tuple of URLs
Baseline.empty(...).identified_as("s", "https://b")            -> TypeError
Baseline.load(<file with "endpoints": "https://one.invalid">)  -> ('h', 't', 't', 'p', 's', ':', '/', '/', 'o', …)
```

The comment says "Refused at construction because that covers `empty`, `load` and `identified_as` alike", and the reply says the same; it covers two of the three, and `load` is the one that turns a file into the 32-element list the comment describes. Not reachable from anything shipped — the old format keyed `endpoint`, so `raw.get("endpoints")` returns `()` for those files — so this is the claim rather than a live defect. Checking the value in `load` before the `tuple()` call, or checking the tuple's contents rather than the field's type, closes it and lets the comment stand as written.

**The mount-strip fixture is still as it was**, and I have replied on that thread with the re-verification: deleting `strip` from all three editor specs still leaves the fidelity suite at zero failures, and the two-line fixture change still passes as written on this head and fails when the strip goes.

**"Three further tests keep that honest" is two tests carrying three checks.** `test_every_mount_selects_something_backlot_serves` is one; the route-gone check and the under-a-mount check are both assertions inside `test_no_uncompared_declaration_outlives_its_route`.

Three smaller things:

The insert left mid-paragraph orphan lines in three places, where every other line in each paragraph runs 90 to 100 characters: `comparisons.py:7` is `S3, whose` at nine characters between two 95-character lines, `docs/fidelity.md:25` is `public, so these` at sixteen, and `docs/fidelity.md:57` is the four words `or, with --update-baseline,` at twenty-nine. `ruff format` rewraps neither a docstring nor markdown, so all three pass the gates.

The opening sentence still reads "Seventeen vendor routes were compared against nothing". The second half is fixed, but four of those seventeen are S3's, which the probe compares and which this PR gave a `mount` precisely so the coverage check could see it. "Reached by no document" is true of all seventeen and is what the paragraph below already says ("11 paths and 12 operations that no document reached before").

`docs/fidelity.md:56` now reads "`endpoints` (a list, one per document)". One per compared contract, rather — `GraphQLComparison.endpoints` is an introspection URL, and its own docstring is explicit that "introspection is a single live schema, not a set of documents".

Agreed on #170 as a follow-up PR, and on the bidirectional check belonging with that fix rather than here.

khj809:
All six in 0483280. Both of the first two were claims of mine that outran what the code did, which is the more useful kind of catch.

**The guard missed `load`.** You are right that it covered two of the three the comment named, and that `load` is the one that turns a file into the spread. The check now runs on the raw value before `tuple()`, so all three paths refuse:

```
Baseline.empty("s", "https://one.invalid")                     -> TypeError
Baseline.empty(...).identified_as("s", "https://b")            -> TypeError
Baseline.load(<file with "endpoints": "https://one.invalid">)  -> TypeError
```

Split into `_reject_one_url` and `_endpoints` so `__post_init__` and `load` share the rule rather than restate it, and the comment now says which path each covers instead of claiming all three.

**The mount-strip fixture.** Applied as you wrote it — declined by oversight, not on purpose; I updated that test mechanically for `.specs` and never noticed its `served` had nothing the editor specs could select. Verified both directions on this head: green as written, and dropping `strip` from the sheets spec now fails `test_the_mount_comes_off_only_where_the_vendor_does_not_repeat_it` and nothing else. The docstring is rewritten too — it said Google's document spells `drive/v3` so nothing comes off, which is now true of one of four, and the other three are the reason `strip` moved onto `Spec`.

**"Three further tests"** is now "two tests carry three checks between them".

The three smaller ones: the orphan lines at `comparisons.py:7`, `docs/fidelity.md:25` and `:57` are rewrapped; the opening sentence reads "reached by no document", which as you say is true of all seventeen; and `endpoints` is described as one per compared *contract* — "a published document, or the URL introspection was read from" — since `GraphQLComparison`'s is the latter.

Gates: `2645 passed, 3 skipped`, `ruff check` and `ruff format --check` clean, jira/google_drive/hubspot still `0 new`.

Thanks for re-running rather than reading — the `load` gap and the strip fixture were both invisible from the diff.

nuri-yoo (approved):
All six land on `0483280`. Re-ran rather than read, including the non-regression side of the guard, which is the part that worried me more than the rejection.

```
Baseline.empty("s", "https://one.invalid")                     -> TypeError
Baseline.empty(...).identified_as("s", "https://b")            -> TypeError
Baseline.load(<file with "endpoints": "https://one.invalid">)  -> TypeError

round-trip ("https://a", "https://b")   -> ('https://a', 'https://b')
file with no "endpoints" key            -> ()
file with the legacy "endpoint" key     -> ()
all shipped baselines                   -> load
```

Splitting it so `__post_init__` and `load` share `_reject_one_url` rather than restate it is the right shape, and the comment now says which path each covers instead of the three.

The mount fixture: dropping `strip` from the sheets spec fails `test_the_mount_comes_off_only_where_the_vendor_does_not_repeat_it` and nothing else, and so does dropping it from all three at once. The rewritten docstring earns the fixture — it now says why `strip` is per-document rather than per-source, which is the thing the old one contradicted.

Gates here: `2644 passed, 4 skipped`, the extra skip being Docker, so your 2645/3 matches. `ruff check`, `ruff format --check` and `gen_docs.py --check` clean. `jira` 1243, `google_drive` 224, `hubspot` 23, all `0 new`. In the body, the opening now reads "reached by no document", the count is "Two further tests carry three checks between them", and `endpoints` is one per compared contract. `comparisons.py:7` reflows cleanly.

Approving. Two leftovers, both cosmetic, either now or whenever:

**`docs/fidelity.md:26` — the orphan moved two lines rather than went away.** `bodies are out of scope` is 23 characters between a 94 and a 97; the rewrap stopped before the paragraph did, which still has three lines to run.

**The `load` rejection has no test.** Reverting `_reject_one_url(raw)` inside `_endpoints` leaves `tests/test_fidelity.py` at zero failures — `test_a_baseline_refuses_one_endpoint_passed_unwrapped` exercises only `Baseline.empty`. Two lines on that test, a tmp file whose `endpoints` is a string, cover the path the whole change was about. Worth naming the shape, since it is the mirror of the thing you caught: the fix for a claim that outran the code is now code that no test would notice losing.

Resolving my six threads. On the two you called the more useful kind of catch — both were only findable by running the thing, which is the argument for doing it on a diff this size rather than the other way round.

nuri-yoo:
Re-ran on `dde0163` — approval stands, and the `/batch` rewrite is a better claim than the one it replaces, which I had let through. Checked it against all five documents rather than the two the comment names:

```
gmail        batchPath='batch'            servicePath=''          ops=79  batch as an operation: none
drive/v3     batchPath='batch/drive/v3'   servicePath='drive/v3/' ops=64  batch as an operation: none
docs/v1      batchPath='batch'            servicePath=''          ops=3   batch as an operation: none
sheets/v4    batchPath='batch'            servicePath=''          ops=17  batch as an operation: none
slides/v1    batchPath='batch'            servicePath=''          ops=5   batch as an operation: none
```

So "every discovery document names a batch endpoint in its top-level `batchPath`" holds for all five, the two spellings the comment gives are right, and `from_google_discovery` yields no batch operation from any of them — which is the half that matters, since operations are what a path diff pairs. `2644 passed, 4 skipped` here, ruff and `gen_docs.py --check` clean.

Both leftovers from the approval are still open, unchanged and still cosmetic: `docs/fidelity.md:26` is `bodies are out of scope` at 23 characters mid-paragraph, and reverting `_reject_one_url(raw)` inside `_endpoints` still leaves the suite at zero failures.

khj809:
Both leftovers in 0eb52a1.

**The untested `load` rejection.** Caught by my own mutation pass and then let through by it, which is the more interesting half: I mutated `_reject_one_url`'s body, so both call sites died at once and `test_a_baseline_refuses_one_endpoint_passed_unwrapped` failed, and I read that as "guarded". Mutating the shared helper is not mutating the guard — each **site** has to go independently. Re-run that way:

```
caught by  load's call site removed              -> test_a_baseline_refuses_one_endpoint_passed_unwrapped
caught by  __post_init__'s call site removed     -> test_a_baseline_refuses_one_endpoint_passed_unwrapped
caught by  the shared helper made a no-op        -> test_a_baseline_refuses_one_endpoint_passed_unwrapped
```

The test now walks all three construction paths and, since you flagged the non-regression side as the part that worried you more, the shapes a real file carries: a two-element list round-trips, and an absent `endpoints` key — which is every shipped baseline written before this branch — still loads as `()`.

**`docs/fidelity.md:26`.** Reflowed to the end of the paragraph this time rather than to the line I had edited; 90/99/57 where it was 94/23/97/97/28.

Gates: `2645 passed, 3 skipped`, `ruff check`, `ruff format --check` and `gen_docs.py --check` clean.

On `/batch`: you confirmed the rewrite against all five documents, and the reason is now accurate as far as it goes — but reviewing it again I do not think it is the *whole* reason, and the entry may not deserve to be in `UNCOMPARED` at all. Each API has its own host with `batch` at the root, Drive alone sitting on the shared `www.googleapis.com` and needing the `batch/drive/v3` discriminator; Backlot's two routes are exactly those two shapes collapsed onto one origin. So the structural obstacle is that a path diff pairs operations selected by a per-**source** mount while Google's batch endpoints are per-**host** — `/batch` stands in for Drive's, Sheets' and Gmail's at once, and belongs to no one source's mount. What *is* checkable is that Backlot's two route shapes cover the `batchPath` every document declares. So `/batch` gets its own comparison, the way S3 has one: a kind that answers "does this hold" without being a path diff, moving the endpoint from "compared against nothing" to "compared, differently". S3 is already the precedent for a surface whose contract a path map cannot express — there it is the query string, here it is a per-host endpoint on a mock that collapses five hosts onto one origin.

That is a follow-up PR, not a line here: it grows the comparison machinery a fourth kind, and `tests/test_fidelity.py` is explicitly no-network so the check belongs on the `backlot diff` path. The `UNCOMPARED` entry stays in this PR as the accurate record until it lands.

PR #164, opened 2026-09-09
Slack reaction corpus

grf53:
## Description

A slack corpus had to invent the Slack user ids in `reactions.users`, values it has no way to compute, and state `count` beside them where the two could disagree; a record now names the reactors by address and the router renders both.

## What changed

```
AS-IS   corpus    {"name": "eyes", "count": 4,
                   "users": ["UD425829D56", "U025E855846", "UAE3444BFEB", "UDB00FA9703"]}
        response  the same object, passed through

TO-BE   corpus    {"name": "eyes", "users": ["ava@acme.com", "bob@acme.com"]}
        response  {"name": "eyes", "users": ["UC20FA2B1C0", "UD425829D56"], "count": 2}
```

- **The contract.** The schema declares the reaction item instead of accepting any object, on a root message and on a reply alike: `name` and `users` required, each entry of `users` an address, and the list non-empty. The router builds the served object from those addresses rather than passing the stored one through, in Slack's own field order.
- **The corpora.** Every reaction in the repo now names its reactors by address, so no invented id is left anywhere.

Breaking for the BYO contract: a corpus outside the repo carrying `"users": ["U01"]` stops at import. Deliberate, because accepting both spellings leaves one field meaning two things, and reading an old file as addresses that resolve to nobody is worse than refusing loudly.

What Backlot returned already matched Slack, and this change doesn't make a difference in API responses.

## Verification

- **Tests** — `pytest` on a `.[…]` install: `.[all]` plus `llama-index-readers-hubspot<0.6`, the zero-skip install CONTRIBUTING.md names. 2635 passed, 2 skipped.
- **Optional surfaces that ran rather than skipped** — `tests/test_llamaindex.py::test_slack`, which drives the official LlamaIndex Slack reader through `conversations.history`, the response this changes. It needs `.[llamaindex]` and sits out of a `.[dev,examples]` run. The 2 remaining skips are the mirage Google shim, which no install here reaches.
- **Lint** — `ruff check . && ruff format --check .`: clean, 138 files.

- [ ] ~~A new endpoint serving corpus content is ACL-scoped, proved for both an admin and a scoped token (no new endpoint: ...)~~

## Notes

- **Breaking for the BYO contract**: a corpus outside the repo carrying `"users": ["U01"]` stops at import. Deliberate, because accepting both spellings leaves one field meaning two things, and reading an old file as addresses that resolve to nobody is worse than refusing loudly.
- A reactor the corpus names nowhere else serves an id `users.info` cannot resolve, since it resolves message authors only. Same limitation `users.list` already carries for display-only speakers.


nuri-yoo (commented):
Superseded by the inline version of this review: https://github.com/brekkylab/backlot/pull/164#pullrequestreview-5150166151 (same findings, anchored to the lines they concern).

backlot/schemas/slack.schema.json
```diff
@@ -68,9 +68,30 @@
...
+        "description": "One reaction. Slack's own OpenAPI (`objs_reaction`) requires `name`, `users` and `count` together, and types each entry of `users` as a `defs_user_id` matching `^[UW][A-Z0-9]{2,}$` — a value a corpus author cannot know, since it is what `synth.slack_user_id` will make of an address. So the corpus writes the ADDRESS and the router renders the id, the same way `author_email` becomes a message's `user`.",
+        "properties": {
+          "name": {
+            "type": "string",
+            "minLength": 1,
+            "description": "The emoji name, without colons: `eyes`, `+1`, `tada`."
+          },
+          "users": {
+            "type": "array",
+            "minItems": 1,
```
nuri-yoo, line 81: `users` accepts the same address twice, and the served reaction is one Slack never sends:

```python
record_errors(rec(reactions=[{"name": "eyes", "users": ["bo@x.com", "bo@x.com"]}]))
# -> []
_message(row)["reactions"]
# -> [{'name': 'eyes', 'users': ['U8DFFC8F825', 'U8DFFC8F825'], 'count': 2}]
```

Slack refuses the second reaction by the same person with `already_reacted`: "The specified item already has the user/reaction combination." So `count: 2` over one id is the disagreement the PR sets out to make impossible, produced by the derivation itself. `"uniqueItems": true` here (and on the `replies` copy) refuses at import the way the rest of this PR refuses; no other schema in the repo uses the keyword yet, but the validator is already `Draft202012Validator` so it is available. The alternative is deduping in `_reactions` (noted there), which serves a correct pair but reads a duplicate the author probably did not mean to write. This one matches the stance the body takes on the old spelling.

Ref: https://docs.slack.dev/reference/methods/reactions.add (Errors, `already_reacted`)

backlot/routers/slack.py
```diff
@@ -1121,6 +1121,39 @@ def _message(
...
+    Field order is Slack's own: its `conversations.history` reference shows
+    `{"name": …, "users": [...], "count": N}`.
+    """
+    out = []
+    for reaction in store.jcol(row, "reactions"):
+        users = list(reaction.get("users") or [])
+        out.append(
+            {
+                "name": reaction.get("name"),
+                "users": [synth.slack_user_id(e) for e in users],
```
nuri-yoo, line 1150: If the schema does not take `uniqueItems`, `dict.fromkeys(users)` here keeps `count` honest for a duplicated address (order-preserving). Trade-off against the schema route is in the comment on `slack.schema.json`: this serves a correct pair silently, that one names the duplicate at import.

backlot/schemas/slack.schema.json
```diff
@@ -68,9 +68,30 @@
...
+            "items": {
+              "type": "string",
+              "format": "email"
+            },
+            "description": "Who reacted, by address. A reaction nobody made is not one Slack returns, so the list may not be empty."
+          }
+        },
+        "required": [
```
nuri-yoo, line 89: The item still reads and drops `count`, and any other key:

```python
record_errors(rec(reactions=[{"name": "eyes", "users": ["ava@x.com"], "count": 9}]))      # -> []
record_errors(rec(reactions=[{"name": "eyes", "users": ["ava@x.com"], "count": "nine"}]))  # -> []
record_errors(rec(reactions=[{"name": "eyes", "users": ["ava@x.com"], "bogus": 1}]))       # -> []
```

Walking every schema in `backlot/schemas/` for an object that declares `properties` without `additionalProperties`, this item and its `replies` copy are the only hits in the eleven files; the 13 other typed item objects (`confluence` comments, `fireflies` sentences and meeting_attendees, `github` comments and branches, `gmail` messages, `hubspot` associations, `jira` comments, `linear` relations, comments and attachments, `notion` comments, `slack` replies) all carry `additionalProperties: false`, and `schemas/README.md` states the rule as "a key no schema names is a validation error rather than a value read and dropped". The body argues that reading an old file quietly is worse than refusing loudly; a `count: 9` that is read and discarded is the same class. Suggest `additionalProperties: false` on both items. A corpus that kept `count` then stops at import with the key named, which is what the body says it wants for the old shape.

tests/test_slack.py
```diff
@@ -1396,3 +1396,49 @@ def test_slack_one_person_has_one_handle_across_the_surface(tmp_path):
...
+                "source_type": "slack",
+                "channel": "inc",
+                "content": "gateway is flapping",
+                "author_email": "ava@x.com",
+                "visibility": "public",
+                "reactions": [
+                    {"name": "eyes", "users": ["ava@x.com", "bo@x.com"]},
+                    # A `count` a record still carries is ignored. Deriving it is what keeps the
+                    # served pair from disagreeing, whatever the record says.
+                    {"name": "+1", "users": ["bo@x.com"], "count": 9},
```
nuri-yoo, line 1426: With `additionalProperties: false` on the item (comment on `slack.schema.json`), this case moves to `test_schema.py` as a refusal (`'count' was unexpected`) rather than pinning the silent drop here.

examples/bring-your-own-corpus/sample_corpus.jsonl
```diff
@@ -1,8 +1,8 @@
...
+{"source_type": "slack", "content": "Reminder: production deploy freeze starts Friday 5pm through Monday.", "channel": "eng-announcements", "group": "engineering", "author_email": "ava@acme.com", "author_name": "Ava Ng", "visibility": "public", "created": "2026-02-05T17:00:00Z", "reactions": [{"name": "eyes", "users": ["ava@acme.com", "bob@acme.com", "hana@acme.com", "raj@acme.com"]}, {"name": "+1", "users": ["ava@acme.com", "bob@acme.com"]}], "edited": {"user": "UC20FA2B1C0", "ts": "1770311280.000000"}, "participants": ["ava"], "doc_id": "sl-deploy-freeze"}
```
nuri-yoo, line 4: This line now carries the one case the Notes flag as unresolvable, and on main it did not. Served this file at this head and called `users.info` on each reactor:

```
ava@acme.com   UC20FA2B1C0 -> ok
bob@acme.com   UD425829D56 -> ok
hana@acme.com  UAE3444BFEB -> ok
raj@acme.com   UB191607BE3 -> user_not_found
```

`raj@acme.com` appears nowhere else in the file. Served main (1246b35) the same way: its seven reactor ids are `synth.slack_user_id` of principals the corpus already names (`mia`, `hana`, `ava`, `bob`, `ceo`, `sec`, `cfo` at acme.com), every one resolves through `users.info`, and every one is in `users.list`. So for the repo's own sample the ids were computed, not invented, and the AS-IS block in the body describes a BYO author's position rather than this file. Swapping `raj@acme.com` for one of the addresses the old file resolved to (`mia@acme.com`, say) keeps the example demonstrating the contract without demonstrating the limitation.

Separately, "no invented id is left anywhere" in the body is wider than the diff: `"edited": {"user": "UC20FA2B1C0", …}` stays on this line and on a reply of the next, a hand-written Slack id of the same class, typed as a bare object in the schema and passed through at `slack.py:1094`. The argument in `_reactions`'s docstring applies to it verbatim; that is follow-up material rather than this PR, but the sentence should say "every reaction" and stop.

backlot/routers/slack.py
```diff
@@ -1121,6 +1121,39 @@ def _message(
...
+def _reactions(row) -> list[dict]:
+    """A message's ``reactions``, built from the addresses the corpus stated.
+
+    Not pass-through, and the vendor's own spec is why. `objs_reaction` requires `name`, `users`
+    and `count` together, and each entry of `users` is a `defs_user_id` (`^[UW][A-Z0-9]{2,}$`) —
+    which is `synth.slack_user_id` of a person, a value no corpus author can know. Writing it by
+    hand meant inventing ids for people the corpus already names by address everywhere else.
+
+    `count` is DERIVED rather than stated. Slack computes one from the other, so a corpus that
+    wrote both could put them in disagreement and produce a reaction the real API never sends; with
```
nuri-yoo, line 1133: This is the opposite of what Slack documents. From the `reactions.get` reference: "The users array in the reactions property will always contain the authenticated user, but might not always contain all users that have reacted. The value of count, however, will always represent the count of all users who made that reaction (i.e. it may be greater than users.length)." And `conversations.history`: "If you need a full list of reactions for a message, use the reactions.get method." So the real API does send `count > len(users)`. Deriving `count` is a sound choice for a corpus Backlot never truncates, and the response still validates against `objs_reaction`; only the justification is off. Suggest saying that (Backlot serves every reactor, so `len(users)` is the count) here and in the two schema `description`s, in place of the claim that Slack computes one from the other. The field order cited below does match the `reactions.get` example response (`name`, `users`, `count`).

Ref: https://docs.slack.dev/reference/methods/reactions.get (Usage info)
Ref: https://docs.slack.dev/reference/methods/conversations.history (Usage info)

backlot/schemas/slack.schema.json
```diff
@@ -68,9 +68,30 @@
...
+          "name",
+          "users"
+        ]
       },
-      "description": "[{name, count, users:[U…]}]."
+      "description": "Reactions to this message, as `[{name, users: [<address>, …]}]`. `count` is derived from `users` rather than read, so the two can never disagree."
```
nuri-yoo, line 94: Same as the docstring comment in `slack.py`: Slack's `reactions.get` reference says `count` "may be greater than users.length", so "the two can never disagree" holds for Backlot's derivation, not for Slack. Worth phrasing as Backlot's guarantee here and on the `replies` copy.

backlot/routers/slack.py
```diff
@@ -1121,6 +1121,39 @@ def _message(
...
+    only `users` written down there is nothing for it to disagree with.
+
+    The address is not required to belong to a principal. `users.info` resolves a message AUTHOR's
```
nuri-yoo, line 1136: Small: `users.info` resolves every principal first (`slack.py:719-721`, `all_user_emails`) and falls back to authors, so "a message AUTHOR's id" is narrower than the code. `hana@acme.com` in the sample never posted to Slack and resolves for that reason. "a principal's or a message author's id" is the accurate span.

nuri-yoo (commented):
Reviewed at edab70d0 (2 commits on 1246b35). Ran `pytest tests/test_schema.py tests/test_slack.py tests/test_importer_byo.py` (402 passed), `ruff check` / `ruff format --check` on the touched files (clean), and `scripts/gen_docs.py --check` (clean). Fetched Slack's OpenAPI: `objs_reaction` requires `name`, `users`, `count` and `defs_user_id` is `^[UW][A-Z0-9]{2,}$`, exactly as the docstring and schema quote them. The direction is right; the inline comments are about what the new contract still lets through and about prose that describes the old one. Two spots outside the diff that now describe the refused shape:

- `backlot/importer/byo.py:38` — the module docstring's slack reply example is `"reactions": [{"name": "eyes", "count": 1}]`, which this schema rejects (`[replies/0/reactions/0]: 'users' is a required property`, run through `record_errors`). A reader who copies it gets an import error.
- `backlot/schemas/README.md:127` — `reactions` is listed among the "loosely typed per-service extras … stored as JSON without a fixed shape". That stays true for github's `{emoji: count}` map and is no longer true for slack.

Ref: https://github.com/slackapi/slack-api-specs/blob/master/web-api/slack_web_openapi_v2.json

backlot/schemas/slack.schema.json
grf53: Thank you for pointing this out.
Adopted `uniqueItems`.

backlot/routers/slack.py
grf53: Adopted `uniqueItems`.

backlot/schemas/slack.schema.json
grf53: added `"additionalProperties": false` on both.

examples/bring-your-own-corpus/sample_corpus.jsonl
grf53: Maybe there was a mistake.
Fixed them into ones based on existing user ids in the corpus.

grf53: In the meanwhile, I found that there is no `deactivated` user(`deleted: True`) in the sample.
Maybe this can be another issue.

backlot/schemas/slack.schema.json
nuri-yoo: Confirmed at b30ef32: `["bo@x.com", "bo@x.com"]` is refused with `has non-unique elements` on a root message and on a reply, and with the keyword removed from the schema `test_slack_reaction_refuses_the_same_person_twice` fails. Resolving.

examples/bring-your-own-corpus/sample_corpus.jsonl
nuri-yoo: Confirmed at b30ef32: served this file and called `users.info` on each of the seven reactors (`ava`, `bob`, `ceo`, `cfo`, `hana`, `mia`, `sec` at acme.com): 7/7 `ok`, all seven in `users.list`. The rendered `users` and `count` on the four reactions equal the ids main's file wrote by hand, so the wire response is unchanged.

On the `deleted: True` observation: agreed, and it is an issue rather than a sample fix. The BYO contract has no field that marks a person deactivated, and `users.list` renders `"deleted": False` for every member (`slack.py:318`), so the sample cannot carry one until the router and schema can. No open issue covers it (searched 2026-09-11).

nuri-yoo (approved):
Approving at b30ef32 (7 commits on bef7320). Re-ran what the 09-09 review measured: `pytest tests/test_schema.py tests/test_slack.py tests/test_importer_byo.py` 436 passed; `ruff check` and `ruff format --check` on the four touched Python files clean; `gen_docs.py --check` clean. Removing `uniqueItems` and `additionalProperties` from the schema fails the two tests that pin them. The vendor quotes in `_reactions` and the schema descriptions match the `reactions.get` and `reactions.add` pages word for word (fetched today).

The body lags the branch in four places; none blocks the merge:

- **The contract** bullet does not say the item is closed: a stated `count`, a repeated address and a misspelled key are refused at import now (b833b49, ff29c7a).
- **Breaking** names `"users": ["U01"]` as what stops at import. A file that carries `count`, which is what the AS-IS block shows, stops too, on `count` before it reaches the ids.
- **Notes**: "`users.info` … resolves message authors only" is the reading b30ef32 corrected in the docstring: principals first, message authors as the fallback.
- **Verification**: "2635 passed, 2 skipped" predates the rebase and the two tests added since.

"no invented id is left anywhere" is still wider than the diff: `"edited": {"user": "UC20FA2B1C0"}` stays on the sample's first message and on a reply of its second. "every reaction names its reactors by address" is the sentence the diff supports; `edited.user` is follow-up material.

nuri-yoo:
Two follow-ups filed off this review, neither for this PR:

- #197 — `edited.user` is the Slack id a corpus still writes by hand, on the message and on a reply. The sample's two `edited` blocks resolve because their value is `synth.slack_user_id("ava@acme.com")`; a record written by address the way `reactions` now teaches is accepted at import and serves `user_not_found`. Rewriting them is what lets the body's "no invented id is left anywhere" hold; until then "every reaction names its reactors by address" is the sentence the diff supports.
- #198 — the `deleted: True` observation from the `sample_corpus.jsonl` thread. `_user_obj` serves `"deleted": False` for every member (`slack.py:318`) and no record or roster entry can change it, so it is a router and roster gap rather than a sample fix.

PR #165, opened 2026-09-09
a Jira comment read serves the page its envelope describes, ordered and clamped the way the real API does

### LINKED ISSUE #163: jira: comment reads ignore startAt, maxResults and orderBy, serving the whole collection as page one
**Source and commit** — `jira`, at `1246b35` (`backlot 0.0.3.dev13+g1246b35af`).

**How you know** — a live call to Jira Cloud on 2026-09-09 for every parameter's validation and echo, and Atlassian's published v2 (`swagger.v3.json`) and v3 (`swagger-v3.v3.json`) documents for what the parameters mean. Both documents declare `startAt` (integer, default `0`, "the index of the first item to return in a page of results (page offset)"), `maxResults` (integer, default `100`, "the maximum number of items to return per page"), `orderBy` (enum `created` / `+created` / `-created`) and `expand` on `GET /rest/api/{2,3}/issue/{issueIdOrKey}/comment`, with a `PageOfComments` response of `startAt` / `maxResults` / `total` / `comments`.

The live site's issues have no comments on them, and creating one is a write to a real instance — but that turns out not to matter for most of this: **parameter validation happens before any comment lookup**, so the accepted values, the clamps, the caps and the two error shapes are all observable on an empty issue. What is *not* measured is the ordering itself — that `-created` reverses a non-empty list is taken from the documents, not seen.

```
GET /rest/api/3/issue/SCRUM-1/comment        (a real issue, zero comments)

  (no params)          200  {"startAt": 0,   "maxResults": 100, "total": 0}
  orderBy=created      200  {"startAt": 0,   "maxResults": 100, "total": 0}
  orderBy=+created     200  {"startAt": 0,   "maxResults": 100, "total": 0}
  orderBy=-created     200  {"startAt": 0,   "maxResults": 100, "total": 0}
  orderBy=bogus        400  {"errorMessages": ["정렬할 필드는 [created] 중 하나여야 합니다. …"], "errors": {}}
  orderBy=updated      400  {"errorMessages": ["정렬할 필드는 [created] 중 하나여야 합니다. …"], "errors": {}}
  startAt=100          200  {"startAt": 100, "maxResults": 100, "total": 0}
  startAt=-1           200  {"startAt": 0,   "maxResults": 100, "total": 0}
  maxResults=0         200  {"startAt": 0,   "maxResults": 1,   "total": 0}
  maxResults=-1        200  {"startAt": 0,   "maxResults": 1,   "total": 0}
  maxResults=100000    200  {"startAt": 0,   "maxResults": 100, "total": 0}
  maxResults=abc       400  application/problem+json
                            {"type": "about:blank", "title": "Bad Request", "status": 400,
                             "detail": "Failed to convert 'maxResults' with value: 'abc'",
                             "instance": "/rest/api/3/issue/SCRUM-1/comment"}
  expand=renderedBody  200  {"startAt": 0,   "maxResults": 100, "total": 0}
  expand=bogus         200  {"startAt": 0,   "maxResults": 100, "total": 0}
```

Four things in there are not in any document and would have been invented otherwise: `maxResults` is **capped at 100**, not merely defaulted to it; `maxResults` below 1 clamps **up to 1**, not down to 0; a negative `startAt` clamps to 0 while one past the end is echoed back unchanged with an empty page; and the two failures use **different error shapes** — semantic validation answers Jira's classic `errorMessages` envelope, type conversion answers RFC 7807 `application/problem+json`. `expand` is not validated at all, so an unknown value is accepted silently.

The `orderBy` message is localised to the account's language, so only the status and the body's shape are portable — not its wording.

This surfaced while adding the Jira v2 document to `backlot diff` (#162), which reports `startAt`, `maxResults`, `orderBy` and `expand` as `missing_param` here at both versions.

**Reproduction** — one record, one issue, seven comments:

```json
{"source_type": "jira", "project": "payments", "doc_id": "pay-1", "title": "Checkout latency spike", "content": "p95 jumped to 2.1s.", "issuetype": "Bug", "status": "In Progress", "author_email": "bob@acme.com", "reporter": "bob@acme.com", "created": "2026-02-08T20:00:00Z", "updated": "2026-02-08T20:00:00Z", "comments": [{"content": "comment 1", "author_email": "ava@acme.com", "created_ts": 1770000060}, {"content": "comment 2", "author_email": "ava@acme.com", "created_ts": 1770000120}, {"content": "comment 3", "author_email": "ava@acme.com", "created_ts": 1770000180}, {"content": "comment 4", "author_email": "ava@acme.com", "created_ts": 1770000240}, {"content": "comment 5", "author_email": "ava@acme.com", "created_ts": 1770000300}, {"content": "comment 6", "author_email": "ava@acme.com", "created_ts": 1770000360}, {"content": "comment 7", "author_email": "ava@acme.com", "created_ts": 1770000420}]}
```

```
backlot import corpus.jsonl && backlot serve
curl -H "Authorization: Bearer $ADMIN" \
  "$BASE/atlassian/rest/api/3/issue/$KEY/comment?startAt=3&maxResults=2"
```

| # | Divergence | Silent? |
|---|---|---|
| 1 | `startAt` and `maxResults` ignored; the whole collection is served as if it were page one | yes |
| 2 | `orderBy` ignored, and a value the vendor refuses is accepted | yes |

## 1. `GET /issue/{key}/comment` advertises a page but serves the whole collection

**Request** — the three below, against the seven-comment issue above.

**Backlot serves** — identical bodies for all three:

```
(no params)               {"startAt": 0, "maxResults": 7, "total": 7}
                          -> ['comment 1', … , 'comment 7']
?startAt=3&maxResults=2   {"startAt": 0, "maxResults": 7, "total": 7}
                          -> ['comment 1', … , 'comment 7']
?startAt=100              {"startAt": 0, "maxResults": 7, "total": 7}
                          -> ['comment 1', … , 'comment 7']
```

Three things are wrong at once. `startAt` and `maxResults` are read from nowhere — [`jira_issue_comments`](https://github.com/brekkylab/backlot/blob/1246b35/backlot/routers/atlassian.py#L385) takes only `key` and `request`, and never touches `request.query_params`. `maxResults` is echoed as the size of the collection rather than the page size, so it re-states `total` instead of describing the page. And `startAt` is hard-coded `0`, so the envelope asserts the caller is on the first page whatever they asked for.

**The real API serves** — comments `[startAt, startAt + maxResults)`, with `startAt` and `maxResults` echoing the values in force after clamping and `total` counting the collection. Measured above: no parameters gives `maxResults: 100`; `maxResults` is capped at 100 and floored at 1; `startAt` is floored at 0 and echoed unchanged past the end; a non-integer is `400 application/problem+json`.

**Can a client tell?** Not from the status, and often not at all — this is the expensive case. Every response is `200` with a well-formed `PageOfComments`, so nothing downstream notices. A client paging with `while startAt + maxResults < total` happens to terminate, because Backlot's echoed `maxResults` always equals `total` — it just silently reads page one as the whole collection, which is correct only while an issue has fewer comments than the page size. Ask for page two and you get page one, labelled page one. Against real Jira the same client gets page two.

The reverse is worse for a corpus with long threads: a client asking for `maxResults=25` to bound its memory or its token budget is handed every comment on the issue.

## 2. `orderBy` is ignored, and an invalid value is accepted rather than refused

**Request** — `?orderBy=-created`, and `?orderBy=bogus`.

**Backlot serves** — `200` for both, comments in corpus order (`seq`), identical to a request with no `orderBy` at all.

**The real API serves** — `created`, `+created` and `-created` are accepted and sort by the comment's created date; anything else is `400` with `{"errorMessages": ["<localised>"], "errors": {}}`. Both measured above.

**Can a client tell?** No, and it fails in both directions. A client asking for newest-first gets oldest-first and has no way to know the sort was dropped — the response carries no statement of the order it is in. A client that sends a value real Jira rejects — `updated`, say — is told `200` here and `400` there, so a bug that would have surfaced immediately against Jira surfaces only in production.

## Not in this issue

`expand=renderedBody`, which returns the comment body rendered as HTML, needs ADF-to-HTML rendering and is its own piece of work. Note that accepting and ignoring an *unknown* `expand` is not a divergence: real Jira does not validate the parameter either, and answers `200` to `expand=bogus` (measured above).

The two other pagination envelopes in the Jira router — `fields.comment` and `changelog` inside `GET /issue/{key}` ([atlassian.py#L771](https://github.com/brekkylab/backlot/blob/1246b35/backlot/routers/atlassian.py#L771), [#L792](https://github.com/brekkylab/backlot/blob/1246b35/backlot/routers/atlassian.py#L792)) — are **not** the same bug. Those are embedded pages on an endpoint whose own spec declares no `startAt` or `maxResults`, so no input is being ignored. Whether real Jira caps an embedded comment list is unmeasured and would be its own issue.

khj809:
Closes #163.

## What changed

`GET /rest/api/{2,3}/issue/{key}/comment` serves the page its envelope has always described. It reads `startAt` and `maxResults`, slices, and echoes the values actually in force — `total` staying the count of the whole collection, which is what makes the envelope a page rather than a restatement of its own length. `orderBy` sorts by the comment's created date in either direction, and any field but `created` is refused the way Jira refuses it. Sent without one, the comments keep whatever order the corpus stated — see below.

Before this, all three parameters were read from nowhere: the handler took `key` and `request` and never touched `request.query_params`. Every response was the whole collection labelled `startAt: 0, maxResults: <collection size>`, so a client asking for page two got page one and had nothing in the response to tell it apart. That is the expensive shape of wrong — a `200` with a well-formed `PageOfComments` that nothing downstream questions.

The clamps are the vendor's, not invented: `maxResults` is capped at 100 and floored **up** to 1, `startAt` is floored at 0, and a `startAt` past the end is echoed back unchanged with an empty page rather than refused.

`expand` is deliberately **not** declared. Real Jira takes one here and `expand=renderedBody` returns each comment's body as HTML, which Backlot does not produce — and `qp` exists only for parameters Backlot honours, because advertising one it ignores makes a client ask for data that never arrives. Declaring it would also have made `backlot diff` clean while the behaviour was unchanged, and put the parameter in front of an agent through `GET /_meta/openapi/atlassian`. The gap stays where it is legible instead: two `missing_param` entries in the jira baseline, still reported.

Measured against Jira Cloud on 2026-09-09, on a real issue. Parameter validation runs before any comment lookup, so an issue with no comments settles the accepted values, the clamps, the caps and the refusals:

```
GET /rest/api/3/issue/SCRUM-1/comment        (a real issue, zero comments)

  (no params)          200  {"startAt": 0,   "maxResults": 100, "total": 0}
  orderBy=created      200  {"startAt": 0,   "maxResults": 100, "total": 0}
  orderBy=+created     200  {"startAt": 0,   "maxResults": 100, "total": 0}
  orderBy=-created     200  {"startAt": 0,   "maxResults": 100, "total": 0}
  orderBy=bogus        400  {"errorMessages": ["…must be one of [created]…"], "errors": {}}
  startAt=100          200  {"startAt": 100, "maxResults": 100, "total": 0}
  startAt=-1           200  {"startAt": 0,   "maxResults": 100, "total": 0}
  maxResults=0         200  {"startAt": 0,   "maxResults": 1,   "total": 0}
  maxResults=-1        200  {"startAt": 0,   "maxResults": 1,   "total": 0}
  maxResults=100000    200  {"startAt": 0,   "maxResults": 100, "total": 0}
  expand=bogus         200  {"startAt": 0,   "maxResults": 100, "total": 0}
```

Four of those are in no document and would have been invented otherwise: that 100 is a **cap** and not merely a default, that a sub-1 `maxResults` clamps **up**, that a `startAt` past the end is echoed rather than reset, and that `expand` is unvalidated.

The `orderBy` grammar is measured too, and is wider than the enum the document prints:

```
orderBy=+created     200   a literal + decodes to a space, and is still accepted
orderBy=%20created   200   leading space          orderBy=created%20   200   trailing
orderBy=-%20created  200   sigil, then space      orderBy=Created      200   any case
orderBy=--created    400   echoes `-created` — exactly ONE sigil is stripped
orderBy=             400   an empty value is refused, not read as absent

GET /issue/NOPE-1/comment?orderBy=bogus   400      ...without it                404
```

That last pair settles the order of checks: the parameter is validated **before** the issue is resolved. It separates nothing — the 400 is identical for a key that exists, one hidden from the caller, and one that never existed.

One thing is still not measured: what `-created` does to a non-empty list, since the site available has no issue carrying a comment and creating one is a write to a live instance. That comes from the documents.

The no-`orderBy` default is a decision rather than a reproduction, and the document it declines to follow is named beside it in the code. Atlassian's REST intro says responses are "listed in ascending order by default", but that is a general statement its own operations contradict — project classification says "If not provided, values will not be sorted" — and it cannot be checked here for the same reason. So a read without `orderBy` leaves the corpus's own order alone, and a test names that so it is not later reversed on the assumption that Jira sorts.

Jira's `orderBy` message is localised to the account's language, so the refusal reproduces its status and envelope and not its wording — the same call the Sheets enum refusals make.

Two things measured here are deliberately left alone. `maxResults=abc` answers `400 application/problem+json` on real Jira, a different envelope from the `errorMessages` one used for semantic validation, and a repeated `?orderBy=a&orderBy=b` is joined and validated as `a,b` there where Starlette hands over only the last. Both are router-wide — the same `_int` and the same `.get` serve `search/jql` and both Confluence listings — so they are #180 rather than this endpoint's business. And the `fields.comment` and `changelog` envelopes inside `GET /issue/{key}` are not the same bug — those are embedded pages on an endpoint whose own spec declares no such parameters, so no input is being ignored there.

## Verification

- **Tests** — `pytest` on a `uv sync --all-extras` install: **2893 passed, 3 skipped**
- **Optional surfaces that ran rather than skipped** — the S3, SDK, MCP, LlamaIndex reader and mirage suites all ran. Correcting this line from an earlier revision: the two mirage skips are not "a module absent from the published package". `mirage.core.google._client` ships in mirage-ai 0.0.5, inside the `~=0.0.5` pin, and is gone from 0.0.6, which is what the resolver picks — so the tripwires skip on exactly the version where the integration is broken. That is its own issue, not this PR's. The third skip is `llama_index.readers.hubspot`, which no extra carries because its `hubspot-api-client<9` pin conflicts with the `>=12` the `examples` extra needs.
- **Lint** — `ruff check . && ruff format --check .`: clean, 138 files

- [ ] A new endpoint serving corpus content is ACL-scoped, proved for both an admin and a scoped token — **n/a**: no new endpoint, and the ACL path is untouched. The issue is still resolved through `_resolve_jira_key(request, conn, key, ids)` before any comment is read, and comments are not individually scoped, so paging exposes nothing an unpaged read did not.


backlot/routers/atlassian.py
```diff
@@ -88,11 +88,31 @@ class ConfluencePage(_ALoose):
...
+_P_JIRA_COMMENTS = {
+    "parameters": [
+        qp("startAt", "integer"),
+        qp("maxResults", "integer"),
+        qp("orderBy"),
+        # Declared and not read. Measured against real Jira: it does not validate this either —
+        # `expand=bogus` answers 200 — so accepting a value and ignoring it is what the vendor
+        # does. What `expand=renderedBody` returns needs ADF-to-HTML rendering, which Backlot has
+        # not.
+        qp("expand"),
```
nuri-yoo: This declaration contradicts the rule `qp` states for itself: "only for parameters Backlot actually honours: advertising one it ignores makes a client ask for data that never arrives, which is worse than not offering it." A client that reads the spec and sends `expand=renderedBody` gets a 200 with no `renderedBody` on any comment, which is exactly that case; that real Jira does not validate the value says nothing about whether Backlot serves it.

The honest home for this gap already exists: `backlot/fidelity/baseline/jira.json` acknowledges `GET /rest/api/{2,3}/issue/{}/comment?expand` as "the vendor accepts it; Backlot does not". Leaving `expand` undeclared keeps those two entries true and keeps `backlot diff` reporting the gap; declaring it makes the diff clean while the behaviour is unchanged.

If you would rather close the gap instead: "needs ADF-to-HTML rendering" overstates it for Backlot's own ADF. `_adf` only ever emits `paragraph` nodes carrying one plain `text` node each, so `renderedBody` for what this server produces is `"".join(f"<p>{html.escape(p)}</p>" for p in paras)`. Either way, one of the two rather than declared-and-ignored.

backlot/routers/atlassian.py
```diff
@@ -392,13 +431,37 @@ async def jira_issue_comments(key: str, request: Request):
             status_code=404,
             detail="Issue does not exist or you do not have permission to see it.",
         )
+    params = request.query_params
+    # An absent or empty `orderBy` is "not asked", and leaves the corpus's own order alone. What
+    # real Jira returns without one is NOT measured — the site available had no issue carrying a
```
nuri-yoo: "NOT measured" is right, but the PR body goes further ("what real Jira returns with no `orderBy` is not established at all", "in no document"), and there is a document. Atlassian's REST intro, section "Expansion, pagination, and ordering" → "Ordering":

> Some operations support ordering the elements of a response by a field. Check the documentation for the operation to confirm whether ordering of a response is supported and which fields can be used. Responses are listed in ascending order by default. You can change the order using the `orderby` query parameter with a `-` or `+` symbol.

(https://developer.atlassian.com/cloud/jira/platform/rest/v3/intro/#ordering). It is a general statement, and at least one operation contradicts it for its own `orderBy` (project classification: "If not provided, values will not be sorted"), so it does not settle the comment endpoint on its own. But the decision pinned by `test_jira_comments_without_order_by_keep_the_order_the_corpus_states` is "corpus order, not ascending", which is the opposite of the one default any Atlassian document states. Two ways to make that defensible: cite the intro here and say why it is discounted, or spend the one write on the sandbox issue that measures it — a single comment on SCRUM-1 also turns "`-created` reverses" from a document reading into a measurement.

backlot/routers/atlassian.py
```diff
@@ -392,13 +431,37 @@ async def jira_issue_comments(key: str, request: Request):
...
+    # comment — so sorting by default would be inventing an order rather than reproducing one.
+    order = params.get("orderBy") or ""
+    if order and order not in _JIRA_ORDER_BY:
+        raise HTTPException(
+            status_code=400,
+            detail=f"The field to order by must be one of [created]. Instead, : {order}",
```
nuri-yoo: The docstring says the wording is not reproduced, only the refusal, but this string reads as a transcription of Jira's English (`Instead, : {order}` is not punctuation anyone writes fresh). If it is Jira's English string, say so and where it was seen; if it is Backlot's own sentence, drop the stray comma. As written, a reader cannot tell which it is, and the localised message in #163 hides the tail behind `…`.

backlot/routers/atlassian.py
```diff
@@ -392,13 +431,37 @@ async def jira_issue_comments(key: str, request: Request):
...
+        )
+    start = max(0, _int(params.get("startAt"), 0))
```
nuri-yoo, line 476: `_int` swallows a non-integer to the default, so `maxResults=abc` and `startAt=abc` answer 200 with the first hundred here, where real Jira answers `400 application/problem+json` (measured in #163). The body says this is deliberately left alone and "noted on #163" — but this PR closes #163, so after merge nothing open records it. It wants its own issue (router-wide: the same `_int` serves `search/jql` and both Confluence listings). Happy to file it if you would rather not.

backlot/routers/atlassian.py
```diff
@@ -392,13 +431,37 @@ async def jira_issue_comments(key: str, request: Request):
...
             detail="Issue does not exist or you do not have permission to see it.",
         )
+    params = request.query_params
+    # An absent or empty `orderBy` is "not asked", and leaves the corpus's own order alone. What
+    # real Jira returns without one is NOT measured — the site available had no issue carrying a
+    # comment — so sorting by default would be inventing an order rather than reproducing one.
+    order = params.get("orderBy") or ""
+    if order and order not in _JIRA_ORDER_BY:
+        raise HTTPException(
+            status_code=400,
```
nuri-yoo: Order of checks: the key is resolved (404) before `orderBy` is validated (400). The body's "parameter validation runs before any comment lookup" was inferred from an existing issue with zero comments, which shows validation does not need comments, not that it precedes issue resolution. `GET /rest/api/3/issue/NOPE-1/comment?orderBy=bogus` on the sandbox settles which of 404 and 400 real Jira answers, and is one read. Not blocking; just an unmeasured cell the body reads as measured.

nuri-yoo (commented):
Reviewed at head `5d883fc` in its own worktree, base `165e019`. The paging, the clamps and the refusal all check out against the v3 document (`swagger-v3.v3.json`: `startAt` default 0, `maxResults` default 100, `orderBy` enum `created`/`-created`/`+created`, `expand` "accepts `renderedBody`") and against the cells measured in #163, and the code reads the way the body says it does. Three things are worth settling before merge; the rest are small.

**1. The baseline still acknowledges what this PR resolves.** `backlot/fidelity/baseline/jira.json` carries eight `missing_param` entries for `GET /rest/api/{2,3}/issue/{}/comment?{startAt,maxResults,orderBy,expand}`, and the PR does not touch the file. At this head:

```
$ python -m backlot diff -s jira
   1235 divergence(s) · 1235 acknowledged · 0 new
🧹 acknowledged but no longer diverging (8) — drop from the baseline
     GET /rest/api/2/issue/{}/comment?expand
     GET /rest/api/2/issue/{}/comment?maxResults
     GET /rest/api/2/issue/{}/comment?orderBy
     GET /rest/api/2/issue/{}/comment?startAt
     GET /rest/api/3/issue/{}/comment?expand
     GET /rest/api/3/issue/{}/comment?maxResults
     GET /rest/api/3/issue/{}/comment?orderBy
     GET /rest/api/3/issue/{}/comment?startAt
```

#131 pruned 14 such entries from `github.json` in the same PR that closed them, and #161 pruned 21 from `google_drive.json`, so this is the convention rather than a new ask. Which entries go depends on the `expand` decision below: six if `expand` stays undeclared, eight if it is served.

**2. `expand` declared but not read** — inline at `atlassian.py:100`. It is the one thing in the diff that goes against `qp`'s own rule, and the baseline already has the honest place for it.

**3. The no-`orderBy` default has a document after all** — inline at `atlassian.py:436`. Atlassian's intro states "Responses are listed in ascending order by default"; the body says no document does. Cite and discount, or measure with one comment on the sandbox.

Smaller, each inline: the non-integer `maxResults`/`startAt` divergence loses its only record when #163 closes (`:444`); 404-before-400 is an unmeasured ordering the body reads as measured (`:441`); the refusal string's provenance (`:442`); "Every expectation below is measured" against two that are not (`test_atlassian.py:818`); sort-before-slice and the `seq` tie-break have no test (`:917`).

Outside the diff: `docs/supported-sources.md:217` still shows `issue/{key}/comment` with an empty Notes cell. The table's own convention lists honoured paging and ordering parameters there (`v3/objects/{objectType}`: "`limit` max 100, `after`"; `/drive/v3/files`: the `orderBy` values), so this row wants `startAt`, `maxResults` (max 100), `orderBy` `created`/`+created`/`-created`. The block is hand-maintained; `gen_docs.py --check` passes at this head and would not catch it.

**Verification (this env, `uv sync --all-extras` venv):** full suite at `5d883fc` — `2874 passed, 4 skipped in 50.08s` (the fourth skip is Docker, absent here; the other three match the body). Mutation: with `backlot/routers/atlassian.py` taken from `origin/main` and the tests from this head, all 16 new tests fail; router restored from HEAD afterwards. `ruff` is CI-green on this head and was not re-run.

tests/test_atlassian.py
```diff
@@ -810,3 +811,140 @@ def test_confluence_child_page_and_restriction_match_a_nonexistent_id_for_an_out
         made_up = client.get(f"/atlassian/wiki/rest/api/content/999999999/{path}", headers=h)
         assert hidden.status_code == made_up.status_code == 404
         assert hidden.content == made_up.content
+
+
+# --- Jira: a page of comments is a page -------------------------------------------------------
+#
+# Every expectation below is measured against Jira Cloud (2026-09-09) on a real issue. Parameter
```
nuri-yoo: "Every expectation below is measured" is contradicted twice in this block: the direction of `-created` is taken from the document (the next sentence says so), and `test_jira_comments_without_order_by_keep_the_order_the_corpus_states` says in its own docstring that its case is not measured at all. "The clamps, caps and refusals below are measured …; the ordering comes from Atlassian's document, and the no-`orderBy` case from neither" is what the tests actually claim.

tests/test_atlassian.py
```diff
@@ -810,3 +811,140 @@ def test_confluence_child_page_and_restriction_match_a_nonexistent_id_for_an_out
...
+    "order,want",
+    [
+        # `comment 7` is the OLDEST in this corpus and `comment 1` the newest, so an ascending
+        # sort has to reorder the rows rather than pass them through.
+        ("created", ["comment 7", "comment 6"]),
+        ("+created", ["comment 7", "comment 6"]),
+        ("-created", ["comment 1", "comment 2"]),
+    ],
+)
+def test_jira_comments_order_by_created(paged, order, want):
```
nuri-yoo, line 923: Two properties the router comments promise but no test pins: that the sort happens before the slice (every `orderBy` case here uses the default `startAt`, so a sort applied to the sliced page would pass), and the `seq` tie-break for two comments with the same `created_ts`. One parametrize row with `startAt=2, orderBy=-created` and one fixture pair sharing a timestamp would cover both. Optional.

backlot/routers/atlassian.py
```diff
@@ -392,13 +431,37 @@ async def jira_issue_comments(key: str, request: Request):
             status_code=404,
             detail="Issue does not exist or you do not have permission to see it.",
         )
+    params = request.query_params
+    # An absent or empty `orderBy` is "not asked", and leaves the corpus's own order alone. What
+    # real Jira returns without one is NOT measured — the site available had no issue carrying a
+    # comment — so sorting by default would be inventing an order rather than reproducing one.
+    order = params.get("orderBy") or ""
```
nuri-yoo: A literal `+` in a query string decodes to a space, so the spelling Atlassian's own document writes (`?orderBy=+name`) arrives here as `" created"` and is refused:

```
GET /atlassian/rest/api/3/issue/PAY-7/comment?orderBy=+created&maxResults=2
400 {"errorMessages": ["The field to order by must be one of [created]. Instead, :  created"]}

GET /atlassian/rest/api/3/issue/PAY-7/comment?orderBy=%2Bcreated&maxResults=2
200 {"startAt": 0, "maxResults": 2, "total": 7}  -> [comment 7, comment 6]
```

(Probed at this head; both API versions behave the same.) The `jira` package builds the URL through `requests`, which encodes `+` as `%2B`, so that client is fine; a curl user or anything that pastes the documented spelling gets the 400. Which of the two real Jira does with a literal `+` is the cell that decides it, and it is one read on the sandbox: `curl -g "$BASE/rest/api/3/issue/SCRUM-1/comment?orderBy=+created"`. If real answers 200, `order = (params.get("orderBy") or "").replace(" ", "+", 1)` (or a `lstrip`) is the fix and the doubled space in the message goes with it; if real answers 400, the current behaviour matches and only the message wants tidying. Either way a test with the literal spelling, sent as a raw URL rather than through `params=`, pins it.

Two more cells from the same probe that #163 did not measure, for the record rather than for this PR: Starlette's `QueryParams.get` returns the LAST value, so `?orderBy=bogus&orderBy=created` is 200 and `?startAt=3&startAt=5` serves page five; and `?orderBy=` (present, empty) is treated as absent while `?orderBy=Created` is refused.

tests/test_atlassian.py
```diff
@@ -810,3 +811,140 @@ def test_confluence_child_page_and_restriction_match_a_nonexistent_id_for_an_out
...
+    assert d["comments"] == []
+
+
+@pytest.mark.parametrize(
+    "order,want",
+    [
+        # `comment 7` is the OLDEST in this corpus and `comment 1` the newest, so an ascending
+        # sort has to reorder the rows rather than pass them through.
+        ("created", ["comment 7", "comment 6"]),
+        ("+created", ["comment 7", "comment 6"]),
```
nuri-yoo, line 919: This row reaches the server as `orderBy=%2Bcreated`, because httpx percent-encodes the value passed through `params=`. The spelling Atlassian's document writes, a literal `+` in the URL, is not exercised by any test here and answers 400 at this head (see `atlassian.py:438`). Worth one raw-URL case once real's answer for it is known.

nuri-yoo (commented):
Second pass at the same head `5d883fc`, re-running what the first pass concluded and probing the angles it did not: both API versions, the literal `+` spelling, repeated and odd-typed parameters, a scoped token, and the surface `backlot mcp` derives from the spec. Then the whole suite with nothing skipped.

**One new thing worth settling before merge**, inline at `atlassian.py:438`: `?orderBy=+created` written literally, as Atlassian's document writes it, is a 400 here — the `+` decodes to a space. The tests pass it through `params=`, which encodes it as `%2B`, so they never meet the spelling. Whether real Jira accepts the literal form is one read on the sandbox; the fix is a one-liner either way.

**Re-verified, holds:**
- v2 and v3 answer identically for the default read, `startAt=3&maxResults=2`, `orderBy=-created&startAt=2&maxResults=2` and `orderBy=created&maxResults=7`.
- The sort happens before the slice (`-created&startAt=2&maxResults=2` serves the third and fourth newest), and two comments sharing a `created_ts` come out in `seq` order ascending and in reverse `seq` order descending — the stable order the tie-break comment at `:456` promises, with the direction following the sort.
- ACL: a scoped user who cannot see the issue gets 404 for `?orderBy=bogus` and for `?startAt=1` alike, the same body as for a key that does not exist, and the user who can see it pages it. So the 404-before-400 order I asked about in the first pass is also the order that leaks nothing; if real turns out to validate first, matching it is still safe, since a 400 for a hidden key and a 400 for a missing key are equally uninformative. The n/a checkbox holds.
- `GET /_meta/openapi/atlassian`, the slice `backlot mcp` hands `FastMCP.from_openapi`, lists `startAt`, `maxResults`, `orderBy` and `expand` on `jira_issue_comments`. That is the concrete cost of the `expand` declaration from the first pass: an agent is offered an input that changes nothing in the answer.
- `ruff check` and `ruff format --check` on the two touched files: clean.

**Verification with every skip run.** A separate venv: `uv pip install -e '.[all]'`, then `mirage-ai[fuse]==0.0.5` and `llama-index-readers-hubspot --no-deps` (its `hubspot-api-client<9` pin conflicts with the `>=12` the `examples` extra needs, which is why no extra carries it), with Docker up for the mcp-atlassian test:

```
2878 passed, 12 warnings in 51.02s      (0 skipped)
```

The four that skipped in the body's run — `test_mcp_atlassian_acl_enforced`, the two mirage tripwires, `test_hubspot` — all pass at this head. One correction to the body's account of them: the two mirage skips are not "third-party modules absent from the published packages". `mirage.core.google._client` is in mirage-ai 0.0.5, which is inside the `~=0.0.5` pin; it is absent from 0.0.6, which is what the resolver picks today, and on 0.0.6 `point_google_at` raises `ModuleNotFoundError: No module named 'mirage.core.google._client'` at the first call. So the tripwires skip on exactly the version where the integration is broken. That is its own issue and not this PR's; I can file it if you would rather not.

backlot/routers/atlassian.py
khj809: Undeclared. `qp`'s rule decides it, and the two `expand` entries stay in the baseline reporting the gap.

Not taking the render-it option: `renderedBody`'s HTML shape is not something I can measure. The sandbox has no issue carrying a comment, so `<p>…</p>` would be a guess at the vendor's markup rather than a reproduction of it — and a wrong shape declared as honoured is worse than an honest gap.

backlot/routers/atlassian.py
khj809: Corrected. The intro is now cited in the comment and discounted for the reason you give — it is general, and project classification contradicts it for its own `orderBy`. The PR body's "in no document" is gone.

Not spending the write: measuring it means creating a comment on a live instance that is not mine to write to. The decision stays "corpus order", now with the document it is declining to follow named beside it.

backlot/routers/atlassian.py
khj809: It was mine, mimicking the structure of the Korean string — which is exactly the ambiguity you are pointing at. Rewritten as Backlot's own sentence (`Instead: {field}`, no stray comma) with a comment saying so and saying why Jira's own English is not available from that account.

backlot/routers/atlassian.py
khj809: Filed as #180 rather than left on a closing issue. Router-wide, as you say — the same `_int` serves `search/jql` and both Confluence listings.

backlot/routers/atlassian.py
khj809: Measured, and it goes the other way: `GET /issue/NOPE-1/comment?orderBy=bogus` is **400** on real Jira while the same key without the parameter is 404. So the parameter is validated first and Backlot had it backwards.

Reordered, and your ACL reasoning is why it is safe to: the 400 is identical for a key that exists, one hidden from the caller, and one that never existed, so it separates none of the three. `test_jira_validates_order_by_before_resolving_the_issue` pins it.

tests/test_atlassian.py
khj809: Rewritten to say what the block actually claims: the clamps, caps, accepted and refused spellings and the 400-before-404 order are measured; `-created`'s effect on a non-empty list comes from the document; the no-`orderBy` case from neither.

tests/test_atlassian.py
khj809: Both pinned. `test_jira_sorts_the_whole_collection_before_slicing_it` uses `startAt=2, orderBy=-created`, and `test_jira_orders_comments_sharing_a_timestamp_by_seq` builds two comments on one timestamp and asserts `seq` ascending and reversed descending.

backlot/routers/atlassian.py
khj809: Real answers **200**, so Backlot was wrong. Probing around it found the rule is wider than the `+`:

```
orderBy=+created     200   (literal +, decodes to a space)
orderBy=%20created   200   leading space
orderBy=created%20   200   trailing space
orderBy=-%20created  200   sigil, then space
orderBy=Created      200   case-insensitive
orderBy=CREATED      200
orderBy=--created    400   echoes `-created`
orderBy=%2B-created  400   echoes `-created`
orderBy=             400
```

So: exactly one leading `+`/`-` is stripped (the two 400s echoing `-created` prove it is one, not all), whitespace around it is ignored, and the field matches case-insensitively — which also makes the empty value a 400 rather than "absent", where Backlot had been reading it as absent. `_jira_order_desc` now does that, and the accepted spellings are parametrized as raw query strings.

Left alone, as you suggested: the repeated-parameter join. Starlette's `.get` taking the last value is router-wide, the same class as the `_int` one, and it is on #180.

tests/test_atlassian.py
khj809: Covered by `test_jira_order_by_takes_the_spellings_the_real_api_takes`, which builds the URL as a raw string so the `+` reaches the server literally.

khj809:
Two passes' worth of findings are in at `3d700eb`, each answered in its own thread. Three changed behaviour, and probing one of them found three more divergences neither pass had measured.

**`orderBy` was wrong in five ways, not one.** The literal `+created` you flagged led to measuring the whole grammar, and real Jira is looser than the enum its document prints: exactly one leading `+`/`-` is stripped (`--created` is a 400 that echoes `-created`, which is what proves it is one and not all), whitespace around the sigil is ignored, and the field matches case-insensitively. An **empty** value is a 400 there, where Backlot had been reading it as absent. `_jira_order_desc` now implements that, parametrized over raw query strings so the `+` reaches the server literally.

**The check order was backwards.** `GET /issue/NOPE-1/comment?orderBy=bogus` is 400 on real Jira where the same key alone is 404, so the parameter is validated before the issue is resolved. Your ACL reasoning is what makes matching it safe, and it is now in the code comment: the 400 is identical for a key that exists, one hidden from the caller, and one that never existed.

**`expand` is undeclared**, on `qp`'s own rule. Not the render-it option — `renderedBody`'s HTML shape is not measurable from a sandbox with no comments, so it would be a guess at the vendor's markup dressed as a reproduction.

**The baseline is pruned**, six entries rather than eight: undeclaring `expand` returns its two to being true.

Also done: the intro's "ascending order by default" is cited and discounted in the code rather than denied in the body; the refusal string is now plainly Backlot's own sentence with a comment saying why Jira's English is unavailable from that account; the test block header says which of its expectations are measured and which are not; and `docs/supported-sources.md` gets the Notes cell in the table's own convention.

Deferred to **#180**, both router-wide as you said: the swallowed non-integer, and the repeated parameter Starlette reads as its last value where Jira joins and validates all of them.

One correction to take back to your own note — you were right that my skip account was wrong, and the version detail you supplied is now in the body.

`2893 passed, 3 skipped`, ruff clean, `backlot diff --source jira` reports 1237 acknowledged and nothing stale.

backlot/routers/atlassian.py
```diff
@@ -380,9 +425,43 @@ async def jira_get_issue(key: str, request: Request):
...
+    response_model=JiraComments,
+    openapi_extra=_P_JIRA_COMMENTS,
+)
 async def jira_issue_comments(key: str, request: Request):
+    """One PAGE of an issue's comments.
+
+    `total` counts the whole collection while `startAt` and `maxResults` describe the slice, which
+    is what makes the envelope a page rather than a restatement of its own length.
+
+    Ordering is by `orderBy`, which real Jira accepts only as `created`, `+created` or `-created`
```
nuri-yoo: Non-blocking, and the last of the prose lag: this sentence is the rule from before `_jira_order_desc`, and the same commit makes it false. `Created`, `CREATED`, `%20created`, `created%20` and `-%20created` are each 200 on real Jira by your own table, and each 200 here now. The helper's docstring at `:115` states the measured grammar; this one still states the enum the document prints, so a reader who takes it learns a rule the file disproves.

"accepts `created` with an optional single leading `+`/`-`, case-insensitively and ignoring whitespace — see `_jira_order_desc`" keeps the two in step.

backlot/routers/atlassian.py
```diff
@@ -88,11 +88,56 @@ class ConfluencePage(_ALoose):
...
+# (0 and -1 both answer 1). `startAt` floors at 0, and one past the end is echoed back unchanged
+# with an empty page rather than refused.
+_JIRA_COMMENT_PAGE_MAX = 100
+_JIRA_ORDER_FIELD = "created"
+
+
+def _jira_order_desc(raw: str) -> bool:
+    """Whether ``orderBy`` asks for the reverse, or Jira's 400 for a field it does not order by.
+
+    Measured against Jira Cloud (2026-09-09). Exactly ONE leading sigil is stripped: `--created`
```
nuri-yoo: Non-blocking. The sigil, whitespace and case rules were measured on 2026-09-10, not 09-09: your reply upthread says the literal `+created` is what led to measuring the grammar, and that flag went up this morning. #180 dates its own calls "2026-09-09 and 2026-09-10", and the baseline in this very commit moves `measured` to 2026-09-10, so the file next to this one already says today.

`tests/test_atlassian.py:819` has the same seam from the other side: it puts "the accepted and refused `orderBy` spellings and the 400-before-404 order" under 09-09, and both of those came from today's session. The clamps and the caps are the 09-09 ones. Since a date here says which instance state a claim reflects, splitting them is worth the line.

nuri-yoo (approved):
Approving. Every thread from both passes is addressed, and I re-ran each claim at `3d700eb` in its own worktree rather than reading the diff for it. Two prose nits inline, neither blocking.

**Verified by execution**

- **The `orderBy` grammar matches the table you measured.** I replayed all nine spellings from your reply plus the three plain ones against Backlot: `+created`, `%20created`, `created%20`, `-%20created`, `Created`, `CREATED` answer 200 with the right direction, and `--created`, `%2B-created`, the empty value answer 400. Both API versions agree. The refusal echoes the field after one sigil is stripped, so `--created` reports `-created` the way real does, and `%2B%2Bcreated` reports `+created` on the same rule.
- **400 before 404, and it separates nothing.** With `orderBy=bogus` a key that exists, one hidden from the caller and one that never existed all answer the identical 400 body; without the parameter the same three answer 200, 404, 404. That is what makes matching real safe here, and the test pins the first half of it.
- **`expand` is gone from the surface a client sees**, not just from `qp`: neither `/openapi.json` nor `GET /_meta/openapi/atlassian` lists it now, so the MCP tool no longer offers an input that changes nothing.
- **The baseline is right and the run is clean.** `backlot diff --source jira` gives `1237 divergence(s) · 1237 acknowledged · 0 new` with no stale section, where the same command at the previous head listed eight. Six pruned, `expand` × 2 kept — the arithmetic works out.
- **Sort before slice and the `seq` tie-break** behave as the comments promise: `-created&startAt=2&maxResults=2` serves the third and fourth newest, and two comments on one timestamp come back in `seq` order ascending and reversed descending.
- **Suite, with nothing skipped.** All extras, Docker up, and mirage pinned to 0.0.5 so the tripwires run: `2896 passed, 0 skipped in 52.27s`. That is your `2893 passed, 3 skipped` with the three environment skips filled in. Mutation: with `backlot/routers/atlassian.py` taken from `origin/main` and the tests from this head, 30 tests fail; router restored from HEAD afterwards. `ruff check` and `ruff format --check` on the two touched files are clean, and `gen_docs.py --check` passes over the hand-edited docs row.
- **#180 carries both deferrals**, and it measures the repeated-parameter case rather than inheriting my probe: real joins the values and validates the pair, where Starlette hands over the last. That is a better finding than the one I left.

The `expand` call is the right one. A guessed `<p>…</p>` would have been the vendor's markup invented rather than reproduced, and the baseline entry says the true thing at no cost.

backlot/routers/atlassian.py
khj809: Fixed at `2ebcee5`, with your wording — the docstring now says `created` with an optional single leading `+`/`-`, case-insensitively and ignoring whitespace, and points at `_jira_order_desc` so the two cannot drift apart again. It was the pre-`_jira_order_desc` rule left standing next to the code that disproves it.

backlot/routers/atlassian.py
khj809: Split, in all three places rather than just the two you named — the third had the same seam.

`_jira_order_desc` is now 2026-09-10, and the clamps block above it stays 2026-09-09 and says so. The 400-before-404 comment in the handler had no date at all, which the file's own convention (`_jira_caller`: "Measured … on 2026-09-04") says it should; it now carries 2026-09-10.

The test block header splits the same way: 09-09 for the defaults, the clamps, the caps and `created`/`+created`/`-created` against `bogus`; 09-10 for the rest of the grammar, for the empty value being refused rather than read as absent, and for the check order.

PR #166, opened 2026-09-09
github: a HEAD is answered as the GET without its body, and the served spec states real's "(max 100)" on per_page (#151, #153)

### LINKED ISSUE #151: github: HEAD is a 405 on every route where real answers it as the GET without its body
**Source and commit** — `github`, at `bab49cb` (`main`; #144 and #147 do not touch this and their charset and version-header rules would apply to a HEAD the same way they apply to the GET).

Every GitHub route on the router is declared `GET` alone, and FastAPI's `APIRoute` does not add `HEAD` to a GET route the way Starlette's `Route` does, so a `HEAD` to any of the 32 routes is Starlette's 405 with `allow: GET`. Real answers a `HEAD` as the GET with the body left off: the same status, the same headers, `content-length` of the body the GET would have carried, `Link` where the GET has one, and nothing in the body. The S3 router already declares `HEAD` where S3 has it (`object_get`, `api_route(methods=["GET", "HEAD"])`); the GitHub router declares it nowhere.

| # | Request | Real | Backlot at `bab49cb` | Silent? |
|---|---|---|---|---|
| 1 | `HEAD /repos/{o}/{r}` | 200, `application/json; charset=utf-8`, no body | 405 | no |
| 2 | `HEAD /repos/{o}/{r}/issues?per_page=2` | 200, `content-length: 9038`, `Link` present | 405 | no |
| 3 | `HEAD /search/code?q=…&per_page=1` | 200, `application/json`, `Link` present | 405 | no |
| 4 | `HEAD /repos/{o}/{r}` (no such repo) | 404, `content-length: 132` | 405 | no |
| 5 | `HEAD /user` (no credential) | 401, `content-length: 120` | 405 | no |
| 6 | `HEAD /search/issues?q=` | 422, `content-length: 219` | 405 | no |
| 7 | `HEAD /search/code?q=…&per_page=abc` | 400, `text/plain; charset=utf-8`, `content-length: 75` | 405 | no |

None of this is in `backlot/fidelity/baseline/github.json`: the baseline is read off the OpenAPI document, real's spec declares no `head` operations, and Backlot's declares none either, so `backlot diff --source github` sees two documents that agree about a method neither mentions.

**How you know** — measured against api.github.com on 2026-09-07 with `curl -I`, each row against its GET the same minute.

- `/repos/psf/requests` and `/repos/psf/requests/issues?per_page=2`: 200 with the GET's headers, `content-length: 9038` on the listing and `Link` carrying `next` and `last`, the `x-ratelimit-*` headers present as on the GET, zero bytes of body.
- `/search/code?q=repo:psf/requests+def&per_page=1`: 200, `content-type: application/json` with no charset (the code search backend's header, as #146 records it for the GET), `Link` present.
- `/repos/psf/ghost-zz-9876`: 404, `content-type: application/json; charset=utf-8`, `content-length: 132`, which is the length of the GET's `Not Found` envelope.
- `/user` with no credential: 401, `content-length: 120`; `/search/issues?q=`: 422, `content-length: 219`; `/search/code?q=…&per_page=abc`: 400, `text/plain; charset=utf-8`, `content-length: 75`, the length of `Failed to deserialize query string: per_page: invalid digit found in string`. The errors answer `HEAD` exactly as they answer `GET`, body length included.
- Seven endpoints, one rule, no exception found, the code search backend included.

```python
import backlot, requests

with backlot.serve() as m:
    api = f"{m.base_url}/github"
    for url, headers in (
        (f"{api}/repos/{org}/pipeline", h),
        (f"{api}/repos/{org}/pipeline/issues?per_page=1", h),
        (f"{api}/repos/{org}/ghost-zz-9876", h),
        (f"{api}/user/repos", {}),
        (f"{api}/search/code?q=extension:md&per_page=1", h),
        (f"{api}/search/code?q=extension:md&per_page=abc", h),
    ):
        get, head = requests.get(url, headers=headers), requests.head(url, headers=headers)
        print(url.rsplit("/github/", 1)[1], "GET", get.status_code, "HEAD", head.status_code, head.headers.get("allow"))
```

Every line prints `HEAD 405 GET` beside a GET that is 200, 200, 404, 401, 200 and 400 in turn: the 405 does not depend on what the GET would have answered, because the method is refused before the route runs.

**What it costs a client** — an existence check. `requests.head(url)` and `curl -I` are how a script asks whether a repository, a file at `/contents/{path}` or an issue is there without downloading it, and how a client reads `Link` or `content-length` before committing to a GET; against real that is a 200 or a 404, against Backlot it is a 405 for both, so the check cannot tell them apart and a client that treats anything but 404 as "exists" reports every missing repo as present.

**Where it lands** — `backlot/routers/github.py`, at route registration rather than in any one handler: the 32 `@router.get` routes need `HEAD` alongside `GET`, and the body dropped, which Starlette already does when the method is declared (checked on FastAPI 0.141.1 / Starlette 1.6.0 in this environment: with `HEAD` added to an `APIRoute`'s `methods`, `HEAD` answers 200 with the GET's `content-length` and an empty body). `object_get` in `backlot/routers/s3.py` is the router that already does this by hand, for the one route S3 defines `HEAD` on; here it is every route, so a loop over `router.routes` after the definitions, or a `HEAD` handler in a middleware that runs the GET and discards the body, is the shape, and not 32 `api_route` decorators.

**What a fix has to settle**

- The OpenAPI document. Adding `HEAD` to a route's `methods` makes FastAPI write a `head` operation for it into `/openapi.json`, with a duplicate operation id (checked: the document for a one-route app gains `['get', 'head']` and FastAPI warns `Duplicate Operation ID`), where real's spec has no `head` operations at all. Left as is, that hands `backlot diff --source github` 32 new operations real lacks and `backlot mcp` 32 tools that do nothing a GET does not. The `HEAD` has to be served without being documented, which points at the middleware shape or at a hidden second route rather than at the methods set.
- Whether the answer is the GET's whole pipeline or only its headers. Real's `content-length` is the GET body's length on every row, so the body has to be computed to be measured, and a `HEAD` costs what the GET costs; that matches real and is the only way the number is right. The alternative, `HEAD` answered from the handler with no body and no length, is cheaper and wrong on the one header a client reads a `HEAD` for.
- The middlewares. `echo_github_api_version`, `resolve_github_id_paths` and #147's `vendor_json_media_type` see a `HEAD` the way they see a `GET`, so its headers follow theirs by construction; a test should say so for the version echo and the charset, and for the code search 400 keeping `text/plain` with `content-length: 75` on the bundled corpus's own body length.
- Whether the other routers are next. Jira, Confluence, Linear, Slack, HubSpot, Drive and Gmail are all `GET`-only in the same way and none of their `HEAD` answers is measured; this issue is the GitHub measurement only, and a fix that lands as a router-wide loop should be written so each vendor opts in once its own `HEAD` is.

### LINKED ISSUE #153: github: the served spec states no ceiling for `per_page`, where real states "(max 100)" in the parameter's own description
**Source and commit** — `github`, at `7ef30e4` (`github-page-size-30-100`, PR #152, where `b6046b7` declared the two page defaults and `7ef30e4` made the 30 the router's own constant). The runtime side of #145 is closed there and this is the half of the same spec line that is still open: the served document now declares `default: 30` / `default: 1`, and states neither parameter's ceiling, which real states in the parameter's own description.

Backlot's served spec is not decoration here — `backlot mcp` builds an agent's tool arguments from it (`/_meta/openapi/{source}`), and a generated client reads it for what it may send. A `per_page` that declares a default of 30 and no ceiling tells a client 30 is the shape and leaves 500 looking legal, when 500 comes back as 100.

| # | Divergence | Silent? |
|---|---|---|
| 1 | `per_page` carries no `description`, where real's states "(max 100)" — the only place real states the cap | yes |
| 2 | `page` carries no `description` | yes |
| 3 | The absence of a schema `maximum` is what makes `per_page=100_000` a 200, and nothing pins it | agrees today, unpinned |

**How you know** — GitHub's OpenAPI description, `descriptions/api.github.com/api.github.com.yaml`, `components/parameters`, read 2026-09-07:

```yaml
per-page:
  name: per_page
  description: The number of results per page (max 100). For more information, see "[Using pagination in the REST API](https://docs.github.com/rest/using-the-rest-api/using-pagination-in-the-rest-api)."
  in: query
  schema:
    type: integer
    default: 30
page:
  name: page
  description: The page number of the results to fetch. For more information, see "[Using pagination in the REST API](https://docs.github.com/rest/using-the-rest-api/using-pagination-in-the-rest-api)."
  in: query
  schema:
    type: integer
    default: 1
```

Sixteen of the seventeen routes Backlot pages reference those two shared parameters and none declares its own, so one description each covers the surface; the seventeenth is the legacy per-sha statuses read, `GET /repos/{owner}/{repo}/statuses/{sha}`, which the description does not declare — that path declares `post` alone, and `/commits/{ref}/statuses`, which does reference the two, names the read only in its prose ("This resource is also available via a legacy route: `GET /repos/:owner/:repo/statuses/:ref`."). Backlot serves it as an acknowledged extra, `backlot/fidelity/baseline/github.json`. The cap they describe is live, measured on api.github.com on 2026-09-07: `/orgs/github/repos?per_page=101` and `?per_page=500` each answer 100 items, as does `/search/issues?q=repo:psf/requests+timeout` at both sizes. So the ceiling is real and simply undeclared.

Note where real puts it: in the prose, over a bare `{type: integer}` with **no** `maximum`. That is not an oversight to improve on — a schema bound is the wrong tool twice over. Real serves `per_page=101` at the cap rather than refusing it (#145, and measured again above), so a declared `maximum` would invite a generated client to refuse what the server accepts; and `Query(le=100)` on the annotation cannot work at all here, measured on FastAPI 0.141.1 / pydantic 2.13: the runtime default is `None`, and an unsent value raises `TypeError: Unable to apply constraint 'le' to supplied value None` — a 500 on every unsent request.

**Reproduction** — the corpus is irrelevant; the spec is served over any:

```python
import backlot, requests

records = [
    {
        "source_type": "github",
        "doc_id": "gh-1",
        "repo": "wide",
        "title": "Issue 1",
        "content": "body",
        "visibility": "public",
        "state": "open",
        "created": "2026-01-02T03:04:05Z",
    }
]
with backlot.serve(records) as m:
    spec = requests.get(f"{m.base_url}/openapi.json").json()
    params = spec["paths"]["/github/repos/{owner}/{repo}/issues"]["get"]["parameters"]
    for name in ("page", "per_page"):
        p = next(x for x in params if x["name"] == name)
        print(name, "description:", p.get("description"), "| schema:", p["schema"])
```

prints

```
page description: None | schema: {'anyOf': [{'type': 'integer'}, {'type': 'null'}], 'title': 'Page', 'default': 1}
per_page description: None | schema: {'anyOf': [{'type': 'integer'}, {'type': 'null'}], 'title': 'Per Page', 'default': 30}
```

where real declares each one's description above. (The `anyOf` in place of real's bare `{type: integer}` is deliberate and older than this: `backlot/pagination.py::_absorb_page` keeps the parameter an integer in the contract while absorbing an unparseable value at runtime, which is what real does.)

**Where it lands** — `backlot/openapi.py::github_page_defaults(spec, defaults)`, which already walks exactly the two parameters this concerns and takes its numbers from its caller, `backlot/main.py`; and `tests/test_github.py::test_github_the_spec_declares_reals_page_defaults`, which already asserts on both. The annotation, `backlot/pagination.py::PageParam`, is the other place it could go.

**What a fix has to settle**

- **Where.** Two routes, each measured green at the commit it was built on:
  - the hook `b6046b7` added, which is one line in its loop and no churn elsewhere. It already takes the numbers it writes from `main.py`, so the prose can arrive the same way — a second mapping beside `{"per_page": github.PER_PAGE_DEFAULT, "page": 1}`, or one mapping of name to `(default, description)`:

    ```python
                    param["schema"]["default"] = default
                    param["description"] = descriptions[param["name"]]
    ```

    Measured at `2942372` with the mapping held in `openapi` rather than passed in, which is the shape that commit had: the description lands at parameter level, where real has it, and reaches the MCP slice (`/_meta/openapi/github` → `/github/search/code` → `per_page` → "The number of results per page (max 100).", `default` 30). Suite 2570 passed, 2 skipped; `ruff check` clean.
  - the annotation, via `Query(description=...)`, measured at `637e69f`, before the hook existed. This needs a second `Annotated` type, since the two parameters' descriptions differ — a `page_param(description)` factory in `backlot/pagination.py` plus a `PerPageParam` built in the router from `PER_PAGE_MAX`, and `per_page: PageParam` → `PerPageParam` at 17 signatures. Also verified green (suite passed, ruff clean, and `per_page` 101/500 → 100 and `abc`/0/-3 → 30 all still 200), and mutation-checked: reverting one route to `PageParam` fails on `assert 'max 100' in 'The page number of the results to fetch.'`. Larger than the hook now that the hook exists, and it puts a second mechanism where one already does the job.
- **Whose prose.** Real's description verbatim, tail included ("For more information, see "[Using pagination in the REST API](…)."), or Backlot's own concise line — which is what `backlot/routers/s3.py:57` does for `max-keys`. Verbatim is the fidelity answer; the repo has precedent for either.
- **One source for each number.** A description carrying "(max 100)" states 100 a second time, next to `github.PER_PAGE_MAX` — the duplication PR #152 removed from sixteen call sites, and the one `7ef30e4` removed from the defaults after review. The pattern to follow is that fix: `main.py` holds the router's constants and passes them in, so the description can be built from `PER_PAGE_MAX` there rather than spelled out in `openapi`, which imports no router (six routers import it for `qp`).
- **Pinning the absent bound.** `assert not {"maximum", "exclusiveMaximum"} & set(p["schema"])` belongs beside the default assertions: it is the declaration that keeps `per_page=100_000` a 200, and today it holds only because nobody has written one.

nuri-yoo:
Closes #151 and closes #153. On `main` at `1246b35` and merged forward twice (`832a8da` to `364e071`, `22f2dbd` to `165e019`), the change is six commits, one per issue and then their neighbours: `cfe8328` the spec's page-parameter descriptions (#153), `5c30f7e` the `HEAD` answer (#151), `32ac135` Backlot's own routes and `f130236` the `state` parameter (the two sections before the review), `cc2b05d` prose only and `eb6e0ce` one assertion tightened. The 2026-09-10 review adds three: `e744ee4` and `9a98090`, its two suggestions applied as written, and `1f60a3b`, the docstrings and docs the first of them made stale (the section before Verification). None adds or removes an operation in the OpenAPI document, so `backlot diff --source github` compares the paths, methods and parameter names it compared before.

**What GitHub does.** Two facts, one per issue. A `HEAD` on a GitHub route is the `GET` with the body left off: the same status and headers, `content-length` of the body the GET would have carried, `Link` where the GET has one, and nothing in the body, measured against api.github.com on 2026-09-07 with `curl -I` on seven endpoints beside their GETs (#151's table): the 200s of `/repos/psf/requests`, its issue listing and `/search/code`, the 404 for a repository that does not exist at `content-length: 132`, the 401 for `/user` with no credential at 120, the 422 for a blank search `q` at 219 and code search's `text/plain` 400 at 75. And GitHub's OpenAPI description declares the shared `per-page` and `page` parameters each with a description on the parameter itself, `per_page`'s being "The number of results per page (max 100). For more information, see …", over a bare `{type: integer, default: 30}` with no `maximum` (`components/parameters`, github/rest-api-description, read 2026-09-09, the same text #153 quotes from its 2026-09-07 read). That description is the only place real states the cap, and the absence of a schema bound is deliberate on real's side: it serves `per_page=101` at the cap rather than refusing it (#145, #152), so a `maximum` would have a generated client refuse what the server accepts. The same description declares no `head` operation on any path (0 in the 2026-09-09 read).

**What Backlot did at `1246b35`.** Every GitHub route is declared `GET` alone and FastAPI's `APIRoute` does not add `HEAD` to a GET route the way Starlette's `Route` does, so a `HEAD` to any of the 32 routes was Starlette's 405 with `allow: GET`, whatever the GET would have answered (#151's reproduction on the bundled corpus: `HEAD 405` beside GETs of 200, 200, 404, 401, 200 and 400). The served document declared `default: 30` and `default: 1` (#152) and no description on either parameter, so a client or an agent reading the MCP slice saw 30 as the shape and nothing to say that 500 comes back as 100.

**What changed for #153.** `openapi.github_page_defaults(spec, defaults)` is `openapi.github_page_parameters(spec, parameters)`, name to `(default, description)`, and writes the description onto the parameter beside the default it already wrote into the schema, on GitHub's operations alone as before. The mapping is the router's: `github.PAGE_PARAMETERS`, beside `PER_PAGE_DEFAULT` and `PER_PAGE_MAX`, its `per_page` text built from `PER_PAGE_MAX` and its default from `PER_PAGE_DEFAULT`, so the document cannot state one number while the route applies another; `main.py` passes it through unchanged and `openapi` still imports no router. Real's text verbatim, tail included, which is the fidelity answer #153 named first; `s3.py`'s own line for `max-keys` stays the other precedent. No `maximum` is written, and the test pins that.

**What changed for #151.** A middleware in `backlot/main.py`, `answer_head_as_the_get_without_its_body`, registered after `resolve_github_id_paths` and before the Jira ones, so `vendor_json_media_type` stays the outermost as its docstring says. For a `HEAD` under `/github` it rewrites the scope's method to `GET` before routing, so the GET runs in full (the router's dependencies, the handler, the version echo and the id-path rewrite see a GET, and the charset middleware outside it rewrites the copied `content-type` as it does the GET's), reads the body to the end to measure it, and answers the GET's status and headers with `content-length` set to that length and no body. Not `HEAD` in each route's `methods`, which #151 measured writes a `head` operation into `/openapi.json` per route under a duplicate operation id; the document is untouched, and `drop_head_operations`' "S3 is the only source with HEAD routes" stays true. The body is read and not sent by the middleware, and the method goes back to `HEAD` on the scope once the GET has answered (the review's first suggestion, `e744ee4`): uvicorn's httptools protocol frames the response by the scope's method, and with it left saying `GET` the empty body under the GET's `content-length` was `RuntimeError: Response content shorter than Content-Length` in the server log and a reset connection for the client's next request, which no TestClient test can see, since Starlette's `TestClient` drops a HEAD body itself and frames nothing (`testclient.py`, `if request.method != "HEAD"`). That is why the test reads the ASGI messages directly for one cell. GitHub alone among the vendors, by a prefix tuple, `_HEAD_IS_THE_GET_WITHOUT_ITS_BODY`, that a vendor joins once its own `HEAD` is measured, which is the opt-in #151's last bullet asked for (Backlot's own `/health` and `/_meta` join it in `32ac135`, below, for a reason of their own); a Slack `HEAD` stays the 405 it was, asserted.

**What #151 asked a fix to settle.** The OpenAPI document: served without being documented, by the middleware; `/openapi.json` carries no `head` under `/github`, asserted. The GET's whole pipeline or only its headers: the whole pipeline, the body computed to be measured, so `content-length` is the GET body's length on every row and a `HEAD` costs what its GET costs, as on real. The middlewares: the version echo, the id-path rewrite and the charset are each asserted on a `HEAD`, and code search's 400 keeps `text/plain; charset=utf-8` at `content-length: 75` with no version echo. The other routers: opt-in by prefix, none added.

**What #153 asked a fix to settle.** Where: the hook, as the issue's first route and its measurement at `2942372` had it, one mapping of name to `(default, description)`; the annotation route is not taken. Whose prose: real's, verbatim. One source for each number: the mapping lives in the router beside the constants it reads, so 100 and 30 are each spelled once (`PER_PAGE_MAX`, `PER_PAGE_DEFAULT`) and `main.py` passes the mapping rather than building prose of its own. Pinning the absent bound: `_schema_bounds(schema)` in the test returns any `maximum` or `exclusiveMaximum` at the schema's top level or in an `anyOf` branch, which is where FastAPI puts the integer half of `int | None`, and the test asserts it empty on all 34 parameters.

**Tests.** `test_github_the_spec_declares_reals_page_defaults` is `test_github_the_spec_declares_reals_page_parameters`: on all 34 `page`/`per_page` parameters across the seventeen routes that page, the default, the integer type, real's description verbatim and no bound; that the `per_page` prose contains `(max {PER_PAGE_MAX})` and the default is `PER_PAGE_DEFAULT`; Slack's `page` untouched; and the MCP slice's `/search/code` `per_page` carrying the default and the description. One new test for #151, `test_github_a_head_is_the_get_with_the_body_left_off`, ten `HEAD`s beside their `GET`s on the test corpus: a repository, the cursor-paged issue listing at `per_page=1` (its `Link` and version echo asserted present), `/contents/README.md` as `application/vnd.github.raw`, the id-keyed `/repositories/{id}`, the 404, the 401 with no credential, the blank-`q` 422, code search's 200 and its `per_page=abc` 400 (`text/plain; charset=utf-8`, `content-length: 75`, no version echo), and the version 400; each answers the GET's status, `content-type`, `Link` and version header with an empty body and the GET's `content-length`, which is `len(get.content)`. Then the ASGI-layer cell, that `/openapi.json` has no `head` under `/github`, and that `HEAD /slack/api/auth.test` is still 405. `test_github_a_wrong_method_is_not_dressed_as_a_measured_answer`'s docstring says `HEAD` is no longer a wrong method.

Mutation-checked, each from the committed tree and restored with `git checkout`: the description line removed fails the spec test on the first parameter (`KeyError: 'description'`); a `maximum: 100` written into the integer branch fails it on the bound; `PER_PAGE_MAX = 50` fails it on `(max 50)` against real's text and fails the page-size test too; the prefix tuple emptied fails the HEAD test on its first row (`assert 405 == 200`); the middleware passing the GET response through fails it at the ASGI cell and nowhere else (`assert 1853 == 0`, the repository body's length on the test corpus), which is that cell's reason to exist; the three source files at `main` with the tests kept fail the spec test on `description` and the HEAD test on `405 == 200`.

**Backlot's own routes, `32ac135`.** Two things found while writing the HEAD test. `/health`, `/_meta/users`, `/_meta/credentials` and `/_meta/openapi/{source}` read `app.state` and `app.openapi()` off the module-level name rather than off `request.app`. That shows when `backlot.main` is re-imported while a first server is still up, which the test helper's `client_for(reload=True)` does: a reload re-executes the module in the same namespace, so the first app's handlers found `app` rebound to the second one, and once the second lifespan had ended and closed its connection, the first server's `/_meta/users` was `sqlite3.ProgrammingError: Cannot operate on a closed database` while its vendor routes, which read `request.app`, kept answering. In the suite this bit a module-scoped client whose first `/_meta/users` call came after a reload, which a `-k` subset of `tests/test_github.py` arranges (the page-size test reloads) and the full loadfile run happened not to. The four handlers read the app off the request now, and `test_backlots_own_routes_read_the_app_that_serves_them` holds a first server up across a reload and reads all four afterwards (mutation: the handlers back on the module-level `app` fail it with that `ProgrammingError`). And the same routes answered a `HEAD` with the 405 the GitHub routes did, for the same `APIRoute` reason; they are Backlot's own, with no vendor to measure against, and `HEAD /health` is the shape a liveness probe takes, so `/health` and `/_meta` join the prefix tuple the middleware answers, with the comment saying which prefix is there for which reason. `test_backlots_own_routes_answer_head_as_the_get` compares the four against their GETs, the 404 for an unknown source included, with the warm-up thread joined first because `/health`'s body grows once the counts land (mutation: the tuple back to `("/github",)` fails it on `/health`, `assert 405 == 200`).

**`state` on the two listings, `f130236`.** The sibling of #153 one parameter over. GitHub's description declares `state` on the issue and the pull listing as `{type: string, enum: [open, closed, all], default: open}`, described "Indicates the state of the issues to return." on the one and "Either `open`, `closed`, or `all` to filter by state." on the other (read 2026-09-09); Backlot declared `{type: string, default: open}`, no enum, no description. Outside the enum the two routes part, measured on api.github.com on 2026-09-09 against `psf/requests`: the issue listing refuses `state=bogus`, `state=OPEN` and `state=` (empty) with a 422, `Validation Failed`, one `errors` entry carrying the value sent with `resource: Issue`, `field: state`, `code: invalid`, under `documentation_url` `https://docs.github.com/v3/issues/#list-issues`, which is not the route's anchor in `ROUTE_DOCS`; the repository is checked first, so the same value on `psf/ghost-zz-9876` is the 404 with the route's own anchor. The pull listing does not refuse: `state=bogus` and `state=OPEN` answer the 87 rows `state=open` answers, where `closed` and `all` answer other, larger sets. Backlot filtered on the value as sent and answered an empty 200 on both routes, which is neither answer. So: `ISSUE_STATES` and the two descriptions beside the handlers, the enum and description written into each route's schema through `Query(json_schema_extra=...)` rather than a `Literal` (a `Literal` has FastAPI refuse before the handler, with one answer for both routes), `_invalid_issue_state(value)` raising real's 422 as the seventh hand-shaped envelope (`errors.github.http_body`'s docstring counts it and names its two reasons), and the pull listing applying `open` for a value it does not know. `test_github_state_is_reals_enum_refused_on_issues_and_absorbed_on_pulls` builds a corpus with an open and a closed pull, since no repository in the sample holds both, and holds the two schemas to real's declarations, the three values to the 422 body and to the open set, and the missing repository to its 404 (mutations: the raise removed, `assert 200 == 422`; the pull listing filtering on the value as sent, `assert [] == [...]`; the enum left out of the schema, `KeyError: 'enum'`).

**Review, 2026-09-10.** Five inline notes and one on the body, each re-measured before it was taken. The first was a blocker the suite could not see: with the scope's method left as `GET` after the GET ran, uvicorn's httptools protocol counted the declared `content-length` against the empty body, logged `RuntimeError: Response content shorter than Content-Length` and reset the connection, so a `requests.Session` that sent a `HEAD` had its next request fail with `ConnectionResetError` (measured at `832a8da` over uvicorn on a one-record corpus; with the suggestion applied the next request is a 200 and the log is clean). Applied as written in `e744ee4`, and the docstring and the test's ASGI comment, which had said the server could not drop the body, say in `1f60a3b` what the framing did instead. The second, reST double backticks on two `state` tokens in `errors.github`, applied as written in `9a98090`. The third, passing `response.background` through, is not taken: `BaseHTTPMiddleware`'s `_StreamingResponse` sets `background = None` unconditionally (starlette 1.6.0, `middleware/base.py:218`) and the handler's own `BackgroundTask` runs inside the wrapped app, so it already runs on a `HEAD` (one `GET` and one `HEAD` through a handler with a recording task leave two entries), and the line would pass a constant `None`; answered in the thread. The fourth and fifth, the two docs files outside the diff, are in `1f60a3b`: `docs/fidelity.md` gains the "Served, not declared" paragraph beside "Declared, not served", and `docs/supported-sources.md`'s GitHub table records `state` on the two listings, with two words added to the proposed rows, "on the listing", since the `[/{n}]` routes take no `state`. The body note, that Verification predated the merge, is answered below.

**Verification.**

```
2862 passed, 4 skipped   (2026-09-10, at 1f60a3b with main merged to 165e019; the skips are Docker, llama-index's hubspot reader and mirage's google client, none touched here)
ruff format --check .   139 files already formatted
PYTHONPATH=. python scripts/gen_docs.py --check   exit 0
```

The two issues' reproductions over uvicorn: #151's prints `HEAD 200`, `200`, `404`, `401`, `200`, `400` beside the same GETs, where it printed `HEAD 405` on every line at `1246b35`; #153's prints real's two descriptions where it printed `None`.

**Not in this PR.** Two neighbours measured on the way and left, each a PR of its own size rather than a line here, filed as #167 and #168 with the measurements. `sort`, `direction` and `type`, which real declares on the issue and pull listings and on `/orgs/{org}/repos` and `/user/repos`, Backlot neither declares nor honours; `?sort=bogus` on the issue listing is a 200 on real (2026-09-09), and what each value orders by, and which direction each defaults to, is the measurement a fix needs. And real sends `x-ratelimit-*` on a `HEAD` as on a `GET`; Backlot sends them on neither, and the values (the unauthenticated and authenticated limits, `used`, `reset`) are their own measurement.


backlot/main.py
```diff
@@ -208,6 +207,59 @@ async def resolve_github_id_paths(request: Request, call_next):
...
+    through here would go out with its body after the headers (measured over uvicorn on the bundled
+    corpus with a raw socket: the repository's 1839 bytes followed the headers with the body left
+    in, none with it read here).
+    """
+    if request.method != "HEAD" or not request.url.path.startswith(
+        _HEAD_IS_THE_GET_WITHOUT_ITS_BODY
+    ):
+        return await call_next(request)
+    request.scope["method"] = "GET"
+    response = await call_next(request)
```
khj809, line 264: The scope's method is left as `GET` after the GET has run, so uvicorn's httptools protocol — the default under `uvicorn[standard]` — counts the `content-length` this middleware declares against the zero body bytes it sends and raises `RuntimeError: Response content shorter than Content-Length` once the headers are already on the wire, which drops the connection on every `HEAD`.

```suggestion
    response = await call_next(request)
    request.scope["method"] = "HEAD"
```

backlot/main.py
```diff
@@ -208,6 +207,59 @@ async def resolve_github_id_paths(request: Request, call_next):
...
+    length = 0
+    async for chunk in response.body_iterator:
+        length += len(chunk)
+    head = Response(status_code=response.status_code, headers=response.headers)
```
khj809, line 269: `response.background` is dropped when the GET's response is rebuilt, so a `BackgroundTask` a handler attached would run on the `GET` and silently not on the `HEAD`. Nothing under `backlot/` sets one today, so this is pre-emptive rather than a live bug — but the asymmetry is invisible if one ever arrives.

```suggestion
    head = Response(
        status_code=response.status_code,
        headers=response.headers,
        background=response.background,
    )
```

backlot/errors/github.py
```diff
@@ -177,12 +177,13 @@ def docs_url(path: str) -> str:
...
     sites do, each for one of three reasons no generic rendering could supply: an ``errors`` member
-    (the version 400 and the two Validation Failed search 422s); a ``documentation_url`` that is not
-    the route's anchor in :data:`ROUTE_DOCS` (the two depth 422s, the issue search's the bare
-    ``/v3/search/`` with its trailing slash and code search's the ``#search-code`` anchor, and the
-    ref 404 the contents routes raise, whose is ``/v3/repos/contents/``); or a ``message`` the
+    (the version 400, the two Validation Failed search 422s and the issue listing's `state` 422); a
+    ``documentation_url`` that is not the route's anchor in :data:`ROUTE_DOCS` (the two depth 422s,
+    the issue search's the bare ``/v3/search/`` with its trailing slash and code search's the
+    ``#search-code`` anchor, the ref 404 the contents routes raise, whose is ``/v3/repos/contents/``,
+    and that `state` 422 again, whose is ``/v3/issues/#list-issues``); or a ``message`` the
```
khj809: These two are the only single-backtick tokens in a docstring that spells every other identifier with reST double backticks — ``errors``, ``documentation_url``, ``/v3/search/``, ``#search-code`` all in the same paragraph.

```suggestion
    (the version 400, the two Validation Failed search 422s and the issue listing's ``state`` 422);
    a ``documentation_url`` that is not the route's anchor in :data:`ROUTE_DOCS` (the two depth
    422s, the issue search's the bare ``/v3/search/`` with its trailing slash and code search's the
    ``#search-code`` anchor, the ref 404 the contents routes raise, whose is ``/v3/repos/contents/``,
    and that ``state`` 422 again, whose is ``/v3/issues/#list-issues``); or a ``message`` the
```

backlot/main.py
```diff
@@ -208,6 +207,59 @@ async def resolve_github_id_paths(request: Request, call_next):
     return await call_next(request)
 
 
+# The path prefixes whose `HEAD` is the GET with the body left off. GitHub because it is measured to
+# be, and none of the other vendors' `HEAD` answers is, so a vendor is added here once its own is
+# rather than by a rewrite that assumes they share GitHub's. `/health` and `/_meta` are Backlot's
+# own routes, with no vendor to measure against: a `HEAD /health` is the shape a liveness probe
+# takes, and `FastAPI`'s `APIRoute` refused it with the same 405 for the same reason.
+_HEAD_IS_THE_GET_WITHOUT_ITS_BODY = ("/github", "/health", "/_meta")
```
khj809, line 222: Documentation drift, `docs/fidelity.md`. This tuple creates a category the Coverage section does not have: **served, not declared**. #151 named it exactly — "the baseline is read off the OpenAPI document, real's spec declares no `head` operations, and Backlot's declares none either, so `backlot diff --source github` sees two documents that agree about a method neither mentions" — and closing the issue without writing that down leaves a served surface the path diff structurally cannot see, with no record of why it is allowed to be invisible. The section already spends a paragraph on the inverse case ("Declared, not served"), so the shape is there to follow.

The file is not in this diff, so no `suggestion` block. Proposed, as a new paragraph after the "Declared, not served" one (`docs/fidelity.md:238`):

```markdown
Served, not declared: a `HEAD` under `/github`, `/health` or `/_meta` is answered as the `GET` with
the body left off (`backlot.main.answer_head_as_the_get_without_its_body`), and no `head` operation
is written for it, because real's own description declares none either. Real GitHub answers a `HEAD`
that way on every route measured, so the divergence was the method missing, not the method being
undocumented — but a path diff reads methods off the two documents, and neither mentions this one,
so nothing here would catch it going away. The middleware's prefix tuple is the record of which
vendors it covers; a vendor joins it once its own `HEAD` is measured.
```

backlot/routers/github.py
```diff
@@ -1077,6 +1139,8 @@ async def list_issues(
     caller = _require(request)
     ids = auth.visible_ids(request, caller)
     _require_repo(conn, repo, ids)
+    if state not in ISSUE_STATES:
```
khj809, line 1142: Documentation drift, `docs/supported-sources.md`. The GitHub table's `Notes` column is where route-specific quirks are recorded — `repos/{o}/{r}/commits/{sha}` carries "a ref naming no commit is real's 422" a few rows down — and `state` splitting into a 422 on the issue listing and a silent fall back to `open` on the pull listing is exactly that kind of thing a caller trips over. Both rows are empty today (`docs/supported-sources.md:78` and `:81`), and this is the only user-visible behaviour change in the PR that a reader cannot find from the docs.

The file is not in this diff, so no `suggestion` block. Proposed, for those two rows:

```markdown
| `repos/{o}/{r}/issues[/{n}]` | `state`: `open`\|`closed`\|`all`; any other value is real's 422, and the repository is checked first, so an unknown repo is the 404 instead |
| `repos/{o}/{r}/pulls[/{n}]` | `state`: `open`\|`closed`\|`all`; any other value is served as `open` — real refuses it on the issue listing and absorbs it here |
```

khj809 (commented):
Reviewed against the two linked issues, with the measurements re-run rather than taken on trust.

**#153 is fully settled.** All four of the issue's "what a fix has to settle" points land: the hook route, real's prose verbatim, one source for each number (`PAGE_PARAMETERS` reads `PER_PAGE_MAX`/`PER_PAGE_DEFAULT`, so 100 is spelled once), and `_schema_bounds` pinning the absent ceiling through the `anyOf` branch.

**#151's design is right** — the document is untouched, the whole GET pipeline runs so `content-length` is the GET body's length, and the prefix tuple is the opt-in the issue asked for — but the middleware has a blocker on the wire that the suite cannot see, because `TestClient` never exercises server framing. See the inline comment.

Re-measured against api.github.com today, all three confirmed: `HEAD /repos/psf/requests` → 200 at `content-length: 6676`, equal to the GET's 6676 bytes; `issues?state=bogus` and `?state=` → the exact 422 envelope and `documentation_url` this PR hard-codes; `pulls?state=bogus` and `?state=` → the `state=open` set. `backlot diff` really is untouched: `fidelity/operations.py` reduces an operation to (method, path, parameter names), so descriptions, enums and defaults are not compared.

One item with no diff line to anchor an inline comment to: **the Verification block is from before the merge commit.** At `832a8da` the suite is `2649 passed, 3 skipped` — the two extra skips are the optional llama-index and mirage extras — not the `2637 passed, 1 skipped` the body records. `ruff check`, `ruff format --check` and `PYTHONPATH=. python scripts/gen_docs.py --check` are all clean at that commit.

backlot/main.py
nuri-yoo: Measured rather than applied, and the task already runs on the `HEAD`. Under `BaseHTTPMiddleware` the response `call_next` hands back is a `_StreamingResponse` whose `__init__` sets `self.background = None` unconditionally (starlette 1.6.0, `middleware/base.py:218`), so `background=response.background` would pass `None` on every call. The handler's own `BackgroundTask` is not lost either: the inner `Response.__call__` runs it inside the wrapped app after streaming its body, which is inside the middleware's `call_next`, not after it. Checked with a FastAPI app carrying this middleware and a `GET /x` returning `JSONResponse(..., background=BackgroundTask(record))`: one `GET` then one `HEAD` leaves two entries in `record`'s list, and the middleware sees `response.background is None` on the `HEAD`. So the asymmetry does not exist on this stack, and the line would be a constant `None`; leaving it out.

The first suggestion is applied as-is (`e744ee4`), and it was a real blocker: at `832a8da`, a `HEAD` over uvicorn's httptools protocol logged `RuntimeError: Response content shorter than Content-Length` and reset the connection, so a `requests.Session` that sent a `HEAD` had its next request fail with `ConnectionResetError`; with the method restored the next request is a 200 and the log is clean. My raw-socket measurement had counted the body bytes and missed the close. The docstring and the test's ASGI comment said the server could not drop the body; they say the opposite now, with that measurement.

backlot/main.py
nuri-yoo: Added in `1f60a3b`, as proposed, after the "Declared, not served" paragraph in `docs/fidelity.md`; one edit to the wording, the em dash in "undocumented — but a path diff" is a full stop, the rest is the text as written.

backlot/routers/github.py
nuri-yoo: Both rows are in `1f60a3b`, with two words added: "`state` on the listing", since `[/{n}]` covers the single-item routes too and those take no `state`. The pull row reads "any other value is served as `open`, which is what real does here where the issue listing refuses it"; the issue row is as proposed.

backlot/main.py
khj809: You're right. Sorry for the noise on this one — I read `middleware/base.py:218` after your reply and `_StreamingResponse` takes no `background` argument at all, so the line would have passed a constant `None`; and a recording task on the handler fires once for the `GET` and once for the `HEAD` through this middleware, so nothing was being lost. Withdrawn.

khj809 (approved):
Approving at `1f60a3b`. The blocker is settled on the wire: over uvicorn's httptools protocol a `requests.Session` sends six `HEAD`s in a row without a reset, the access log records the `HEAD`, and there is no `RuntimeError` left in the server log. The docstring and the two docs files now carry the reason. Suite here is `2863 passed, 3 skipped` (Docker was up, so the mcp test ran), `gen_docs.py --check` exit 0, ruff check and format clean.

PR #175, opened 2026-09-10
github: the four listings order and filter by sort, direction, type and visibility as measured, and every answer carries real's five x-ratelimit-* headers with GET /rate_limit behind them (#167, #168)

### LINKED ISSUE #167: github: sort, direction and type are undeclared and ignored on the four listings that take them, so every order and filter a client asks for comes back as the store's
**Source and commit** — `github`, at `1246b35` (`main`; #166 declares `state` on the issue and pull listings and touches none of the parameters here).

Real orders the issue, pull and repository listings by `sort` and `direction`, filters the two repository listings by `type`, and declares each of the three in its OpenAPI description with an enum and a default. Backlot declares none of them, reads none of them, and answers every request in the one order its store yields, with a 200: a client asking for the most recently updated issues, the oldest pulls or an organization's forks gets the rows it gets asking for nothing.

| # | Request | Real | Backlot at `1246b35` | Silent? |
|---|---|---|---|---|
| 1 | `GET /repos/{o}/{r}/issues?sort=updated` | most recently updated first (7620, 7619, 7610) | the unsent order | yes |
| 2 | `GET /repos/{o}/{r}/issues?direction=asc` | oldest first (1, 2, 3) | the unsent order | yes |
| 3 | `GET /repos/{o}/{r}/issues?sort=comments` | another order (2966, 5797, 1573) | the unsent order | yes |
| 4 | `GET /repos/{o}/{r}/pulls?sort=popularity`, `?sort=long-running` | 10, 15, 42 and 3335, 3443, 3900 | the unsent order | yes |
| 5 | `GET /repos/{o}/{r}/pulls?sort=created` | oldest first (4, 8, 10), where no `sort` at all answers newest first (7619, 7616, 7609) | the unsent order | yes |
| 6 | `GET /orgs/{org}/repos?type=forks` | the forks alone | every repository | yes |
| 7 | `GET /orgs/{org}/repos?sort=full_name`, `?sort=pushed` | `.github`, `advisory-database`, `black` and `advisory-database`, `cachecontrol`, `requests` | the unsent order | yes |
| 8 | `GET /user/repos?sort=created` | newest first | the unsent order | yes |
| 9 | the served spec, four routes | `sort`, `direction` and `type` each `{type: string, enum: […], default: …}` with a description; `visibility` and `affiliation` on `/user/repos` besides | `state`, `page`, `per_page` alone | yes |

`backlot/fidelity/baseline/github.json` acknowledges `sort`, `direction` and `type` on all four routes and `visibility` and `affiliation` on `/user/repos` as `missing_param` gaps ("the vendor accepts it; Backlot does not"), so `backlot diff --source github` has been reporting the declarations; what the listing answers to a value it does accept is not something the diff sees.

**How you know** — measured against api.github.com on 2026-09-09, authenticated (`gh api`), three rows a page, against `psf/requests`, the `psf` organization and the token's own repositories; the description read from github/rest-api-description the same day.

- The description: issues `sort` `enum: [created, updated, comments]`, `default: created`, and `direction` `enum: [asc, desc]`, `default: desc`; pulls `sort` `enum: [created, updated, popularity, long-running]`, `default: created` ("`popularity` will sort by the number of comments. `long-running` will sort by date created and will limit the results to pull requests that have been open for more than a month and have had activity within the past month.") and `direction` with no default and the prose "Default: `desc` when sort is `created` or sort is not specified, otherwise `asc`."; organization repositories `type` `enum: [all, public, private, forks, sources, member]`, `default: all`, `sort` `enum: [created, updated, pushed, full_name]`, `default: created`, `direction` "Default: `asc` when using `full_name`, otherwise `desc`."; `/user/repos` `sort` the same enum with `default: full_name`, `type` `enum: [all, owner, public, private, member]`, `visibility` `enum: [all, public, private]`, `affiliation` `default: owner,collaborator,organization_member`, and on `type` the sentence "Will cause a `422` error if used in the same request as **visibility** or **affiliation**." (that 422 is not measured here).
- Issues: no `sort` and `sort=created` both answer 7620, 7619, 7618 (created 2026-09-08, 09-07, 09-05); `sort=updated` 7620, 7619, 7610 (updated 09-08, 09-07, 09-06); `sort=comments` 2966, 5797, 1573, whose `comments` members are 211, 105 and 122, so the order is not by that member alone; `direction=asc` and `sort=created&direction=asc` 1, 2, 3; `sort=bogus`, `direction=bogus` and `sort=updated&direction=bogus` each answer the unsent order, the last of them not `updated`'s.
- Pulls: no `sort` 7619, 7616, 7609 (newest first); `sort=created` 4, 8, 10 (oldest first), so the description's "`desc` when sort is `created`" is the unsent case on the wire and not the explicit one; `sort=updated` 519, 929, 1350 (updated 2014-06-18, 06-18, 06-19); `sort=popularity` 10, 15, 42, each `comments: 0`; `sort=long-running` 3335, 3443, 3900 (created 2016-06-15 on); `direction=asc` 4, 8, 10; `sort=bogus` 4, 8, 10, the explicit `sort=created` order; `direction=bogus` the unsent order.
- Organization repositories: no `sort` `requests`, `cachecontrol`, `pyperf` (created 2011, 2013, 2016); `sort=created` `wiki`, `blog`, `organizer-toolkit` (2026-02-17, 2026-02-13, 2025-08-11), so here too the unsent order is not the explicit `created`'s; `sort=pushed` `advisory-database`, `cachecontrol`, `requests` (pushed 2026-09-09, 09-07, 09-07); `sort=full_name` `.github`, `advisory-database`, `black`, and with `direction=desc` `wiki`, `webassembly`, `user-success-wg`; `type=forks` `matterbridge-configuration`, `httpbin`, `plausible-analytics`, each `fork: true`; `type=bogus` and `direction=bogus` the unsent order; `sort=bogus` the `sort=created` order.
- The token's repositories: no `sort` an order that is not `full_name`'s (`bsk`, `bsk_check`, `agent-k`), where the description declares `default: full_name`; `sort=created` newest first; `sort=bogus` the `sort=created` order; `type=bogus` and `visibility=bogus` the unsent order.
- Every request above is a 200: no value of `sort`, `direction` or `type` was refused on any of the four routes.

```python
import backlot, requests

with backlot.serve() as m:
    api = f"{m.base_url}/github"
    for path, params in (
        (f"repos/{org}/pipeline/issues", {"state": "all"}),
        (f"repos/{org}/pipeline/issues", {"state": "all", "sort": "updated"}),
        (f"repos/{org}/pipeline/issues", {"state": "all", "direction": "asc"}),
        (f"repos/{org}/pipeline/pulls", {"state": "all", "sort": "long-running"}),
        (f"orgs/{org}/repos", {"type": "forks"}),
        (f"orgs/{org}/repos", {"sort": "full_name", "direction": "desc"}),
    ):
        r = requests.get(f"{api}/{path}", headers=h, params=params)
        print(path, params, r.status_code, [x.get("number") or x.get("name") for x in r.json()])
    spec = requests.get(f"{m.base_url}/openapi.json").json()
    issues = spec["paths"]["/github/repos/{owner}/{repo}/issues"]["get"]["parameters"]
    print([p["name"] for p in issues if p["in"] == "query"])
```

The three issue lines print the same eight numbers (287, 2317, 9597, 20536, 30648, 47516, 75342, 87703), the pull line its three whatever `sort` says, the two repository lines both `['docs-site', 'pipeline']`, and the spec `['state', 'page', 'per_page']`.

**What it costs a client** — the "what changed" query. A client fetching `?sort=updated&per_page=5` to see what moved gets the store's first five rows, moved or not; PyGithub's `repo.get_issues(sort="updated", direction="desc")` and `org.get_repos(type="forks")` return the same rows with those arguments as without them, so a client's ordering and filtering logic passes here on rows it never sorted or filtered, and a paginator walking `?sort=created&direction=asc` from the beginning of a repository's history starts wherever the store starts. An agent reading the MCP slice sees no such argument and cannot ask.

**Where it lands** — `backlot/routers/github.py`: `list_issues`, `list_pulls`, `list_repos` and `list_user_repos`, each declaring the parameters the way #166 declares `state` (`Query(json_schema_extra={"enum": […]})` with the route's own description, absorbing a value outside the enum rather than refusing it, as measured) and ordering or filtering its rows before `_clamp` pages them; `_echo`, which writes the caller's parameters back into the `Link` urls, already carries `state` and would carry these. Each baseline entry drops as its parameter lands, the way #117 dropped the branch and tag listings' entries.

**What a fix has to settle**

- The unsent order. On issues it is `sort=created`'s, newest first; on pulls and on organization repositories it is not (newest first unsent, oldest first with `sort=created` spelt out), and real's own description says it should be. The wire is what a client meets, so the routes follow the wire, and the comment that departs from the description's sentence carries the measurement.
- What the corpus can order by. Each sort key needs a column the store holds: `created` and `updated` timestamps, the comment count for `comments` and `popularity`, `pushed_at` and the name for repositories. `long-running` also filters, by the description's own sentence, on being open for more than a month with activity in the last month, which the measured rows (closed pulls from 2016) do not obviously satisfy; a fix measures that filter on its own before serving it.
- Unknown values absorb, on every route measured, the way `state` on pulls does; the order an unknown `sort` yields is the route's: the unsent order on issues, the explicit `sort=created` order on pulls and the two repository listings. An unknown `direction` yields the unsent order everywhere measured, on issues even beside a `sort` it would otherwise honour.
- `direction`'s default follows `sort` (the description's `full_name` rule on repositories is measured to hold: `full_name` ascends and `direction=desc` reverses it), so the test is a cell per `(sort, direction)` pair per route, not a rule per parameter.
- `type`, `visibility` and `affiliation` filter on facts a corpus states or does not: whether a `subtype: "repo"` record says the repository is a fork, private or owned decides which of these can be honoured. A filter the corpus cannot answer stays undeclared rather than declared and ignored, which is the rule `openapi.qp`'s docstring already states for the hand-declared parameters.
- Scope. The four listings only. `/search/issues?sort`, `/search/code?sort`, `/repos/{o}/{r}/pulls/{n}/comments?sort` and `?direction`, and `/repos/{o}/{r}/collaborators?affiliation` are the baseline's other acknowledged entries of this kind and are not measured here.

### LINKED ISSUE #168: github: no response carries the five x-ratelimit-* headers real puts on every answer, and GET /rate_limit is a 404
**Source and commit** — `github`, at `1246b35` (`main`; #166 changes no header here, and the `HEAD` it answers carries none of these either).

Real puts five `x-ratelimit-*` headers on every response it gives, 200 and error alike, and serves `GET /rate_limit`; Backlot sends none of the five on any response and answers the route 404. `backlot/fidelity/baseline/github.json` acknowledges `GET /rate_limit` as a `missing_operation` ("the vendor serves it; Backlot does not"); response headers are not in the OpenAPI comparison at all, so the five are the silent half.

| # | Request | Real | Backlot at `1246b35` | Silent? |
|---|---|---|---|---|
| 1 | `GET /repos/{o}/{r}`, no credential | `x-ratelimit-limit: 60`, `-remaining`, `-used`, `-reset: <epoch seconds>`, `-resource: core` | none of the five | yes |
| 2 | `GET /repos/{o}/{r}`, authenticated | `limit: 5000`, `resource: core` | none | yes |
| 3 | `HEAD /repos/{o}/{r}` | the five, and the call counts (`remaining` 46 → 45) | none (405 at `1246b35`; the GET's headers without these five once #166 lands) | yes |
| 4 | `GET /search/issues?q=…` | `limit: 10` unauthenticated, `30` authenticated, `resource: search` | none | yes |
| 5 | `GET /search/code?q=…`, authenticated | `limit: 10`, `resource: code_search` | none | yes |
| 6 | the 404, the 401s, the blank-`q` 422, the version 400 | the five, counted against `core` (the 422 against `search`) | none | yes |
| 7 | `GET /rate_limit` | 200, `resources.core`, `.search`, `.code_search` each `{limit, remaining, reset, used}` | 404 | no |

**How you know** — measured against api.github.com on 2026-09-09, unauthenticated with `curl` and authenticated with `gh api`, within a few minutes; the docs page "Rate limits for the REST API" read the same day.

- Unauthenticated, in sequence: `GET /repos/psf/requests` 200 with `x-ratelimit-limit: 60`, `x-ratelimit-remaining: 46`, `x-ratelimit-used: 14`, `x-ratelimit-reset: 1788930807`, `x-ratelimit-resource: core`; `HEAD` on the same path 200 with `remaining: 45`, `used: 15`, the same `reset`; `GET /repos/psf/ghost-zz-9876` 404 with `remaining: 44`, `used: 16`; `GET /search/issues?q=repo:psf/requests+timeout&per_page=1` 200 with `limit: 10`, `remaining: 9`, `used: 1`, `resource: search`, `reset: 1788929255`; `GET /search/code?q=…` 401 with `limit: 60`, `remaining: 43`, `resource: core`, so the gateway's 401 on the code search path counts against `core`, not `code_search`; `GET /user` 401 with `remaining: 42`, `resource: core`. The `reset` was `1788930807` on every `core` answer of the sequence, so the window did not move with the requests.
- Authenticated: `GET /repos/psf/requests` 200 with `limit: 5000`, `remaining: 4966`, `used: 34`, `reset: 1788931438`, `resource: core`; `GET /search/issues?…` 200 with `limit: 30`, `remaining: 29`, `resource: search`; `GET /search/code?…` 200 with `limit: 10`, `remaining: 9`, `resource: code_search`; `GET /search/issues?q=` 422 with `limit: 30`, `remaining: 28`, `resource: search`; `X-GitHub-Api-Version: 1999-01-01` on the repository 400 with `limit: 5000`, `remaining: 4965`, `resource: core`. Every status measured carries the five.
- `GET /rate_limit`, authenticated, a second after those: `resources.core` `{"limit": 5000, "remaining": 5000, "reset": 1788932798, "used": 0}`, `resources.search` `{"limit": 30, "remaining": 30, …}`, `resources.code_search` `{"limit": 10, "remaining": 10, …}`. That disagrees with the headers the requests just before it carried (`remaining: 4966`, `reset: 1788931438`), so which window the route reports is itself a thing to measure again before serving it.
- The docs page, verbatim: "The primary rate limit for unauthenticated requests is 60 requests per hour."; "All of these requests count towards your personal rate limit of 5,000 requests per hour."; the header table, `x-ratelimit-limit` "The maximum number of requests that you can make per hour", `x-ratelimit-remaining` "The number of requests remaining in the current rate limit window", `x-ratelimit-used` "The number of requests you have made in the current rate limit window", `x-ratelimit-reset` "The time at which the current rate limit window resets, in UTC epoch seconds", `x-ratelimit-resource` "The rate limit resource that the request counted against."; and "If you exceed your primary rate limit, you will receive a 403 or 429 response, and the x-ratelimit-remaining header will be 0. You should not retry your request until after the time specified by the x-ratelimit-reset header." Exhaustion is docs only; nothing here was driven to it.
- The OpenAPI description declares `GET /rate_limit`; the headers are documented on that page and not declared per operation, which is why `backlot diff` cannot see them.

**Reproduction** — the bundled corpus, repo `pipeline`, with PyGithub 2.10.0 for the last two lines:

```python
import backlot, requests
from github import Auth, Github

with backlot.serve() as m:
    api = f"{m.base_url}/github"
    for path, headers in (
        (f"repos/{org}/pipeline", h),
        (f"search/issues?q=repo:{org}/pipeline", h),
        ("user/repos", {}),
        ("rate_limit", h),
    ):
        r = requests.get(f"{api}/{path}", headers=headers)
        print(path.split("?")[0], r.status_code, sorted(k for k in r.headers if k.lower().startswith("x-ratelimit")))
    g = Github(base_url=api, auth=Auth.Token(m.token))
    print(g.get_rate_limit())
```

prints `200 []`, `200 []`, `401 []`, `404 []` for the four, and the PyGithub call raises `UnknownObjectException: 404 {"message": "Not Found", "documentation_url": "https://docs.github.com/rest", "status": "404"}`.

**What it costs a client** — its pacing. A client that paces by `x-ratelimit-remaining` and sleeps until `x-ratelimit-reset` never sees either header here, so the backoff path it will run in production is unreachable against Backlot; a client that asks `GET /rate_limit` before a crawl, which is what PyGithub's `get_rate_limit()` does, gets a 404 and raises before the first real request; and a client that treats a missing `x-ratelimit-remaining` as exhausted stops before it starts. None of these can be tested here today.

**Where it lands** — a middleware in `backlot/main.py` beside `echo_github_api_version`, since the headers ride on every `/github` answer including the ones the exception handlers return, with the resource chosen by path (`/search/code` is `code_search`, the other `/search/*` routes `search`, everything else `core`, and the 401 on `/search/code` `core` as measured); a `GET /rate_limit` route in `backlot/routers/github.py`, shaped by the description's `rate-limit-overview` schema; and a counter keyed by the resolved credential with an hourly window, if the headers are to count.

**What a fix has to settle**

- Whether the headers count. Fixed values (`limit: 5000`, `remaining: 4999`) satisfy a client that only reads them and mislead one that paces by them; counted values need a per-credential, per-resource window and a `reset` that stays put across the window, as the measured `reset` did. Whether real's window is fixed to the hour or starts at the first request is not measured.
- Whose limits. Real counts an unauthenticated caller by IP at 60 and a token at 5000, `search` at 10 and 30, `code_search` at 10 (unauthenticated code search is the 401, counted against `core`); Backlot resolves a token or an anonymous caller and has the client host for the latter.
- Exhaustion. The docs' 403 or 429 with `remaining: 0` is not measured, and a mock that refuses the 61st anonymous request makes a test suite fail on its own volume; counting and reporting without refusing, or refusing only under a setting, is the choice.
- `/rate_limit`. The body shape from the description, and which window it reports given the disagreement measured above.
- The other vendors. Not measured; GitHub alone, by the same opt-in a `HEAD` gets in #166.

nuri-yoo:
Closes #167 and closes #168. On `main` at `f5611ce`. The parameters follow the `state` declaration #166 added, and the rate-limit count sits inside that PR's `HEAD` middleware so a `HEAD` counts once as it does on real. Opened stacked on #166's branch, rebased onto `main` after that merge, and `main` merged in again after #165; the one conflict, back at the rebase, was the two `Notes` cells in `docs/supported-sources.md` where #166's `state` sentences and this PR's `sort`/`direction` sentences met, kept both. The OpenAPI document gains one operation, `GET /rate_limit`, and ten query parameters, and the eleven baseline entries they resolve are dropped by hand, the way #131 dropped the paging ones.

**What GitHub does.** For #167, the four listings order by `sort` and `direction` and the two repository listings filter by `type` and `visibility`, each declared in the description with an enum and a default (read 2026-09-10, verbatim in the code), and the wire departs from the description in the cells the issue predicted and a few more, measured 2026-09-09 and 2026-09-10 against `psf/requests`, the `psf` organization and the token's own repositories, three rows a page. Pulls: newest first with no `sort` (7619, 7616, 7609) and oldest first the moment one is sent, `sort=created` included (4, 8, 10), where the description says `desc` for exactly that case; `sort=long-running` answered closed pulls from 2016 with `state=all` and pulls idle since 2021 with `state=open`, so the filter the description attaches to it is not applied on the wire. Issues: every sort descends by default, an unknown `sort` or `direction` drops the other parameter (`sort=bogus&direction=asc` and `sort=updated&direction=bogus` both answer the unsent order), and `sort=comments` orders by conversation and review comments added together (5797 at `comments: 105`, `review_comments: 25` sits between 2966 at 211 and 1573 at 122, which no single member orders); a tie keeps the unsent order whatever the direction (`sort=comments&direction=asc` answers 7620, 7619, 7618, three rows at 0 comments). An organization's repositories come oldest first with no sort (`requests`, `cachecontrol`, `pyperf`) and newest first with `sort=created` sent; `full_name` alone ascends by default; an unknown `direction` is `asc` there (`sort=pushed&direction=bogus` answers the least recently pushed) and `desc` on pulls (`sort=updated&direction=bogus` answers the `direction=desc` order); `type=member` answers `[]` for `psf`, `type=forks` its three forks, `type=sources` and `type=bogus` every repository. On the token's own repositories an unknown `direction` is `desc` whether or not a sort was sent: `direction=bogus` alone answers what `direction=desc` answers, the reverse of the unsent order, and a second unknown value answers the same (measured 2026-09-10). That is the one listing of the four whose bare unknown direction is not its unsent order, which is why it is measured rather than carried over. For #168, five `x-ratelimit-*` headers on every response, 200 and error alike: `limit: 60` on `core` and `10` on `search` for a caller with no credential, `5000`, `30` and `10` on `core`, `search` and `code_search` for a token; a `HEAD` counts once (`remaining` 46 → 45); the 401 an anonymous code search gets counts against `core`; `reset` in epoch seconds stays put across a window, and windows differ between credentials and between resources. `GET /rate_limit` answers 200 with `resources.core`, `.search`, `.code_search` each `{limit, used, remaining, reset}`, `rate` BEFORE them under `2022-11-28` and absent under `2026-03-10`, carrying the five itself and not counting (two reads in a row both `used: 0`); it answers a caller with no credential at the anonymous limits and a bad bearer with the 401.

**What Backlot did before this PR.** Declared none of the ten parameters, read none of them, and answered every request in the store's order: number ascending on issues and pulls, name order on both repository listings. No response carried an `x-ratelimit-*` header and `GET /rate_limit` was a 404, so PyGithub's `get_rate_limit()` raised `UnknownObjectException` before a crawl's first request.

**What changed for #167.** Each listing carries an `_Ordering` (`_ISSUE_ORDERING`, `_PULL_ORDERING`, `_ORG_REPO_ORDERING`, `_USER_REPO_ORDERING`): real's enum in real's order, the declared default, whether the unsent order descends, which sorts descend when sent with no direction, what an unknown `sort` is read as and what an unknown `direction` yields. Each field's value is a measured cell and the comment on each constant names the rows that measured it. `_order(request, spec)` reads the pair off the query, because sent and unsent differ on the wire and FastAPI hands the handler the default for both; `_ordered` sorts into the unsent order first and stably by the sort second, which is how a tie keeps the unsent order whatever the direction, as measured. The parameters are declared with `_enum_param`, the `Query(json_schema_extra={"enum": …})` route #166 took for `state` (`_state_param` now calls it), with `None` for the two direction parameters real declares without a default, so the schema carries the enum and the description and no `default`. `comments` and `popularity` read `store.github_comment_counts`, one GROUP BY over the repository's comments, conversation and review together, scoped to the caller: a review comment whose `path` names no file that caller can read is not counted, because `_resolved_review_comments` drops it from the pull's `review_comments` and from the list, so the key counts what the caller is served, and its docstring says that rather than the reverse. `long-running` is `created` with no filter. The repository timestamps and the issue timestamps the keys order by are the ones the bodies serve, factored into `_repo_timestamps`, `_created_ts` and `_updated_ts` and read by `_repo_obj` and `_shared_obj` from the same functions. `type` on the organization listing and `visibility` on the token's own select on the one fact a corpus states about a repository's kind, its ACL: `public` and `private` are the org-wide grant or its absence, `forks` and `member` keep nothing (every repository is the organization's own and the object says `fork: false`), `sources`, `all` and a value outside the enum keep every one, as measured. The two that keep nothing return `_keeps_nothing`, which `_repo_page` recognises, so those requests answer their empty page without reading one repository's ACL rather than reading every visible repository's to discard every row. `type` and `affiliation` on `/user/repos` stay undeclared and their two baseline entries stay: they select on what the caller is to each repository, which no corpus states, so the 422 real answers for `type` beside `visibility` is not reachable here, and the route's docstring says which parameters are in and which out and why. The `Link` urls carry the caller's own spelling of the parameters they sent (`_sent`) and none the handler applied, `_echo`'s rule. The unsent order of the issue and pull listings is now newest first and of the organization's repositories oldest first, so one existing visibility test that compared two listings' names in name order now compares them sorted.

**What changed for #168.** `report_github_rate_limit` in `backlot/main.py`, registered after the version echo and before the id-path rewrite, so it is inside the `HEAD` middleware and a `HEAD` counts as the GET it is rewritten to; the middleware docstring that counted the two middlewares inside the `HEAD` one now counts three. It calls `github.rate_limit_headers(request, status_code)`, which resolves the caller (`rate_limit_caller`: the token when it resolves, the client's address otherwise, a bad bearer counted with the anonymous callers from its address since which window real counts one against is not measured), the resource (`rate_limit_resource`: `code_search` for code search, `core` for its 401, `search` for the other search route, `core` for everything else) and the limit (`RATE_LIMITS`, `core` 60/5000, `search` 10/30, `code_search` 10/10), and counts in `RateLimitWindows`, kept on `app.state` and opened on first use: a window per `(credential, resource)` opens at the first request counted or reported under it and closes an hour later, `reset` is that second on every answer inside it, `remaining` stops at 0 and `used` keeps counting. `GET /rate_limit` reports the same windows without counting, and the path is compared with its trailing slash stripped, so the 307 FastAPI answers `/github/rate_limit/` with is not counted either. `rate` is gated by `_HAS_RATE_ALIAS` beside the two version sets that already exist and is serialized first, as real orders it. The route serves a caller with no credential: `_validate_path_owner` skips the credential check on that one path, its docstring and `_require`'s say so, and the route checks a bearer it is given itself. Its `ROUTE_DOCS` entry is the description's `externalDocs` url, the one entry in that table not measured off a 404, because the route's only error on the wire is the credential 401 with the root; the module docstring records the exception.

**What #167 asked a fix to settle.** The unsent order: the wire's, per listing, and every departure from the description's sentence carries its measurement in the constant's comment. What the corpus can order by: `created` and `updated` from the row, the comment count from one query, `pushed_at` and the name from what `_repo_obj` derives, so every key in every enum is served. `long-running`: measured on its own, with `state=all` and `state=open`, and found to filter nothing, so it filters nothing. Unknown values: absorbed everywhere, in the per-listing order the issue predicted for `sort` and in three different orders for `direction` (dropped on issues, `desc` on pulls, `asc` on the organization's repositories, `desc` on the token's own), each a field, each measured both beside a sort and alone. `direction`'s default follows `sort`: the cells are `(sort, direction)` pairs per route in the test, not a rule per parameter. `type`, `visibility`, `affiliation`: the two the ACL can answer are declared, the two it cannot are not. Scope: the four listings; the search routes' `sort` and `order`, the review comments' `sort` and `direction` and the collaborators' `affiliation` stay acknowledged in the baseline, unmeasured.

**What #168 asked a fix to settle.** Whether the headers count: they count, per credential and per resource, with a `reset` that stays put, and the two measured resets (`core` and `search` 1552 seconds apart under one anonymous caller, different again under the token) are read as windows opened by use rather than one clock hour. Whose limits: a token's by token, an anonymous caller's by address, at real's numbers. Exhaustion: not measured, not refused; `remaining` stops at 0, the docstring says why. `/rate_limit`: the description's `rate-limit-overview` shape over the three resources Backlot counts (of the fifteen real's authenticated answer carries), and it reports the headers' window rather than the fresh one real answered twice, because a client asks the route to learn what the headers would say and the fresh window could only be reproduced by reporting one nothing counts against; the route's docstring carries both measurements. The other vendors: GitHub alone, by the path prefix the middleware checks.

**Tests.** Four new in `tests/test_github.py`. `test_github_sort_and_direction_order_the_issue_and_pull_listings_as_measured` builds six documents whose creation, update and comment orders all differ (one pull with a conversation comment and two review comments anchored to files of the same repo, so the count `comments` orders by is the three that pull's own `comments` and `review_comments` add up to) and holds the two schemas to real's declarations and 23 ordering cells to the measured orders, the `Link` echo included. `test_github_type_visibility_sort_and_direction_on_the_repository_listings` builds three repositories whose derived creation order is not their name order (asserted as a precondition) with one private, and holds the schemas, 29 cells across the two listings, `type` undeclared and inert on `/user/repos`, and the `Link` echo. `test_github_a_type_that_keeps_nothing_reads_no_repository_acl` counts the ACL reads a repository listing makes through `set_trace_callback`, the shape `test_github_a_ref_check_reads_the_repos_pulls_once` already uses: two for a request selecting on nothing and for `type=sources`, one per visible repository for `type=public`, and none at all for `type=forks` and `type=member`. `test_github_every_response_carries_the_five_ratelimit_headers_and_rate_limit_reports_them` walks a fresh corpus through a GET, a HEAD, the 404, the version 400, both search routes at their limits, the anonymous 401s against `core` at 60, a second token's own window, the route's body against the headers just read and its `content-type`, that a second read changes nothing and that `rate` is the first key on the wire, the 307 a trailing slash answers and that it does not count either, `rate` gone under `2026-03-10`, the anonymous 200 and the bad-bearer 401, a window rolled over by moving the clock, and `remaining: 0` with `used: 3` at a limit of 2 and a 200. `test_sdk.py`'s GitHub run adds `get_rate_limit().resources.core.remaining` and `Github.rate_limiting`, the `(remaining, limit)` pair PyGithub parses off `x-ratelimit-remaining` and `x-ratelimit-limit` as ints on every response. `test_github_documentation_url_names_the_route_that_failed` covers the new route's `ROUTE_DOCS` entry by its set equality; `test_github_a_container_only_repo_reaches_the_admin_and_no_scoped_caller` compares names sorted; `test_github_comment_counts_match_the_lists_they_describe` gains the ordering half, that a `sort=comments&direction=asc` page ascends by the counts each caller is actually served.

**What this round changed.** Seven review comments, all taken. Three were applied as suggestions: the trailing-slash comparison, `rate` serialized before `resources`, and the `orgs/{org}/repos` docs cell, which now says the three timestamp sorts are one derived order here rather than promising three. The other four are this round's four commits. `store.github_comment_counts` takes the caller's ids and counts only comments that caller is served, which closes the ordering half of the leak `_resolved_review_comments` closed for the list: with the raw count, a caller served `(0 conversation, 1 review)` on one pull and `(1, 2)` on another was answered them in that order under `direction=asc`, own counts descending 3, 1, because the hidden file's comment counted for the sort and not for the row. The fixture that models "a pull with three comments" now gives the two anchored paths files of the repo, so they resolve and are served, and the measured order it asserts is unchanged. `/user/repos?direction=bogus` was measured with no sort (it had been extrapolated from `sort=created&direction=bogus`) and holds: it is `desc`, so the code is unchanged and the comment and a new cell record the measurement. `_keeps_nothing` removes the per-repository ACL read behind `type=forks` and `type=member`. The `_USER_REPO_ORDERING` comment no longer names the repositories the measurement ran against: that listing is one token's own, so a reader cannot reach it from the names, and the dates carry the claim.

Mutation-checked, each from the committed tree and restored afterwards: the middleware writing no header fails the rate-limit test on its first response (`KeyError: 'x-ratelimit-limit'`); `_ORG_REPO_ORDERING` with `unsent_descending=True` fails the repository test on the first cell (`['portal', 'gateway', 'ledger'] == ['ledger', 'gateway', 'portal']`); `_ordered` sorting once rather than into the unsent order first fails the issue test on `sort=comments`, where two pairs tie (`[2, 3, 5, 6, 1, 4] == [2, 3, 6, 5, 4, 1]`); `/rate_limit` counting itself fails the rate-limit test on the headers the route carries, `used` off by one. This round: the raw comment count fails the new ordering half with the descending pair (`[0, 0, 0, 0, 0, 3, ...] == [0, 0, 0, 0, 0, 2, ...]`); the un-stripped path fails the trailing-slash cell (`'5' == '4'`); the pre-suggestion body construction fails the key-order cell (`['resources', 'rate'] == ['rate', 'resources']`); dropping the `_keeps_nothing` branch fails the read count (`(0, 7) == (0, 0)`).

**Verification.** At `0b0c357`, the head of this branch:

```
PYTHONPATH=. python -m pytest -q -rs      2897 passed, 3 skipped
PYTHONPATH=. python scripts/gen_docs.py --check      exit 0
ruff check .                              All checks passed!
ruff format --check .                     139 files already formatted
PYTHONPATH=. python -m backlot diff --source github    1229 divergence(s) · 1225 acknowledged · 4 new
```

The three skips are extras imports unrelated to this change: the llama-index hubspot reader and two mirage google-client tests. The four new divergences are `GET` and `PATCH /orgs/{}/code-scanning/ai-scan` and the same pair under `/repos/{}/{}`, operations GitHub has added to its published description since the baseline was measured; `origin/main` reports the same four (1240 · 1236 · 4), so they are not this PR's and are left for a baseline refresh.

The two issues' reproductions, run verbatim over `backlot.serve()` on the bundled corpus at `0b0c357`: #167's three issue lines print three different orders (`20536, 47516, 87703, …`, `20536, 47516, 287, …` and `75342, 9597, 2317, …`), the `long-running` line `[9597, 287, 20536]`, the `type=forks` line `[]`, the `sort=full_name&direction=desc` line `['pipeline', 'docs-site']`, and the spec `['state', 'sort', 'direction', 'page', 'per_page']`; #168's four lines each carry the five headers, the anonymous `user/repos` 401 included, and PyGithub's `get_rate_limit()` returns `RateLimitOverview(rate=Rate(reset=…, remaining=4999, limit=5000))`.

**Not in this PR.** The baseline's other `sort`-family entries (`/search/issues?sort` and `?order`, `/search/code?sort` and `?order`, `/repos/{o}/{r}/pulls/{n}/comments?sort` and `?direction`, `/repos/{o}/{r}/collaborators?affiliation`), unmeasured, as #167's scope bullet left them. `since` and `before` on `/user/repos`, `since`, `labels`, `milestone`, `assignee`, `creator`, `mentioned`, `type` and `issue_field_values` on the issue listing and `head` and `base` on the pull listing, filters real declares on these same four routes that neither issue measured. The other vendors' rate-limit headers. The `code-scanning/ai-scan` operations above.

### DIFF
--- a/backlot/routers/github.py
+++ b/backlot/routers/github.py
@@ -11,7 +11,10 @@ import base64
 import hashlib
 import json
 import re
+import time
+from collections.abc import Callable
 from email.utils import formatdate
+from typing import NamedTuple
 from urllib.parse import quote
 
 from fastapi import APIRouter, Depends, HTTPException, Query, Request, Response
@@ -90,8 +93,9 @@ def _require(request: Request) -> Caller:
     Header PRESENT but not a scheme this API takes is the second case, not the first: real ignores
     an `Authorization` it cannot parse and serves the request anonymously (measured: `Basic …` and a
     scheme-less value both answer 200 on a public repo), so the caller reaches an auth-required
-    route with no credential rather than with a rejected one. Backlot serves nothing anonymously,
-    which is why that lands here as the missing-credential 401 rather than as a public read.
+    route with no credential rather than with a rejected one. Backlot serves no document
+    anonymously, which is why that lands here as the missing-credential 401 rather than as a public
+    read (`/rate_limit`, which serves none, is the exception — see :func:`_validate_path_owner`).
     """
     if auth.bearer_token(request) is None:
         raise HTTPException(status_code=401, detail="Requires authentication")
@@ -206,7 +210,10 @@ async def _validate_path_owner(request: Request) -> None:
     client's owner-handling bug pass against Backlot and fail in production. A router-wide
     dependency rather than a call in each handler so a route added later cannot forget it — routes
     with neither path param (``/search/issues``, ``/user/repos``) are unaffected. Credentials are
-    checked first, so a bad token still reports 401 rather than the owner's 404.
+    checked first, so a bad token still reports 401 rather than the owner's 404. `/rate_limit` is
+    the one route a caller with no credential is served, as real serves it (200 at the anonymous
+    limits, measured 2026-09-10); it names no owner and reads no document, and it checks the
+    credential it is given itself (see :func:`get_rate_limit`).
 
     The match is case-insensitive, as GitHub logins are, and real then answers in the canonical
     spelling whatever case was asked for: `/repos/PSF/REQUESTS` answers `full_name: psf/requests`
@@ -223,7 +230,8 @@ async def _validate_path_owner(request: Request) -> None:
     day), and :func:`_page_base_url` hashes the settled spelling rather than the one the request
     arrived with — this dependency's for the org, :func:`_canonical_path_repo`'s for the repo.
     """
-    _require(request)
+    if request.url.path != RATE_LIMIT_PATH:
+        _require(request)
     key = "owner" if "owner" in request.path_params else "org"
     owner = request.path_params.get(key)
     if owner is None:
@@ -251,6 +259,143 @@ async def _canonical_path_repo(request: Request) -> None:
     spelled = store.container_spelling(auth.conn(request), "github", repo)
     if spelled is not None:
         request.path_params["repo"] = spelled
+
+
+# --- rate limits ----------------------------------------------------------------
+#
+# Real puts five `x-ratelimit-*` headers on every response it gives, 200 and error alike, and
+# serves `GET /rate_limit`. Measured against api.github.com on 2026-09-09 and 2026-09-10,
+# unauthenticated with curl and authenticated with `gh api`: an anonymous caller has `limit: 60`
+# on `core` and `10` on `search`, a token `5000`, `30` and `10` on `core`, `search` and
+# `code_search`; the 404 for a repository that does not exist, the 401s, the blank-`q` 422 and the
+# version 400 each carry the five and count (`remaining` 46, 45, 44 across a GET, a HEAD and a 404
+# in a row); `reset` is epoch seconds and stayed put across every answer inside one window. The
+# docs page "Rate limits for the REST API" states the 60 and the 5,000 and the five headers'
+# meanings; the numbers below are the wire's. Nothing here refuses a request, see
+# :class:`RateLimitWindows`.
+
+RATE_LIMIT_PATH = "/github/rate_limit"
+RATE_LIMIT_WINDOW = 3600
+
+
+class _HourlyLimit(NamedTuple):
+    anonymous: int
+    authenticated: int
+
+
+#: resource -> the requests an hour real allows a caller with no credential and one with a token
+RATE_LIMITS: dict[str, _HourlyLimit] = {
+    "core": _HourlyLimit(60, 5000),
+    "search": _HourlyLimit(10, 30),
+    "code_search": _HourlyLimit(10, 10),
+}
+
+
+def rate_limit_resource(path: str, status_code: int) -> str:
+    """The resource a request to ``path`` answered with ``status_code`` counts against.
+
+    `code_search` for code search, `search` for the other search route and `core` for everything
+    else, including the 401 code search answers a caller with no credential: that refusal is the
+    gateway's, not the code search backend's, and counts against `core` (measured: `limit: 60`,
+    `resource: core` on it, where the same route authenticated answers `limit: 10`,
+    `resource: code_search`), the same split ``errors.github.json_media_type`` draws for the
+    charset."""
+    if path == CODE_SEARCH_PATH:
+        return "core" if status_code == 401 else "code_search"
+    if path.startswith("/github/search/"):
+        return "search"
+    return "core"
+
+
+class RateLimitWindows:
+    """The requests counted so far, per credential and per resource, in hourly windows.
+
+    A window opens at the first request counted or reported under a `(credential, resource)` and
+    closes an hour later; `reset` is its closing second, the same on every answer inside it. The
+    windows measured stayed put across the requests inside them and differed between credentials
+    and between resources (an anonymous caller's `core` and `search` resets 1552 seconds apart),
+    which is a
+    window per pair opened by use rather than one clock hour shared by all. `remaining` stops at 0
+    and `used` keeps counting past the limit: nothing here refuses a request, because a test
+    suite's own volume drives an anonymous client past 60 in an hour, and a mock answering the
+    61st request with the 403 or 429 the docs describe fails that suite for pacing it never asked
+    for. Exhaustion is docs only; nothing was driven to it. ``clock`` is `time.time` unless a test
+    hands in another to move a window."""
+
+    def __init__(self, clock: Callable[[], float] = time.time):
+        self.clock = clock
+        self._windows: dict[tuple[str, str], list[int]] = {}
+
+    def _window(self, key: str, resource: str) -> list[int]:
+        now = int(self.clock())
+        window = self._windows.get((key, resource))
+        if window is None or now >= window[0] + RATE_LIMIT_WINDOW:
+            window = self._windows[(key, resource)] = [now, 0]
+        return window
+
+    def count(self, key: str, resource: str, limit: int) -> dict[str, int]:
+        """The window after counting one more request in it."""
+        window = self._window(key, resource)
+        window[1] += 1
+        return self._status(window, limit)
+
+    def status(self, key: str, resource: str, limit: int) -> dict[str, int]:
+        """The window as it stands, nothing counted."""
+        return self._status(self._window(key, resource), limit)
+
+    @staticmethod
+    def _status(window: list[int], limit: int) -> dict[str, int]:
+        start, used = window
+        return {
+            "limit": limit,
+            "used": used,
+            "remaining": max(limit - used, 0),
+            "reset": start + RATE_LIMIT_WINDOW,
+        }
+
+
+def _rate_limit_windows(app) -> RateLimitWindows:
+    """The app's windows, opened on first use so a request needs no lifespan step to be counted."""
+    windows = getattr(app.state, "github_rate_limits", None)
+    if windows is None:
+        windows = app.state.github_rate_limits = RateLimitWindows()
+    return windows
+
+
+def rate_limit_caller(request: Request) -> tuple[str, bool]:
+    """Whose requests these are for counting, and whether that is a credential.
+
+    The token when it resolves; the client's address otherwise, which is how real counts a caller
+    with no credential (the docs' 60 an hour "for unauthenticated requests", measured at
+    `limit: 60` on every anonymous answer measured). A bearer that does not resolve is counted with the
+    anonymous callers from its address: real's answer for one is a 401 carrying the five, and which
+    window it counts against is not measured."""
+    token = auth.bearer_token(request)
+    if token is not None and auth.resolve_bearer(request) is not None:
+        return f"token:{token}", True
+    host = request.client.host if request.client is not None else "anonymous"
+    return f"host:{host}", False
+
+
+def rate_limit_headers(request: Request, status_code: int) -> dict[str, str]:
+    """The five `x-ratelimit-*` headers for a `/github` answer, counting it, except on
+    :data:`RATE_LIMIT_PATH`, which reports its window without counting: two `GET /rate_limit` in
+    a row both answered `remaining: 5000`, `used: 0`, each carrying the five with `resource: core`,
+    and the description's own note says the route does not count."""
+    key, authenticated = rate_limit_caller(request)
+    resource = rate_limit_resource(request.url.path, status_code)
+    limits = RATE_LIMITS[resource]
+    limit = limits.authenticated if authenticated else limits.anonymous
+    windows = _rate_limit_windows(request.app)
+    read = windows.status if request.url.path == RATE_LIMIT_PATH else windows.count
+    window = read(key, resource, limit)
+    return {
+        "x-ratelimit-limit": str(window["limit"]),
+        "x-ratelimit-remaining": str(window["remaining"]),
+        "x-ratelimit-used": str(window["used"]),
+        "x-ratelimit-reset": str(window["reset"]),
+        "x-ratelimit-resource": resource,
+    }
 
 
 router = APIRouter(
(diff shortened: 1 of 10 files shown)


backlot/routers/github.py
```diff
@@ -478,6 +623,273 @@ def _echo(request: Request, **params) -> dict:
...
+def _issue_sort_keys(conn, repo: str, sort: str) -> dict[str, Callable]:
+    """The sort keys of the issue and pull listings, over the store's rows.
+
+    The timestamps are the ones the bodies serve (:func:`_shared_obj` reads the same two
+    functions), and the number breaks a tie the way real's monotonic ids do. `comments` and
+    `popularity` are one key: both order by the conversation and review comments added together,
+    which is the count real orders by (see :data:`_ISSUE_ORDERING`), read in one query for the
+    repository and only when asked for. `long-running` is `created` with no filter (see
+    :data:`_PULL_ORDERING`)."""
+    counts = store.github_comment_counts(conn, repo) if sort in ("comments", "popularity") else {}
```
khj809: `github_comment_counts` counts raw rows while the served `review_comments` counts only comments whose `path` resolves to a file the caller can read (`_resolved_review_comments`), so the order is caller-independent while the counts are not: with three review comments on a group-restricted file, `?sort=popularity` answers `[#4, #2]` to a caller served `(#4: comments 2, review 0)` then `(#2: 0, 0)` — an ascending sort whose own counts descend. Suggest threading `visible_ids` into `github_comment_counts` so the key counts what the caller is served; as it stands a row's position leaks that a hidden file carries a comment, which `_resolved_review_comments`' docstring says was closed deliberately.

backlot/store.py
```diff
@@ -2815,6 +2815,21 @@ def github_comments(conn, repo, number, *, anchored: bool | None = None) -> list
...
 
 
+def github_comment_counts(conn, repo) -> dict[int, int]:
+    """Document number -> how many ``github_comments`` rows it has, conversation and review
+    comments together, for every document of one repo.
+
+    One GROUP BY rather than a :func:`github_comments` call per row: the issue and pull listings
+    order a whole repository by this number when asked to (`sort=comments`, `sort=popularity`),
+    and the count real orders by is the two kinds added, which is why ``anchored`` is not a
+    parameter here (see ``routers.github._issue_sort_keys``)."""
```
khj809: The PR body says this key's docstring records that the count includes a review comment anchored to a path that does not resolve, which the pull's served `review_comments` leaves out — no docstring says it. State it here, or drop the claim from the body if the ACL point above is fixed instead.

```suggestion
    parameter here (see ``routers.github._issue_sort_keys``). The count is of the raw rows: a
    review comment whose ``path`` names no file the caller can read is counted here and left out
    of the pull's served ``review_comments``, so the two numbers can disagree."""
```

backlot/routers/github.py
```diff
@@ -478,6 +623,273 @@ def _echo(request: Request, **params) -> dict:
...
+
+#: The token's own repositories: the description declares `default: full_name`, and with no `sort`
+#: real answers an order that is neither the names' nor `created_at`'s nor `pushed_at`'s (`bsk`,
+#: `bsk_check`, `agent-k`, `agent-zoo`, created 2026-07, 2026-09, 2026-03, 2026-08, pushed 09-02,
+#: 09-05, 08-15, 08-31), and `sort=full_name` the same first three names, which are not in
+#: full-name order either; Backlot answers the declared default, name order, for both.
+#: `sort=created` newest first; `sort=bogus` the same; `sort=created&direction=bogus` newest first
+#: too, so an unknown direction is read as `desc` (one sort measured). `type=bogus` and
+#: `visibility=bogus` keep every repository.
+_USER_REPO_ORDERING = _Ordering(
```
khj809, line 756: `unknown_direction="desc"` was measured only beside a sort (`sort=created&direction=bogus`); with no `sort` it makes `?direction=bogus` answer `full_name` descending — the reverse of this listing's unsent order, where the three listings whose bare `direction=bogus` was measured all answer the unsent order (`['portal', 'ledger', 'gateway']` against an unsent `['gateway', 'ledger', 'portal']`). Worth one measurement against `/user/repos?direction=bogus` and a cell in the test, since it is the only absorbed-value case the new test does not assert.

khj809 (commented):
Re-reviewed at `10e09c4` on `main`. Verification reproduces: the full suite 2866 passed, 3 skipped; `gen_docs.py --check` exit 0; `backlot diff --source github` 1225 acknowledged, 0 new, none resolved; `ruff check` and `ruff format --check` clean.

Checked the constants against a fresh read of `api.github.com.json`: the four listings' enums, defaults and description strings match verbatim, `ROUTE_DOCS["/github/rate_limit"]` is the operation's own `externalDocs` url, and `GET /rate_limit`'s 200 is the only operation in the document declaring `x-ratelimit-*` headers and declares three of the five, so the new `docs/fidelity.md` paragraph holds.

Seven comments below. The first is the one I would want settled before merge; the rest are smaller. One thing outside the diff: the body still cites `5f481f8`, `f34899c`, `1017662` and `832a8da`, all orphaned by the rebase, and the Verification block reads as a per-round log rather than the final state — worth rewriting against `10e09c4`.

backlot/routers/github.py
```diff
@@ -253,6 +261,143 @@ async def _canonical_path_repo(request: Request) -> None:
...
+    """The five `x-ratelimit-*` headers for a `/github` answer, counting it, except on
+    :data:`RATE_LIMIT_PATH`, which reports its window without counting: two `GET /rate_limit` in
+    a row both answered `remaining: 5000`, `used: 0`, each carrying the five with `resource: core`,
+    and the description's own note says the route does not count."""
+    key, authenticated = rate_limit_caller(request)
+    resource = rate_limit_resource(request.url.path, status_code)
+    limits = RATE_LIMITS[resource]
+    limit = limits.authenticated if authenticated else limits.anonymous
+    windows = _rate_limit_windows(request.app)
+    read = windows.status if request.url.path == RATE_LIMIT_PATH else windows.count
```
khj809: FastAPI answers `/github/rate_limit/` with a 307 whose path is not this exact string, so the redirect counts against `core` (measured: `used` 2 → 3) where the route's docstring and the docs table both say a read of it does not count.

```suggestion
    read = windows.status if request.url.path.rstrip("/") == RATE_LIMIT_PATH else windows.count
```

backlot/routers/github.py
```diff
@@ -989,6 +1401,44 @@ async def get_org(org: str, request: Request):
...
+    resources = {
+        resource: windows.status(
+            key, resource, limits.authenticated if authenticated else limits.anonymous
+        )
+        for resource, limits in RATE_LIMITS.items()
+    }
+    body: dict = {"resources": resources}
+    if _version(request) in _HAS_RATE_ALIAS:
+        body["rate"] = resources["core"]
+    return body
```
khj809: The measurement quoted in the body and the test is that real's `2022-11-28` answer carries `rate` before `resources`; this serves them the other way round.

```suggestion
    if _version(request) not in _HAS_RATE_ALIAS:
        return {"resources": resources}
    return {"rate": resources["core"], "resources": resources}
```

docs/supported-sources.md
```diff
@@ -71,14 +71,15 @@ Field names are snake_case, as Fireflies' own schema has them. Full introspectio
 | `search/issues` | `q`: free text + `repo:` `is:` `state:` `type:` `label:` `author:` |
 | `search/code` | `q`: free text over a file's body and path + `repo:` `path:` `filename:` `extension:` `in:file`/`in:path` |
 | `orgs/{org}` | |
-| `orgs/{org}/repos` | |
+| `orgs/{org}/repos` | `type`, `sort`, `direction`: real's enums and defaults, each order as measured |
```
khj809: `_repo_timestamps` derives all three from one epoch plus fixed offsets, so `sort=created`, `sort=updated` and `sort=pushed` are always the same sequence — which the code's docstring admits and the new test asserts — while "each order as measured" reads as three distinct orders.

```suggestion
| `orgs/{org}/repos` | `type`, `sort`, `direction`: real's enums and defaults, each direction as measured; `created`, `updated` and `pushed` are one derived order here |
```

backlot/routers/github.py
```diff
@@ -1028,30 +1478,90 @@ def _visible_repos(conn, ids) -> list[str]:
...
+    """One page of a repository listing: the visible repositories ``keeps`` keeps, in the order
+    ``ordering`` reads off the request, paged after both so a `Link` walk sees one sequence.
+
+    ``keeps`` is ``None`` when the request selects on nothing, so the ACL read that decides
+    `private` happens once per repository on the page, as before, and once per visible repository
+    only when a `type` or `visibility` value asks it of every one."""
     repos = _visible_repos(conn, ids)
+    private: dict[str, bool] = {}
+    if keeps is not None:
+        private = {n: not store.container_has_public(conn, "github", n) for n in repos}
```
khj809, line 1523: `_repo_type_keeps` returns `lambda private: False` for `forks` and `member`, which is not `None`, so this line still runs one `container_has_public` query per visible repository before every row is discarded. A "keeps nothing" sentinel, or reading the value before the map is built, would make `?type=forks` free.

backlot/routers/github.py
nuri-yoo: Fixed in `7ef1031`. `store.github_comment_counts` takes the caller's ids and counts a review comment only where its `path` resolves to a file that caller can read, so the key is the `comments` and `review_comments` the caller is actually served added together.

Reproduced your case first, on the bundled `diffable` corpus: bob's `?sort=comments&direction=asc` answered the unresolvable pull ahead of the declared one, served totals 3 then 1, because the comment on `secret/keys.txt` counted for the sort and not for the row. After the change each caller's page ascends by their own totals, and the two callers differ exactly where the hidden comment is.

The ordering half is asserted in `test_github_comment_counts_match_the_lists_they_describe`; reverting the store change fails it with `[0, 0, 0, 0, 0, 3, ...] == [0, 0, 0, 0, 0, 2, ...]`. One fixture had to follow: the six-document corpus in the ordering test anchored two review comments to paths no file answered, so the count it modelled was never served to anyone. Those paths are files of the repo now, and the measured order that test asserts is unchanged.

backlot/store.py
nuri-yoo: Taken the other way, since the ACL point above is fixed: the docstring now says a review comment on a path the caller cannot read is not counted, and the claim is gone from the PR body. You were right that no docstring said what the body claimed.

backlot/routers/github.py
nuri-yoo: Measured. `/user/repos?direction=bogus` with no sort answers what `?direction=desc` answers, the reverse of the unsent order, and a second unknown value (`?direction=zzz`) answers the same — so `desc` holds and the code is unchanged.

`695fbe8` records the measurement in `_USER_REPO_ORDERING` and adds the cell, `assert listed(user_repos, direction="bogus") == by_name[::-1]`, with `direction="asc"` beside it. You were right that it was the one absorbed-value case with no cell and that it had been extrapolated from a single sort. The comment also stops naming the repositories the measurement ran against: that listing is one token's own, so the names are not a path a reader can take, and the dates carry the claim.

backlot/routers/github.py
nuri-yoo: Applied. Measured before applying it: `/github/rate_limit/` answered 307 and moved `used` 7 to 8. `9e5ee1c` adds the cell — `follow_redirects=False`, 307, `used` unchanged — and reverting the `rstrip` fails it with `'5' == '4'`.

backlot/routers/github.py
nuri-yoo: Applied, and a cell added for it: the existing `status.json() == {...}` is a dict comparison and does not see key order, so it passed either way. `9e5ee1c` asserts `list(status.json()) == ["rate", "resources"]`; reverting the construction fails it with `['resources', 'rate'] == ['rate', 'resources']`.

docs/supported-sources.md
nuri-yoo: Applied. Checked the same phrase elsewhere: the `user/repos` row below makes no claim about distinct orders, and the issue and pull rows' "each order as measured" still holds, because `_updated_ts` reads a stated `updated_ts` from the row where `_repo_timestamps` derives all three from one epoch.

backlot/routers/github.py
nuri-yoo: Fixed in `106b47b`. `_repo_type_keeps` returns a named `_keeps_nothing` that `_repo_page` recognises, so those requests answer their empty page without reading one repository's ACL.

Measured on the bundled corpus at `per_page=2`: `type=forks` and `type=member` went from 7 reads to 0, `type=public` still reads all 7 because its answer does depend on the flag, and a request selecting on nothing reads 2. `test_github_a_type_that_keeps_nothing_reads_no_repository_acl` counts them through `set_trace_callback`, the shape `test_github_a_ref_check_reads_the_repos_pulls_once` already uses; dropping the branch fails it with `(0, 7) == (0, 0)`.

nuri-yoo:
All seven taken, at `0b0c357`. Three were the suggestions, applied as written; the other four are four commits, each with a cell that fails when the change is reverted (the four reverts and their failures are in the body, beside the earlier round's).

The body is rewritten against this head, which also answers your note outside the diff: it cited four commits the rebase had orphaned, and the Verification block read as a per-round log. It is now the final state, and two counts in it were wrong before your review — "26 request cells" and "30 cells" were 23 and 27 — so they are recounted at 23 and 29.

One thing that will differ when you re-run: `backlot diff --source github` now reports 1229 · 1225 acknowledged · **4 new**, not the 0 new you saw this morning. The four are `GET` and `PATCH /orgs/{}/code-scanning/ai-scan` and the same pair under `/repos/{}/{}`, which GitHub has added to its published description since the baseline was measured. `origin/main` reports the same four (1240 · 1236 · 4), so they are not this branch's and I have left them for a baseline refresh rather than acknowledge them here.

```
PYTHONPATH=. python -m pytest -rs                       2897 passed, 3 skipped
PYTHONPATH=. python scripts/gen_docs.py --check         exit 0
ruff check .                                            All checks passed!
ruff format --check .                                   139 files already formatted
PYTHONPATH=. python -m backlot diff --source github     1229 · 1225 acknowledged · 4 new (see above)
```

Both issue reproductions re-run verbatim at this head; their output is in the body.

khj809:
Re-reviewed at `0b0c357`. All seven comments are addressed, and I checked each against the tree rather than taking the replies for it:

- `github_comment_counts` takes `visible_ids` and counts only what the caller is served, so the order and the served `comments`/`review_comments` are the same number again; `test_github_comment_counts_match_the_lists_they_describe` holds it.
- The docstring now states the caveat the body used to attribute to it.
- `/user/repos?direction=bogus` is measured with no sort, and the cell sits in the test beside `direction=asc`.
- `rstrip("/")` on the rate-limit path, with a 307 cell asserting `used` does not move.
- `rate` before `resources`, with `list(status.json())` asserting the order the dict comparison could not see.
- The `orgs/{org}/repos` docs row no longer reads as three distinct orders.
- `_keeps_nothing` short-circuits the ACL reads, counted through `set_trace_callback`.

Full suite 2897 passed, 3 skipped; `gen_docs.py --check` exit 0; `ruff check` and `ruff format --check` clean.

**The four new divergences are not this branch's.** `backlot diff --source github` reports the same four at `main` (`f5611ce`: 1240 divergences, 1236 acknowledged, 4 new) as it does here (1229, 1225 acknowledged, 4 new). GitHub published four `code-scanning/ai-scan` operations upstream — `GET` and `PATCH` on `/orgs/{}` and on `/repos/{}/{}` — surface Backlot has never served and this PR does not touch.

The scheduled run has already filed them: #125 carries exactly this report, updated today. That is the fidelity workflow doing what its own header describes — drift "is never the bug of whichever pull request happens to be open when a vendor ships a change — so a divergence opens an issue rather than failing someone else's merge." So they are out of scope here and get settled on #125, as a fix or as `--update-baseline` gaps. Nothing to do in this PR.

khj809 (approved):

PR #176, opened 2026-09-10
s3: ListMultipartUploads answers the empty page real serves for a bucket with no upload in progress, where it was a 501 

### LINKED ISSUE #169: s3: ListMultipartUploads is refused with 501 on the ordinary browse path, where real answers 200 with an empty ListMultipartUploadsResult
Cyberduck reaches `GET /{bucket}?uploads` on the ordinary browse path, not as an exotic call. Listing a bucket or any prefix issues it immediately after the `ListObjectsV2` call, unprompted, on every listing.

Measured with Cyberduck CLI 9.5.3 (45464) against `backlot serve` over the bundled corpus, path-style, SigV4, a connection profile whose `Context` is `/s3`:

```
GET /s3/acme-artifacts?location=                                              200 OK
GET /s3/acme-artifacts?encoding-type=url&max-keys=1000&prefix=&delimiter=%2F  200 OK
GET /s3/acme-artifacts?delimiter=%2F&uploads=                                 501 Not Implemented
```

The same pair repeats for a prefix: `...&prefix=runbooks%2F&delimiter=%2F&uploads=` is refused the same way.

Real S3 answers this. The API reference for ListMultipartUploads states "If the action is successful, the service sends back an HTTP 200 response", and of the `Upload` element, "A response can contain zero or more `Upload` elements". A bucket with no in-progress multipart uploads therefore answers 200 with a `ListMultipartUploadsResult` carrying no `Upload` children — not an error.

The S3 baseline already records this operation as `missing_operation`, with the note that the 501 "is an honest answer for an operation Backlot does not serve". That reasoning holds for a programmatic caller, which reaches `?uploads` only by asking for it. It does not hold for a GUI client, which asks on every folder open. The gap is on the hot path of the most common interaction a desktop S3 client has.

The answer is cheap and has no state behind it. Data enters Backlot through `backlot import` and never through the served API, so there is never an upload in progress and the faithful response is always the empty one: echo `Bucket`, `Prefix`, `Delimiter` and `EncodingType` from the request, `MaxUploads` defaulting to 1000, `IsTruncated` false, empty `KeyMarker` and `UploadIdMarker`, and no `Upload` elements.

Scope of what this run establishes. Cyberduck tolerated the refusal and rendered the listing anyway, so no flow is broken today; this is a divergence on the hot path that a stricter client need not tolerate. Everything else the client exercised matched: connect and SigV4 auth, `ListAllMyBuckets`, `ListObjectsV2` with prefix and delimiter, `GetBucketLocation`, `HEAD` and `GET` on an object, and a completed download. `HEAD /{bucket}/{prefix}/` returning 404 is what real answers too. The other 40 refusals the baseline records were not exercised by this client — a GUI Info panel would likely reach `?acl`, but that was not measured here.

nuri-yoo:
Closes #169. Fourteen commits on `main` at `50c0419`: `fede480`, the router, its tests, the S3 baseline and the operation table in `docs/supported-sources.md`; `a82adc7` and `be547e3` from a first and second self-review, one guard and prose; `868112d` and `5e07325` from a third, described under **What the third self-review changed** below; `c4cbc83`, `8fe3984`, `70aebb2` and `996eb24` from khj809's review, described under **What the review changed**; `dca9180`, three test comments reworded at the approval so they name the client behind #169 by what it does; and four merges of `main`, `7afedd2`, `ba513cf`, `4b408db` and `361bd69`. No open PR touches `backlot/routers/s3.py` or `tests/test_s3.py`; #165 and #175, which edited other sections of `docs/supported-sources.md`, have merged and are in the branch.

**What real S3 does.** Measured on 2026-09-10 against a general purpose bucket in ap-northeast-2 with no upload in progress, path-style, SigV4, the requests built the way `tests/test_s3.py`'s `_sign_get` builds them (form-decoded, then re-encoded with `quote`), so a value reaches real the way it reaches Backlot's test server. `GET /{bucket}?uploads` and `?uploads=` (Cyberduck's spelling, from #169's trace) are both 200 `application/xml`:

```
<?xml version="1.0" encoding="UTF-8"?>
<ListMultipartUploadsResult xmlns="http://s3.amazonaws.com/doc/2006-03-01/"><Bucket>brekkylab.com</Bucket><KeyMarker></KeyMarker><UploadIdMarker></UploadIdMarker><NextKeyMarker></NextKeyMarker><NextUploadIdMarker></NextUploadIdMarker><MaxUploads>1000</MaxUploads><IsTruncated>false</IsTruncated></ListMultipartUploadsResult>
```

The four markers are present and empty even though nothing is truncated, and there is no `Upload` and no `CommonPrefixes`: "A response can contain zero or more Upload elements" (the API reference on ListMultipartUploads). What was sent comes back in a fixed order, `Delimiter` then `Prefix` (`?uploads&prefix=runbooks/&delimiter=/` answers `<Delimiter>/</Delimiter><Prefix>runbooks/</Prefix>` before `MaxUploads`), each only when sent non-empty (`prefix=&delimiter=&key-marker=` echoes none of them); `key-marker` is echoed in `KeyMarker`; `upload-id-marker` alone is ignored, as the API reference says, and beside a `key-marker` it is refused, `InvalidArgument` "Invalid uploadId marker", for `xyz` and for the API reference's own example id alike; `EncodingType` is echoed as sent (`URL` comes back `URL`) after `MaxUploads` and before `IsTruncated`, and under it `Prefix`, `Delimiter` and `KeyMarker` come back encoded: letters, digits, `-_.*/` unchanged, space as `+`, and `!'()~,;:@="<>|%` and the bytes of `한글` as `%XX` in upper case (`run books/x+y%z` is `run+books/x+y%25z`, `|` is `%7C`, `한글` is `%ED%95%9C%EA%B8%80`), which is Java's `URLEncoder` with `/` added to its safe set and not Python's `quote_plus`, which keeps `~` and encodes `*`. Any other `encoding-type`, the empty value included, is `InvalidArgument` "Invalid Encoding Method specified in Request". `max-uploads` is read for its value and served at 1000 past it (`1001` and `2000` both echo 1000, "The limit of 1,000 multipart uploads is also the default value" in botocore's operation documentation); leading zeros come off and the length of the run does not matter (`05`, `00000000005` and twenty zeros ahead of a 5 all echo 5, five thousand zeros echo 0, `00002147483647` echoes the cap and only `00002147483648` is refused); the sign is read with the digits (`-0` and `-00` echo 0); empty is the default; a value that fits an int32 and is below 0 (`-1`, `-2147483648`) is `InvalidArgument` "Argument max-uploads must be an integer between 0 and 2147483647" naming the value as parsed (`-01` as `-1`), and a word, a leading space or digits that do not fit an int32 in either direction (`2147483648`, `-2147483649`) is `InvalidArgument` "Provided max-uploads not an integer or within integer range" naming the value as sent; each refusal carries `ArgumentName` and `ArgumentValue`. A parameter sent twice is read for its first value: `?uploads&max-uploads=1&max-uploads=abc` is the page at 1 and the same two values the other way round the refusal of `abc`, and `encoding-type`, `key-marker`, `upload-id-marker`, `prefix` and `delimiter` go the same way. The checks are ordered around the bucket lookup: `?uploads&max-uploads=abc` on a bucket that does not exist is the 400, `?uploads&max-uploads=-1`, `?uploads&encoding-type=bogus` and `?uploads&key-marker=a&upload-id-marker=b` on it are `NoSuchBucket`, on a bucket that exists `encoding-type` is checked before the `max-uploads` range and the range before the markers, and `?uploads&versioning` is the conflict, "Conflicting query string parameters: uploads, versioning", bucket or no bucket.

**What Backlot did at `364e071`.** `uploads` sat in `_BUCKET_SELECTORS` with the twenty-odd selectors Backlot refuses and fell to `_not_implemented`, so #169's 501 on every folder open; the S3 baseline acknowledged it as a `missing_operation` with the note that real answers 200 (measured 2026-09-03).

**What changed.** `bucket_get` gains one branch beside `location`: `selected == ["uploads"]` answers `_list_multipart_uploads(request, bucket, max_uploads)`, with `_max_uploads` run before `_bucket_visible` because real parses it before the bucket exists, and its range and the other parameters checked after. `_first` reads a parameter's first value, as real does, where Starlette's `QueryParams.get` returns the last, and the six reads on the route go through it. `_url_encode` is real's `encoding-type=url` as measured (the safe set, `+` for space, `%XX` upper case). `_argument_error` builds the `ArgumentName`/`ArgumentValue` envelope, and `_conflict`, which built it inline, calls it. `_P_BUCKET_GET` declares `uploads` the way it declares `location`, "present, at any value"; the four ListMultipartUploads parameters (`max-uploads`, `key-marker`, `upload-id-marker`, `encoding-type`) are not declared, but not for `list-type`'s reason — three of them come back echoed and three can turn the 200 into an `InvalidArgument`. What none of them does is decide which uploads come back, because there are never any, so declaring them would offer a marker to resume from and a page size over a listing with no second page. The comment above the list says that. The baseline drops the one entry, rewritten by `backlot diff --source s3 --update-baseline`; the `measured` date moves to 2026-09-10 and nothing else in the file changes. `docs/supported-sources.md` gains the row; the generated table is untouched.

**What #169 named and what the measurement added.** The issue named the answer: echo `Bucket`, `Prefix`, `Delimiter` and `EncodingType`, `MaxUploads` defaulting to 1000, `IsTruncated` false, empty `KeyMarker` and `UploadIdMarker`, no `Upload`. All of it is here as measured. The measurement added what the issue did not name: the element order and the two next-markers present and empty; `key-marker` echoed and `upload-id-marker` ignored alone and refused beside it; `max-uploads` parsing, its cap and its two messages; the first value of a repeated parameter; `encoding-type` refused unless `url` (`URL` passes too, the two spellings measured), and what it encodes; the check order across the bucket lookup.

**One judgement call.** `upload-id-marker` beside `key-marker`. Real refused both ids tried on a bucket with no upload in progress; whether it would also refuse an id that does name an upload cannot be told apart on such a bucket, and Backlot never has an upload, so every such request takes the refusal here. The docstring on `_list_multipart_uploads` says exactly that, including what was and was not measured.

**Tests.** Seven new in `tests/test_s3.py`, under a `ListMultipartUploads` header that states where and when the answers were measured. `test_list_multipart_uploads_is_the_empty_page_real_serves_byte_for_byte`: `?uploads` and `?uploads=` against real's body with the bucket name swapped in, byte for byte, XML declaration included. `test_list_multipart_uploads_echoes_what_was_sent_in_reals_order`: the children as `(tag, text)` in document order for Cyberduck's two queries from the issue (`delimiter=%2F&uploads=` and `encoding-type=url&prefix=runbooks%2F&delimiter=%2F&uploads=`), for empty values, for the markers, and `MaxUploads` for ten `max-uploads` values, six of them zero-padded. `test_list_multipart_uploads_encodes_under_encoding_type_url_as_real_does`: `Prefix`, `KeyMarker` and `Delimiter` raw without `encoding-type` and encoded under `URL`, `한글` included. `test_list_multipart_uploads_refuses_what_real_refuses_with_reals_messages`: each refusal with real's message, `ArgumentName` and `ArgumentValue` (a 5000-digit `max-uploads` among them, the row `a82adc7` adds; `00002147483648`, the row `868112d` adds; the negatives split across the two messages and `-01` named as `-1`, the rows `8fe3984` and `996eb24` add), and the check order on a missing bucket and between checks. `test_list_multipart_uploads_reads_the_first_of_a_repeated_parameter_as_real_does`: each parameter sent twice, in both orders where the order decides the status. `test_list_multipart_uploads_on_a_bucket_the_caller_cannot_see_is_no_such_bucket`: an engineer is told `people-vault` does not exist, as the listing tells them, while admin gets the page. `test_boto3_list_multipart_uploads_is_an_empty_page_not_a_client_error`: botocore parses the page with `Prefix` and `Delimiter`, `MaxUploads` 1000, `IsTruncated` False and no `Uploads` key. Three existing tests move: `uploads` leaves `BUCKET_SUBRESOURCES` (it is no longer refused, so the 501 row for it would fail), the conflict test adds `?versioning&uploads`, and the HEAD test adds `?uploads`, still 405 since no sub-resource has a HEAD form.

Mutation-checked from the committed tree, each restored with `git checkout`. At `a82adc7`, over the whole of `tests/test_s3.py`: the router at `main` with the tests kept fails the six and nothing else (`6 failed, 65 passed`), and the digit-length guard removed fails the refusal test on the 5000-digit row with the 500 that `int()` raising would have served. At `fede480`: `Delimiter` and `Prefix` swapped fails the order test; `max-uploads` parsed with `int()` fails the refusal test on ` 5`; `_max_uploads` moved after the bucket lookup fails the refusal test on the missing-bucket row. At `868112d`: the leading-zero strip put back to `raw` fails the echo test on `00000000005`. At `8fe3984`: `_first` reading the last value fails the repeated-parameter test and nothing else; the sign judged by shape fails the echo test on `-0` and the refusal test on `-2147483649`. At `996eb24`: the range check put back before the bucket lookup fails the refusal test on the missing-bucket `-1` row.

**Verification.**

```
2905 passed, 1 skipped   (2026-09-11 at 996eb24 on main 50c0419; skipped: the hubspot reader not installed)
ruff format --check .   139 files already formatted
PYTHONPATH=. python scripts/gen_docs.py --check   exit 0
PYTHONPATH=. python -m backlot diff --source s3
42 divergence(s) · 42 acknowledged · 0 new, and nothing stale (the same run against 364e071's baseline reports the ListMultipartUploads entry as "acknowledged but no longer diverging")
```

The same five commands ran at the earlier commits, against whichever `main` each sat on: `2649 passed, 4 skipped` at `fede480`, `a82adc7` and `be547e3`, where the fourth skip was Docker being down for the one mcp test; `2868 passed, 3 skipped` at `7afedd2`, `868112d` and `5e07325`; `2898 passed, 3 skipped` at `ba513cf` in khj809's worktree; and `2905 passed, 1 skipped` at `70aebb2`. `c4cbc83` was applied from the review and ran only in CI.

**What the third self-review changed.** Two commits, one of them a behaviour change. `868112d`: `_max_uploads` judged `max-uploads` by the length of its digits, `len(raw) > 10`, so `00000000005` — eleven characters worth five — was refused as "Provided max-uploads not an integer or within integer range", which describes neither thing it is. Real judges the value, measured against the same bucket on the same day: `00000000005` and twenty zeros ahead of a 5 both answer 5, five thousand zeros answer 0, `00002147483647` answers the cap, and only `00002147483648` is refused. Stripping the zeros before the range check is also what keeps `int()` off a run of digits it will not parse, which is what the length check was there for, so five thousand nines is still refused. `5e07325`, prose and one call: `bucket_get`'s docstring still called `location` "S3's other GET on this same path" and never named `?uploads`, and that docstring is the operation's description in the OpenAPI document and the MCP tool's description, so an agent was told there was one other GET while `uploads` sat in the parameter list beside `location`; the comment above `_P_BUCKET_GET` gave the four ListMultipartUploads parameters `list-type`'s reason, "none changes what a caller gets", which the tests in this branch already contradicted; and `_conflict` now calls `_argument_error` rather than building the same envelope inline, with `_argument_error` moved up beside it.

**What the review changed.** khj809's review at `ba513cf` made four findings, all measured against real on 2026-09-11 and all taken. `c4cbc83`, applied from the review: `max-uploads` read for its first value. `8fe3984`: `_max_uploads` judged a leading `-` by shape, refusing `-0` and giving `-2147483649` the range message; real judges the value (`-0` is 0, `-2147483648` is the range message, `-2147483649` is "not an integer"), and the parser is the one the review proposed, which matches all 22 inputs measured. The same commit adds `_first` and puts the six reads on the route through it, since Starlette's `QueryParams.get` returns the last of a repeated parameter and real reads the first (the review's rows, measured in both directions: this branch and real were exact mirrors). `70aebb2`, prose: `docs/fidelity.md` and `s3_probe.py` said the four catch-all routes "declare no query parameters", and `GET /{bucket}` declares seven, `uploads` the seventh as of this branch; the sentence now gives the reason that holds, that operations are selected by the query string. `996eb24`, from measuring further along the review's model: real's two `max-uploads` checks straddle the bucket lookup (`abc` on a missing bucket is the 400, `-1` on it is `NoSuchBucket`), and the range message names the value as parsed (`-01` as `-1`); the range refusal moved into `_list_multipart_uploads`, after `encoding-type` and before the markers, as measured. The review's third finding, a negative `max-keys` on the listing being a 500 or a listing with no page cap, is filed as #192 with the real side measured.

**Not in this PR.** Backlot's ListObjectsV2 does not read `encoding-type` (`_list_objects_v2` reads `prefix`, `delimiter`, `start-after`, `max-keys` and `continuation-token`), and Cyberduck sends `encoding-type=url` on every listing (#169's trace), so a client that asked for `url` gets keys raw and no `EncodingType` element; real's encoded listing was not measured here, and `_url_encode` is what a fix would reuse. That, and the listing still reading the last of a repeated parameter where `?uploads` now reads the first, is #178. The listing's `max-keys` parser, which `_max_uploads` is the model for, is #192.


backlot/routers/s3.py
```diff
@@ -469,6 +527,110 @@ def _list_objects_v2(request: Request, conn, bucket: str, visible) -> Response:
...
+    and served at ``_MAX_UPLOADS`` past it; a value with a leading ``-`` draws "Argument
+    max-uploads must be an integer between 0 and 2147483647"; anything else — a word, a leading
+    space, a value whose digits come to more than 2147483647 — draws "Provided max-uploads not an
+    integer or within integer range". Not ``int()``, which accepts `` 5`` and ``+5`` and has no
+    ceiling.
+    """
+    raw = q.get("max-uploads", "")
+    if raw == "":
+        return _MAX_UPLOADS, None
+    if re.fullmatch(r"-[0-9]+", raw):
```
khj809: **`-0` and a negative outside int32 range are judged by the shape of the string, not its value — the same error `868112d` fixed on the positive branch.**

`re.fullmatch(r"-[0-9]+", raw)` routes every leading-`-` value to the range message. Measured against this branch:

```
max-uploads=-0                    -> 400 "Argument max-uploads must be an integer between 0 and 2147483647"
max-uploads=-00                   -> 400  (same)
max-uploads=-2147483649           -> 400  (same)
max-uploads=-99999999999999999999 -> 400  (same)
```

The two messages you measured separate into a model, and the PR body is what supplies it: "Argument max-uploads must be an integer between 0 and 2147483647" is *parsed as an int32, but out of range* (`-1`), while "Provided max-uploads not an integer or within integer range" is *not parseable as an int32* — which is why `2147483648`, a perfectly good integer, draws the second and not the first.

Under that model both ends of the regex are wrong. `-0` parses to 0, which is in range, so it belongs with `05` and `00000000005` as a 200 carrying `<MaxUploads>0</MaxUploads>`. And `-2147483649` does not fit an int32 at all, so it belongs with `2147483648` on the second message, not the range one. That is the `00000000005` case of `868112d` mirrored onto the sign: the leading zeros were fixed to judge the value, and the leading minus was left judging the shape.

Neither input appears in the measurement list, and I cannot measure from here, so this is a question rather than a claim. Two requests against the same ap-northeast-2 bucket settle it:

```
GET /{bucket}?uploads&max-uploads=-0
GET /{bucket}?uploads&max-uploads=-2147483649
```

`-2147483648` is worth a third — it is the one negative that does fit an int32, so it is where the two messages meet.

Deliberately **not** a `suggestion` block: one-click applying a rule about real S3 ahead of the measurement is the thing AGENTS.md warns against. If the model holds, the shape is:

```python
    raw = q.get("max-uploads", "")
    if raw == "":
        return _MAX_UPLOADS, None
    negative = raw.startswith("-")
    digits = (raw[1:] if negative else raw).lstrip("0") or "0"
    # int32, so a negative reaches one further than a positive does.
    ceiling = _INT32_MAX + 1 if negative else _INT32_MAX
    if not re.fullmatch(r"-?[0-9]+", raw) or len(digits) > 10 or int(digits) > ceiling:
        return 0, _argument_error(
            "Provided max-uploads not an integer or within integer range",
            "max-uploads",
            raw,
            resource,
        )
    if negative and digits != "0":
        return 0, _argument_error(
            f"Argument max-uploads must be an integer between 0 and {_INT32_MAX}",
            "max-uploads",
            raw,
            resource,
        )
    return min(int(digits), _MAX_UPLOADS), None
```

The docstring needs the same move: "a value with a leading `-` draws ..." is the shape rule written down, so it would become a rule about the value the digits come to.

khj809 (commented):
Reviewed at `ba513cf`, re-run in a worktree rather than read: full suite `2898 passed, 3 skipped`, `ruff check` and `ruff format --check` clean, `scripts/gen_docs.py --check` exit 0, and `backlot diff --source s3` reports `42 divergence(s) · 42 acknowledged · 0 new`, so the baseline entry drops cleanly and nothing goes stale behind it.

#169 is answered. Both Cyberduck spellings from the trace (`?delimiter=%2F&uploads=` and the prefixed one) come back 200 with real's body; `uploads` stays in `_BUCKET_SELECTORS` so the conflict and the HEAD 405 still hold; and the `_conflict` → `_argument_error` refactor is byte-identical on the wire, which I checked against a live server rather than trusting the tests. Omitting `Prefix`/`Delimiter` when empty departs from the wording of the issue and follows the measurement instead, which is the right way round. The table row in `docs/supported-sources.md` is outside the generated markers (lines 15–29), so hand-editing it is correct, and its text matches the code.

One thing to state plainly: **this machine has no AWS credentials, so I could not measure real S3.** Nothing below claims what real answers. The first comment is a question about two inputs your own measurement list does not cover; everything else is measured against this branch.

Three inline comments follow. A fourth finding has no line in this diff to hang off:

### `docs/fidelity.md:119` and `backlot/fidelity/s3_probe.py:7` — "four catch-all routes that declare no query parameters" is not true, and this PR makes it less true

Both files carry the same sentence:

> Backlot serves all of them from four catch-all routes that declare no query parameters and read the string themselves, so a path-and-parameter diff pairs every S3 operation with the same route and reports a clean match every time.

`GET /s3/{bucket}` declares seven query parameters in the served document — I built it and counted:

```
GET /s3/{bucket}: 7 query params -> ['prefix', 'delimiter', 'start-after', 'max-keys',
                                     'continuation-token', 'location', 'uploads']
```

`uploads` is the seventh, added here. The drift predates this PR (`location` was already declared), but the sentence is load-bearing: it is the stated reason S3 is probed instead of diffed, and a reader who checks it finds the opposite. The real reason is that operations are selected by the query string, not that the routes declare nothing — which survives however many parameters get declared. Dropping four words fixes both copies:

```diff
-Backlot serves all of them from four catch-all routes that declare no
-query parameters and read the string themselves, so a path-and-parameter diff pairs every S3
+Backlot serves all of them from four catch-all routes that read the query
+string themselves, so a path-and-parameter diff pairs every S3
 operation with the same route and reports a clean match every time
```

backlot/routers/s3.py
```diff
@@ -469,6 +527,110 @@ def _list_objects_v2(request: Request, conn, bucket: str, visible) -> Response:
...
+
+    Absent or empty is the default; a run of digits is read for its value, leading zeros and all
+    (``05`` and ``00000000005`` are both 5, twenty zeros and a 5 is 5, five thousand zeros is 0),
+    and served at ``_MAX_UPLOADS`` past it; a value with a leading ``-`` draws "Argument
+    max-uploads must be an integer between 0 and 2147483647"; anything else — a word, a leading
+    space, a value whose digits come to more than 2147483647 — draws "Provided max-uploads not an
+    integer or within integer range". Not ``int()``, which accepts `` 5`` and ``+5`` and has no
+    ceiling.
+    """
+    raw = q.get("max-uploads", "")
```
khj809: **A repeated query parameter turns real's 200 into a 400 on this route, and the measurement that says so is already in the PR body.**

`QueryParams.get` returns the LAST value of a repeated key, and "Not in this PR" reports that real takes the FIRST (`?uploads&prefix=a&prefix=b` echoes `a`). On the listing that only changes which `Prefix` is echoed, which is why #178 is a fair home for it. Here it changes the status code. Measured against this branch:

```
?uploads&max-uploads=1&max-uploads=abc                      -> 400 InvalidArgument, ArgumentValue=abc
?uploads&max-uploads=abc&max-uploads=1                      -> 200 MaxUploads=1
?uploads&encoding-type=url&encoding-type=bogus              -> 400 InvalidArgument, ArgumentValue=bogus
?uploads&encoding-type=bogus&encoding-type=url              -> 200
?uploads&key-marker=k&upload-id-marker=&upload-id-marker=x  -> 400 "Invalid uploadId marker"
```

The first two rows are the same two values in opposite order and land on opposite statuses — a clean demonstration that it is last-wins, and that real serves 200 for both. A client that appends a query parameter rather than replacing it gets a hard failure on a call Cyberduck makes on every folder open, which is the whole reason this route exists.

Since real's behaviour here is already measured, this one can be closed rather than filed:

```suggestion
    raw = next(iter(q.getlist("max-uploads")), "")
```

and the five reads in `_list_multipart_uploads` (lines 605, 610, 611, 616, 617):

```python
    encoding_type = next(iter(q.getlist("encoding-type")), None)
    key_marker = next(iter(q.getlist("key-marker")), "")
    upload_id_marker = next(iter(q.getlist("upload-id-marker")), "")
    prefix = next(iter(q.getlist("prefix")), "")
    delimiter = next(iter(q.getlist("delimiter")), "")
```

Six of those read better as a helper beside `_selected`, which also gives the measurement somewhere to live:

```python
def _first(q, name: str, default=""):
    """The FIRST of a repeated query parameter, which is the one real S3 reads (measured:
    ``?uploads&prefix=a&prefix=b`` echoes ``a``); ``QueryParams.get`` returns the last."""
    values = q.getlist(name)
    return values[0] if values else default
```

That leaves #178 to the listing route alone, where the divergence really is echo-only.

backlot/routers/s3.py
```diff
@@ -338,11 +387,20 @@ async def bucket_get(request: Request, bucket: str):
...
     if not _bucket_visible(conn, bucket, visible):
         return _error("NoSuchBucket", "The specified bucket does not exist", bucket)
     if selected == ["location"]:
         # us-east-1 is represented by an *empty* LocationConstraint element on real S3.
         return _xml(f'<LocationConstraint xmlns="{NS}"></LocationConstraint>')
+    if selected == ["uploads"]:
+        return _list_multipart_uploads(request, bucket, max_uploads)
     if selected:
         return _not_implemented(selected[0], f"/{bucket}")
     return _list_objects_v2(request, conn, bucket, visible)
```
khj809, line 417: **Not this PR's change, but this is its call site, and the correct parser now sits two functions above it: a negative `max-keys` either 500s or drops the page cap.**

`_list_objects_v2` (line 415) does `min(int(q.get("max-keys", _MAX_KEYS)), _MAX_KEYS)` and never rejects a negative, so `max_keys` goes negative and is then used both as a slice index and as the SQL LIMIT. Measured against this branch:

```
?max-keys=-1                  -> 500 Internal Server Error
?max-keys=-3                  -> 500
?max-keys=-2&prefix=design/   -> 500
?max-keys=-2                  -> 200 <MaxKeys>-2</MaxKeys><IsTruncated>true</IsTruncated>
```

The 500 is `overflow_row = rows[max_keys]` reading `rows[-1]` off an empty list, surfacing as a bare `Internal Server Error` with no S3 `<Error>` envelope at all — a boto3 caller retries it and then reports a 500.

The 200 is quieter and worse. `limit = max_keys + 1` is `-1`, and SQLite reads a negative LIMIT as no limit (`select * from t order by k limit -1` returns every row — checked), so `store.list_s3_objects` walks every ACL-visible row in the bucket before `rows[:max_keys]` throws most of them away. The 1000-key cap, and the keyset paging the comment above it is careful to explain, are both simply gone for that request.

This PR writes exactly the parser this wants — `_max_uploads`, with the two-message split and the `ArgumentName`/`ArgumentValue` envelope — so as of this branch the two integer parameters on the same route sit at very different fidelity, and the less faithful one is the one every client sends. Better as a follow-up issue than as growth on this PR: measure `max-keys` against the same bucket for `-1`, `abc`, ` 5`, `05` and `2147483648` (the messages will name this parameter, so they need their own measurement rather than reuse of the `max-uploads` strings), then give `_max_uploads` the parameter name and its message pair as arguments and call it from both.

backlot/routers/s3.py
nuri-yoo: Measured 2026-09-11 against the same ap-northeast-2 bucket, the three you named and three more:

```
max-uploads=-0                    -> 200 <MaxUploads>0</MaxUploads>
max-uploads=-00                   -> 200 <MaxUploads>0</MaxUploads>
max-uploads=-2147483648           -> 400 "Argument max-uploads must be an integer between 0 and 2147483647"
max-uploads=-2147483649           -> 400 "Provided max-uploads not an integer or within integer range"
max-uploads=-99999999999999999999 -> 400 "Provided max-uploads not an integer or within integer range"
max-uploads=-abc, max-uploads=-   -> 400 "Provided max-uploads not an integer or within integer range"
```

The model holds on every row: parsed as an int32 and out of range draws the first message, not parseable as an int32 the second, and `-2147483648` is where they meet. Your parser as written matches all 22 inputs measured so far (the 16 in the body and these 6), so `8fe3984` takes it, with the docstring rewritten as a rule about the value the digits come to; the echo test takes `-0` and `-00`, and the refusal test splits the negatives across the two messages.

backlot/routers/s3.py
nuri-yoo: Taken in `8fe3984`: `_first` beside `_selected`, and the six reads on the route go through it. One correction to the measurement paragraph, since I could measure the real side: real is first-wins, so it does not serve 200 for both of your first two rows. `?uploads&max-uploads=abc&max-uploads=1` is a 400 on real too, `ArgumentValue` `abc`; real and this branch were exact mirrors of each other, which the fix removes either way. Measured 2026-09-11, same bucket:

```
max-uploads=1&max-uploads=abc                     -> 200 MaxUploads=1
max-uploads=abc&max-uploads=1                     -> 400 InvalidArgument, ArgumentValue=abc
encoding-type=url&encoding-type=bogus             -> 200 EncodingType=url
encoding-type=bogus&encoding-type=url             -> 400 InvalidArgument, ArgumentValue=bogus
key-marker=k&upload-id-marker=&upload-id-marker=x -> 200
key-marker=k&upload-id-marker=x&upload-id-marker= -> 400 "Invalid uploadId marker", ArgumentValue=x
prefix=a&prefix=b&delimiter=/&delimiter=|&key-marker=a&key-marker=b -> Prefix=a, Delimiter=/, KeyMarker=a
```

`test_list_multipart_uploads_reads_the_first_of_a_repeated_parameter_as_real_does` sends each of them; with `_first` reading the last value it is the one test that fails. #178's row 5 is edited to name the listing route alone.

nuri-yoo:
The fourth finding, `docs/fidelity.md:119` and `backlot/fidelity/s3_probe.py:7`: confirmed, `_P_BUCKET_GET` declares the seven you counted. The four words are dropped from both copies in `70aebb2`, prose only, and `scripts/gen_docs.py --check` exits 0 after it.

backlot/routers/s3.py
nuri-yoo: One more thing the same measurement turned up, fixed in `996eb24`: the two checks straddle the bucket lookup. On a bucket that does not exist, `?uploads&max-uploads=abc` is the 400 and `?uploads&max-uploads=-1` is `NoSuchBucket`; on one that exists, `-1&encoding-type=bogus` is the encoding-type refusal and `-1&key-marker=a&upload-id-marker=b` the range one, so the order is parse, lookup, encoding-type, range, markers. And the range message names the value as parsed, `-01` as `-1`, where the other message names it as sent. The parse stays before the lookup and hands the value on; `_list_multipart_uploads` refuses a negative in its place. The refusal test has the missing-bucket row for `-1`, the three orderings and `-01`; with the range check put back before the lookup, that row is the one that fails.

backlot/routers/s3.py
nuri-yoo: Reproduced at this head (`-1` and `-3` are the 500, `-2` the 200 with `<MaxKeys>-2</MaxKeys>`), and the real side is now measured: filed as #192 with the table. The messages did need their own measurement, in a way neither of us would have guessed from `max-uploads`: the range message says `Argument maxKeys must be an integer between 0 and 2147483647` with `ArgumentName` `maxKeys`, camelCase, while the other says `Provided max-keys not an integer or within integer range` with `ArgumentName` `max-keys`. Also in the issue: `-0` is 0, `05` is 5, ` 5` and `2147483648` are refused, empty is the default, `1001` is echoed as sent and served at 1000 (on a bucket with more than 1000 keys), `max-keys=0` answers `IsTruncated` false where this branch answers true, and the same parse-before, range-after straddle of the bucket lookup that `996eb24` gives `max-uploads`. The fix is the one you describe, `_max_uploads` taking the name and the message pair, with the echo and the LIMIT clamped separately.

khj809 (approved):
LGTM

one minor thing to ask: could you remove "Cyberduck-related" comments from `test_s3.py`? It's true that I found this issue via Cyberduck, but don't want to emphasize it too much.

nuri-yoo:
Done in `dca9180`: the three comments in `tests/test_s3.py` that named the client now say what it does ("a browsing client sends `?uploads` after every ListObjectsV2", "the two queries in #169's trace", "with and without the `=`").

nuri-yoo:
And the docstring one too, in `320a6da`: `_list_multipart_uploads` now says "a browsing client that sends `?uploads` after every ListObjectsV2". No file under `backlot/`, `tests/` or `docs/` names the client any more.

PR #181, opened 2026-09-10
Fix scheduled fidelity check workflow

khj809:
`Fidelity: <source> diverges from its vendor` was one issue per source for the life of the repository, found again on later runs whether it was open or closed. A source that diverged again after its issue had been triaged and closed had that issue reopened and its body overwritten with the new report. #141 was reopened that way on 2026-09-09, and the comments on it now answer a report that is no longer there.

A run now reads only the open issue. A closed one is left as the account of what was triaged on it — never reopened, never written over. A source that still diverges after its issue is closed gets a new issue instead, titled with the date the divergence was first seen (`Fidelity: linear diverges from its vendor (first seen 2026-09-10)`), whose body links back with `Previously reported in #N.`

The open issue is found by title prefix among the bot's own `fidelity`-labelled open issues, listed under the label rather than searched: the listing is immediate, where issue search is indexed asynchronously. The back-link is recomputed on every run rather than written once at creation, because the body is rewritten on every run — a link that came and went would rewrite the body and comment about the rewrite each night. Scoping that lookup to this source's own titles keeps it answering the same way as closed fidelity issues pile up.

#141 is retitled by hand to carry its own first-seen date. It is still matched by the prefix, so it is neither orphaned nor duplicated.

The schedule moves from 06:17 to 01:00 UTC, so a run lands at the end of a Silicon Valley day — 18:00 in Pacific daylight time, 17:00 in standard time — and at 10:00 the same morning in Seoul, rather than at 15:17 in Seoul and 23:17 the previous night in California.

Verified by running the step's shell against a stubbed `gh` over seven cases: nothing filed yet; a closed issue only, where it stays untouched and the new issue carries the back-link; an open issue whose report changed, and one whose report did not; a person's own issue on the same subject, which is left alone; the issue this step opened before titles carried a date, which is still found; and a report over the 55000-character cap.

### DIFF
--- a/.github/workflows/fidelity.yml
+++ b/.github/workflows/fidelity.yml
@@ -8,7 +8,9 @@ name: Fidelity
 
 on:
   schedule:
-    - cron: "17 6 * * *"
+    # 01:00 UTC is 18:00 the previous day in Pacific daylight time and 17:00 in standard time — the
+    # end of a Silicon Valley day either way — and 10:00 the same morning in Seoul.
+    - cron: "0 1 * * *"
   workflow_dispatch:
     inputs:
       source:
--- a/.github/workflows/fidelity.yml
+++ b/.github/workflows/fidelity.yml
@@ -53,7 +53,7 @@ jobs:
   compare:
     needs: sources
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


nuri-yoo (approved):
Approving. Pretty straightforward and has been known issue.

PR #183, opened 2026-09-10
example: Pin mirage version to >=0.0.6

### LINKED ISSUE #179: mirage: both patchers target a module 0.0.6 removed, so every Google and GitHub example exits 1 while the tripwires skip instead of failing
**What you ran** — the shipped mirage examples on a stock install of current `main` (`1494a42`), the way `examples/using-mirage/README.md` gives them. `pip install -e ".[all]"` resolves `mirage-ai[fuse]~=0.0.5` to **0.0.6**, released 2026-09-03.

```bash
pip install -e ".[all]"
python examples/using-mirage/gdrive.py
python examples/using-mirage/gmail.py
python examples/using-mirage/github.py
python examples/using-mirage/unified.py
```

**What happened** — all four exit 1 before reaching mirage, in `backlot/integrations/mirage.py`:

```
gdrive.py    exit 1   ModuleNotFoundError: No module named 'mirage.core.google._client'
gmail.py     exit 1   ModuleNotFoundError: No module named 'mirage.core.google._client'
github.py    exit 1   ModuleNotFoundError: No module named 'mirage.core.github._client'
unified.py   exit 1   ModuleNotFoundError: No module named 'mirage.core.google._client'
```

`slack.py` still passes, and `notion.py` and `s3.py` call no patcher either. Those three are exactly the scripts whose mirage resources take a base URL in their own config.

mirage 0.0.6 moved the host constants out of the private client module and into a public one. Every name Backlot rebinds still exists and still means the same thing; only the module holding them changed:

| Constants | mirage 0.0.5 | mirage 0.0.6 |
|---|---|---|
| `TOKEN_URL`, `GMAIL_API_BASE`, `DRIVE_API_BASE`, `DRIVE_UPLOAD_BASE`, `DOCS_API_BASE`, `SHEETS_API_BASE`, `SLIDES_API_BASE`, `CALENDAR_API_BASE`, `FORMS_API_BASE` | `mirage.core.google._client` | `mirage.core.google.constants` |
| `API_BASE`, `API_VERSION` | `mirage.core.github._client` | `mirage.core.github.constants` |

`_rebind_mirage_constants` imports its `source_module` by name before sweeping, so the import is what raises. The guard that was built for this — `RuntimeError("mirage's constants moved")` when a name is gone — never runs, because the module is gone before any name is looked up.

**Why CI is green anyway** — the two tripwire tests gate on `pytest.importorskip("mirage.core.google._client")`, which is the very module that disappeared. A rename therefore skips them rather than failing them. On the last green run of `main` (`1494a42`, https://github.com/brekkylab/backlot/actions/runs/34456144209) each of the four Python jobs reports:

```
SKIPPED [1] tests/test_integrations.py:211: could not import 'mirage.core.google._client': No module named 'mirage.core.google._client'
SKIPPED [1] tests/test_integrations.py:244: could not import 'mirage.core.google._client': No module named 'mirage.core.google._client'
2864 passed, 2 skipped, 4 warnings
```

So the gate reports its own subject missing as a reason to sit out. `test_mirage_patchers_rebind_every_constant_they_name` says in its docstring that it exists "since the raise above is only as good as the set it checks", and `test_mirage_patchers_name_every_host_constant_mirage_ships` is meant to fail "when mirage grows one, which is the moment to decide whether Backlot serves it". Both intend to catch a mirage upgrade. Neither catches the upgrade that happened. An `importorskip` is right for a gate on an optional dependency; it is wrong when the module path itself is the thing under test.

**A second break inside the same pin** — `~=0.0.5` admits 0.0.5 and 0.0.6, and `github.py` runs on neither. On 0.0.6 the patcher raises first. On 0.0.5 the patcher works and the script then dies constructing the resource, because `GitHubResource.__init__` took five more required positional arguments before 0.0.6 made them optional:

```
mirage 0.0.5   TypeError: GitHubResource.__init__() missing 5 required positional arguments:
               'owner', 'repo', 'ref', 'default_branch', and 'tree'
mirage 0.0.6   ModuleNotFoundError: No module named 'mirage.core.github._client'
```

`examples/using-mirage/github.py:131` is `GitHubResource(GitHubConfig(...))`, which is 0.0.6's signature. The example and the shim are pinned to different mirage versions, and the floor comment at `pyproject.toml:121` still explains a 0.0.5 floor chosen for constants 0.0.6 has moved.

**Where it lands**

- `backlot/integrations/mirage.py:84` and `:114` — the two module names.
- `backlot/integrations/mirage.py:109` — `point_github_at`'s docstring says patching the source module alone would do, because `API_BASE` is read as a module global at call time. In 0.0.6 `github/client.py:23` is `from mirage.core.github.constants import API_BASE, API_VERSION`, a by-value import, so the sweep the docstring calls symmetry is now load-bearing.
- `tests/test_integrations.py:190`, `:193`, `:211`, `:212`, `:221`, `:222`, `:244`, `:245` — the stubbed names and the two `importorskip` gates.
- `tests/test_packaging.py:135` and `:136` — `_IMPORT_TO_DIST_EXCEPTIONS`, which maps each gate's module to its distribution and is meant to be updated in the same diff as the gate.
- `pyproject.toml:121` and `:124` — the floor and the comment justifying it.

**The fix, measured** — renaming the module in those three files and raising the floor to `~=0.0.6`. The floor has to move: 0.0.5 has no `mirage.core.google.constants` at all, and its `mirage.core.github.constants` holds only `SCOPE_WARN` and `SCOPE_ERROR`, so one spelling cannot serve both releases. Raising it also gives `github.py` the constructor it is already written against. With that change applied locally against mirage 0.0.6:

```
examples/using-mirage/{gdrive,gmail,github,unified}.py      exit 0
pytest (all extras, Docker up, mirage 0.0.6)                2878 passed, 0 skipped
```

The alternative of accepting both spellings buys support for a release whose own `github.py` example cannot run, so it looks like cost without a reader.

**What a fix has to settle**

- Whether the gate stays an `importorskip` on the seam module. A hard `import` after `importorskip("mirage")` fails loudly on the next rename, which is what these tests were written to do; the current form converts that failure into a skip and a green run.
- mirage 0.0.6 grew first-class seams that would retire the monkeypatch: `GoogleConfig.api_base`, whose own comment says it is "used to point backends at a fake server", and `GitHubConfig.base_url`. They do not line up with Backlot's paths yet — 0.0.6 derives `{base}/v1` for both Docs and Slides, `{base}/v4` for Sheets and `{base}/token` for OAuth, while Backlot serves `/docs/v1`, `/slides/v1`, `/sheets/v4` and `/oauth2/token`, and one base cannot distinguish Docs from Slides. Moving to the config seam therefore needs either alias routes here or per-service bases upstream, which is a separate piece of work from getting the examples running again.
- Whether the version floor is worth a CI job that installs the lowest supported mirage, given the pin now has to track a young dependency's module layout.

**Where you ran it** — `main` at `1494a42`, Python 3.14.4, macOS 26.5.2 arm64, mirage-ai 0.0.6 and 0.0.5 in separate environments, Docker via colima 29.5.2 for the full-suite figure. Found while reviewing #165, whose body reads the two skips as third-party modules missing from published packages; they are present in 0.0.5 and absent from 0.0.6, so the skips mark the break rather than a gap in packaging.

I am happy to take this one if you would rather assign it to me than pick it up yourself.

khj809:
Closes #179.

mirage 0.0.6 moved the host constants both patchers rebind out of `mirage.core.{google,github}._client` into `mirage.core.{google,github}.constants`. Every name still exists and still means the same thing, so `_rebind_mirage_constants` reaches its `RuntimeError("mirage's constants moved")` guard for a renamed constant but never for a moved module: the `importlib.import_module(source_module)` on its first line raises first. Both patchers now name `…constants`, and `examples/using-mirage/{gdrive,gmail,github,unified}.py` exit 0 again.

The two tripwires written to catch exactly this upgrade gated on `pytest.importorskip("mirage.core.google._client")` — the module that disappeared — so the rename skipped them instead of failing them, and `main` stayed green with both patchers broken. `importorskip` is right for an optional distribution and wrong for the module path under test. A new `_mirage_constant_modules()` helper keeps the `importorskip` on `mirage` itself and hard-imports the two seam modules, so the next move fails where it is measured. `_IMPORT_TO_DIST_EXCEPTIONS` in `tests/test_packaging.py` follows the gate to `mirage` -> `mirage-ai`.

The floor is `>=0.0.6`. Earlier releases are not supported: 0.0.5 has no `mirage.core.google.constants` at all and its `mirage.core.github.constants` holds only `SCOPE_WARN`/`SCOPE_ERROR`, so one spelling cannot serve both. 0.0.6 is also the release that takes `GitHubResource(config)` alone — the signature `examples/using-mirage/github.py:131` already calls, against which 0.0.5 additionally required `owner`, `repo`, `ref`, `default_branch` and `tree`.

The extra is now `mirage-ai[fuse,s3]`, because `s3.py` could not run on the install its own header prescribes either. mirage's S3 backend does a module-scope `import aioboto3`, mirage ships that behind its own `s3` extra, and Backlot pulled only `[fuse]` — so the documented install never brought what the example imports, on 0.0.5 and 0.0.6 alike.

That extra is also why CI now installs with `uv sync --all-extras --locked` instead of `pip install -e ".[all]"`. aioboto3's newest release hard-pins `aiobotocore[boto3]==2.25.1`, which admits only botocore 1.40.46-1.40.61, while `boto3>=1.40` here has no upper bound; pip starts at the newest boto3 and backtracks through 215 releases toward that window before exiting `resolution-too-deep`, where uv derives the conflict and resolves in seconds. This is not a preference between tools — plain pip cannot install this project's `[all]`, and `pip install -e ".[examples,mirage]"` fails the same way, so CONTRIBUTING, the mirage README and every `examples/using-mirage/` header now name one command: `uv sync --all-extras --locked`. `--locked` holds uv.lock and pyproject.toml to each other, which a bare `uv sync` would hide by relocking; `--all-extras` is what keeps every gated test installed, and `uv run --no-sync pytest` runs the suite inside that environment. The Linear example's `PYTHON` now points at `.venv/bin/python`, since backlot is no longer on the ambient interpreter. `.[all]` stays defined and tested as the one-shot install for anyone taking Backlot off PyPI. Install went from failing at 7m51s to succeeding at ~25s, and the test job from 7.9 to 2.6 minutes.

Worth knowing before the boto3 floor next moves: it cannot pass 1.40.x while aioboto3 pins aiobotocore that way. `s3fs` 2026.7.0 accepts `aiobotocore>=2.19,<4.0` and holds its version, and `tests/test_s3.py`, `tests/test_sdk.py`, every `examples/using-fsspec/` script and every `examples/using-official-sdk/` script pass on the older stack.

Three prose claims about mirage were wrong wherever they appeared, and are corrected against the installed package rather than restated. That `GitHubConfig` has no `base_url` field: it has one, and had one in 0.0.5 too (`examples/using-mirage/github.py`'s module docstring and the comment in `build`, and the mirage README). That `API_BASE` is read as a module global at call time, so patching the source alone would do and the sweep is kept only for symmetry: `mirage/core/github/client.py:23` imports it by value and reads that copy in `github_url`, so the sweep is load-bearing for GitHub exactly as it is for Google (`point_github_at`'s docstring and the README). And that Google has no host config at all, its helpers returning the constants verbatim: `GoogleConfig.api_base` exists — its own comment calls it "used to point backends at a fake server" — and every helper reads `f"{base}/…" if base else CONSTANT` (the README plus `gdrive.py`, `gmail.py` and `unified.py`). What actually rules `api_base` out is the *single*: one base composes Docs, Slides and Forms onto the same `{base}/v1` while Backlot serves `/docs/v1` and `/slides/v1` apart. `GitHubConfig.base_url`, by contrast, is a real seam — mirage threads it to all 36 call sites of the four client functions that accept it — so there the rebind is what redirects a resource built without that field rather than the only way through.

Verified against mirage 0.0.6 with all extras synced and Docker up: all seven `examples/using-mirage/` scripts exit 0, and the suite is 2894 passed with the documented `llama_index.readers.hubspot` skip as the only skip — the two mirage skips that used to sit in `tests/test_integrations.py` are gone. The 0.0.5 comparisons above were measured from that release's wheel, and the pip failures from pip 26.2.1, not taken from the issue.

Retiring the monkeypatch for `GoogleConfig.api_base` / `GitHubConfig.base_url` is left for its own change: it needs either alias routes here or per-service bases upstream, which #179 separates from getting the examples running.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,19 +17,26 @@ jobs:
       - uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
         with:
           python-version: ${{ matrix.python-version }}
-          cache: pip
-          cache-dependency-path: pyproject.toml
       - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
           node-version: "22"
           cache: npm
           cache-dependency-path: examples/using-official-sdk/linear/package-lock.json
       - uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
+        with:
+          enable-cache: true
 
+      # `uv sync`, not `pip install -e ".[all]"`: mirage's S3 backend needs aioboto3, whose newest
+      # release hard-pins `aiobotocore[boto3]==2.25.1` and so admits only botocore 1.40.46-1.40.61,
+      # while `boto3>=1.40` here has no upper bound. pip starts at the newest boto3 and backtracks
+      # through 215 releases toward that window, giving up with `resolution-too-deep`; uv derives
+      # the conflict instead of searching for it and resolves in seconds. `--locked` fails the run
+      # when uv.lock and pyproject.toml disagree, so the lock cannot drift unnoticed, and
+      # `--all-extras` is what keeps every gated test installed.
       - name: Install
-        run: pip install -e ".[all]"
+        run: uv sync --all-extras --locked --python ${{ matrix.python-version }}
       - name: Install the HubSpot reader past its stale pin
-        run: pip install --no-deps "llama-index-readers-hubspot<0.6"
+        run: uv pip install --no-deps "llama-index-readers-hubspot<0.6"
       - name: Fetch the AWS MCP server the S3 test drives
         # tests/test_mcp.py runs this through `uvx …@latest`, which is the point — Backlot is
         # proved against the server awslabs ships today. Downloading it here rather than there
@@ -37,7 +44,7 @@ jobs:
         # that has not seen this version otherwise loses.
         run: uv tool install --with "mcp<2" "awslabs.aws-api-mcp-server@latest"
       - name: Unit tests
-        run: pytest -rs
+        run: uv run --no-sync pytest -rs
 
       - name: Install the Linear example's dependencies
         working-directory: examples/using-official-sdk/linear
@@ -48,7 +55,9 @@ jobs:
       - name: Run the Linear example against a server it starts itself
         working-directory: examples/using-official-sdk/linear
         env:
-          PYTHON: python
+          # The interpreter `uv sync` installed backlot into, not whatever `python` resolves to:
+          # index.ts spawns this to run `-m backlot import`, and the venv is never on PATH here.
+          PYTHON: ${{ github.workspace }}/.venv/bin/python
         run: npx tsx index.ts
 
   lint:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,13 +26,6 @@ jobs:
         with:
           enable-cache: true
 
-      # `uv sync`, not `pip install -e ".[all]"`: mirage's S3 backend needs aioboto3, whose newest
-      # release hard-pins `aiobotocore[boto3]==2.25.1` and so admits only botocore 1.40.46-1.40.61,
-      # while `boto3>=1.40` here has no upper bound. pip starts at the newest boto3 and backtracks
-      # through 215 releases toward that window, giving up with `resolution-too-deep`; uv derives
-      # the conflict instead of searching for it and resolves in seconds. `--locked` fails the run
-      # when uv.lock and pyproject.toml disagree, so the lock cannot drift unnoticed, and
-      # `--all-extras` is what keeps every gated test installed.
       - name: Install
         run: uv sync --all-extras --locked --python ${{ matrix.python-version }}
       - name: Install the HubSpot reader past its stale pin
@@ -55,8 +48,6 @@ jobs:
       - name: Run the Linear example against a server it starts itself
         working-directory: examples/using-official-sdk/linear
         env:
-          # The interpreter `uv sync` installed backlot into, not whatever `python` resolves to:
-          # index.ts spawns this to run `-m backlot import`, and the venv is never on PATH here.
           PYTHON: ${{ github.workspace }}/.venv/bin/python
         run: npx tsx index.ts
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -77,16 +77,21 @@ file with them:
 | `git` | the tests in `tests/test_github.py` that build a real repo |
 
 Install what covers the surface you touched, or all of it at once, and let `-rs` confirm nothing
-you meant to run skipped. The zero-skip install is two commands, the way CI's is — `.[all]`
-is every extra above, kept in step with them by `tests/test_packaging.py`, and the HubSpot
-reader pins `hubspot-api-client<9` against the `>=12` that `examples` needs — over-restrictive
-rather than a real incompatibility, so it goes in past its own dependencies, at the version CI
-installs:
+you meant to run skipped. The zero-skip install is two commands, the way CI's is — `--all-extras`
+is every extra above, and the HubSpot reader pins `hubspot-api-client<9` against the `>=12` that
+`examples` needs — over-restrictive rather than a real incompatibility, so it goes in past its own
+dependencies, at the version CI installs:
 
 ```bash
-uv pip install -e ".[all]"
+uv sync --all-extras
 uv pip install --no-deps "llama-index-readers-hubspot<0.6"
 ```
+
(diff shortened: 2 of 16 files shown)


nuri-yoo (approved):
LGTM

PR #186, opened 2026-09-10
deps: rename `examples` dependency group to `official-sdk`

khj809:
Two changes to `[project.optional-dependencies]`, and the test that keeps the first one honest.

**`pip install backlot[official-sdk]`** installs the clients the vendors publish themselves — slack_sdk, PyGithub, atlassian-python-api, google-api-python-client with google-auth, notion-client, boto3 and hubspot-api-client — and the extra now says so. It was called `examples`, a name that claims a category four other extras also sit in: `mcp`, `llamaindex`, `mirage` and `fsspec` each carry an example set of their own, and each names the thing it installs. This one names the directory it installs for, `examples/using-official-sdk/`, whose ten scripts and README are where the install line is read.

The name is written down in 26 places: the definition and the `all` self-reference in `pyproject.toml`, plus the comment there that explains why the HubSpot reader's `hubspot-api-client<9` cannot be reconciled with this extra's `>=12`; CONTRIBUTING's gate table row; the install header in each of the ten scripts under `examples/using-official-sdk/` and in its README; the twelve `.[official-sdk,llamaindex]` lines under `examples/using-llamaindex-readers/`, whose scripts need both extras; `tests/test_sdk.py`'s module docstring and the gate on its Notion section; and the certifi comment in `backlot/server.py`, which now says where certifi actually comes from — the core install, through httpx.

**A third test holds the places the other two cannot see.** `test_the_all_extra_aggregates_every_other_one` and `test_the_contributing_extras_are_the_pyproject_extras` hold a triangle: pyproject's definitions, the `all` aggregate, CONTRIBUTING's table. Every other install line in the repository was unheld, and an extra renamed in pyproject leaves them naming one that no longer exists — quiet at install time, because pip and uv answer an unknown extra with a warning and install the base package at exit 0, and loud later, when the script dies on the import that extra was carrying. `test_no_tracked_file_installs_an_extra_pyproject_does_not_define` walks `git ls-files` for `.[...]` and asserts every name in it is defined: 52 occurrences, 0 undefined on `main` and on this head, and reverting any one install line fails it with that file and line number.

`backlot[examples]` is gone rather than kept as a `backlot[official-sdk]` self-reference. pip's warning-then-ignore is the visible failure, and an alias is an extra like any other, which the two enumeration tests would require to appear in the `all` aggregate and in CONTRIBUTING's table — the dead name documented alongside the live one for as long as it stood.

**No extra restates a core dependency.** Three lines named a distribution `[project.dependencies]` already carries, at the identical floor: `httpx>=0.27` in `dev` and in this extra, and `pyyaml>=6.0` in `mcp`. Each has a direct importer — `tests/test_fidelity.py`, `test_google.py` and `test_mcp.py` call httpx, `examples/using-official-sdk/fireflies.py` posts GraphQL over raw HTTP because Fireflies publishes no SDK, and `tests/test_mcp.py` reads the token file with `yaml.safe_load` — but the package installs both regardless of which extras are asked for, so the restatement decided nothing. Resolution is measured identical without them: `uv pip install --dry-run -e ".[<extra>]"` returns the same set for `dev` (51 packages), `official-sdk` (75), `mcp` (85) and `all` (209), each counting backlot itself. `dev` keeps `httpx2`, which is not a core dependency: starlette 1.6's TestClient wants it, and its comment now says where the httpx beside it comes from.

`uv.lock` is regenerated across both changes. Only the extra's name and those three memberships move in it; no package changes version.

**Verification.** On `main` at `50c0419`.

```
uv sync --all-extras --locked                    resolved, no version change
uv run pytest -rs                                2900 passed, 1 skipped
ruff check .                                     All checks passed!
ruff format --check .                            139 files already formatted
python scripts/gen_docs.py --check               exit 0
uv pip install --dry-run -e ".[official-sdk]"    75 packages
uv pip install --dry-run -e ".[examples]"        warning: does not have an extra named `examples`
```

The one skip is an optional-extra import this branch does not touch, `llama_index.readers.hubspot`.

**Not in this PR.** Which versions the environment resolves is unchanged, and one of them is worth its own work: in the all-extras environment `llama-index-readers-confluence` caps `atlassian-python-api<5` and `aioboto3`'s pinned `aiobotocore==2.25.1` caps `boto3<1.40.62`, so CI runs the SDK tests at atlassian-python-api 4.0.7 and boto3 1.40.61 while `pip install backlot[official-sdk]` alone resolves to 5.0.4 and 1.43.91 (measured 2026-09-10; boto3 ships most days). Both capping packages are at their latest release and still impose the caps, so no bump lifts them — the combination users actually install is simply never exercised.


examples/using-llamaindex-readers/README.md
```diff
@@ -43,7 +43,7 @@ exactly as against the real API.
 - **GitHub** (`github.py`): `GitHubIssuesClient(base_url=...)` is a first-class constructor arg —
   no shim needed.
 - **HubSpot** (`hubspot.py`): the reader is **not** in the `[llamaindex]` extra — it pins
-  `hubspot-api-client<9`, which no resolver can reconcile with the `>=12` that `[examples]` needs, so
+  `hubspot-api-client<9`, which no resolver can reconcile with the `>=12` that `[official-sdk]` needs, so
   declaring it would make `.[examples,llamaindex]` uninstallable. The pin is over-restrictive (the
```
nuri-yoo: The rename reaches the two prose mentions in this directory and stops there. Every install line under `examples/using-llamaindex-readers/` still names the extra this PR deletes, 12 occurrences across 11 files:

```
README.md:7,47   confluence.py:4   gdrive.py:22   github.py:4   gmail.py:15
hubspot.py:12    jira.py:4         linear.py:13   notion.py:7   s3.py:8   slack.py:10

    pip install -e ".[examples,llamaindex]"
```

Measured on this head:

```
uv pip install -e ".[examples,llamaindex]"       warning + Installed 157 packages, exit 0
uv pip install -e ".[official-sdk,llamaindex]"   Installed 161 packages
absent from the first: hubspot-api-client, notion-client, pygithub (+pynacl)
```

I built the 157-package environment and ran all ten scripts in this directory. Nine pass. `hubspot.py` exits 1 on `ModuleNotFoundError: No module named 'hubspot'`, because `point_hubspot_at()` imports `hubspot` when it is called and the only thing that carries `hubspot-api-client` is the extra this line no longer names. In the 161-package environment the same script exits 0 and prints its three Documents.

The fix is one substitution in all 12 places, `.[examples,llamaindex]` to `.[official-sdk,llamaindex]`. `git grep -F '.[examples'` at this head returns those 12 and nothing else in the tree.

examples/using-llamaindex-readers/hubspot.py
```diff
@@ -5,7 +5,7 @@
 point_hubspot_at() rebinds `hubspot.HubSpot` to inject Backlot's host before the reader runs.
 
 The reader is deliberately NOT in the [llamaindex] extra: it pins hubspot-api-client<9, which no
-resolver can reconcile with the >=12 that [examples] needs. The pin is over-restrictive — the reader
+resolver can reconcile with the >=12 that [official-sdk] needs. The pin is over-restrictive — the reader
```
nuri-yoo, line 8: This line was renamed and line 12, four lines below it in the same docstring, was not:

```python
resolver can reconcile with the >=12 that [official-sdk] needs.    # line 8, renamed
...
    pip install -e ".[examples,llamaindex]"                         # line 12, still the old name
    pip install --no-deps llama-index-readers-hubspot
```

This is also the script that breaks. Following its own install block produces an environment with no `hubspot-api-client`, and `point_hubspot_at()` does `import hubspot` inside the call. Verified both ways in the README comment.

backlot/server.py
```diff
@@ -95,7 +95,7 @@ class Server:
 def ensure_cert_bundle() -> None:
     # Talking to an HTTPS url (a deployment behind an ACM cert) can fail with
     # CERTIFICATE_VERIFY_FAILED on macOS, where Python's default SSL context has no CA bundle.
-    # certifi ships with the [examples] extra; point OpenSSL at it unless already configured.
+    # certifi ships with the [official-sdk] extra; point OpenSSL at it unless already configured.
```
nuri-yoo: The claim in this comment is not true, and it was not true as `[examples]` either, so this is a rename landing on top of an older mistake rather than one this PR makes. certifi arrives with the core install:

```
uv pip install -e .     Installed 41 packages
uv pip show certifi     Required-by: httpcore, httpx
```

`httpx>=0.27` is a `[project.dependencies]` entry, so a plain `pip install backlot` already has certifi and `ensure_cert_bundle` finds it. Removing `httpx` from the extra in this PR also takes away the extra's own direct path to it. Since the line is already in the diff, something like "certifi arrives with the core install (httpx depends on it)" would settle it.

pyproject.toml
```diff
@@ -88,7 +87,6 @@ mcp = [
     "anthropic[mcp]>=0.40",
     "openai>=1.40",
     "openai-agents>=0.1",
```
nuri-yoo, line 89: Removing `pyyaml` here is right and the floors do match: `pyyaml>=6.0` at line 34 is the same specifier the `mcp` block carried.

The importer the body credits does not exist, though. It says "the MCP examples read YAML", and `examples/using-mcp-with-agents/` has no occurrence of `yaml` or `yml` in any of its twelve files, nor does `backlot/mcp.py`. What reads YAML behind this gate is `tests/test_mcp.py`: `import yaml` at module scope above `pytest.importorskip("mcp")`, then `yaml.safe_load(settings.tokens_path.read_text())` at nine call sites. Same conclusion, different file, and the test is the stronger argument anyway since it is the thing that would break.

nuri-yoo (commented):
Two changes here, and the dependency removals are solid. `dev`, `official-sdk` and `mcp` resolve to identical package sets on `main` and on this head, `uv lock --check` is clean, and the lockfile diff is 50 lines of extra name and membership with no version or url touched. The suite is green here too, `2899 passed, 1 skipped`, matching the body.

The rename is the half that is not finished. `examples/using-llamaindex-readers/` still installs the deleted extra in 12 places, and one of its scripts stops working because of it. Details on the README and `hubspot.py` comments below.

**Nothing in the suite catches that.** The body says the two enumeration tests "fail a half-done rename". I measured what they actually cover by reverting the rename one place at a time on this head:

| reverted | result |
|---|---|
| CONTRIBUTING's table row | `1 failed` (`test_the_contributing_extras_are_the_pyproject_extras`) |
| the `all` self-reference | `1 failed` (`test_the_all_extra_aggregates_every_other_one`) |
| the install line in `examples/using-official-sdk/slack.py` | `150 passed` |

The triangle of pyproject, `all` and CONTRIBUTING is held. Every install line in `examples/` and in the READMEs is not, which is why this branch passes CI on four Pythons with 12 dead install lines in it. A third test closes the gap. Walking `git ls-files` for `.[...]` finds 52 occurrences on both revisions and reports 0 undefined on `main`, exactly the 12 below on this head, and no false positives:

```python
extras = set(tomllib.load(open(REPO_ROOT / "pyproject.toml", "rb"))["project"]["optional-dependencies"])
for path in tracked_files():
    for i, line in enumerate(path.read_text().splitlines(), 1):
        for group in re.findall(r"\.\[([\w,\s-]+)\]", line):
            for e in (x.strip() for x in group.split(",")):
                assert e in extras, f"{path}:{i} installs undefined extra {e!r}"
```

**One number in the body.** `uv pip install --dry-run -e ".[all]"`, the command the body names, reports 209, not 208. The other three (51, 75, 85) count backlot itself and the 208 does not, so the four came from two different counts rather than from a release landing in between: pinning the index with `--exclude-newer 2026-09-10` still gives 208 without backlot and 209 with it. `boto3 1.43.91` was right when you measured it, since 1.43.92 was uploaded at 2026-09-10T19:21Z, after this PR opened. boto3 ships daily, so the `measured <date>` note used elsewhere in the repo would keep that paragraph readable.

Verification, on this head at `177c521`, Python 3.14, uv 0.11.8:

```
pytest -rs                                    2899 passed, 1 skipped
ruff check .                                  All checks passed!
ruff format --check .                         139 files already formatted
python scripts/gen_docs.py --check            exit 0
uv lock --check                               exit 0
pip install --dry-run -e ".[official-sdk]"    no warning, all eight SDKs in the install set
```

The hyphen in the new name gives neither resolver any trouble.

examples/using-llamaindex-readers/README.md
khj809: Fixed in 89d565b. All 12 lines are `.[official-sdk,llamaindex]` now, and `git grep -F ".[examples"` returns nothing on the head.

Confirmed the hubspot.py breakage the same way you scoped it, without building both environments: `[llamaindex]` alone resolves with no `hubspot-api-client`, `[official-sdk,llamaindex]` with it, and `point_hubspot_at` does `from hubspot import HubSpot` inside `load_data()` — so the script following its own install block dies at call time, not import time, which is why nothing here noticed.

examples/using-llamaindex-readers/hubspot.py
khj809: Both lines say `official-sdk` now. The docstring at line 8 and the install block at line 12 were renamed by two different passes of mine — the first grepped `[examples]`, which never matches `[examples,llamaindex]`. That is the hole the new test closes.

backlot/server.py
khj809: Verified and rewritten. `uv pip install --dry-run -e "."` brings certifi in through httpx, so the comment was wrong before this branch and more wrong after it removed the extra's own httpx line. It now reads "certifi arrives with the core install (httpx depends on it)".

pyproject.toml
khj809: Checked and you are right about the importer: `examples/using-mcp-with-agents/` has no yaml at all, and `tests/test_mcp.py` imports it at module scope above the `mcp` gate with `yaml.safe_load` on the token file. The PR body now credits that file instead of the examples. The removal itself stands on the floors matching, which it does.

khj809:
All four are fixed in `89d565b`, and the two measurement corrections are in the body.

**The missing test.** Your table is what the body claimed and what it actually held, and the gap is exactly where the 12 lines slipped through. `test_no_tracked_file_installs_an_extra_pyproject_does_not_define` is in `tests/test_packaging.py` beside the other two, close to your sketch: `git ls-files`, `\.\[([\w,\s-]+)\]` per line, every name checked against `[project.optional-dependencies]`, reported as `path:lineno installs <name>`. I reproduced your counts before trusting it — 52 occurrences on `main` and on this head, 0 undefined on both — and ran it on a `main` worktree to confirm it is not a branch-shaped test. Reverting one install line fails it with that file and line; reverting the whole set reports all 12.

**`.[all]` = 209.** Measured again on the head: 51, 75, 85, 209, all four counting backlot itself. The body said 208 because that one number came from a pre-merge run before `mirage-ai[fuse,s3]` landed, and I did not re-measure it after the merge — the other three were unaffected. Fixed, and the boto3/atlassian paragraph now carries `measured 2026-09-10` for the reason you give.

Suite on the head at `89d565b`: `2900 passed, 1 skipped`, ruff and `gen_docs --check` clean.

nuri-yoo (approved):
LGTM

PR #193, opened 2026-09-11
docs: add released version badge in readme

khj809:
A PyPI version badge shows the latest release at the top of README.md and README.pypi.md. The badges sit one per line, with the community links in their own group.

PR #185, opened 2026-09-10
github: the AI Scan setting the vendor reads and writes per organization and per repository is an acknowledged gap (#125)

khj809:
`backlot diff --source github` reported four new gaps, all of them the AI Scan enablement setting GitHub added under `code-scanning`: `GET` and `PATCH` on `/orgs/{org}/code-scanning/ai-scan` and on `/repos/{owner}/{repo}/code-scanning/ai-scan`. The baseline now acknowledges the four, each carrying the same line saying why — the vendor calls them public preview, so the report that retires them will name the paths one by one, and a note on one of the four would leave three of those lines unexplained.

They stay a gap. What a caller gets today is a 404 with real's body, `{"message":"Not Found","documentation_url":"https://docs.github.com/rest","status":"404"}`, both `PATCH`es included rather than a 405, which is what the already acknowledged `/code-scanning/default-setup` pair answers too. Serving a 200 instead would mean inventing state: the response is a single required `pr_scan: "enabled" | "disabled"`, no corpus record states a scanning posture, and the organization setting answers to an enterprise policy no deployment of Backlot has. The two `PATCH` routes are writes besides, and Backlot serves none — all 33 of its github routes are `GET`, because a corpus enters through `backlot import`. Every one of the 25 code-scanning operations the vendor declares is now in the baseline.

Read from GitHub's published spec on 2026-09-10, which marks all four in public preview and subject to change.

The second commit corrects two sentences that describe a step the tool does not take. `docs/fidelity.md` said every entry carries a note, which is false for 4002 of the 4098 entries across the eleven baselines, and the scheduled issue's intro said a gap gets its note from `--update-baseline`, which writes the entry alone. Both now say what happens: the flag writes the entry, and the note is the hand edit made in the review that follows, where a decision needed recording.

Verified with `backlot diff --source github`: 1229 divergences, 1229 acknowledged, 0 new. #125 reports 1240 / 1236 / 4 against the same vendor document; the 11 entries between the two runs are #175's, which implemented `GET /rate_limit` and ten `sort`/`direction`/`type`/`visibility` params. The suite is 2897 passed, 3 skipped in this checkout — the three skips are missing extras, `llama_index.readers.hubspot` and `mirage.core.google._client`, not tests agreeing.

Closes #125.

PR #182, opened 2026-09-10
linear: the SSH-address and view-preference roots real serves are acknowledged gaps, and the agent-session family's reason is written down

nuri-yoo:
Closes #141. Seven commits and five merges of `main` the branch took through GitHub: `9e2b6f8` adds two entries to `backlot/fidelity/baseline/linear.json` and writes a note onto a third; `3fa1d15` merges `main` at `f5611ce`, which is #165; `f64f46d` corrects three counts in those notes, prose only, and is described under **What the self-review changed** below; `52af94a` merges `main` at `3a047d0`, which is #175; `22c7709`, `13cdec5` and `166873a` each apply one review suggestion, one note apiece, described under **What the review changed** below; `cb6eea9` merges `main` at `50c0419`, which is #183; `bc231ac` corrects one clause the `agentSessionSshAddress` suggestion carried, described under **What the review changed** as well; `7459593` merges `main` at `c74a820`, which is #176; `a0473fb` takes the approval's nit, so that note states the bare `null` once, in the sentence that carries the signature and the three inputs; `fdb5ec0` merges `main` at `1c17868`, which is #185, and touches nothing in this file. No code changes, so nothing about the served surface moves. #141's body carries the 2026-09-09 scheduled run, 775 divergences with 774 acknowledged and one new; a run on 2026-09-10 finds two, Linear having added `Query.userViewPreferences` in between. No open PR touches this file: #181 changes the workflow that files the issue and `docs/fidelity.md`, #176 and #175 are the S3 and GitHub baselines beside it, #164 is Slack.

[...]

PR #195, opened 2026-09-11
linear: the short summary the vendor now serves on a project update is an acknowledged gap, and the ProjectUpdate stub's reason is written down

### LINKED ISSUE #194: Fidelity: linear diverges from its vendor (first seen 2026-09-11)
`backlot diff --source linear` reports divergences the baseline does not acknowledge. Each is either a bug in Backlot or a deliberate gap; `backlot diff --source linear --update-baseline` writes the entry for a gap, and a note saying why it is deliberate is added by hand in the review that follows. This body is the latest scheduled run's report, rewritten in place. Closing this issue keeps it as the account of what was triaged; a run that still diverges opens a new one. Previously reported in #184.

```
   777 divergence(s) · 776 acknowledged · 1 new

     missing_field  ProjectUpdate.shortSummary
       vendor serves shortSummary: String; Backlot does not

   1 new. Fix them, or run --update-baseline to acknowledge the gaps.
```

nuri-yoo:
Closes #194. Two commits and one merge of `main` the branch took through GitHub, touching `backlot/fidelity/baseline/linear.json` alone: `2e3006c` runs `backlot diff --source linear --update-baseline`, which adds the `ProjectUpdate.shortSummary` entry and moves `measured` to 2026-09-11, and writes a note by hand; `1e597d9` merges `main` at `2f6913e`, which is #193; `55f3f7c` moves the stub's reason onto `Query.projectUpdates`, described under **What the review changed** below. No code changes, so nothing about the served surface moves. #194's body carries the 2026-09-11 scheduled run, 777 divergences with 776 acknowledged and one new; #182's verification saw that same one on its final run and left it for the scheduled run to file, which it did. No open PR touches this file.

**What Linear serves.** Measured against api.linear.app on 2026-09-11 with a personal API key. `ProjectUpdate` carries 22 fields, none deprecated; `shortSummary: String` is the one added since the 2026-09-10 measurement, it takes no argument, and Linear's own description reads `A short AI-generated summary of the project update. Null if no short summary is available.` The type's own description reads `A status update posted to a project. Project updates communicate progress, health, and blockers to stakeholders. Each update captures the project's health at the time of writing and includes a rich-text body with the update content.` The workspace measured has no project and no update: `projectUpdates(first: 5)` and `projects(first: 5)` both answer an empty connection, so what the field answers for an update that has a summary is unmeasured.

**Why it is a gap.** No corpus states a project update. A Linear record here is an issue with its team, its comments and the people on it, and nothing in it says how a project stood on a given day or who wrote that up. `ProjectUpdate` is one of the id-only stubs at the top of `backlot/graphql/linear.graphql`, declared so `@linear/sdk`'s fragments validate. Three fields are typed with it: `Project.lastUpdate` and `Comment.projectUpdate` resolve to null, and `Reaction.projectUpdate` is never reached because `reactions` is an empty list on every issue and comment `linear_resolvers.py` builds. The roots that would list updates, `Query.projectUpdate`, `Query.projectUpdates` and `Project.projectUpdates`, are gaps in the same file already, and so are the 20 other fields the vendor declares on the type beyond `id`; `shortSummary` is the 21st. The reason for the whole stub is written on `Query.projectUpdates`, the root that would list them, and this entry's note opens with the pointer back to it and keeps what is the field's own: the description, that it takes no argument, that it is the newest field on the type, and that what it answers for an update with a summary is unmeasured. Serving it would mean giving the stub a field no owning field ever leads to; a constant `null` would match the description's own case for an update without a summary and state nothing.

**What the review changed.** khj809's one finding, taken as its suggestion in `55f3f7c`. `2e3006c` had anchored the stub's reason on this leaf, where every other note in the file anchors on a `Query.*` root and later joiners point back to it, as `Query.inboxNotifications` and `Query.agentSessionSshAddress` do. The cost is in the code: `Baseline.write` builds the rewritten file from the findings live on the run alone, and the `--update-baseline` branch of the command returns before the line that reports resolved entries, so had Linear removed or renamed `shortSummary` the next scheduled rewrite would have dropped the entry and the stub's only written reason with it, unreported, and a reader arriving at `Project.lastUpdate` or `Query.projectUpdates` would have found no pointer either way. The two notes are the review's two paragraphs, re-checked against the 2026-09-11 measurement.

**Verification.** At `55f3f7c`, on `main` at `2f6913e`, which is #193.

```
PYTHONPATH=. python -m pytest -rs                  2906 passed, 1 skipped
ruff check .                                       All checks passed!
ruff format --check .                              139 files already formatted
python scripts/gen_docs.py --check                 exit 0
backlot diff --source linear                       777 divergence(s), 777 acknowledged, 0 new
backlot diff --source linear --update-baseline     rewrites the file byte-identical; the note carries across
```

The one skip is `llama_index.readers.hubspot`, an optional extra this branch does not touch. No test comes with this, as with #182 and #185: nothing executable changes, and the property the note depends on, that `--update-baseline` rewrites the whole file and carries a gap's note across by key rather than dropping it, is held by `test_rewriting_a_baseline_keeps_its_notes_and_takes_the_new_identity`. The last line above is that property on this file rather than on a fixture.

**Not in this PR.** Nothing about project updates is served, and the 20 entries beside this one keep the entries they had without a note of their own; the reason for the whole stub is on `Query.projectUpdates`, which a vendor removing or renaming `shortSummary` would not take away. Serving `ProjectUpdate`, its roots and `ProjectUpdateConnection` is the wider work this does not begin.

### DIFF
--- a/backlot/fidelity/baseline/linear.json
+++ b/backlot/fidelity/baseline/linear.json
@@ -3,7 +3,7 @@
   "endpoints": [
     "https://api.linear.app/graphql"
   ],
-  "measured": "2026-09-10",
+  "measured": "2026-09-11",
   "acknowledged": [
     {
       "kind": "missing_field",
@@ -2782,6 +2782,13 @@
       "severity": "gap",
       "path": "ProjectUpdate.reactions",
       "detail": "vendor serves reactions: [Reaction!]!; Backlot does not"
+    },
+    {
+      "kind": "missing_field",
+      "severity": "gap",
+      "path": "ProjectUpdate.shortSummary",
+      "detail": "vendor serves shortSummary: String; Backlot does not",
+      "note": "No corpus states a project update. A Linear record here is an issue with its team, its comments and the people on it; nothing in it says how a project stood on a given day or who wrote that up, so `ProjectUpdate` is one of the id-only stubs at the top of `backlot/graphql/linear.graphql`: `Project.lastUpdate` and `Comment.projectUpdate` resolve to null, `Reaction.projectUpdate` is never reached because `reactions` is an empty list on every issue and comment, and `Query.projectUpdate`, `Query.projectUpdates` and `Project.projectUpdates` are gaps in this file. Measured against api.linear.app on 2026-09-11 with a personal API key: the vendor's `ProjectUpdate` carries 22 fields, none deprecated, and the 21 beyond `id` are the `ProjectUpdate.*` entries here. This one is the newest, added after the 2026-09-10 measurement. Linear's own description: \"A short AI-generated summary of the project update. Null if no short summary is available.\" It takes no argument. The workspace measured has no project and no update, so `projectUpdates` answers an empty connection there and what the field answers for an update that has a summary is unmeasured; a constant `null` would match the description's own case for an update without one and state nothing."
     },
     {
       "kind": "missing_field",
--- a/backlot/fidelity/baseline/linear.json
+++ b/backlot/fidelity/baseline/linear.json
@@ -2788,7 +2788,7 @@
       "severity": "gap",
       "path": "ProjectUpdate.shortSummary",
       "detail": "vendor serves shortSummary: String; Backlot does not",
-      "note": "No corpus states a project update. A Linear record here is an issue with its team, its comments and the people on it; nothing in it says how a project stood on a given day or who wrote that up, so `ProjectUpdate` is one of the id-only stubs at the top of `backlot/graphql/linear.graphql`: `Project.lastUpdate` and `Comment.projectUpdate` resolve to null, `Reaction.projectUpdate` is never reached because `reactions` is an empty list on every issue and comment, and `Query.projectUpdate`, `Query.projectUpdates` and `Project.projectUpdates` are gaps in this file. Measured against api.linear.app on 2026-09-11 with a personal API key: the vendor's `ProjectUpdate` carries 22 fields, none deprecated, and the 21 beyond `id` are the `ProjectUpdate.*` entries here. This one is the newest, added after the 2026-09-10 measurement. Linear's own description: \"A short AI-generated summary of the project update. Null if no short summary is available.\" It takes no argument. The workspace measured has no project and no update, so `projectUpdates` answers an empty connection there and what the field answers for an update that has a summary is unmeasured; a constant `null` would match the description's own case for an update without one and state nothing."
+      "note": "It joins the family whose gap is recorded on `Query.projectUpdates`, and it carries a note of its own because the 2026-09-11 scheduled run read it as new surface: it is the newest field on the type, added after the 2026-09-10 measurement. Linear's own description: \"A short AI-generated summary of the project update. Null if no short summary is available.\" It takes no argument. What it answers for an update that has a summary is unmeasured, since the workspace measured has none; a constant `null` would match the description's own case for an update without one and state nothing."
     },
     {
       "kind": "missing_field",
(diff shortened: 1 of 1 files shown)


backlot/fidelity/baseline/linear.json
```diff
@@ -2783,6 +2783,13 @@
...
...e is an issue with its team, its comments and the people on it; nothing in it says how a project stood on a given day or who wrote that up, so `ProjectUpdate` is one of the id-only stubs at the top of `backlot/graphql/linear.graphql`: `Project.lastUpdate` and `Comment.projectUpdate` resolve to null, `Reaction.projectUpdate` is never reached because `reactions` is an empty list on every issue and comment, and `Query.projectUpdate`, `Query.projectUpdates` and `Project.projectUpdates` are gaps in this file. Measured against api.linear.app on 2026-09-11 with a personal API key: the vendor's `ProjectUpdate` carries 22 fields, none deprecated, and the 21 beyond `id` are the `ProjectUpdate.*` entries here. This one is the newest, added after the 2026-09-10 measurement. Linear's own description: \"A short AI-generated summary of the project update. Null if no short summary is available.\" It takes no argument. The workspace measured has no project and no update, so `projectUpdates` answers an empty connection there and what the field answers for an update that has a summary is unmeasured; a constant `null` would match the description's own case for an update without one and state nothing."
```
khj809: The family reason for the whole `ProjectUpdate` stub is anchored here, on a leaf, and it is the most volatile leaf the family has: an AI-generated field that appeared between the 2026-09-10 and 2026-09-11 measurements. Every other note in this file anchors on a `Query.*` root, and the two later joiners point back to it rather than carry the reason themselves — `Query.inboxNotifications` opens with "The gap on `Query.notifications` covers the family", and `Query.agentSessionSshAddress` with "It joins the family whose gap is recorded on `Query.agentSession`", written in #182 for exactly this situation, a scheduled run naming a leaf.

Concrete cost. `Baseline.write` (`backlot/fidelity/findings.py:106-112`) appends only findings live on the run, and the `--update-baseline` branch in `cli.py` computes `stale` but never echoes it. If Linear removes or renames `shortSummary`, the next scheduled `--update-baseline` drops this entry and the stub's only written reason with it, silently, and the 20 sibling `ProjectUpdate.*` entries plus `Query.projectUpdate`, `Query.projectUpdates` and `Project.projectUpdates` are left with no note. A reader arriving at `Project.lastUpdate` or `Query.projectUpdates` today finds no pointer either.

Suggested fix: move the family reason onto `Query.projectUpdates` and keep only what is specific to this field here.

`Query.projectUpdates` note:

> No corpus states a project update. A Linear record here is an issue with its team, its comments and the people on it; nothing in it says how a project stood on a given day or who wrote that up, so `ProjectUpdate` is one of the id-only stubs at the top of `backlot/graphql/linear.graphql`: `Project.lastUpdate` and `Comment.projectUpdate` resolve to null, `Reaction.projectUpdate` is never reached because `reactions` is an empty list on every issue and comment, and `Query.projectUpdate`, `Project.projectUpdates` and this root are gaps in this file. Measured against api.linear.app on 2026-09-11 with a personal API key: the vendor's `ProjectUpdate` carries 22 fields, none deprecated, and the 21 beyond `id` are the `ProjectUpdate.*` entries here. The workspace measured has no project and no update, so this root answers an empty connection there.

`ProjectUpdate.shortSummary` note:

> It joins the family whose gap is recorded on `Query.projectUpdates`, and it carries a note of its own because the 2026-09-11 scheduled run read it as new surface: it is the newest field on the type, added after the 2026-09-10 measurement. Linear's own description: "A short AI-generated summary of the project update. Null if no short summary is available." It takes no argument. What it answers for an update that has a summary is unmeasured, since the workspace measured has none; a constant `null` would match the description's own case for an update without one and state nothing.

Both survive `--update-baseline` the same way the current note does — carried across by key — and I checked that the PR head already rewrites byte-identical, so nothing else in the file moves.

nuri-yoo: Taken as suggested in `55f3f7c`: the family reason now sits on `Query.projectUpdates`, and `ProjectUpdate.shortSummary` opens with the pointer back to it and keeps only what is this field's. Both texts are your two paragraphs verbatim, with the one sentence about this root answering an empty connection re-checked against the 2026-09-11 measurement (`projectUpdates(first: 5)` answered `nodes: []`).

Reproduced the cost before moving it. `Baseline.write` (`findings.py:106-112`) builds `kept` from the run's `findings` alone, so an entry the vendor no longer serves is not carried; and the `--update-baseline` branch in `cli.py` returns at :790 before the `if stale:` echo at :810, so the drop would go unreported. `backlot diff --source linear` is still 777 / 777 / 0 new, and a second `--update-baseline` on this head rewrites the file byte-identical with both notes in place. Pushing after this reply.

khj809 (approved):
LGTM

PR #196, opened 2026-09-11
slack: add minimal write APIs for chats and reactions

khj809:
## What changed

Slack serves nine write and write-adjacent methods: `chat.postMessage`, `chat.postEphemeral`, `chat.update`, `chat.delete`, `chat.getPermalink`, `reactions.add`, `reactions.remove`, `reactions.get`, `reactions.list`. Every one answers both `GET` and `POST`, and the six the spec types POST read a JSON body as well as a form — which is what `consumes` declares, and where `slack_sdk` puts three of them.

The corpus is still never written. The serving connection stays `mode=ro`, and `backlot/overlay.py` attaches a writable in-memory database to it; SQLite applies read-only to `main` alone, so the immutability is enforced rather than agreed to. A test hashes the corpus file before and after a run of writes. The overlay belongs to one server process and dies with it.

Reads merge the two behind `store.merged_source`, one SQL fragment aliased to the corpus table's own name so `_acl_clause` keeps its clause verbatim. A posted message appears in `conversations.history`, `conversations.replies` and `search.messages`; an edited one reads back edited everywhere, search included; a deleted one leaves all of them, including a message the corpus shipped — a corpus row cannot be removed, so it is subtracted at read time. `chat.postMessage` honours `thread_ts`.

The substrate is keyed on `source_type` from the start, because other vendors get write surfaces after this one. The overlay's tables are generated per source from `store.WRITABLE` the way `store.SCHEMA` already generates the eleven `<source>_acl` tables, and the document mirror is read out of the corpus with `PRAGMA table_info` so it cannot drift from the table it mirrors. What is per-vendor stays per-vendor and is not generalised on speculation: id minting, channel membership, refusal vocabularies. A source outside `WRITABLE` gets no overlay tables and reads the SQL it read before.

Two endpoints support an evaluation run: `GET /_meta/overlay` reports everything written since the last reset, and `POST /_meta/overlay/reset` throws it away, tombstones included, so a second run starts where the first did. Reset takes the admin token — it is the one `/_meta` route that destroys something, and an agent graded from `/_meta/overlay` must not be able to erase the record of what it just did.

Channel membership is one tri-state — in, out, derived — that `is_member`, `num_members`, `conversations.members` and the warm-cache builder share. Three states rather than a set because "removed" is a fact the corpus cannot express: `slack_membership_violations` called a speaker who cannot read a private channel a corpus that cannot be true, noting that leaving is the one state real Slack reaches it from. Nothing writes `in` yet; the read side is complete so `conversations.join` and `conversations.kick` are endpoint work rather than a semantics change.

`docs/overlay.md` is new. `AGENTS.md`'s "Data flows one way" is rewritten around the narrower invariant, its Tests rule reconciled with the "or a quote from the vendor's spec" it already allowed, and `backlot/overlay.py` added to where-a-change-goes. Four docstrings that justified something with "this is a read-only server" now say what is actually true — including Fireflies', which declares no `Mutation` type because Fireflies has none, not because the server is read-only.

Measured against slack.com on 2026-09-06, with the read-only token in `.env`:

- **Both verbs.** All nine methods answer `missing_scope` — not `unknown_method` — over `GET` and `POST` alike, so the method and the verb are recognised. `chat.getPermalink`, whose scope that token holds, reaches argument validation on both.
- **Query arguments on a POST.** `POST chat.getPermalink` with no arguments answers `invalid_arguments`; with `channel` and `message_ts` in the query string it answers `channel_not_found`, the same answer it gives with them in the body. Slack's spec documents only the body location, so the resulting `extra_param` findings are the spec being incomplete rather than an invented parameter.

Refusals are quoted from the `default` response `error` enum in `slack_web_openapi_v2.json` — the document `backlot diff --source slack` already fetches. Each endpoint docstring names which errors are quoted and which were measured, and which trigger conditions are Backlot's own decision because the spec enumerates a string without describing when it fires. Measuring the refusals directly needs write scopes on a real workspace; the available token has none, and `AGENTS.md` permits the vendor's spec where a measurement is not possible.

Search ranking rests on its own measurement. bm25's inverse document frequency is a property of the index a term is in, the argument `build_fts` already makes for giving each source its own index — so an overlay holding a handful of rows scores everything at about `-1e-06` where the corpus index scores comparable text `-2.70`, and a newly posted message would sort behind every corpus hit however well it matched. Overlay rows are matched by an FTS5 index of their own and scored against the corpus index's statistics, read through `fts5vocab`. `store._corpus_bm25` reproduces SQLite's own `bm25()` to `rel=1e-9` on corpus rows, which is how it is tested, over single terms, a multi-term query, a term common enough to drive the IDF clamp, and a repeated term — fts5 sums per phrase, so `"the the"` scores double `"the"`.

Timed on the 11 GB corpus, biggest Slack channel at 768,839 messages:

| Operation | Overhead |
|---|---|
| history | +3.4% |
| search | +3.7% |
| count | +2% |

**Not modelled: OAuth scopes.** Real Slack splits some answers on scope rather than identity — posting to a public channel you have not joined succeeds with `chat:write.public` and answers `not_in_channel` without it. Backlot takes the permissive branch. Recorded in the baseline and in `docs/overlay.md` rather than left to be discovered.

`backlot diff --source slack` reports 228 divergences, 228 acknowledged, 0 new. Every `missing_operation` entry now carries a note; they were all empty before. The 26 breaking entries are hand-written with the measurement behind them, as acknowledging a breaking divergence requires.

## Verification

- **Tests** — `pytest` on a `.[all]` install: 3384 passed, 1 skipped.
- **Optional surfaces that ran rather than skipped** — `tests/test_s3.py`, `tests/test_sdk.py`, `tests/test_mcp.py`, `tests/test_llamaindex.py`, and the reader, Google and mirage tests in `tests/test_integrations.py`. The single skip is `tests/test_llamaindex.py:127`, which imports `llama_index.readers.hubspot` — not published for the pinned llama-index. `tests/test_sdk.py` drives Slack's write methods through `slack_sdk`, which is the only place the request's media type is exercised.
- **Lint** — `ruff check . && ruff format --check .`: clean on ruff 0.16.6.
- **Packaging** — a built wheel installed into a venv outside the checkout, driven over HTTP from a cwd the source tree cannot shadow: post, threaded reply, search, reactions, update, delete, `/_meta/overlay`, and reset behind the admin token.
- **Fidelity** — `backlot diff --source slack`: 0 new.


backlot/routers/slack.py
```diff
@@ -920,6 +966,541 @@ async def search_all(request: Request):
     return {"ok": True, "query": query, "messages": block, "files": empty}
 
 
+# --- writes ---------------------------------------------------------------------------------
+#
+# Every method here answers BOTH verbs. Measured against slack.com on 2026-09-06 with a read-only
```
nuri-yoo: **[block]** `slack_sdk` 3.44.1 sends three of these nine as `application/json`: `WebClient.chat_postMessage`, `chat_update` and `chat_postEphemeral` end in `self.api_call(..., json=kwargs)`; the other six go as form or query. `parse_slack_form` in `backlot/main.py` stashes a body only when the content type is `application/x-www-form-urlencoded`, so through the SDK all three answer `invalid_arguments`. Measured at 449f403, `backlot serve` over the SAMPLE corpus, `WebClient(token=<user token>, base_url=".../slack/api/")`:

```
chat_postMessage    SlackApiError {'ok': False, 'error': 'invalid_arguments'}
chat_postEphemeral  SlackApiError {'ok': False, 'error': 'invalid_arguments'}
chat_update         SlackApiError {'ok': False, 'error': 'invalid_arguments'}
chat_delete, reactions_add, reactions_remove, reactions_get, reactions_list, chat_getPermalink   ok
```

Real takes the JSON: docs.slack.dev's chat.postMessage page lists `Content types application/x-www-form-urlencoded application/json`, and the spec's `consumes` says the same for the six POST-typed methods. The three GET-typed ones (`chat.getPermalink`, `reactions.get`, `reactions.list`) declare form only, and measured real ignores a JSON body there: `POST chat.getPermalink` with `{"channel", "message_ts"}` as JSON answers `invalid_arguments` with `missing required field: channel`. So the fix is a JSON branch in `parse_slack_form` (a dict body into `request.state._form`), and `tests/test_sdk.py`, which today drives Slack read methods only, is where a `chat_postMessage` would have caught this and keeps it caught. `backlot diff` cannot see it: it compares parameters, not media types.

backlot/routers/slack.py
```diff
@@ -920,6 +966,541 @@ async def search_all(request: Request):
...
+            "ts": ts,
+            "author_email": author,
+            "content": text,
+            "created_ts": now,
+            "thread_ts": thread_ts,
+            "thread_seq": thread_seq,
+            "subtype": "bot_message" if is_bot else None,
+        },
+    )
+    # Membership on a public channel is derived from having spoken, so the row just written would
```
nuri-yoo: **[block]** The premise here does not hold, and the write it justifies records something the agent did not do. `slack_channel_has_author` and `_slack_public_members_sql` read `slack_messages` unqualified, and SQLite resolves an unqualified name in `temp`, then `main`, then the attached schemas, so a row in `ov.slack_messages` is never a speaker anywhere public membership is derived. Measured on the serving connection at 449f403 after one post: `SELECT COUNT(*) FROM slack_messages` is 5, `main.slack_messages` 5, `ov.slack_messages` 1. Deleting these two lines leaves the suite green, `test_posting_does_not_silently_join_a_public_channel` included (M1 in the review body), and so does deleting the explicit-state branch of `_is_member` (M21): nothing in the suite reaches `out` through the router.

What the write does do is visible to a grader. After one `chat.postMessage` by a non-member, `GET /_meta/overlay` answers `slack_membership: [{"channel": "eng-announcements", "email": "bob@acme.com", "state": "out"}]`, and `docs/overlay.md` tells that grader the list is "the memberships changed". An agent that posted once is on record as having left the channel it posted to. Dropping the write is enough, since the derivation already gives this comment the answer it wants; if the row is meant to stay, the docs have to say that `out` also means "posted without joining", and `_is_member`'s explicit branch needs a test that reaches it.

backlot/routers/slack.py
```diff
@@ -920,6 +966,541 @@ async def search_all(request: Request):
...
+        return err
+    ts = _param(request, "message_ts") or ""
+    if store.document_by_key(conn, "slack", (name, ts), auth.visible_ids(request, caller)) is None:
+        return _err("message_not_found")
+    cid = synth.slack_channel_id(name)
+    return {
+        "ok": True,
+        "channel": cid,
+        "permalink": (
+            f"https://{get_settings().org_name}.slack.com/archives/{cid}/p{ts.replace('.', '')}"
```
nuri-yoo: **[note]** Real changes the permalink for anything in a thread. Measured against slack.com on 2026-09-18 with the read-only token in `.env`, over the ten visible messages in the workspace's public channels: a root with replies gets `/p<ts>?thread_ts=<its own ts>`, a reply gets `/p<ts>?thread_ts=<root ts>&cid=<channel id>`, and only a root with no replies is the bare `/archives/<C>/p<ts>` minted here. docs.slack.dev says the same for the reply case: "When the provided message_ts is part of a thread, the permalink format changes", with `?thread_ts=1358546515.000008&cid=C1H9RESGL` as the example. The row already carries `thread_ts` and `slack_reply_count` already answers whether a root has replies, so both shapes are within reach. `_message`'s permalink on line 1532 has the same gap and is pre-existing. `test_get_permalink_returns_an_archives_url` asserts the bare shape with `endswith`, so it moves with the fix. Fine as its own change if you would rather keep this PR to the write surface; say so and I will file it.

backlot/routers/slack.py
```diff
@@ -109,6 +122,22 @@ class SlackSearch(_SlackOk):
 ]
 _P_SEARCH_FILES = [qp("query", required=True), qp("count", "integer")]
 
+# The write surface. Each lists what Backlot actually READS, the way every other `_P_` here does:
+# a param the vendor documents but Backlot ignores stays an acknowledged gap in the fidelity
+# baseline rather than a declaration that promises behaviour it does not have.
+_P_POST_MESSAGE = [qp("channel", required=True), qp("text", required=True), qp("thread_ts")]
```
nuri-yoo: **[note]** The spec declares `text` optional on `chat.postMessage`, `chat.update` and `chat.postEphemeral` (`required: false` on the `formData` parameter in `slack_web_openapi_v2.json`) while listing `no_text` in each method's error enum, and docs.slack.dev glosses `no_text` as "No message text provided." That reads as the vendor's answer for an absent `text`, where this declares it required and `_missing_argument` answers `invalid_arguments`. The docstring calls the line Backlot's decision; the spec has an opinion on it. Not measurable with the token in `.env` (the scope check answers first), so spec-sourced rather than measured, and invisible to `backlot diff` because these are form fields.

backlot/store.py
```diff
@@ -3258,6 +3815,102 @@ def slack_private_channel_members(conn, channel) -> list[str] | None:
     return sorted(members - deactivated)
 
 
+# --- slack channel membership ----------------------------------------------------------------
+#
+# Slack's own, deliberately not generalised: no other served vendor has a channel to be in. The
+# three readers below (`slack_channel_has_author`, `slack_channel_member_emails`,
```
nuri-yoo: **[note]** Four names in the parentheses, "three readers" before them and "three answers" after. The PR body counts four (`is_member`, `num_members`, `conversations.members` and the warm-cache builder).

backlot/store.py
```diff
@@ -1098,6 +1113,301 @@ def _acl_clause(
...
+) -> str:
+    """What a query over this source's documents should read FROM.
+
+    For a source outside :data:`WRITABLE`, or with no overlay attached, this is the table name and
+    every reader is character-for-character what it was before writes existed. Otherwise it is the
+    corpus and the overlay unioned, tombstoned documents subtracted and patches applied,
+    **aliased to the corpus table's own name**.
+
+    The alias is not cosmetic: :func:`_acl_clause` emits ``_acl.<col> = <table>.<col>``,
+    hard-referencing whatever name the caller gave it, so the alias is what lets fourteen readers
```
nuri-yoo: **[note]** Counted at 449f403: `merged_source(` has seventeen call sites in `store.py`, spread over fifteen functions, thirteen of which also emit `_acl_clause`. None of those is fourteen. Naming the set (or the number that matches it) keeps the sentence checkable.

backlot/overlay.py
```diff
@@ -0,0 +1,226 @@
...
+
+from __future__ import annotations
+
+import itertools
+import sqlite3
+import threading
+
+from backlot import store
+
+# One lock for every overlay write. FastAPI runs sync endpoints on a threadpool over the single
```
nuri-yoo: **[note]** Every handler in `backlot/routers/slack.py` is `async def`, so none of the nine write methods runs on the threadpool: they run on the event loop one at a time, which is why sixteen concurrent `chat.postMessage` calls to one channel came back with sixteen distinct `ts` and no error in three trials at 449f403, even though `slack_next_ts` is called outside the lock. The lock still earns its place against the two `/_meta/overlay*` handlers, which are plain `def` and do run on the threadpool (a `reset` can interleave with an in-flight handler's statements), so that pair is the reason to keep it rather than "sync endpoints". The same premise is in `edit_document`'s docstring (`store.py:1251`) and `_fts_query_terms`'s.

backlot/store.py
```diff
@@ -1098,6 +1113,301 @@ def _acl_clause(
...
+    from backlot import overlay
+
+    if field not in PATCHABLE.get(source_type, frozenset()):
+        raise ValueError(
+            f"{source_type}.{field} is not patchable — see store.PATCHABLE. An identifier column "
+            f"is never patchable: it would move the row out from under its own ACL grant"
+        )
+    names = overlay.table_names(source_type)
+    cols = id_columns(source_type)
+    if True:
```
nuri-yoo: **[note]** `if True:` reads like the remains of a lock block that moved to the callers; the body can lose the indent. `_patch_locked` also repeats the `PATCHABLE` check `patch_document` just made, so the refusal text lives twice.

backlot/store.py
```diff
@@ -1098,6 +1113,301 @@ def _acl_clause(
...
+    channel. A posted message's second is *now* and the corpus is historical, so a collision is
+    almost always with another posted message — but "almost always" is why this probes rather than
+    assumes.
+
+    Slack's own, not a generic minter: what a served id looks like is a fact about a vendor, and
+    the sources that already synthesize one do it several different ways.
+    """
+    from backlot import overlay
+
+    tomb = overlay.table_names("slack")["tombstone"] if overlay.is_attached(conn) else None
```
nuri-yoo: **[note]** `tomb` is only ever read as "is the overlay attached" and never names the tombstone table it holds. `overlay.is_attached(conn)` in the `if` says what the branch is about, and the comment below already explains why the tombstone table is not what is consulted.

backlot/routers/slack.py
```diff
@@ -920,6 +966,541 @@ async def search_all(request: Request):
...
+                "type": "message",
+                "channel": synth.slack_channel_id(p["channel"]),
+                "message": _message(row),
+            }
+        )
+    count = _int(request, "count", 100)
+    return {
+        "ok": True,
+        "items": items[:count],
+        "paging": {"count": count, "total": len(items), "page": 1, "pages": 1},
```
nuri-yoo: **[note]** With two reacted items and `count=1` this answers one item and `paging: {"count": 1, "total": 2, "page": 1, "pages": 1}` (measured at 449f403). `pages` contradicts `total`; it should follow from `total` and `count`.

backlot/routers/slack.py
```diff
@@ -1172,18 +1770,25 @@ def _reactions(row) -> list[dict]:
         out.append(
             {
                 "name": reaction.get("name"),
-                "users": [synth.slack_user_id(e) for e in users],
+                # `_uid`, not `synth.slack_user_id`: the service account is not a corpus address
+                # and its id is fixed at USERVICE0, the same value `auth.test` and a message it
+                # posted report. Hashing its sentinel would give a client two ids for one reactor.
+                "users": [_uid(e) for e in users],
```
nuri-yoo: **[note]** Reverting this to `synth.slack_user_id(e)` leaves the suite green, and so does the same revert in `_edited` on line 1791 (M9 and M10 in the review body): no test reacts or edits as the service token and reads the id back. Both are one assertion each on the message `test_the_service_token_posts_as_a_bot` already posts.

backlot/routers/slack.py
```diff
@@ -920,6 +966,541 @@ async def search_all(request: Request):
...
+
+@router.api_route("/chat.delete", methods=["GET", "POST"], openapi_extra={"parameters": _P_DELETE})
+async def chat_delete(request: Request):
+    """Delete a message.
+
+    Errors quoted from the spec's ``default`` response: `message_not_found`, `channel_not_found`,
+    `cant_delete_message`, `compliance_exports_prevent_deletion`. The last is a workspace
+    compliance setting no corpus states.
+
+    A CORPUS message can be deleted. The row stays in the read-only database and is subtracted at
```
nuri-yoo: **[note]** Deleting a root that has replies (measured at 449f403 over the SAMPLE corpus) hides the root from `conversations.history`, leaves its replies in the channel with a `thread_ts` that now names nothing, makes `conversations.replies?ts=<root>` answer `thread_not_found`, and `chat.getPermalink` on one of those replies still answers `ok`. I could not find real's rule in the spec (no enum covers it) and could not measure it without a write scope, so this is a consistency question rather than a fidelity claim: three readers now disagree about whether the thread exists.

backlot/routers/slack.py
```diff
@@ -920,6 +966,541 @@ async def search_all(request: Request):
...
+    "/chat.postMessage", methods=["GET", "POST"], openapi_extra={"parameters": _P_POST_MESSAGE}
+)
+async def chat_post_message(request: Request):
+    """Post a message to a channel.
+
+    Errors quoted from `slack_web_openapi_v2.json`,
+    ``paths./chat.postMessage.post.responses.default``: `channel_not_found`, `not_in_channel`,
+    `is_archived`, `msg_too_long`, `no_text`, `too_many_attachments`. Backlot answers
+    `channel_not_found` and `no_text`; `not_in_channel` needs a scope model, `is_archived` needs a
+    channel that can be archived, and the two ceilings are workspace limits no corpus states.
```
nuri-yoo: **[note]** `msg_too_long` is filed here under "workspace limits no corpus states", but the length rule is Slack's own, not a workspace's: docs.slack.dev's chat.postMessage page says "Slack will truncate messages containing more than 40,000 characters." Measured at 449f403, a 50,000-character `text` posts `ok` and reads back at 50,000. A platform rule with a number in the vendor's docs is reproducible where a workspace setting is not, so this one does not belong in that sentence.

backlot/routers/slack.py
```diff
@@ -920,6 +966,541 @@ async def search_all(request: Request):
...
+    null counts is wrong" — so clearing it put the server in exactly that state for its lifetime
+    on the first write. A served write does not change how many documents the CORPUS holds, which
+    is what that number means.
+    """
+    members = getattr(request.app.state, "channel_members", None)
+    if members is not None:
+        members.pop(channel, None)
+
+
+def _writable_channel(request: Request, conn, caller: Caller):
```
nuri-yoo: **[note]** docs.slack.dev's chat.postMessage page describes `channel` as "An encoded ID or channel name that represents a channel, private group, or IM channel to send the message to." At 449f403, `channel=incidents` and `channel=#incidents` both answer `channel_not_found`. `chat.update` and `chat.delete` describe theirs as "Channel containing the message" with no name form, so the name belongs to the post alone; `_channel_name` would need a by-name branch for this caller only. Docs-sourced, not measured (the scope check answers first with the token in `.env`).

backlot/routers/slack.py
```diff
@@ -920,6 +966,541 @@ async def search_all(request: Request):
...
+    """Edit a message's text.
+
+    Errors quoted from `slack_web_openapi_v2.json`,
+    ``paths./chat.update.post.responses.default``: `message_not_found`, `cant_update_message`,
+    `channel_not_found`, `edit_window_closed`, `msg_too_long`, `too_many_attachments`, `no_text`,
+    `is_inactive`. Backlot answers the first three and `no_text`; `edit_window_closed` is a
+    workspace retention setting no corpus states, and the rest are limits it does not model.
+
+    That Slack answers `cant_update_message` rather than a permission error is quoted from that
+    enum. WHICH condition fires it is Backlot's decision, since the spec enumerates the string
```
nuri-yoo: **[note]** The spec's enum is bare, but docs.slack.dev's error table for chat.update is not: "cant_update_message: Authenticated user does not have permission to update this message." chat.delete's says the same for `cant_delete_message` ("Authenticated user does not have permission to delete this message."). That is the condition this docstring calls Backlot's decision; quoting it makes both refusals vendor-sourced rather than chosen.

backlot/routers/slack.py
```diff
@@ -920,6 +966,541 @@ async def search_all(request: Request):
...
+    if err is not None:
+        return err
+    name, _ts, row, err = _reaction_target(request, conn, caller)
+    if err is not None:
+        return err
+    return {
+        "ok": True,
+        "type": "message",
+        "channel": synth.slack_channel_id(name),
+        "message": _message(row),
```
nuri-yoo: **[note]** Real's reactions.get message carries a `permalink` (docs.slack.dev's response example: `"permalink": "https://xxx.slack.com/archives/C123456/p1648602352215969"` inside `message`, next to `reactions`). `_message` does not build one, so this answers the message without it; `_search_match` on line 1532 already mints the same shape for search hits.

backlot/store.py
```diff
@@ -1098,6 +1113,301 @@ def _acl_clause(
...
+        (channel, thread_ts),
+    ).fetchone()
+    return (row[0] or 0) + 1
+
+
+def slack_next_ts(conn, channel: str, epoch_sec: int) -> str:
+    """A free ``ts`` for a new Slack message in this channel at this second.
+
+    Same shape and same collision rule as an imported message's (``byo._Loader._slack_ts``): the
+    integer part is the second, the six-digit fraction is probed until it is free within the
```
nuri-yoo: **[note]** The fraction is hash-salted rather than sequential, so posts inside one second come back from `conversations.history` in hash order rather than posting order. Measured at 449f403: six `chat.postMessage` calls in one second minted fractions `954839, 911825, 092356, 066681, 091152, 712912`, and history (sorted `created_ts DESC, ts`) returned them shuffled. The importer's `_slack_ts` has the same property and a corpus is historical, so it was invisible until an agent could post a numbered sequence and read it straight back. Not a fidelity claim about real's minting, which I could not measure; a client-facing consequence of sharing the importer's rule.

backlot/store.py
```diff
@@ -2473,16 +3013,27 @@ def count_search(
...
-            f"SELECT COUNT(*) FROM (SELECT 1 FROM {fts} JOIN {tbl} t "
+            f"SELECT COUNT(*) FROM (SELECT 1 FROM {fts} JOIN {src} "
             f"ON {on} WHERE {fts} MATCH ?"
-            f"{cont_sql.format(a='t')}{clause} LIMIT ?)"
+            f"{cont_sql.format(a='t')}{clause}{fresh} LIMIT ?)"
+        )
+        total = conn.execute(sql, [m, *cont_p, *cparams, cap]).fetchone()[0]
+        # A total the pages contradict is worse than no total: search.messages reports it verbatim
+        # as `total` and derives its page count from it.
+        hits = _overlay_hits(
```
nuri-yoo: **[note]** `search.messages` computes the overlay hits twice per request (here and in `search_documents`), and `_corpus_bm25` runs one `COUNT(*)` per term per hit plus one per hit for the document length. Measured at 449f403 over the SAMPLE corpus: with 1,500 posted messages matching the query, `search.messages` takes 1.18s for `gateway` and 1.45s for `deploy note`, against 0.005s before the posts and 0.004s for a query that matches nothing; `conversations.history` stays at 0.012s. An evaluation run that posts a few thousand messages and searches a common term pays this on every search. Passing the hits from `count_search` into `search_documents` (or the reverse) halves it, and one grouped query over the `instance` vocabulary for all terms and rowids would take out the per-hit loop.

docs/supported-sources.md
```diff
@@ -25,7 +25,7 @@ Generated from `backlot/schemas/*.schema.json` and the app's own `/openapi.json`
 | `linear` | Linear | `/linear/graphql` | GraphQL (one `POST`) | [`linear.schema.json`](../backlot/schemas/linear.schema.json) | A Linear issue. |
 | `notion` | Notion | `/notion/v1` | 12 | [`notion.schema.json`](../backlot/schemas/notion.schema.json) | A Notion page or database. |
 | `s3` | Amazon S3 | `/s3` | 4 | [`s3.schema.json`](../backlot/schemas/s3.schema.json) | An S3 object. |
-| `slack` | Slack | `/slack/api` | 12 | [`slack.schema.json`](../backlot/schemas/slack.schema.json) | A Slack message. |
+| `slack` | Slack | `/slack/api` | 21 | [`slack.schema.json`](../backlot/schemas/slack.schema.json) | A Slack message. |
```
nuri-yoo: **[note]** The count moved to 21, but the hand-written Slack table under "Per-service detail" (line 346 onward) still lists the twelve read methods and none of the nine this PR adds. Its lead sentence, "Each method answers on `GET` and `POST` alike", is what the new baseline notes cite, so that section is where a reader would look for the write surface.

nuri-yoo (changes requested):
Reviewed in a worktree at `449f403` (`main` at `1a62604`), `.[all]` installed: `pytest` gives `3384 passed, 1 skipped` (the hubspot reader), `ruff check .` and `ruff format --check .` are clean, `backlot diff --source slack` reports `228 divergence(s) · 228 acknowledged · 0 new`. I re-measured the verb claim against slack.com on 2026-09-18 with the read-only token: the nine methods over `GET` and `POST` answer `missing_scope` on eight and `invalid_arguments` on `chat.getPermalink`, as the body says. I also drove `slack_sdk` 3.44.1 at a `backlot serve` process over the SAMPLE corpus, and re-measured `chat.getPermalink` on real messages.

Two are worth fixing before this leaves draft; the rest are notes. Both blocks are inline: the SDK sends `chat.postMessage`, `chat.update` and `chat.postEphemeral` as JSON and all three answer `invalid_arguments` here, and the explicit `out` a post writes for a non-member records a departure the agent never made while the comment justifying it names a mechanism that does not run (the public-membership derivation reads `main.slack_messages` only).

The notes, all inline: the permalink shape real gives anything in a thread (measured), `channel` accepting a name on chat.postMessage and the 40,000-character truncation (both from docs.slack.dev, both answered differently here), `text` being optional in the spec on the three posting methods, the docs' own condition for `cant_update_message`, `reactions.get` without a `permalink`, two numerals that do not match what they count, the threadpool premise behind the lock (every Slack handler is `async def`), an `if True:` and a misnamed `tomb`, `reactions.list`'s `pages`, same-second posts reading back in hash order, the cost of `search.messages` once a few thousand rows are in the overlay (1.2s to 1.5s against 5ms), what deleting a thread root leaves behind, the hand-written Slack table in `docs/supported-sources.md`, and two `_uid` sites no test reaches.

Outside the diff, prose this PR makes false: `backlot/cli.py:155` (`backlot --help` still says "Enterprise SaaS read APIs (Slack, …)"), `backlot/__init__.py:1` (the same sentence), `tests/test_sdk.py:1` ("Read-only coverage"), and `examples/using-mcp-with-agents/s3.py:76` ("Backlot has no write endpoints, so dropping the read-only guard is safe here"). The first two are user-facing, which is what the new AGENTS.md rule is about. The other routers' "(read-only)" module docstrings are facts about their own sources and can stay.

Mutations, each applied to a copy of the file and restored from it, run over `tests/test_slack.py tests/test_overlay.py tests/test_store.py tests/test_openapi.py` at `449f403`:

```
M1   chat.postMessage: drop the explicit 'out' write                     SURVIVED
M2   chat.delete: drop _invalidate                                        SURVIVED
M3   _corpus_not_reindexed returns ''                                     caught  test_editing_a_corpus_message_leaves_one_search_hit
M4   count_search ignores overlay hits                                    caught  test_count_search_agrees_with_what_search_pages (+1)
M5   _literal_tier always 0                                               SURVIVED
M6   overlay.reset without rollback                                       caught  test_reset_empties_the_overlay_and_keeps_it_usable
M7   /_meta/overlay/reset without _require_admin                          caught  test_meta_overlay_reset_needs_the_admin_token
M8   _message: no bot_id on bot_message                                   caught  test_a_bot_message_carries_bot_id_in_history
M9   _reactions: synth.slack_user_id instead of _uid                      SURVIVED
M10  _edited: synth.slack_user_id instead of _uid                         SURVIVED
M11  _slack_author_by_uid: no USERVICE0 entry                             caught  test_a_bot_messages_author_resolves_through_users_info
M12  _member_count: cache.get(name, 0) again                              caught  test_a_write_drops_the_stale_member_count
M13  slack_next_ts: probe main only                                       caught  test_next_ts_does_not_collide_within_a_channel_second (+1)
M14  private member page: keep deactivated explicit 'in'                  SURVIVED
M15  thread_ts reply under the named message, not the root                caught  test_replying_to_a_reply_lands_in_the_same_thread
M16  _own_message: case-sensitive authorship                              caught  test_an_author_with_a_differently_cased_address_can_edit_their_own_message
M17  reactions.remove: keep an emptied reaction                           caught  test_the_last_person_removing_a_reaction_removes_it_entirely
M18  _fts_query_terms: deduplicate terms                                  caught  test_python_bm25_agrees_with_sqlite_on_corpus_rows[the the]
M19  search merge: recency ignores direction                              SURVIVED
M20  reset: keep app.state.channel_members                                SURVIVED
M21  _is_member: ignore the explicit state                                SURVIVED
M22  count_slack_channel_members: ignore overrides (public)               caught  test_an_explicit_out_removes_a_speaker_from_every_membership_answer (+1)
M23  slack_channel_member_counts: never recount an overridden channel     caught  test_an_explicit_out_removes_a_speaker_from_every_membership_answer
M24  chat.postMessage: drop _invalidate                                   SURVIVED
M25  _patch_locked: never re-index a text patch                           caught  test_a_patched_corpus_row_is_not_counted_twice_by_search (+1)
M26  tombstone_document: leave the overlay fts entry                      SURVIVED
M27  insert_document: copy corpus grants only, not overlay 'in' grants    SURVIVED
M28  ACL view: drop overlay grants                                        caught  test_post_message_is_readable_back_by_the_poster (+1)
```

Sixteen caught, twelve survived. M1, M21, M2 and M24 share one root, the block on line 1098: because the derivation never sees an overlay row, neither the `out` write nor the `_is_member` branch that reads it nor the two cache drops has an observable effect today. M5 and M19 are the Python side of the merge: no test searches with overlay hits present and a punctuated query, or with `sort=timestamp` in both directions. M9 and M10 are inline. M14 is the deactivated filter on an explicit `in` for a private page. M20 is the cache drop on reset. M26 is harmless (`merged_source` already subtracts the row, so the stale index entry matches nothing) and M27 is expected, since nothing writes an overlay grant yet; both are worth saying so in a comment rather than a test.

On the body: the Verification block is the 2026-09-06 run (`2998 passed`); at `449f403`, after the merges from `main`, it is `3384 passed, 1 skipped`. The other claims I checked hold: both verbs (re-measured), `0 new`, eleven ACL tables, sixty-seven `_acl_clause` call sites, twelve mirrored columns, and "every `missing_operation` entry now carries a note" (the fourteen empty notes left in the baseline are all `missing_param`).

nuri-yoo:
The review above is a pre-review: I saw the PR was back in draft and went ahead anyway, so take it whenever you mark it ready. Nothing in it needs an answer before then.

khj809:
> The review above is a pre-review: I saw the PR was back in draft and went ahead anyway, so take it whenever you mark it ready. Nothing in it needs an answer before then.

This was accidentally marked as ready while I'm testing backlot-loop skill. It tried to touch PRs that were not generated by the loop and this was affected. Sorry for the confusion, and thanks for your pre-review anyway.

backlot/routers/slack.py
khj809: Fixed in 58b444d. `parse_slack_form` reads a JSON body too, but only for the six methods whose `consumes` declares one — a new `slack.JSON_BODY_METHODS`. A JSON body on the three GET-typed methods is ignored, which is what you measured real doing. Values are stringified on the way in so a JSON `true` or `7` reaches `_param` as the text the form path would have carried. `tests/test_sdk.py` now drives all nine through `slack_sdk`; removing the JSON branch from a copy of `main.py` makes it fail (`1 SDK check failed`) and restoring it green.

backlot/routers/slack.py
khj809: You are right, and the premise was mine to check. Confirmed: `_slack_public_members_sql` and `slack_channel_has_author` both read `slack_messages` unqualified, and the only `temp` view is the ACL one, so an overlay row is never a speaker. The write is gone in 58b444d, along with the comment and the now-unused `was_member`; the docstring says instead that nothing has to arrange this because the derivation reads `main`. The explicit branch of `_is_member` stays for `conversations.join`/`kick` and now has a test that reaches it in both directions, folded into `test_posting_does_not_silently_join_a_public_channel` (M21).

backlot/store.py
khj809: Fixed in 58b444d — four readers, four answers.

backlot/store.py
khj809: Dropped the number in 58b444d rather than tracking it; the sentence now says every caller pastes the fragment in and none restates the rule, which is the part that has to stay true.

backlot/overlay.py
khj809: Corrected in 58b444d, in all three docstrings. The lock's reason is the two `/_meta/overlay*` handlers being plain `def` where every Slack handler is `async def`, not a threadpool the Slack writes never touch.

backlot/store.py
khj809: Both fixed in 58b444d: the `if True:` is gone and the body unindented, and the `PATCHABLE` check moved into `_require_patchable`, called by `patch_document` and `edit_document` before either takes the lock.

backlot/store.py
khj809: Renamed to `attached` in 58b444d, holding `overlay.is_attached(conn)`.

backlot/routers/slack.py
khj809: Fixed in 58b444d: `pages` is `(total + count - 1) // count` rather than 1.

backlot/routers/slack.py
khj809: Both covered in 58b444d, folded into `test_the_service_token_posts_as_a_bot`: it now reacts and edits as the service token and asserts `USERVICE0` on the reaction's `users` and on `edited.user`. M9 and M10 fail without `_uid`.

backlot/routers/slack.py
khj809: Reclassified in 58b444d. `msg_too_long` is out of the workspace-limits sentence; the docstring quotes the 40,000-character rule from docs.slack.dev and states that Backlot does not truncate yet. Implementing the truncation is behaviour rather than prose, so it is on the follow-up list below.

backlot/routers/slack.py
khj809: Requoted in 58b444d. Both docstrings now carry the vendor's own condition from docs.slack.dev's error tables, so neither refusal reads as Backlot's choice.

docs/supported-sources.md
khj809: Fixed in 58b444d: the Per-service detail table lists all nine, and a paragraph under it points at `docs/overlay.md` and states which six take a JSON body.

backlot/routers/slack.py
khj809: Agreed, and it is pre-existing on `_message`'s permalink as well as this one — please file it. Keeping this PR to the write surface; leaving the thread open so the issue can close it.

backlot/routers/slack.py
khj809: Taking this as a separate change: `text` becoming optional moves `no_text` from `_missing_argument`'s absent-argument path on three methods, which is a refusal change rather than the write surface. Leaving open.

backlot/routers/slack.py
khj809: Left as you found it, deliberately: with no rule in the spec and no write scope to measure one, any of the three answers I could make consistent would be invented. Worth its own issue against a scratch workspace with `chat:write`. Leaving open.

backlot/routers/slack.py
khj809: Deferred: `_channel_name` would need a by-name branch for one caller, and docs.slack.dev is the only source (the scope check answers first with the token in `.env`). Its own change. Leaving open.

backlot/routers/slack.py
khj809: Deferred with the permalink thread shape above — both are `_message`/`_search_match` permalink work and belong in one change. Leaving open.

backlot/store.py
khj809: Real's minting is unmeasured, so this is a consequence of sharing the importer's rule rather than a divergence I can close against a measurement. A sequential fraction for posted messages is a small change but it makes served ids depend on write order, which is worth deciding on its own. Leaving open.

backlot/store.py
khj809: Confirmed the double computation by reading it: `count_search` and `search_documents` each call `_overlay_hits`. Passing one result into the other, and grouping the vocabulary lookups, is a change with its own measurement to bring; it is the one follow-up I would put first, since an evaluation run is exactly the workload that hits it. Leaving open.

PR #199, opened 2026-09-12
gdrive: the MCP slice is Drive, Docs, Sheets and Slides, and no prefix in the table selects nothing

### LINKED ISSUE #170: gdrive: the MCP slice is Drive only, so the Docs, Sheets and Slides routes have no tool and no /_meta/openapi entry
**What you ran** — asked Backlot for the `gdrive` OpenAPI slice the way the shipped skill tells an agent to. `skills/backlot/SKILL.md` says "`GET /_meta/openapi/<key>` returns a source's typed OpenAPI slice, where `<key>` is the **OpenAPI slice** column of the table below", and that table's `google_drive` row lists the URL prefixes `/drive/v3` `/docs/v1` `/sheets/v4` `/slides/v1` and the slice key `gdrive` on the same line.

```python
import json, urllib.request, backlot

with backlot.serve() as s:
    served = json.load(urllib.request.urlopen(f"{s.base_url}/openapi.json"))["paths"]
    slice_ = json.load(urllib.request.urlopen(f"{s.base_url}/_meta/openapi/gdrive"))["paths"]
    for prefix in ("/drive/v3", "/docs/v1", "/sheets/v4", "/slides/v1"):
        n = sum(1 for p in served if p.startswith(prefix))
        m = sum(1 for p in slice_ if p.startswith(prefix))
        print(f"{prefix:12} served {n}   in /_meta/openapi/gdrive {m}")
```

**What happened**

```
/drive/v3    served 6   in /_meta/openapi/gdrive 6
/docs/v1     served 1   in /_meta/openapi/gdrive 0
/sheets/v4   served 3   in /_meta/openapi/gdrive 0
/slides/v1   served 1   in /_meta/openapi/gdrive 0
```

The slice is six operations, all of them Drive's. The five routes Backlot serves for Docs, Sheets and Slides — `/docs/v1/documents/{document_id}`, `/sheets/v4/spreadsheets/{spreadsheet_id}`, `/sheets/v4/spreadsheets/{spreadsheet_id}/values/{a1_range}`, `/sheets/v4/spreadsheets/{spreadsheet_id}/values:batchGet` and `/slides/v1/presentations/{presentation_id}` — are absent. `backlot mcp --source gdrive` builds its tools from the same slice, so it offers six Drive tools and no way to read a document, a spreadsheet or a presentation.

**What you expected instead** — the four prefixes the skill's own row lists, which is what every other description of this source in the repository says. `README.md:82` gives Google Drive as `/drive/v3` `/docs/v1` `/sheets/v4` `/slides/v1`. `scripts/gen_docs.py:53-56` holds the same four in `SOURCES["google_drive"]`, and its module docstring states the reason a mapping lives there at all: "one `google_drive` spans `/drive`, `/docs`, `/sheets` and `/slides`". `docs/supported-sources.md` documents all five routes with their parameters. So a client following any of them asks for a surface the slice does not carry.

**Where it is** — `backlot/openapi.py:35-44`. `SOURCE_PREFIXES["gdrive"]` is `["/drive"]`, and `slice_spec` keeps only paths under one of a source's prefixes, so `/docs/v1`, `/sheets/v4` and `/slides/v1` are dropped before `build_mcp_spec` ever sees them. The comment above the dict — "source -> the path prefix(es) whose operations that source's MCP server should expose" — reads as if the narrowing were deliberate; nothing else in the repository says it is, and `gen_docs.py:16-18` describes this table only as being scoped to the MCP bridge's naming ("it merges Jira with Confluence under `atlassian` and knows `google_drive` as `gdrive`"), not to Drive alone.

This was incomplete from the start rather than made stale later: `git log -S` puts `"gdrive": ["/drive"]` and the five editor routes in the same commit, `936fb10` (#44, 2026-08-11).

The same dict has a second, harmless defect worth fixing in the same edit. `SOURCE_PREFIXES["atlassian"]` carries `/wiki`, which matches no served path — Confluence is served under `/atlassian/wiki/...`, which `/atlassian` already covers. It has selected nothing since #44. The atlassian slice is complete, so this costs a reader rather than a client, but a dead prefix in a table like this is the kind of thing that reads as coverage.

**The fix** is `["/drive", "/docs", "/sheets", "/slides"]` (or the version-qualified spellings `gen_docs.SOURCES` already uses), and dropping `/wiki`. A dead-prefix check already exists for the other two prefix tables and not for this one: `gen_docs.validate()` fails when a `SOURCES` prefix matches no path in `app.openapi()`, and #162 adds the same check for the fidelity registry's mounts in `test_every_mount_selects_something_backlot_serves`. Asserting the other direction as well — that every served path outside Backlot's own surface is under some source's prefixes — is what would have caught the three missing here, since `/drive` on its own selects something and so passes a dead-prefix check.

**Where you ran it** — `main` at `1246b35`, Python 3.14.4, macOS 26.5.2 arm64, FastAPI 0.141.1, Starlette 1.6.0. Not behind an optional dependency: `/_meta/openapi/{source}` is core, and the repro above needs nothing installed beyond the package. `backlot mcp` needs the `mcp` extra, but it reads the same slice. Found while reviewing #162, which fixes the same blind spot in the fidelity registry — `google_drive` compared against Drive's document alone — and left this table untouched.

nuri-yoo:
Closes #170. Three commits off `main` at `6490f3b`, nothing merged in: `7ce8474` is the table and the checks that keep it honest, `72df52e` the four sentences elsewhere that stated the old values, `e965fcd` one sentence in the skill that has been wrong since #109 and that #170 quotes as the instruction an agent follows. No served route changes, so `backlot diff` sees nothing — `--source google_drive` is `213 divergence(s) · 213 acknowledged · 0 new` on `main` at `6490f3b` and the same here.

**What the slice answered.** `SOURCE_PREFIXES["gdrive"]` was `["/drive"]`, and `slice_spec` keeps only paths under one of a source's prefixes, so Docs, Sheets and Slides were dropped before `build_mcp_spec` ever saw them. The issue's repro, re-run at `e965fcd`, with `main`'s answers beside it:

| prefix | served | in `/_meta/openapi/gdrive` on `main` | here |
|---|---|---|---|
| `/drive/v3` | 6 | 6 | 6 |
| `/docs/v1` | 1 | 0 | 1 |
| `/sheets/v4` | 5 | 0 | 5 |
| `/slides/v1` | 1 | 0 | 1 |

One number moved since the issue was filed: `/sheets/v4` was 3 there and is 5 now, because #161 added `values:batchGetByDataFilter` and `:getByDataFilter` (`git log -S` puts both in `165e019`). Neither reached the slice either, which is why the test names the routes rather than counting them.

**What a client got.** `backlot mcp --source gdrive` builds from the same slice, so it offered six tools, all Drive's, and no way to open what a listing had just named. It now offers thirteen, driven in-process through `fastmcp.Client` against the bundled corpus: `docs_get`, `sheets_get`, `sheets_get_by_data_filter`, `sheets_values_batch_get`, `sheets_values_batch_get_by_data_filter`, `sheets_values_get` and `slides_get` join Drive's six — `drive_about`, `drive_files_export`, `drive_files_get`, `drive_files_list`, `drive_files_permissions`, `drive_shared_drives`. Two calls, admin credential: `drive_files_list` with `q=mimeType='application/vnd.google-apps.spreadsheet'` answers `1Hkbyz5pNvA32DZLy21MMkLi1iq-fOFmi` "Pricing Sheet 2026", and `sheets_get` on that id answers its `Sheet1` grid; the same pair over `…apps.document` and `docs_get` answers "Northwind Proposal". Under the composite server's `<source>_` namespace they read `gdrive_docs_get`, `gdrive_sheets_get`, `gdrive_slides_get` and so on — no stutter, because `mcp._HANDLER_PREFIX` folds only the prefix that repeats the namespace, and the longest, `gdrive_sheets_values_batch_get_by_data_filter` at 45 characters, stays inside the 64-character cap `test_bridged_operation_ids_are_the_route_names` holds.

I took the version-qualified spellings #170 offers rather than `/docs`, `/sheets`, `/slides`, so the entry reads the same as `gen_docs.SOURCES`, which is the repository's other statement of the same mapping.

**Why `/drive` alone passed every check the table had.** `slice_spec` raises when a source slices to nothing, which a dead prefix beside a live one never reaches, and `gen_docs.validate()`'s dead-prefix check is over `SOURCES`, not this table. Nothing asked the other direction at all, which is the direction that would have caught this: `/drive` selects six operations, so a dead-prefix check passes while three roots go unbridged. `tests/test_openapi.py` now asserts both, plus the case that started it:

- `test_the_gdrive_slice_carries_the_docs_sheets_and_slides_routes` — the regression test, and not implied by the coverage test below it: declaring `/docs/v1` in `NOT_BRIDGED` would satisfy that one and leave the editors unreachable, so this is the cell that says which bucket they belong in. Reverting the entry to `["/drive"]` fails both, nothing else.
- `test_every_source_prefix_selects_something` — the dead-prefix direction, which this table did not have. Restoring `/wiki` fails it alone.
- `test_every_served_path_is_bridged_or_says_why_not` — every served path is under some source's prefixes or declared in `NOT_BRIDGED`. Deleting the `/batch` entry fails it alone.
- `test_no_not_bridged_declaration_outlives_its_route` — an entry may not outlive its route or claim a bridged one. Adding a `/gmail` entry fails it alone.

**`/wiki`.** `SOURCE_PREFIXES["atlassian"]` carried it, and it selected nothing for the whole of its life: `936fb10` (#44) is both the commit that added the prefix and the one in which `routers/atlassian.py` already reads `prefix="/atlassian"`, so Confluence has always been served under `/atlassian/wiki/...`, which `/atlassian` covers. The atlassian slice is nineteen operations either way, measured on `main` at `6490f3b` and here: ten `confluence_*` tools and nine `jira_*`. So this costs a reader rather than a client, which is what #170 says.

**`NOT_BRIDGED`.** Four entries, in the shape `fidelity.comparisons.UNCOMPARED` set, because the coverage check needs an escape hatch and an escape hatch needs to be a reason a reviewer reads. `/health`, `/oauth2/token` and `/_meta` are Backlot's own surface, which is not a vendor's. `/batch` is a vendor's: it is Google's `multipart/mixed` transport, each part an `application/http` sub-request with its own method and target that `routers.google.batch` dispatches back through the app, and the route declares no request body, so the only arguments a tool derived from it would get are the `api` and `version` query params it shares with `/batch/{api}/{version}`, neither of which can carry a sub-request.

**The prose in `72df52e`.** `examples/using-mcp-with-agents/gdrive.py` said the bridge "slices to `/drive`". The example README said `gdrive.py` is "likewise for Google Drive (`/drive/*`)", and that the atlassian entry "covers both the `/atlassian` (Jira) and `/wiki` (Confluence) path roots". `mcp._HANDLER_PREFIX`'s comment said `routers/google.py` "holds Gmail and Drive together", which is three surfaces short and leaves a reader of `gdrive_sheets_get` wondering why `sheets` was not folded into the namespace the way `drive` is. One is older than this branch: `tests/test_skills.py` said six of the eleven `source_type`s 404 at `/_meta/openapi/`, and #109 made it five by giving S3 a key.

**`e965fcd` is a separate commit because it is a separate bug.** `skills/backlot/SKILL.md` is the paragraph #170 quotes, and it says "S3 has neither" a slice nor GraphQL introspection. Since #109 `/_meta/openapi/s3` answers three operations, so an agent following the skill would not ask for them. Its count is dropped rather than corrected, because the rule — a `source_type` 404s wherever it is spelled differently from its key — is what a reader needs and does not go stale on the next key. Say the word and I will split it out; it is outside #170's subject, and it is in #170's own quotation.

**Not in this PR.** Whether `gdrive` should be four MCP sources rather than one is not settled here, but the table already answers it by precedent: an MCP source is a `source_type`, which is how `docs/supported-sources.md`, the README table and `gen_docs.SOURCES` all describe `google_drive`, and `atlassian` already hands a model Jira and Confluence as one source of 19 tools while `github` alone is 33. At 13, `gdrive` is neither the largest toolset nor the first to span products. Nothing here touches the fidelity registry — #162 closed the same blind spot there, and this PR closes it in the table beside it.

**Verification.** At `e965fcd`, in a worktree off `main` at `6490f3b`: `pytest -rs` is `2915 passed, 1 skipped`, the skip being `tests/test_llamaindex.py`'s when `llama_index.readers.hubspot` is absent, which it is here; nothing else self-skipped, so mirage, fastmcp, fsspec, boto3 and the vendor SDKs all ran. `ruff check`, `ruff format --check` and `PYTHONPATH=$PWD python scripts/gen_docs.py --check` each exit 0; the generated blocks do not move, because the docs table reads `gen_docs.SOURCES` and the skill's slice column derives the key by prefix overlap, which `/drive/v3` satisfies exactly as `/drive` did. `backlot diff --source google_drive` is `213 divergence(s) · 213 acknowledged · 0 new`. The four mutations above were each applied to a committed tree and reverted from a copy, and each one fails exactly the tests named beside it.

### DIFF
--- a/backlot/mcp.py
+++ b/backlot/mcp.py
@@ -75,8 +75,11 @@ PROBE_TIMEOUT = 2
 NAMED_SERVER_TIMEOUT = 10
 
 # The prefix a source's route handlers carry to stay unique inside one module (`routers/google.py`
-# holds Gmail and Drive together), which under the `<source>_` namespace would be said twice:
+# holds all five Google surfaces), which under the `<source>_` namespace would be said twice:
 # `gmail_gmail_messages_list`, `gdrive_drive_files_list`. `mounted_name` folds it into the namespace.
+# Only the prefix that repeats the namespace is folded. `gdrive` bridges Docs, Sheets and Slides
+# as well as Drive, and their handlers are `docs_get`, `sheets_get` and `slides_get`: `gdrive` says
+# which source, and the handler's own prefix says which of its four surfaces answers.
 _HANDLER_PREFIX: dict[str, str] = {"gmail": "gmail", "gdrive": "drive"}
 
 # The one per-source knob, measured (figures in examples/using-mcp-with-agents/README.md). Linear
--- a/backlot/openapi.py
+++ b/backlot/openapi.py
@@ -32,15 +32,47 @@ import warnings
 warnings.filterwarnings("ignore", message="Duplicate Operation ID", category=UserWarning)
 
 # source -> the path prefix(es) whose operations that source's MCP server should expose.
+#
+# A source is named here for the MCP bridge, so one entry can span several vendor APIs: `gdrive`
+# is the Drive, Docs, Sheets and Slides roots, because one `google_drive` record is a file that
+# Drive lists and one of the three editors reads. Every served path outside `NOT_BRIDGED` has to
+# be under one of these, and every prefix has to select something — `test_openapi.py` asserts both
+# directions, which a prefix that selects something on its own does not give (`/drive` alone did,
+# while `/docs/v1`, `/sheets/v4` and `/slides/v1` went unbridged from #44 until #170).
 SOURCE_PREFIXES: dict[str, list[str]] = {
     "github": ["/github"],
     "slack": ["/slack/api"],
     "gmail": ["/gmail"],
-    "gdrive": ["/drive"],
+    "gdrive": ["/drive/v3", "/docs/v1", "/sheets/v4", "/slides/v1"],
     "notion": ["/notion/v1"],
-    "atlassian": ["/atlassian", "/wiki"],
+    "atlassian": ["/atlassian"],
     "hubspot": ["/hubspot"],
     "s3": ["/s3"],
+}
+
+# Served paths no source's MCP server exposes, and why. Matched by prefix, so `/batch` covers
+# `/batch/{api}/{version}`.
+#
+# The escape hatch the coverage check named above is built around, in the shape
+# `fidelity.comparisons.UNCOMPARED` set: an entry is a reason a reviewer reads, not a silence.
+# Three are Backlot's own surface, which is not a vendor's and so has no place in a toolset built
+# to answer as one; `/batch` is a vendor's, and says its own reason.
+NOT_BRIDGED: dict[str, str] = {
+    "/health": "Backlot's own liveness endpoint, not a vendor's surface",
+    "/oauth2/token": (
+        "Backlot's own token exchange. `backlot.mcp` reads the whole credential set off "
+        "`/_meta/users` and puts it on every tool call, so an agent has nothing to exchange."
+    ),
+    "/_meta": (
+        "Backlot's own introspection — the corpus's principals and the per-source OpenAPI. This "
+        "table is what the second of those answers, so a tool built from it would be circular."
+    ),
+    "/batch": (
+        "Google's `multipart/mixed` batch transport, which carries operations rather than being "
+        "one: a part is an `application/http` sub-request with its own method and target, which "
+        "`routers.google.batch` dispatches back through this app. The route itself declares no "
+        "request body, so a tool derived from it would take no arguments at all."
+    ),
 }
 
 _METHODS = ("get", "post", "put", "delete", "patch")
--- a/examples/using-mcp-with-agents/gdrive.py
+++ b/examples/using-mcp-with-agents/gdrive.py
@@ -3,8 +3,9 @@
 
 Official and community Drive MCP servers hard-wire `googleapis.com` and require real Google OAuth,
 so none can be pointed at a self-hosted server. Instead `backlot mcp --source gdrive` turns
-Backlot's typed `/openapi.json` into MCP tools: it slices to `/drive`, dedupes operation aliases,
-and serves them over stdio with a `Bearer <token>` header — retrieval is ACL-scoped by
+Backlot's typed `/openapi.json` into MCP tools: it slices to `/drive/v3`, `/docs/v1`, `/sheets/v4`
+and `/slides/v1` — the editors a Drive file opens in are the same source — dedupes operation
+aliases, and serves them over stdio with a `Bearer <token>` header — retrieval is ACL-scoped by
 `--user` (default: the admin; any email from GET /_meta/users).
 
 Prereqs: `pip install -e ".[mcp]"` (installs fastmcp); an LLM key for --agent
--- a/examples/using-mcp-with-agents/README.md
+++ b/examples/using-mcp-with-agents/README.md
@@ -24,7 +24,7 @@ service (like the other `examples/` dirs) — run the one you want:
   (`/slack/api/*`) as tools instead.
 - **`gmail.py`** via the same **OpenAPI→MCP bridge** — Gmail MCP servers hard-wire `googleapis.com`
   and need real Google OAuth, so the bridge serves Backlot's Gmail API (`/gmail/*`) as tools.
-- **`gdrive.py`** via the same **OpenAPI→MCP bridge** — likewise for Google Drive (`/drive/*`).
+- **`gdrive.py`** via the same **OpenAPI→MCP bridge** — likewise for Google Drive, and for the three editors a Drive file opens in: the `gdrive` slice is `/drive/v3`, `/docs/v1`, `/sheets/v4` and `/slides/v1`, so an agent that finds a spreadsheet can read its cells in the next call.
 - **`hubspot.py`** via the same **OpenAPI→MCP bridge** — no HubSpot MCP server takes a base-URL
   override. Because the CRM API is polymorphic over `{object_type}`, the agent gets five tools that
   each work across every object type (list, read, search, batch-read, associations) rather than a set
@@ -236,8 +236,8 @@ backlot mcp --source atlassian --url <url> --user someone@acme.com
 ```
 
 Atlassian authenticates with HTTP Basic, the resolved user's own email as the username and their
-Backlot token as the password, and its `SOURCE_PREFIXES` entry covers both the `/atlassian` (Jira)
-and `/wiki` (Confluence) path roots.
+Backlot token as the password, and its `SOURCE_PREFIXES` entry is the single `/atlassian` root,
+which covers Jira and Confluence alike — Confluence is served under `/atlassian/wiki/…`.
 
 **S3 goes through the same path, signed rather than bearer-authenticated.** SigV4 signs each
 request, so a fixed `Authorization` header cannot serve it — the bridge signs every call with
--- a/skills/backlot/SKILL.md
+++ b/skills/backlot/SKILL.md
@@ -146,9 +146,9 @@ the naming does not generalise: Google Drive is `gdrive` throughout, but only
 
 With a server already up, prefer the served surface over any description of it.
 `GET /_meta/openapi/<key>` returns a source's typed OpenAPI slice, where `<key>` is the **OpenAPI
-slice** column of the table below — not the `source_type`, which 404s for six of the eleven. A `—`
-in that column means no slice exists: Linear and Fireflies answer full GraphQL introspection at
-their own endpoints instead, and S3 has neither. The 404 body lists every key it does accept.
+slice** column of the table below — not the `source_type`, which 404s wherever the two spellings
+differ. A `—` in that column means no slice exists: Linear and Fireflies answer full GraphQL
+introspection at their own endpoints instead. The 404 body lists every key it does accept.
 
 ## 9. Check the answer against the corpus
 
--- a/tests/test_openapi.py
+++ b/tests/test_openapi.py
@@ -333,3 +333,92 @@ def test_drop_head_operations_drops_a_path_it_empties():
         }
     }
     assert openapi.drop_head_operations(spec)["paths"] == {"/both": {"get": {"operationId": "b"}}}
+
+
+def test_the_gdrive_slice_carries_the_docs_sheets_and_slides_routes():
+    """One `google_drive` record is a file Drive lists and one of the three editors reads, so a
+    slice holding Drive alone offers an agent no way to open what it just found.
+
+    `SOURCE_PREFIXES["gdrive"]` was `["/drive"]` from #44 until #170, which dropped the Docs,
+    Sheets and Slides routes before `build_mcp_spec` saw them: `/_meta/openapi/gdrive` answered
+    Drive's six operations and `backlot mcp --source gdrive` offered six tools. The routes are
+    named rather than counted, because the count moved while the gap was open — #161 added the two
+    `…ByDataFilter` routes after #170 was filed, and neither reached the slice either.
+
+    Not covered by `test_every_served_path_is_bridged_or_says_why_not` below, which asks only that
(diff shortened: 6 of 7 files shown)


examples/using-mcp-with-agents/gdrive.py
```diff
@@ -3,8 +3,9 @@
 
 Official and community Drive MCP servers hard-wire `googleapis.com` and require real Google OAuth,
 so none can be pointed at a self-hosted server. Instead `backlot mcp --source gdrive` turns
-Backlot's typed `/openapi.json` into MCP tools: it slices to `/drive`, dedupes operation aliases,
-and serves them over stdio with a `Bearer <token>` header — retrieval is ACL-scoped by
+Backlot's typed `/openapi.json` into MCP tools: it slices to `/drive/v3`, `/docs/v1`, `/sheets/v4`
+and `/slides/v1` — the editors a Drive file opens in are the same source — dedupes operation
+aliases, and serves them over stdio with a `Bearer <token>` header — retrieval is ACL-scoped by
```
khj809, line 8: The docstring now advertises the Sheets and Slides slice, but `CORPUS` below is two `subtype: "document"` records and `QUESTION` asks only for those, so `sheets_get`, `slides_get` and the rest can only ever come back empty in the one script the README points a reader at.

One `subtype: "spreadsheet"` record and a clause asking for its cells would make the example demonstrate what the README bullet now claims.

khj809 (commented):
Reproduced the table, the thirteen tools and the four mutations — they land exactly as described, and `backlot diff --source google_drive` is `213 · 213 acknowledged · 0 new` here too. Three things below.

backlot/openapi.py
```diff
@@ -32,17 +32,49 @@
...
+    ),
+    "/_meta": (
+        "Backlot's own introspection — the corpus's principals and the per-source OpenAPI. This "
+        "table is what the second of those answers, so a tool built from it would be circular."
+    ),
+    "/batch": (
+        "Google's `multipart/mixed` batch transport, which carries operations rather than being "
+        "one: a part is an `application/http` sub-request with its own method and target, which "
+        "`routers.google.batch` dispatches back through this app. The route itself declares no "
+        "request body, so a tool derived from it would take no arguments at all."
```
khj809: The no-body half is right, but the conclusion isn't: `/batch` declares `api` and `version` as optional query params (FastAPI derives them from the signature it shares with `/batch/{api}/{version}`, and `dedupe_operations` keeps the bare path), so a derived tool would take two arguments rather than none.

```suggestion
        "`routers.google.batch` dispatches back through this app. The route declares no request "
        "body, so the only arguments a tool would get are the `api` and `version` query params it "
        "shares with `/batch/{api}/{version}`, neither of which can carry a sub-request."
```

backlot/mcp.py
```diff
@@ -75,8 +75,11 @@
 NAMED_SERVER_TIMEOUT = 10
 
 # The prefix a source's route handlers carry to stay unique inside one module (`routers/google.py`
-# holds Gmail and Drive together), which under the `<source>_` namespace would be said twice:
+# holds all five Google surfaces), which under the `<source>_` namespace would be said twice:
 # `gmail_gmail_messages_list`, `gdrive_drive_files_list`. `mounted_name` folds it into the namespace.
+# Only the prefix that repeats the namespace is folded. `gdrive` bridges Docs, Sheets and Slides
+# as well as Drive, and their handlers are `docs_get`, `sheets_get` and `slides_get`: `gdrive` says
+# which source, and the handler's own prefix says which of its four surfaces answers.
```
khj809, line 82: Of the three handlers named here, `sheets_get` carries a docstring and `docs_get` / `slides_get` do not, so those two reach a model with `description: ""` and nothing but FastAPI's `"Docs Get"` / `"Slides Get"` summary and an opaque id parameter.

Could you add a one-line docstring to each in `routers/google.py`? This PR is what first puts them in front of an agent.

backlot/openapi.py
nuri-yoo: Applied as f8b9e73. The PR body's `NOT_BRIDGED` paragraph repeated the same sentence, so it now says the two query params too.

backlot/mcp.py
nuri-yoo: Added in 73866ea, one line each: `docs_get` names the one-`paragraph`-per-line shape, `slides_get` the one-`TEXT_BOX`-slide-per-block shape. Checked through the bridge with `list_tools()`: both now carry that line as `description`. One correction to the framing: FastMCP does not send `""` for a route without a docstring, it copies FastAPI's summary, so what a model saw was `description: "Docs Get"` and `"Slides Get"`; the conclusion is the same. Of the thirteen tools, three more have only that summary, `drive_files_get`, `drive_files_export` and `drive_files_permissions`; they were in the slice before this PR, so I left them for a separate change.

examples/using-mcp-with-agents/gdrive.py
nuri-yoo: Done in 8f21197. The corpus gains one `subtype: "spreadsheet"` record stating a 4x3 grid as `sheets` (no `content`, as `docs/corpus.md` requires when `sheets` is present), and the question gains a clause asking which region's p95 rose the most. Driven through the bridge without a model: `drive_files_list` with `q=mimeType='application/vnd.google-apps.spreadsheet'` answers the one file, `sheets_values_get` on `p95!A1:C4` answers its four rows, and `fullText contains 'checkout'` finds it alongside the postmortem, because Drive's search reads the `content` derived from the first sheet, whose header row carries the word. With a model as well: the same runner (`_agent._run_anthropic`, tool runner with adaptive thinking) on `us.anthropic.claude-fable-5-1` through Bedrock finds all three files, reads the `p95` sheet through the bridge and answers us-east, +1,690 ms (410 to 2,100), naming eu-west (+1,520) and ap-south (+120) beside it, then summarizes the postmortem and the runbook and cites the three titles.

backlot/openapi.py
```diff
@@ -32,17 +32,50 @@
 warnings.filterwarnings("ignore", message="Duplicate Operation ID", category=UserWarning)
 
 # source -> the path prefix(es) whose operations that source's MCP server should expose.
+#
+# A source is named here for the MCP bridge, so one entry can span several vendor APIs: `gdrive`
+# is the Drive, Docs, Sheets and Slides roots, because one `google_drive` record is a file that
+# Drive lists and one of the three editors reads. Every served path outside `NOT_BRIDGED` has to
+# be under one of these, and every prefix has to select something — `test_openapi.py` asserts both
+# directions, which a prefix that selects something on its own does not give (`/drive` alone did,
+# while `/docs/v1`, `/sheets/v4` and `/slides/v1` went unbridged from #44 until #170).
```
khj809: The parenthetical is development history — which prefix was wrong, and between which two issues — so it reads as narration of this change rather than as the rule a reader needs. The invariant is worth keeping; the reason a dead-prefix check cannot see it stands on its own:

```suggestion
# be under one of these, and every prefix has to select something — `test_openapi.py` asserts both
# directions, because a source whose other prefix selects something still slices to a non-empty
# spec, so a dead-prefix check alone cannot see a root that goes unbridged.
```

khj809 (approved):
All three check out here: the `/batch` reason (and the PR body) now names the two query params, `docs_get` and `slides_get` come through the bridge carrying their new line as `description`, and the example's spreadsheet answers `sheets_values_get` on `p95!A1:C4` with the four rows the question needs. Full suite, `ruff`, and `gen_docs.py --check` are clean. One comment-only nit below — approving either way.

PR #200, opened 2026-09-12
google: the batch endpoint is compared against the batchPath every discovery document declares

### LINKED ISSUE #171: google: batch endpoint needs its own comparison kind, the way S3 has a probe
**What's missing** — `backlot diff` has no way to check `POST /batch` and `POST /batch/{api}/{version}` against the vendor, so #162 lists them in `UNCOMPARED`. They are the only served vendor surface in that table; the other three entries there are Backlot's own (`/health`, `/oauth2/token`, `/_meta`).

Google does document the endpoint. Every discovery document Backlot compares against names one in its top-level `batchPath`, measured 2026-09-09:

```
gmail v1     batchPath='batch'            rootUrl=https://gmail.googleapis.com/
drive v3     batchPath='batch/drive/v3'   rootUrl=https://www.googleapis.com/
docs v1      batchPath='batch'            rootUrl=https://docs.googleapis.com/
sheets v4    batchPath='batch'            rootUrl=https://sheets.googleapis.com/
slides v4    batchPath='batch'            rootUrl=https://slides.googleapis.com/
```

Each API has its own host with `batch` at the root; Drive alone sits on the shared `www.googleapis.com` and so carries the `batch/drive/v3` discriminator. Backlot's two routes ([`routers/google.py:126-127`](https://github.com/brekkylab/backlot/blob/main/backlot/routers/google.py#L126)) are exactly those two shapes, collapsed onto one origin.

**Why the existing machinery cannot do it**, which is what makes this a new comparison kind rather than a registry entry:

- `from_google_discovery` builds operations from `resources[].methods[]`. `batchPath` is a top-level string, and no document declares a batch operation, so a path diff has nothing on the vendor side to pair. Verified across all five.
- A path diff pairs operations selected by a per-**source** mount, and Google's batch endpoints are per-**host**. Backlot's single `/batch` stands in for Drive's, Sheets', Gmail's, Docs' and Slides' at once, so it belongs to no one source's mount. Pairing it with any one document's `batchPath` would assert a correspondence that is not one-to-one.

**What it unblocks** — the endpoint stops being invisible to the check that exists to notice vendor drift. Concretely: if Google moved a `batchPath` to a third shape, or dropped batch for an API, nothing today would report it, and the `UNCOMPARED` entry would go on reading as a deliberate exemption. The same argument #162 makes for the Sheets response shape applies here.

**Where it lands** — `backlot/fidelity/comparisons.py`, as a fourth comparison kind beside `OpenAPIComparison`, `GoogleDiscoveryComparison`, `GraphQLComparison` and `ProbeComparison`.

`ProbeComparison` is the precedent and the argument for the shape. Its docstring already draws the line this needs: "What separates them is not whether a document exists but whether it enumerates operations as PATHS." S3 is compared behaviourally because it dispatches on the query string; batch needs its own kind for the same reason at a different point — the contract is a document field, not an operation.

What it would assert is narrow: that every document Backlot compares against declares a `batchPath`, and that each declared value is covered by one of Backlot's two route shapes (`batch` or `batch/<api>/<version>`). That is checkable today and would have to be re-checked on every run, since the point is drift.

Two constraints on where the check runs:

- `tests/test_fidelity.py` opens with "No network: every comparison here is built from a document or an SDL in this file", so a live-fetching check does not belong there. It goes on the `backlot diff` path, which is already the scheduled, network-touching one.
- Whatever it reports has to fit `Finding` and be acknowledgeable in a baseline like every other kind, so `--update-baseline` and the breaking/gap split keep working unchanged.

Once it lands, `/batch` comes out of `UNCOMPARED` and `docs/fidelity.md`'s coverage table gains a row, the way S3's "probed" row reads now.

Found while reviewing #162. That PR leaves the `UNCOMPARED` entry in place as the accurate record until this lands; its reason text describes the obstacle rather than claiming the endpoint is undocumented.

nuri-yoo:
Closes #171. Four commits: `c784f5b` is the comparison, and `2ec207a`, `2d8037b` and `3b762a9` are fixes from review, described under **What the review changed** below. `backlot/fidelity/google_batch.py` compares Google's batch endpoint against the `batchPath` every discovery document declares it in, `BatchPathComparison` registers that as a kind in `backlot/fidelity/comparisons.py`, and `/batch` leaves `UNCOMPARED` and joins the coverage check. It lands as the fifth kind rather than the fourth the issue counts: `OpenAPIComparison`, `GoogleDiscoveryComparison`, `GraphQLComparison` and `ProbeComparison` are already four.

**What the vendor declares.** Measured 2026-09-12, over the same five documents the Gmail and Drive comparisons read. The issue's 2026-09-09 reading reproduces unchanged.

| Document | `batchPath` | `rootUrl` | Answered by |
|---|---|---|---|
| `gmail:v1` | `batch` | `https://gmail.googleapis.com/` | `/batch` |
| `drive:v3` | `batch/drive/v3` | `https://www.googleapis.com/` | `/batch/{api}/{version}` |
| `docs:v1` | `batch` | `https://docs.googleapis.com/` | `/batch` |
| `sheets:v4` | `batch` | `https://sheets.googleapis.com/` | `/batch` |
| `slides:v1` | `batch` | `https://slides.googleapis.com/` | `/batch` |

None of the five declares that endpoint as a method, so there is nothing for a path diff to pair Backlot's routes with. The methods that do carry the word are a different thing and are already compared as operations: `messages.batchModify` and `messages.batchDelete` on Gmail, `documents.batchUpdate` on Docs, seven `batch*` methods on Sheets, `presentations.batchUpdate` on Slides. Measured across the five documents, no path `from_google_discovery` builds is a `batchPath` value.

**Why it is a kind and not a mount on an existing source.** Each API answers batch on its own host, and Drive alone sits on the shared `www.googleapis.com`, which is what the `drive/v3` in its value discriminates. Backlot collapses those hosts onto one origin, so `/batch` stands in for Gmail's, Docs', Sheets' and Slides' at once and `/batch/{api}/{version}` for Drive's. Mounting the endpoint under one of those sources would assert a correspondence that is not one to one; mounting it under each would put one served path in several sources' baselines and report it once per source. `ProbeComparison` is the precedent and the docstring says where the two part company: S3 is compared behaviourally because what the document does not enumerate as a path is an *operation*, and here it is a *field*.

**What it reports.** Three kinds, and the vocabulary follows the existing reading rule that `missing_*` is reliable and `extra_*` means undocumented by the vendor rather than proof of a bug.

| Kind | Severity | Fires when | What drift it catches |
|---|---|---|---|
| `extra_batch_api` | breaking | a document declares no `batchPath` | Google drops batch for one API while Backlot goes on answering it |
| `missing_batch_path` | gap | a declared value no Backlot route answers | a value moves to a third shape |
| `extra_batch_route` | breaking | a Backlot route no declared value selects | Drive moves to its own host, leaving `/batch/{api}/{version}` served for nobody |

A `missing_batch_path` finding is identified by the document and the value together — `drive:v3 v3/batch`, not `drive:v3`. A gap is acknowledged by identity alone (`Baseline.unacknowledged` reads a gap's detail for nothing), so keyed on the document, the entry `--update-baseline` wrote for one move would answer for the next one too, and a second move would pass as `0 new`. The path diffs have that property for free, since there the moved path is the identity.

The first two are what #171 asks for. The third is the other direction and is not in the issue, so it is worth saying why it is here: without it, Drive moving to its own host would have every document declare `batch`, every value answered, and `/batch/{api}/{version}` standing for nothing forever with a green check — the same failure `test_every_mount_selects_something_backlot_serves` exists to stop for a mount, at a place no mount reaches. The repo's own rule is that a one-direction comparison is worse than none.

Two of the three are breaking, so `--update-baseline` cannot acknowledge them and a scheduled run stays red until someone writes the note by hand. That is the existing rule for a breaking finding, not a new one.

**Where each side comes from.** The vendor's side is `_documents_of(GOOGLE_DISCOVERY.values())` — every document the Google path diffs already read, in their order, deduplicated by URL — rather than a second list of the same URLs, which would go on reading a document the registry had repointed and would leave a sixth Google API's `batchPath` unchecked. Backlot's side is `operations.from_backlot(app.openapi(), comparison.mount)`, the same reader the path diffs use, so the route shapes come out of the app rather than out of a constant. A finding is identified by what the document calls itself (`gmail:v1`), not by the URL it was fetched from: measured 2026-09-12, Google answers that same document at both `gmail.googleapis.com/$discovery/rest?version=v1` and `www.googleapis.com/discovery/v1/apis/gmail/v1/rest`, so repointing the registry from one to the other must not read as a new divergence. That is also why the comparison passes its findings through `_one_report`, where every other kind's report is keyed: a registry holding both of those URLs reads one document twice, and a baseline keyed on `kind:path` cannot hold the second copy.

The cost of comparing this separately is five more public GETs on the nightly, since the Gmail and Drive jobs fetch the same documents for their own path diffs. Folding the check into those two jobs would save them and give up what the separate kind is for: the endpoint would then sit in one source's baseline or in both, and `backlot diff --source google_batch` would not exist as the thing that answers for it.

**What this moves in the registry.** `COMPARISONS` gains one key that is not a `source_type`, so `test_the_registry_names_exactly_the_source_types_backlot_serves` now reads `set(store.SOURCE_TABLE) | {"google_batch"}`, with the exception spelled out rather than excused by a rule so a second one cannot arrive as a typo. `cli.FIDELITY_SOURCES` gains the same name — `--help` and the validated set stay one list — and the CI matrix picks it up from `COMPARISONS` with no change to the workflow. The `UNCOMPARED` entry for `/batch` is gone; the comment above the dict now records that every remaining entry is Backlot's own surface and that `/batch` stopped qualifying because its documents do describe it, in a field rather than as an operation.

**What the tests hold.** Thirteen new tests, `tests/test_fidelity.py` collecting 100 cases in all, none of them touching the network — the file's opening promise. Each rule was checked by reverting it and watching the suite go red rather than by reading the code, counted here in tests rather than in parametrised cases:

```
answers() drops its segment-count check      3: the matcher, and both whole-comparison route cases
the extra_batch_route loop is removed        2: the moved-value case and the own-host case
the moved value is left out of the gap's
  identity (keyed on the document alone)     2: the moved-value case and the second-move case
a route some document declares is reported
  anyway (selected is never updated)         5
a document with no batchPath is skipped      2: the dropped-API case, and the one-document-
                                                at-two-URLs case, whose fixture is two without one
batch_path() stops canonicalising            1 test, 2 cases: the leading slash and the trailing one
a document is identified by its URL          4
the comparison claims no mount               7, one of them test_every_served_path_is_compared_or_says_why_not
the report is not keyed through _one_report  1: the one-document-at-two-URLs case
_documents_of is not deduplicated by URL     1: the shared-document case
the batch registry lists its own URLs        2, one of them the drift check against GOOGLE_DISCOVERY
BatchPathComparison is left out of the
  dispatcher's isinstance guard              1: the new dispatcher test
/batch is left in UNCOMPARED as well         1: test_no_uncompared_declaration_outlives_its_route
```

The dispatcher test and the shared-document test are the two of the thirteen that are not about `batchPath` itself, and the last two rows are why they exist. Leaving the new class out of the dispatcher's guard broke nothing in the whole suite before `test_the_dispatcher_hands_over_for_every_kind_the_registry_holds` existed: the guard is an isinstance check over a hand-written tuple, so the miss would have surfaced on a nightly run as `BatchPathComparison is not a kind of comparison this knows how to run` and nowhere else. And `_documents_of` deduplicates a case no Google document exercises today, so it takes an argument instead of reading the registry, which makes the case reachable from a test rather than left as an unfireable check.

**What the review changed.** Four findings, each reproduced before it was fixed: three from a self-review, one from review.

`2ec207a`, two of them. `test_a_batch_path_that_moved_is_a_gap_and_leaves_its_route_standing_for_nothing` said it was simulating Drive getting its own host while its fixture declared `v3/batch`, and those are different scenarios: Drive on its own host would declare `batch`, like the four APIs already there, and that reports one breaking finding and no gap. So the docstring now says what the fixture is — a value moved to a third shape — and the case it had been claiming got the test it did not have, `test_a_route_no_document_declares_any_more_is_breaking_on_its_own`, which is the scenario `docs/fidelity.md` names for `extra_batch_route` and was the one bullet of the three with nothing behind it. Measured over five documents all declaring `batch`: exactly one finding, `extra_batch_route /batch/{}/{}`.

`2d8037b`, the third. Two documents that call themselves the same thing reported twice — measured, four documents of which two answered `{"id": "gmail:v1"}` produced two `extra_batch_api gmail:v1` findings — which is the harm `_one_report` is written to prevent for the path diffs, and this kind was not going through it. It does now, at the comparison rather than in the module, so there is no second copy of that rule; the module still reports what it read, and the new test asserts both halves of that difference. Fixing it made the sentence explaining why `_documents_of` deduplicates by URL false — the duplicate report is no longer what that guard prevents — so that comment and the test on it now say what the URL dedup actually saves, a vendor round trip per run, and point at the keying for the rest.

`3b762a9`, the fourth, from review. `missing_batch_path` was keyed on the document alone, and a gap is acknowledged by identity alone — reproduced by acknowledging the finding for `drive:v3` at `v3/batch` and running the comparison with the value moved again to `v4/batch`: `unacknowledged` answered `[]`. The value is now part of the identity, `test_a_value_that_moves_again_is_not_covered_by_the_first_move_s_acknowledgement` holds the reproduction, and the docstring, `docs/fidelity.md` and the sample report below say so.

**Verification.** At `3b762a9`, on `main` at `6490f3b`.

```
PYTHONPATH=. python -m pytest -rs                          2938 passed, 1 skipped
ruff check .                                               All checks passed!
ruff format --check .                                      140 files already formatted
python scripts/gen_docs.py --check                         exit 0
backlot diff --source google_batch                         0 divergence(s), 0 acknowledged, 0 new
backlot diff --source google_batch --update-baseline       acknowledges 0; the shipped baseline is unchanged
```

The one skip is `llama_index.readers.hubspot`, an optional extra this branch does not touch. The live run is against Google, so it is the measurement in the table above rather than a fixture. What the report looks like when it fires, with Drive's answer patched to `v3/batch` and everything else the real command:

```
🔍 google_batch · https://gmail.googleapis.com/$discovery/rest?version=v1, …
   2 divergence(s) · 0 acknowledged · 2 new

❌ breaking (1) — Backlot contradicts the vendor
     extra_batch_route  /batch/{}/{}
       Backlot serves it, and no compared document declares a batchPath it answers

     missing_batch_path  drive:v3 v3/batch
       the vendor batches at 'v3/batch'; no Backlot route answers that shape
```

**Not in this PR.** Nothing about what the batch endpoint *does* is compared: the multipart envelope, the `Content-ID` pairing, the outer credential applying to a sub-request that carries none. Those are behaviour, they are held by `tests/test_google.py`, and reaching them would need the probe kind rather than this one. Nor is the endpoint asked of Google — that needs a credential, which is the line every comparison but the two GraphQL ones stays on the public side of.

One thing found while measuring and left alone: `POST /batch` declares `api` and `version` as query parameters in Backlot's own `/openapi.json`, because the handler serves both routes and FastAPI reads the two defaults as query parameters on the route whose template does not name them. No discovery document declares any parameter for `batchPath`, and what real Google does with `?api=&version=` on `gmail.googleapis.com/batch` is unmeasured here, so this comparison — which is over paths — says nothing about it either way. Deciding it needs a measurement against a real Google account.

### DIFF
--- a/backlot/cli.py
+++ b/backlot/cli.py
@@ -39,11 +39,15 @@ from backlot import mcp as _mcp
 # `erb` is what this codebase calls it among itself, and a caller reading `--type erb` learns nothing
 # about what is being downloaded. The cost of the abbreviation lands on the person who did not write
 # the code, so it is not offered.
-# The sources `backlot diff` can compare, named here so `--help` lists them without importing
+# What `backlot diff` can compare, named here so `--help` lists them without importing
 # backlot.fidelity (and httpx's TLS stack) on every other command.
+#
+# Every `source_type` Backlot serves, and `google_batch`, which is not one: Backlot's batch routes
+# stand in for five Google APIs, so they sit under no source's mount and are compared on their own.
 FIDELITY_SOURCES = (
     "confluence",
     "fireflies",
+    "google_batch",
     "google_drive",
     "github",
     "gmail",
--- a/backlot/fidelity/baseline/google_batch.json
+++ b/backlot/fidelity/baseline/google_batch.json
@@ -1,0 +1,12 @@
+{
+  "source": "google_batch",
+  "endpoints": [
+    "https://gmail.googleapis.com/$discovery/rest?version=v1",
+    "https://www.googleapis.com/discovery/v1/apis/drive/v3/rest",
+    "https://www.googleapis.com/discovery/v1/apis/docs/v1/rest",
+    "https://www.googleapis.com/discovery/v1/apis/sheets/v4/rest",
+    "https://www.googleapis.com/discovery/v1/apis/slides/v1/rest"
+  ],
+  "measured": "2026-09-12",
+  "acknowledged": []
+}
--- a/backlot/fidelity/comparisons.py
+++ b/backlot/fidelity/comparisons.py
@@ -1,13 +1,15 @@
 """Which comparison each source gets, and what it needs to run.
 
-Four kinds, and they differ in what each side of the comparison is. For the two GraphQL sources,
+Five kinds, and they differ in what each side of the comparison is. For the two GraphQL sources,
 Backlot's side is the SDL the server builds its engine from and the vendor's is a live introspection
 response, which needs a credential. For the eight document sources, Backlot's side is the app's own
 ``app.openapi()`` and the vendor's is one or more documents it publishes, which needs none. For
 S3, whose operations are selected by query string rather than by path, both sides are answers from
-a running server that ``backlot.serve()`` starts.
+a running server that ``backlot.serve()`` starts. For Google's batch endpoint, which every discovery
+document names in a top-level field rather than declaring as an operation, Backlot's side is the
+paths it serves under ``/batch`` and the vendor's is that field.
 
-Nine of the eleven need no credential. All eleven run on a schedule and never on a pull request:
+Ten of the twelve need no credential. All twelve run on a schedule and never on a pull request:
 drift is this project's bug, but it is never the bug of whichever pull request happens to be open
 when a vendor ships a change.
 """
@@ -17,9 +19,10 @@ from __future__ import annotations
 import os
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Any, Callable, Mapping
+from typing import Any, Callable, Iterable, Mapping
 
 from backlot.fidelity import (
+    google_batch,
     google_discovery_diff,
     graphql_diff,
     hubspot_catalog,
@@ -246,6 +249,45 @@ class ProbeComparison:
         return s3_probe.divergences(self, timeout=timeout)
 
 
+@dataclass(frozen=True)
+class BatchPathComparison:
+    """Google's batch endpoint, compared against the field every discovery document declares it in.
+
+    Its own kind because the contract here is a document FIELD rather than an operation. Every
+    document names a batch endpoint in its top-level ``batchPath`` and none declares that endpoint
+    as a method, so a path diff has nothing on the vendor's side to pair Backlot's routes with —
+    the same argument :class:`ProbeComparison` makes for S3 at a different point, where what a
+    document does not enumerate as a path is an operation rather than a field.
+
+    And the endpoint belongs to no one source. Each API answers batch on its own host, which
+    Backlot collapses onto one origin: its ``/batch`` stands in for Gmail's, Docs', Sheets' and
+    Slides' at once. Mounting it under one of those sources would assert a correspondence that is
+    not one to one; mounting it under each would put one served path in several sources' baselines
+    and report it once per source.
+
+    ``documents``, not ``specs``: a document's own ``mount`` says which served paths that document
+    speaks for, and those are not the paths this comparison covers. ``mount`` below is.
+    """
+
+    name: str
+    documents: tuple[Spec, ...]
+    # Which served paths this comparison speaks for, as every kind answers. Both batch routes are
+    # under this one prefix, so what the coverage check reads here is also what
+    # `google_batch.divergences` selects Backlot's side of the comparison with.
+    mount: tuple[str, ...] = ("/batch",)
+    # None: the documents are the same public ones the Google path diffs read.
+    credentials: tuple[Credential, ...] = ()
+
+    @property
+    def endpoints(self) -> tuple[str, ...]:
+        return tuple(d.endpoint for d in self.documents)
+
+    def divergences(
+        self, credentials: Mapping[str, str] | None = None, *, timeout: float = 120.0
+    ) -> list[Finding]:
+        return google_batch.divergences(self, timeout=timeout)
+
+
 OPENAPI = {
     "slack": OpenAPIComparison(
         name="slack",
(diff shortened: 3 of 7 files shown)


khj809:
When I filed this issue, the direction I thought was extending current `google_discovery_diff` and `GoogleDiscoveryComparison` to cover the batch endpoints too, not adding the extra comparison and separate baseline file. If you don't mind, can I make a change to align with my direction?

nuri-yoo:
> When I filed this issue, the direction I thought was extending current `google_discovery_diff` and `GoogleDiscoveryComparison` to cover the batch endpoints too, not adding the extra comparison and separate baseline file. If you don't mind, can I make a change to align with my direction?

Absolutely. Feel free to change

khj809:
Closing in favour of #283, which compares the same `batchPath` field without a fifth comparison kind: `GoogleDiscoveryComparison` gains a `batch_mount` naming the batch routes a source speaks for, so the check runs inside the two Google sources' existing baselines. `COMPARISONS` keeps holding exactly the `source_type`s Backlot serves, and there is no `google_batch.json`.

#171 stays open until #283 lands.

PR #201, opened 2026-09-12
drive: files.list parses q as the reference's grammar and refuses what it cannot evaluate; sheets: a values read accepts R1C1 notation (#173, #174)

### LINKED ISSUE #173: drive: files.list serves the unfiltered listing for q clauses it does not parse (name =, name !=, mimeType contains, modifiedTime <, or, not, unknown terms)
**Source and commit** — `google_drive`, at `364e071` (main).

`files.list` matches `q` clause by clause, with one regex per clause it knows (`trashed`, `'x' in parents`, `mimeType =`/`!=`, `name contains`, `modifiedTime >`, `sharedWithMe`, `'x' in owners`), and by its own docstring ignores the rest. A clause it does not parse therefore drops out of the filter rather than out of the result, and the caller gets the unfiltered listing under a 200. #23 §1 found the same mechanism behind `sharedWithMe` and that clause was added; the rows below are operators the reference lists for terms Backlot already handles, sent in a form the regexes do not match, plus the two connectives.

| # | Clause | Real (reference) | Backlot | Silent? |
|---|---|---|---|---|
| 1 | `name = 'First Week Checklist'` | `name` takes `contains`, `=`, `!=` | every visible file, 16 | yes |
| 2 | `name != 'First Week Checklist'` | as above | every visible file, 16 | yes |
| 3 | `mimeType contains 'spreadsheet'` | `mimeType` takes `contains`, `=`, `!=` | every visible file, 16 | yes |
| 4 | `modifiedTime < '2026-01-10T00:00:00Z'` | `modifiedTime` takes `<=`, `<`, `=`, `!=`, `>`, `>=` | every visible file, 16 | yes |
| 5 | `name contains 'Checklist' or name contains 'Runbook'` | `or` composes | the first clause alone, 1 | yes |
| 6 | `not name contains 'Checklist'` | `not` negates | the clause un-negated, 1 | yes |
| 7 | `bogusField = 'x'` | not a query term | every visible file, 16 | yes |

None of this is in `backlot/fidelity/baseline/google_drive.json`: the discovery diff compares parameters, and `q` is one string parameter Backlot declares. Nothing reports these today.

**How you know** — Drive's query terms reference (developers.google.com/workspace/drive/api/guides/ref-search-terms), read 2026-09-10. The real API was not called for this issue, so what real answers for row 7 is spec only.

- The reference lists the operators per term as in the table: `name` and `mimeType` take `contains`, `=`, `!=`; `modifiedTime` takes `<=`, `<`, `=`, `!=`, `>`, `>=`; `fullText` takes `contains` only; `trashed` takes `=`, `!=`.
- `name` usage there: "Name of the file. Surround with single quotes (`'`). Escape single quotes in queries with `\'`, such as `'Valentine\'s Day'`." Backlot's `[^']+` capture stops at an escaped quote; not measured further here.
- The Backlot column is measured against `backlot.serve()` at 364e071 on 2026-09-10 with the reproduction below.

**Reproduction** — the bundled corpus:

```python
import backlot, requests

with backlot.serve() as m:
    for q in (
        "name = 'First Week Checklist'",
        "name != 'First Week Checklist'",
        "mimeType contains 'spreadsheet'",
        "modifiedTime < '2026-01-10T00:00:00Z'",
        "name contains 'Checklist' or name contains 'Runbook'",
        "not name contains 'Checklist'",
        "bogusField = 'x'",
    ):
        r = requests.get(f"{m.base_url}/drive/v3/files", headers=h, params={"q": q, "fields": "files(name)"})
        print(q, "->", len(r.json()["files"]))
```

Prints 16, 16, 16, 16, 1, 1, 16. For comparison `name contains 'Checklist'` alone prints 1 and `mimeType = 'application/vnd.google-apps.spreadsheet'` prints 3, which is the shape the unparsed clauses should have taken.

**What it costs a client** — a client that resolves a file by exact name sends `name = '...'`, the documented equality form, and gets the whole corpus back with a 200. One that takes the first hit has the wrong file and no signal; one that asserts a single match fails on every lookup. mirage resolves Drive files with `name contains`, so it is unaffected; a client written from the reference is not. `or` and `not` are worse in kind: the answer is a plausible subset, so nothing looks wrong.

**Where it lands** — `_drive_q_match_facts` in `backlot/routers/google.py` (the per-clause regexes) and the `name contains` SQL push-down in `_drive_q_rows`. `_DRIVE_MIME_RE` already handles two operators for one term and is the shape to extend.

**What a fix has to settle**

- Whether an unparsed clause is refused or honoured. Refusing (400 on any clause the matcher does not recognise) closes the silent case for every term at once and is small; honouring needs a `q` grammar (terms, operators, `and`/`or`/`not`, parentheses, `\'` escaping) rather than seven regexes. Real's answer to an unknown term is unmeasured, so a refusal's status and wording have no source until someone measures it. Recommendation: refuse now, then grow the grammar term by term with a measurement each.
- Which operators to honour first. `name =`/`!=` and the full `modifiedTime` set are what a client sends without thinking; `mimeType contains` is rarer.
- `or` and `not` cannot be added per clause; they need the expression parsed. Whether that belongs here or in its own issue.

### LINKED ISSUE #174: sheets: values.get refuses R1C1 notation, which the discovery document names beside A1 and the LlamaIndex reader sends
**Source and commit** — `google_drive`, at `364e071` (main).

`values.get` and `values:batchGet` take A1 notation only. The discovery document describes the `range` parameter as "The A1 notation or R1C1 notation of the range to retrieve values from", and the parser refuses every R1C1 form with the `Unable to parse range` message it gives unparseable input. #161 reworks the A1 parser for multi-sheet workbooks and inherits the gap unchanged.

| # | Range | Real (discovery) | Backlot | Silent? |
|---|---|---|---|---|
| 1 | `R1C1:R2C2` | R1C1 notation accepted | 400 `Unable to parse range: R1C1:R2C2` | no |
| 2 | `Sheet1!R1C1:R2C2` | as above | 400 | no |
| 3 | `R1C1` | as above | 400 | no |

Not in the baseline: the discovery diff sees `range` as one path parameter Backlot declares, and the notation it accepts is not a parameter-level fact.

**How you know** — the Sheets v4 discovery document (`sheets.googleapis.com/$discovery/rest?version=v4`), read 2026-09-10. The real API was not called for this issue.

- `spreadsheets.values.get`, parameter `range`: "The [A1 notation or R1C1 notation](https://developers.google.com/workspace/sheets/api/guides/concepts#cell) of the range to retrieve values from."
- What real echoes as `range` for an R1C1 request, A1 or the form sent, is not measured. A fix has to measure it before the test asserts one.
- The Backlot column is measured against `backlot.serve()` at 364e071 on 2026-09-10.

**Reproduction** — the bundled corpus, spreadsheet `First Week Checklist`:

```python
import backlot, requests

with backlot.serve() as m:
    sid = requests.get(f"{m.base_url}/drive/v3/files", headers=h,
                       params={"q": "name contains 'First Week Checklist'", "fields": "files(id)"}).json()["files"][0]["id"]
    for rng in ("R1C1:R2C2", "Sheet1!R1C1:R2C2", "R1C1", "Sheet1!A1:B2"):
        r = requests.get(f"{m.base_url}/sheets/v4/spreadsheets/{sid}/values/{rng}", headers=h)
        print(rng, r.status_code, r.json().get("range") or r.json()["error"]["message"])
```

The three R1C1 forms are 400; the A1 form answers `Sheet1!A1:B2`.

**What it costs a client** — the installed LlamaIndex `GoogleSheetsReader` (`llama-index-readers-google` 0.8.0) reads every sheet as `R1C1:R{rowCount}C{columnCount}`, built from `spreadsheets.get`. Pointed at Backlot the way `examples/using-llamaindex-readers/gdrive.py` points the Drive reader, `load_data` on the bundled spreadsheet fails on its first values call:

```
HttpError 400 ... /sheets/v4/spreadsheets/<id>/values/R1C1%3AR1000C26?alt=json returned "Unable to parse range: R1C1:R1000C26"
```

So the README's LlamaIndex row holds for Drive and not for Sheets, on every spreadsheet. `backlot.integrations.llamaindex` also has no `point_sheets_at`; the Drive shim's pattern carries over once the range parses.

**Where it lands** — `_A1_END`, `_a1_endpoint` and `_a1_range` in `backlot/routers/google.py`. On #161's branch the split between sheet name and cell part moves to `_a1_sheet` and `_a1_looks_like_a_range`, so a second notation has to be recognised at that split as well as where the endpoints resolve.

**What a fix has to settle**

- The echoed `range`. Real's answer for an R1C1 request is unmeasured; A1 is the likely form, but a real spreadsheet should say so before the test asserts it.
- Which R1C1 forms. `R1C1:R2C2` is what the reader sends; whether whole-row and whole-column forms (`R1:R2`, `C1:C2`), a bare `R1C1` cell, and a mixed range (`A1:R2C2`) are accepted or refused are measurements too.
- A sheet titled like an R1C1 reference (`R1C1`, `R2`) meets the ambiguity #159 A.4 measured for A1, where bare `A1` is a cell even when a sheet is named `A1`. The fix should apply the same rule.
- The LlamaIndex reader as a test client. Once this parses, the `gdrive.py` example or a sibling should read a spreadsheet through `GoogleSheetsReader`, which is the client that exposed it.

nuri-yoo:
Closes #173, closes #174. Drive's `files.list` parses `q` as the reference's grammar and refuses what it cannot evaluate; Sheets' `values.get` and `values:batchGet` accept R1C1 notation; and the LlamaIndex `GoogleSheetsReader`, the client that exposed #174, has a shim, a test and an example.

**Drive `files.list` parses `q` instead of pattern-matching it (#173).** A clause is either evaluated or refused; none drops out of the filter while the listing answers 200. The issue's seven rows against the bundled corpus, on `main` (`6490f3b`) and on this branch:

| `q` | main | this branch |
|---|---|---|
| `name = 'First Week Checklist'` | 16 | 1 |
| `name != 'First Week Checklist'` | 16 | 15 |
| `mimeType contains 'spreadsheet'` | 16 | 3 |
| `modifiedTime < '2026-01-10T00:00:00Z'` | 16 | 5 |
| `name contains 'Checklist' or name contains 'Runbook'` | 1 | 2 |
| `not name contains 'Checklist'` | 1 | 15 |
| `bogusField = 'x'` | 16 | 400 |

`name contains 'Checklist'` answers 1 and `mimeType = 'application/vnd.google-apps.spreadsheet'` answers 3 on both. The client this bit is not hypothetical: mirage's Drive folder listing resolves a file with `name='…'` and bounds a sync with `modifiedTime >= '…'` and `modifiedTime < '…'` (read off `mirage/core/google/drive.py`), and its `escape_query_value` writes the `\'` the reference asks for.

The grammar is the reference's ([ref-search-terms](https://developers.google.com/workspace/drive/api/guides/ref-search-terms)): a term is `<field> <operator> <value>` or `'<value>' in <collection>`, terms join with `and` and `or`, `not` negates, parentheses group, and an apostrophe inside a value is `\'` ("Escape single quotes in queries with `\'`"). The operators per term are the reference's table: `name` and `mimeType` take `contains`, `=`, `!=`; `fullText` takes `contains`; `modifiedTime` and `createdTime` take `<=`, `<`, `=`, `!=`, `>`, `>=`; `trashed` and `sharedWithMe` take `=`, `!=`; `parents` and `owners` take `in`, and `'me' in owners` is the caller, resolved through the identity `sharedWithMe` reads (the admin token is not a Drive user and owns nothing). Whether `and` binds before `or` the reference does not say; Backlot reads `and` first. How `name` compares is measured (2026-09-14, four titles created in a real Drive and deleted after): `name =` and `!=` fold case for ASCII letters and nothing else — `'ÉLAN VITAL'` finds `Élan Vital`, `'élan vital'` and `'strasse plan'` do not — and `name contains` folds case across the alphabet and compatibility forms (`'finance'` finds `ﬁnance deck`, `'istanbul'` finds `İstanbul`) but neither `ß` nor accents. The seventeen measured rows are `MEASURED_NAME` in `tests/test_google.py`, each sent through the handler. One thing is Backlot's rule rather than measured: a time value is compared as an instant, so `'2026-01-21T09:00:00'` and `'…Z'` are equal, while one that does not parse as RFC 3339 is refused, since real's answer for it is unmeasured.

A term Backlot cannot evaluate is a 400 on `q`, never a silence. Two kinds. A term the reference does not list (`bogusField = 'x'`), a clause that is not the grammar, or a query nested more than 32 levels deep (the parser is recursive; unbounded, 320 parentheses answered a 500) is refused with the parameter envelope's bare `Invalid Value` (`reason: invalid, location: q`, the envelope the measured `orderBy` refusal uses); real Drive refuses these too and its wording is unmeasured, so nothing fuller is claimed. A documented term Backlot holds no fact for — `starred`, `viewedByMeTime`, `writers`, `readers`, `properties`, `appProperties`, `visibility` — is refused with a message that names it, the way an unmodelled `orderBy` key is: honouring `starred = true` as "everything" would be the listing the parser exists to stop. `{` and `}` are tokens so that the reference's own spelling of the two map-valued terms, `properties has { key='x' and value='y' }`, reaches that message rather than the bare one.

The SQL narrowing survives. The query is parsed once in the handler, and the terms every match has to satisfy — those joined by `and` at the top, with nothing under an `or` or a `not` — are what the candidate set is built from: a `fullText contains` conjunct takes the index's ranked rows; a `name contains` or `name =` conjunct the title `LIKE` (measured on 20,000 records: 8 ms against 120 ms for the whole-corpus scan the `=` form took before); a `sharedWithMe` conjunct the owner filter; `'root' in parents` the empty row stream; a plain folder scope `list_drive_folder`. For `name` the candidate set is the answer for that term, not a superset the evaluator narrows: `=` is SQLite's `LIKE` with no wildcards, which folds case for ASCII letters only, the rule real Drive measured to; `contains` is a `LIKE` over `gdrive_files.title_fold`, the `drive_name_fold` of the title the importer stores, against the same fold of the needle, so SQL and `_drive_q_eval` compare the same text and nothing the evaluator accepts is outside the candidate set. The column is the first one a build reads that an older DB lacks, and `CREATE TABLE IF NOT EXISTS` adds no column to a table it finds, so the startup check that named a missing table now names a missing column too: `store.missing_schema` reads every table's columns off `SCHEMA`'s text and a DB built before `title_fold` is refused at startup with `has no gdrive_files.title_fold`, the same refusal and the same remedy (re-import) a missing table gets, rather than a bare 500 on every `name contains`. The whole tree is then evaluated in Python over each candidate, which is what makes `or` and `not` correct: a `fullText` term under an `or` becomes a membership test against the index's answer rather than a pre-filter, so `fullText contains 'palette' or name = 'Q1 Deck'` returns both. The title `LIKE` answers non-trashed rows only, so it is used only when `trashed = false` is itself a conjunct — a query asking for trash through `trashed = true` or `not trashed = false` goes to the full listing. The synthesized folders go through the same evaluator, so `mimeType = '…folder' or name = 'Q1 Deck'` returns the folders and the file. A query with no `trashed` term gets `and trashed = false` appended at parse, as the plain listing excludes trash through `exclude_trashed`.

`createdTime` is an evaluated term. `test_google_serves_a_corpus_declared_epoch_zero` sends `createdTime < '1970-01-02T00:00:00'` and passes on the evaluation.

**Sheets `values.get` and `values:batchGet` accept R1C1 notation (#174).** The discovery document describes `range` as "The A1 notation or R1C1 notation of the range to retrieve values from". The LlamaIndex `GoogleSheetsReader` reads every sheet as `R1C1:R{rowCount}C{columnCount}` built from `spreadsheets.get`, so its first values call against any spreadsheet is an R1C1 one.

The grammar is measured, not read off a document: 217 requests to the live Sheets API on 2026-09-12 against a workbook created for it (sheets `Sheet1`, `Data`, `R1C1`, `RC` and `A`, 1000x26 each), every one pinned as a row of `tests/test_google.py::MEASURED_R1C1` — the range as sent, the status, the echoed `range` or the error message — and `test_sheets_values_answer_every_measured_r1c1_request_as_real_does` serves a corpus with the same five sheets and holds Backlot to all 217. The six rules those rows yield are stated once, in the comment above `_R1C1_END` in `routers/google.py`, and the helpers that apply them point there. In one line each: `R` and `C` take an absolute 1-based number, a bracketed 0-based offset from A1, or no number; either letter may be absent, an absent axis being unbounded beside a numbered half; both halves of a range are one notation, and a token is read as A1, then R1C1, then as a sheet name; a reversed range swaps as A1's does except that same-kind numbers with start == end + 1 as written are refused; the echo and the grid rules are A1's; whitespace anywhere is refused and an empty title before the bang is the first sheet.

Three A1 behaviours the same measurement settled are fixed here, because the same code paths carry them: row 0 is unparseable (`A0`, `A1:A0`, `0:1`); whitespace is unparseable wherever it stands (around the whole, around the bang, around the colon, inside a token, inside a quoted title, a tab as much as a space); and a whole column or row past the grid is named by its letters or numbers alone in the refusal (`ZZ:ZZ` reports `Sheet1!ZZ`, `1001:1002` reports `Sheet1!1001:1002`, `C[26]` reports `Sheet1!AA`).

`point_sheets_at` joins `point_gmail_at` and `point_drive_at` in `backlot.integrations.llamaindex`, through the same `build` wrapper (the reader calls `discovery.build("sheets", "v4", …)` on the imported module, so the wrapper is what that lookup finds), with `api_endpoint = base + "/sheets"` — no version, the Sheets document spells it in the path, the split `examples/using-official-sdk/gdrive.py` already makes. `tests/test_llamaindex.py::test_gsheets` reads the SAMPLE spreadsheet through `GoogleSheetsReader`, resolving it through Drive with `name = '…'` and patching `_get_credentials` the way `test_gmail` does, since this reader has the same disk-based OAuth flow and no injection hook. `examples/using-llamaindex-readers/gsheets.py` does the same against a throwaway server or `--url`, and the README's redirect table has the Sheets row. What the example prints is the reader's own: its `load_data` sends `R1C1:R{rowCount}C{columnCount}` with no sheet title in the range, so every block after the first repeats the first sheet's grid, against real Sheets as much as here, since an unqualified range is the first sheet's on both; the example keeps two sheets so that shows, and its docstring says why.

**What the tests hold.** `tests/test_google.py` and `tests/test_llamaindex.py`, none touching the network. The 217-row measurement is one test over a corpus with the probe's five sheets. The Drive tests run over a corpus of their own (`_Q_RECORDS`) rather than the SAMPLE fixture, so each answer is a named set — and because a `corpus_client` earlier in the file starts a second lifespan on the shared app, which closes the module client's connection for any DB-touching test placed after it. Each rule below, reverted in place, fails at least one test:

```
`not` is not parsed
the default `trashed = false` is not appended
`or` is parsed as `and`
conjuncts are read from under an `or`
an unknown term is ignored, not refused
a time value is not checked as RFC 3339
`\'` is not unescaped
`name =` compares case-sensitively
the title LIKE is used with trash asked for
`'me' in owners` is compared as a literal address
`name =` folds beyond ASCII letters
`name contains` compares the raw title in SQL
the startup check ignores a missing column
the nesting bound is removed
`{` is not a token
the reversed rule drops its same-kind clause
the reversed rule is removed
an absent R or C is never unbounded
absolute 0 is accepted
R1C1 is tried before A1
a lone A1 column is accepted as a range
mixed notation is resolved per half
surrounding whitespace is stripped
A1 row 0 is accepted
whole columns past the grid are named with rows
```

**Not in this PR.** `name contains` stays a substring match: the reference says "The `contains` operator only performs prefix matching for a `name` term", and mirage's name lookups lean on the substring, so narrowing it is its own change with its own measurement. Unmodelled terms stay refused rather than approximated: `writers`/`readers` could be read off the ACL and `starred` could match nothing, but each is a fidelity claim wanting a measurement. The Sheets reader's title-less range is the reader's to fix (its pandas path qualifies the range with the title); Backlot answers it as real does.

**Verification.** At `625c583`, on `main` at `9c2e96a`, merged into the branch.

```
PYTHONPATH=. python -m pytest -rs                    2969 passed, 1 skipped
```

The one skip is `llama_index.readers.hubspot`, an optional extra this branch does not touch. The table at the top is the issue's reproduction script against `backlot.serve()` on the bundled corpus at `625c583`; the `main` column is the issue's own measurement, re-run at `6490f3b` with the same numbers (the branch has since merged `main` at `9c2e96a`, which touches no Drive route). The same server answers `R1C1:R2C2` and `A1:B2` on the bundled spreadsheet with byte-identical bodies. On a 20,000-record corpus `name = '…'` answers in 8 ms and `name contains '…'` in 9 ms (120 ms before the `=` form took the title `LIKE`); 320 nested parentheses answer 400 (500 before the bound). The seventeen `name` rows in `MEASURED_NAME` are what real Drive answered on 2026-09-14 for four titles created for the measurement. `examples/using-llamaindex-readers/gsheets.py` run against a throwaway server prints the `Revenue` grid under both sheet titles, and `--user mia@acme.com` the same.


khj809 (commented):
Ten inline comments below. The two issues themselves land: I re-ran #173's seven-row script against this branch and got the PR's numbers, and walked the R1C1 grammar from the issue comments term by term (`R1C1:R2`, `C[1]`, the same-kind reversed clause, `ZZ:ZZ`, `!A1`, whitespace, `A` vs `RC` vs `R1C1` as titles) against `values.get`, `spreadsheets.get?ranges=`, `values:batchGet` and `batchGetByDataFilter` — all consistent.

Four of the comments are code (one silent wrong answer, one push-down the PR's own motivating client misses, one 500, one unreachable message), four are prose that repeats itself or records how the PR was made, two are drift.

One more, not inline: the description's opening commit inventory and the whole **What the self-review and the measurement changed** section describe the passes that produced the work rather than its final state, and rewriting from that state would also collapse the #174 grammar bullets, which currently exist three times over — issue comments, PR body, `_R1C1_END`.

backlot/routers/google.py
```diff
@@ -844,10 +844,317 @@ def _gmail_message(row, fmt: str, caller_email: str | None = None) -> dict:
...
+    if field == "fullText":
+        return f["id"] in fulltext.get(value, ())
+    if field == "modifiedTime":
+        return _drive_q_compare(op, f["modified"], value)
+    if field == "createdTime":
+        return _drive_q_compare(op, f["created"], value)
+    if field == "parents":
+        return value in f["parents"]
+    if field == "owners":
+        return value.strip().lower() in f["owners"]
```
khj809: `'me'` is Drive's alias for the authenticated user in an `owners` term, but it is compared here as a literal address, so `'me' in owners` answers `[]` under a 200 — measured on this branch, user `ava.chen@acme.com` gets 0 files where `sharedWithMe = false` gets 4. Suggest resolving it against `me`, the identity the `sharedWithMe` branch twenty lines up already has in hand; refusing it would do too, but answering nothing silently is the class #173 exists to close.

backlot/routers/google.py
```diff
@@ -1196,84 +1476,94 @@ def _drive_sort(files: list[dict], specs: list[tuple]) -> list[dict]:
...
+    name = next((t for t in conjuncts if t.field == "name" and t.op == "contains"), None)
+    # `list_drive_by_name` answers non-trashed rows only, so it can be the candidate set only when
+    # every match is non-trashed — `trashed = false` a conjunct, not merely present: under a `not`
+    # it asks for the trash.
+    non_trashed = _QTerm("trashed", "=", "false") in conjuncts or (
+        _QTerm("trashed", "!=", "true") in conjuncts
+    )
+    if fulltext is not None:  # the index's candidates, in rank order, then the other terms
+        candidates = hits[fulltext.value]
+    elif name is not None and non_trashed:
```
khj809, line 1576: The title-LIKE candidate set is gated on `op == "contains"`, so `name = '…'` — the shape this PR names as mirage's own, and #173's row 1 — falls through to the 100k-row `list_documents` scan: measured 110 ms against 3 ms for `contains` on a 20,000-record corpus. Suggest taking the `name` conjunct for `=` as well, since `list_drive_by_name`'s `%needle%` is a superset of exact equality and `_drive_q_eval` below already narrows whatever it returns.

backlot/routers/google.py
```diff
@@ -844,10 +844,317 @@ def _gmail_message(row, fmt: str, caller_email: str | None = None) -> dict:
...
+                raise _drive_q_refused()
+            break
+        pos = m.end()
+        kind = m.lastgroup or ""
+        text = m.group(kind)
+        out.append((kind, re.sub(r"\\(.)", r"\1", text) if kind == "str" else text))
+    return out
+
+
+def _drive_q_parse(q: str):
```
khj809, line 934: The recursive descent has no depth bound, so a nested query exhausts Python's recursion limit and the request 500s where the section's stated contract is a 400 — measured, `q = "(" * 400 + "name contains 'a'" + ")" * 400` (817 characters) returns `500 Internal Server Error`, and `"not " * 600` likewise. Suggest a depth counter in `unary()` that raises `_drive_q_refused()` past a fixed nesting limit.

backlot/routers/google.py
```diff
@@ -844,10 +844,317 @@ def _gmail_message(row, fmt: str, caller_email: str | None = None) -> dict:
...
+            node = disjunction()
+            if take() != ("paren", ")"):
+                raise _drive_q_refused()
+            return node
+        return term()
+
+    def unmodelled(field: str) -> gerr.GoogleError:
+        evaluated = ", ".join(sorted(_DRIVE_Q_OPERATORS))
+        return _drive_q_refused(
+            f"'{field}' is not evaluated by Backlot: a corpus record carries nothing to answer it "
```
khj809, line 996: `properties` and `appProperties` can only be written the way the reference spells them (`properties has { key='x' and value='y' }`), which fails in `_drive_q_tokens` on the `{` before the field is ever read — so both always get the bare `Invalid Value` and never this message. Suggest tokenizing `{`/`}` so the field is seen first, or dropping the two entries from `_DRIVE_Q_UNMODELLED` since nothing can reach them.

backlot/routers/google.py
```diff
@@ -2198,8 +2496,132 @@ def _sheets_format(cell) -> dict:
...
+#   token is read as A1 first (`R1` is the cell R1, `RC1` the cell RC1, `RC:RC` the columns RC),
+#   then as R1C1 (`RC` alone is A1, and so is `R1C1` even in a workbook with a sheet named `R1C1`),
+#   then — bang-less — as a sheet name (`A` is the sheet `A` when there is one).
+# * a reversed R1C1 range is swapped the way an A1 one is (`R3C3:R1C1` is A1:C3), except that on
+#   an axis where both halves carry a number OF THE SAME KIND — both absolute, or both offsets —
+#   start == end + 1 on the numbers AS WRITTEN is unparseable: `R2C2:R1C1`, `R1C2:R1C1`,
+#   `R[1]C[1]:R[0]C[0]` and `R2:R1C1` 400 while `R3C3:R1C1` and `R[2]C[2]:R[0]C[0]` answer. An
+#   axis that mixes the kinds swaps freely: `R[1]C[1]:R1C1`, `R[2]C[2]:R1C1` and `R1C1:R[0]C[0]`
+#   all answer. That reads as a half-open interval on the raw numbers coming out empty, checked
+#   before offsets are resolved; the rule was fitted on 22 reversed ranges, the 17 predictions made
```
khj809: These lines record how the rule was arrived at — the fitting, the 17 predictions, what the first version got wrong — rather than the rule the code answers to; same for the `q` header at 848-854 (four lines on what the deleted matcher did), `MEASURED_R1C1`'s "the edges asked for in review", and `test_drive_q_evaluates_the_operators_the_reference_lists`'s opening paragraph. Suggest keeping the measured rule and the date, and letting the PR body carry the derivation, which it already does.

backlot/routers/google.py
```diff
@@ -2198,8 +2496,132 @@ def _sheets_format(cell) -> dict:
 _A1_RENDER = ("FORMATTED_VALUE", "UNFORMATTED_VALUE", "FORMULA")
 _A1_DATETIME = ("SERIAL_NUMBER", "FORMATTED_STRING")
 _SHEETS_ENUM = "type.googleapis.com/google.apps.sheets.v4"
-# One endpoint of an A1 range: a full cell (`B2`), a bare column (`B`) or a bare row (`2`).
+# One endpoint of an A1 range: a full cell (`B2`), a bare column (`B`) or a bare row (`2`). A bare
+# column or row is an endpoint only INSIDE a range — measured 2026-09-12, `Sheet1!B`, `Sheet1!2` and
+# a bang-less `Z` are all "Unable to parse range" while `A:A` and `1:1` answer — and row 0 is not
+# one at all (`A0`, `A1:A0` are unparseable too).
 _A1_END = re.compile(r"(?:(?P<col>[A-Za-z]{1,3})(?P<row>\d+)?|(?P<rowonly>\d+))\Z")
+# One endpoint in R1C1 notation, which the discovery document names beside A1 for `values.get`'s
```
khj809, line 2533: This 35-line block states rules that are then restated, with overlapping example lists, in `_a1_sheet` (2709-2718), `_a1_title` (2806-2808) and `_a1_axis_name` (2823-2830): whitespace twice, the empty title twice, "a lone column is no range" four times (2499, 2519, 2703, 2807), the whole-column refusal naming twice with the same `ZZ:ZZ` / `1001:1002` / `C[26]` examples. Suggest stating each measured rule once here and pointing at it from the code that applies it, the way `_a1_range`'s "see `_R1C1_END`" at 2766 already does.

backlot/integrations/llamaindex.py
```diff
@@ -207,6 +208,33 @@ def point_drive_at(base_url: str) -> None:
...
+    Sheets reads accept R1C1 notation for.
+    """
+    from googleapiclient import discovery
+
+    if not hasattr(discovery, "build"):
+        raise RuntimeError(
+            "point_sheets_at: googleapiclient.discovery.build is gone — update the shim"
+        )
+    _ensure_google_build_wrapped()
+    _BACKLOT_SERVICE_ENDPOINTS["sheets"] = f"{base_url.rstrip('/')}/sheets"
```
khj809, line 237: `point_sheets_at` is in `__all__` and `GoogleSheetsReader` works end to end, but `examples/using-llamaindex-readers/README.md`'s "How it's pointed at Backlot" table has a row for every other redirected reader and none for Sheets, and there is no `gsheets.py` beside `gdrive.py` — which is #174's last open item ("the `gdrive.py` example or a sibling should read a spreadsheet through `GoogleSheetsReader`"). Suggest adding the row and the script, or saying in the description that they are deliberately left out.

backlot/integrations/llamaindex.py
```diff
@@ -7,8 +7,8 @@
 
   - Slack: `slack_reader_at` — the reader calls `api_test()` DURING construction, so the client
     has to arrive already pointed at Backlot.
-  - Gmail/Drive: `point_gmail_at` / `point_drive_at` — wrap `build` to inject
-    `client_options(api_endpoint=...)`.
+  - Gmail/Drive/Sheets: `point_gmail_at` / `point_drive_at` / `point_sheets_at` — wrap `build`
```
khj809, line 10: This line grew the third helper, but `point_gmail_at` (166) and `point_drive_at` (196) still say the wrap point is SHARED with each other and that "both can be active at once, in either order" — a reader of either concludes `point_sheets_at` uses a different seam. Suggest making both read three.

backlot/routers/google.py
```diff
@@ -901,52 +1210,23 @@ def _drive_obj_facts(obj: dict) -> dict:
...
-        return False
-    for who in re.findall(r"'([^']+)'\s+in\s+owners", q):
-        if who.strip().lower() not in f["owners"]:
-            return False
-    return True
-
-
-def _drive_q_match(row, q: str, me: str | None = None) -> bool:
-    return _drive_q_match_facts(_drive_facts(row), q, me)
+def _drive_q_match(row, query, me: str | None = None, fulltext: dict | None = None) -> bool:
```
khj809: Nothing under `backlot/` calls this any more — `drive_files_list` and `_drive_q_rows` both go straight to `_drive_q_eval` — so the string-or-tree `isinstance` dispatch exists only for two assertions at `tests/test_google.py:2554-2555`. Suggest dropping the helper and having those two call `_drive_q_parse` + `_drive_q_eval` directly.

tests/test_google.py
```diff
@@ -1794,11 +1806,31 @@ def test_sheets_values_get_omits_values_when_the_range_is_empty(base, admin_h, s
...
+        "A1:R2C2",  # both halves of a range are one notation
+        "R2C2:R1C1",  # reversed by one, on the numbers as written
+        "A0",
+        "Z",  # a lone column is no range, so a bang-less one is a sheet name — and there is none
+        "Sheet1!B",
+        " A1",  # not stripped
+    ],
+)
 def test_sheets_values_get_rejects_an_unusable_range(base, admin_h, sheet_id, rng):
     """Backlot has exactly one sheet, `Sheet1`; naming another is as unresolvable as a malformed
```
khj809: The docstring still explains the test as "naming another sheet", but the list above grew from 5 cases to 14 and only `Other!A1:B2` is that one — the rest are `R0C1`, `R[-1]C[0]`, `A1:R2C2`, `R2C2:R1C1`, `A0`, `Z`, `Sheet1!B`, ` A1`, ` !A1`, nine distinct measured refusal rules. Suggest rewriting it around what the cases now hold, so someone adding a row knows what belongs here.

backlot/routers/google.py
nuri-yoo: Fixed in 7fb2fe0: `me` resolves to the caller through the identity `sharedWithMe` reads, so the admin token, which is not a Drive user, owns nothing under it too. Reproduced first: as ava.chen, `'me' in owners` answered 0 and `not 'me' in owners` 13 of 13 visible, where `sharedWithMe = false` says 4 are hers. `test_drive_q_me_is_the_caller_and_a_non_ascii_name_still_matches` holds both directions for a user, and the client-shapes test the admin's; the fix reverted fails both.

backlot/routers/google.py
nuri-yoo: Taken in 7fb2fe0, with one narrowing. Measured on 20,000 records: `name = '…'` 120 ms against 6 ms for `contains`, and the SQL itself is 14 ms for the scan against 1 ms for the LIKE, so the cost is the Python evaluation over the whole set. `=` now takes the LIKE path. The superset does not hold for a non-ASCII needle, though: SQLite's LIKE folds case for ASCII only, and so does its `lower()`, where `_drive_q_eval` casefolds. Against a title `Élan Vital`, `title LIKE '%élan%'` and `lower(title) LIKE lower('%élan%')` both return 0 rows, so `name = 'élan vital'` would have left the candidate set and never reached the evaluator. The gate is `t.op in ("contains", "=") and t.value.isascii()`; a non-ASCII `=` takes the scan, and the new test sends `name = 'élan vital'` and `'ÉLAN VITAL'`, which fail with the guard removed. `contains` has carried that drift all along (`name contains 'élan'` answers 0 for that title today); it is the same code path, but a separate change with its own measurement of what real Drive does with case there, so I have left it out of this PR.

backlot/routers/google.py
nuri-yoo: Fixed in 7fb2fe0: `unary()` counts depth and refuses past `_DRIVE_Q_MAX_DEPTH = 32` with the same `Invalid Value`. Reproduced first, and bisected: 320 parentheses is the first depth that 500s here (each level costs three frames), and `not` 500s from 960 on, so `"not " * 600` answered 200 on this machine (Python 3.14, limit 1000); the fix is the same either way. Both shapes join `test_drive_q_refuses_a_clause_it_cannot_parse` at 64 deep, and an 8-deep `(((name = 'Q1 Deck')))` in the client-shapes test stays a 200.

backlot/routers/google.py
nuri-yoo: Fixed in 7fb2fe0, the first way: `{` and `}` are tokens, so the field is read before the brace is refused and `properties has { key='x' and value='y' }` gets the message that names `properties`. Both spellings are in `test_drive_q_refuses_a_documented_term_it_holds_no_fact_for`, failing with the brace token removed. One precision: the message was reachable, by a spelling the reference does not use (`properties has 'x'` and `properties = 'x'` both reached it), so the entries stayed; what could not reach it was the reference's own form.

backlot/routers/google.py
nuri-yoo: Done in f79ae81, prose only. The q header now describes the parser and the two mirage shapes, not the matcher this branch removed; `_drive_q_parse`'s docstring says what the plain listing does rather than what the matcher used to; `_R1C1_END`'s reversed-range bullet ends at the rule and points at `MEASURED_R1C1` for the rows; `MEASURED_R1C1`'s header says what the rows cover and drops which review round asked for which; the operators test opens on what it holds. The PR body carries the derivation, or did — see the description, which is rewritten from the final state.

backlot/routers/google.py
nuri-yoo: Done in f79ae81. `_a1_sheet`, `_a1_classify`, `_a1_title` and `_a1_axis_name` each point at `_R1C1_END`'s comment (or `_A1_END`'s, for the lone-column rule) instead of restating the rule with their own example lists, the way `_a1_range` already did; the letters-only examples `_a1_axis_name` carried (`ZZ:ZZ` names `ZZ`, `AA:AB` names `AA:AB`) moved into the naming bullet there, so nothing measured was dropped. The lone-column rule was at 2499, 2614, 2704 and 2807; it is stated at `_A1_END` and pointed at from `_a1_classify`, the other two sentences are gone.

backlot/integrations/llamaindex.py
nuri-yoo: Added in e689940: `examples/using-llamaindex-readers/gsheets.py`, a Sheets row in the README table and a paragraph beside Drive's. It resolves the workbook through Drive by exact name, builds the authorized-user credential the way gmail.py does, overrides `_get_credentials` on the instance, and loads it; run against a throwaway server it prints the grid, and with `--user mia@acme.com` the same. One thing it shows that is the reader's, not Backlot's: the installed reader's `load_data` sends `R1C1:R{rowCount}C{columnCount}` with no sheet title, so every block after the first repeats the first sheet's grid, against real Sheets as much as here (an unqualified range is the first sheet's on both, measured in #174). The example keeps two sheets so that is visible rather than hidden, and its docstring says why. The pandas path in the same reader qualifies the range with the title, so the fix upstream is one line; I found no report of it in run-llama/llama_index.

backlot/integrations/llamaindex.py
nuri-yoo: Done in f79ae81: both say the wrap point is shared with the other two and that all three can be active at once. The module docstring and the README also counted the readers that hardcode a host, as four and five; neither count was right before the Sheets shim (six each), so both now say "the others".

backlot/routers/google.py
nuri-yoo: Done in 7fb2fe0: `_drive_q_match` is gone, the two assertions call `_drive_q_parse` and `_drive_q_eval` on `_drive_facts(d2)`, and the one docstring that named it (`test_drive_folders_match_the_same_q_clauses_as_files`) names `_drive_q_eval`.

tests/test_google.py
nuri-yoo: Done in f79ae81: the docstring now says every case is one representative of a rule `_R1C1_END`'s comment states, lists the rules the fourteen stand for, and says a new case belongs here when it stands for a rule and in `MEASURED_R1C1` when it is a request that was sent.

backlot/integrations/llamaindex.py
nuri-yoo: The reader-side half is filed upstream: run-llama/llama_index#23045 (`_load_sheet` builds the range without the sheet title; the pandas path in the same file qualifies it, and neither quotes the title). Until it lands, what `gsheets.py` prints under the second sheet title is the first sheet's grid, on real Sheets and here alike.

nuri-yoo: And the fix, from the fork: run-llama/llama_index#23046 (one helper spells the range as `'<title>'!R1C1:R{rows}C{cols}` for both loaders, three tests, 0.8.0 to 0.8.1).

backlot/routers/google.py
```diff
@@ -1196,84 +1487,105 @@ def _drive_sort(files: list[dict], specs: list[tuple]) -> list[dict]:
...
+    fulltext = next((t for t in conjuncts if t.field == "fullText"), None)
+    # `contains` or `=`: `list_drive_by_name`'s `%needle%` LIKE is a superset of equality, and the
+    # evaluator narrows what it returns. ASCII needles only — SQLite's LIKE (and its `lower()`)
+    # fold case for ASCII alone, where `_drive_q_eval` casefolds, so `name = 'élan vital'` against
+    # a title `Élan Vital` would leave the candidate set and never reach the evaluator.
+    name = next(
+        (
+            t
+            for t in conjuncts
+            if t.field == "name" and t.op in ("contains", "=") and t.value.isascii()
```
khj809: The guard tests the needle, but the widening can come from the title: `str.casefold()` folds `ß` to `ss` where SQLite's LIKE does not, so an ASCII needle can match in the evaluator and miss in SQL. Measured across the two commits, corpus of one title `Straße Plan` — `name = 'strasse plan'` answers `['Straße Plan']` at `b122aa0` and `[]` at `e689940`, a silent empty that is the same class the guard was added for. Suggest making the invariant explicit (the candidate set has to be a superset of what `_drive_q_eval` accepts) and either matching the evaluator to SQL for the LIKE path or storing a casefolded title to key it on; `contains` has had this since before the PR, `=` acquired it here.

backlot/routers/google.py
```diff
@@ -1196,84 +1487,105 @@ def _drive_sort(files: list[dict], specs: list[tuple]) -> list[dict]:
...
+
+def _drive_q_rows(conn, query, container: str | None, ids, me: str | None, hits: dict) -> list:
+    """Rows matching a non-trivial query: build the smallest candidate set SQL can produce from the
+    terms every match has to satisfy, then evaluate the whole query in Python.
+
+    ``hits`` is ``_drive_q_fulltext``'s answer for this query, so the index is asked once."""
+    conjuncts = _drive_q_conjuncts(query)
+    fulltext = next((t for t in conjuncts if t.field == "fullText"), None)
+    # `contains` or `=`: `list_drive_by_name`'s `%needle%` LIKE is a superset of equality, and the
+    # evaluator narrows what it returns. ASCII needles only — SQLite's LIKE (and its `lower()`)
```
khj809: The `contains` half of the gate is a behaviour change, not a no-op: `name contains 'élan'` against `Élan Vital` answers `[]` at `b122aa0` and `['Élan Vital']` at `e689940`, because a non-ASCII needle now skips the LIKE and reaches the casefolding evaluator — so the drift the reply says is left for a separate change is in fact fixed here, untested and unmentioned. It looks like the right answer (real Drive's `name contains` is case-insensitive), so suggest keeping it and adding the case beside the two `=` ones in `test_drive_q_me_is_the_caller_and_a_non_ascii_name_still_matches`, rather than leaving it as a silent rider.

khj809 (commented):
Verified all ten at `e689940`, each by reproducing rather than reading: `'me' in owners` now answers 4 for ava.chen where `sharedWithMe = false` says 4 and `not 'me' in owners` 9 (admin still 0); 32 parentheses and 2000 `not`s are 400s and 8 deep still answers; `properties has { key='x' and value='y' }` names `properties`; `name = '…'` is 3.1 ms against 110 ms before on 20,000 records; `_drive_q_match` is gone from the tree; the Sheets helpers point at `_R1C1_END` instead of restating it; `gsheets.py` runs and prints the `Revenue` grid under both sheet titles. Suite 2946 passed / 1 skipped, ruff clean. The description reads as finished work now.

One correction to my own comment on the map-valued terms: you are right that the message was reachable — `properties has 'x'` and `properties = 'x'` both got it at `b122aa0`. "Always the bare `Invalid Value`" was wrong; what could not reach it was the reference's own spelling, which is what the fix addresses.

Two things on the `name` push-down, both new at `e689940` — the guard is sound in the direction you measured but not in the other one.

backlot/routers/google.py
nuri-yoo: Reproduced (`Straße Plan` answers `[]` to `name = 'strasse plan'` at e689940, and `ﬁnance deck` to `name = 'finance deck'` the same way), and rather than choose which side to bend, I measured what real Drive does — four titles created in a real Drive on 2026-09-14 and deleted after:

| title | `q` | real Drive |
|---|---|---|
| `Straße Plan` | `name = 'STRAßE PLAN'` | match |
| `Straße Plan` | `name = 'strasse plan'`, `contains 'strasse'` | no match |
| `Élan Vital` | `name = 'ÉLAN VITAL'` | match |
| `Élan Vital` | `name = 'élan vital'`, `= 'elan vital'` | no match |
| `Élan Vital` | `contains 'élan'`, `contains 'ÉLAN'` | match |
| `Élan Vital` | `contains 'elan'` | no match |
| `ﬁnance deck` | `contains 'finance'` | match |
| `ﬁnance deck` | `name = 'finance deck'` | no match |
| `İstanbul` | `contains 'istanbul'` | match |

So the evaluator was the wrong side for `=`: real folds case for ASCII letters and nothing else, which is exactly SQLite's `LIKE` with no wildcards. In 4893911 `_drive_q_eval` compares `=` and `!=` with an ASCII-only lower-case and `list_drive_by_name(exact=True)` is `title LIKE ?` — the same comparison, so `name = 'strasse plan'` answers `[]` as real does, and the invariant you name holds by construction rather than by a guard: for the name term the candidate set *is* the evaluator's answer, and the `isascii()` gate is gone. `contains` took your second option; see the other thread. `test_drive_q_name_compares_as_real_drive_does` sends all seventeen rows through the handler and, for each `=`, its `!=` complement; reverting the equality fold to `casefold` fails four of them.

backlot/routers/google.py
nuri-yoo: You are right, and the measurement above settles what the rule should be rather than whether to keep the accident. Real's `name contains` folds case across the alphabet and compatibility forms (`'finance'` finds `ﬁnance deck`, `'istanbul'` finds `İstanbul`) but neither `ß` (`'strasse'` does not find `Straße Plan`) nor accents (`'elan'` does not find `Élan`). `str.casefold` was not that rule either — it folds `ß` to `ss` and leaves the dotted I as `i` plus a combining dot — so at e689940 `name contains 'strasse'` would have matched through the evaluator where real does not. In 4893911 `store.drive_name_fold` is NFKC then `lower()` with that dot removed, which reproduces every measured row; the importer stores it as `gdrive_files.title_fold` and `list_drive_by_name` LIKEs the folded needle against it, so the candidate set and the evaluator compare the same text and no needle is routed around the SQL any more. The eight `contains` rows are in `MEASURED_NAME`; with the SQL back on the raw `title`, five of them fail. The reply that called the `contains` drift a separate change was wrong about what e689940 did — it had already changed it, as you say — and this makes the change deliberate and measured. Real's own limit of `contains` to prefix matching is still not modelled; the description says so.

backlot/store.py
```diff
@@ -362,7 +363,11 @@ def id_seed_scope(source_type: str) -> str | None:
     id TEXT PRIMARY KEY, folder TEXT NOT NULL, author_email TEXT NOT NULL,
     title TEXT NOT NULL, content TEXT NOT NULL,
     subtype TEXT, mime_type TEXT, parents TEXT, created_ts INTEGER NOT NULL, updated_ts INTEGER,
-    trashed INTEGER, owner_display TEXT
+    trashed INTEGER, owner_display TEXT,
+    -- `drive_name_fold(title)`, so a `name contains` lookup can be a LIKE over exactly the text
+    -- the evaluator compares; NULL on a row written before the column existed, which the lookup
```
khj809: This fallback cannot run: `CREATE TABLE IF NOT EXISTS` never adds a column to a DB that already has `gdrive_files`, so the column is absent rather than NULL and `COALESCE(title_fold, title)` raises `no such column: title_fold`. Measured — a DB built at `e689940`, served at `4893911`: `name contains 'Brand'` is a bare 500 per request while `name = '…'` and the plain listing answer 200, and `store.missing_tables` waves it through because the table is there. Suggest dropping the COALESCE and this sentence and extending the startup check to columns, which is what `missing_tables`' own docstring prescribes for exactly this ("a corpus is re-imported, not upgraded in place — so naming the gap IS the remedy"); this is the first column to need it.

khj809 (commented):
Both threads settled, and better than what I suggested — measuring real instead of picking a side turned my "regression" into the correct answer: `name = 'strasse plan'` answers `[]` because real answers `[]`, and the `contains` rider became a measured rule. Verified all seventeen `MEASURED_NAME` rows through a running server on a 20,000-record corpus, `!=` complements included, and SQL and evaluator now agree by construction — `name = 'élan vital'` is 2.3 ms where the `isascii()` gate sent it to a 116 ms scan at `e689940`, so removing the gate bought speed as well as fidelity. 2963 passed / 1 skipped, ruff clean.

One new thing, from the column this brought with it — inline below.

backlot/store.py
nuri-yoo: Reproduced — a DB built at e689940, served at 4893911: the plain listing and `name = 'Brand guidelines'` 200, `name contains 'Brand'` raises `no such column: title_fold`, and `missing_tables` answers `[]` for it. The fallback and its sentence are gone in 4e3cb07, and the check reads columns: `store.missing_schema` replaces `missing_tables`, `schema_columns` parses every table's columns off `SCHEMA`'s own text (parentheses counted, `--` comments stripped — the first regex swallowed a one-line table's neighbours), and a column the DB lacks is named `table.column` in the same startup refusal a missing table gets. That e689940 DB now fails startup with `has no gdrive_files.title_fold`. Two tests: `test_missing_schema_names_what_an_older_db_lacks` drops the column and then the table and expects both named; `test_schema_columns_reads_every_table_off_the_ddl` holds the parsed columns to what `connect_rw` creates, table by table (48 tables). The server test drops the column too and expects the refusal. With the column loop removed, both the store test and the server test fail. One consequence in the DDL: `title_fold` moved onto the `trashed, owner_display` line, because SQLite's `DROP COLUMN` rewrites the CREATE text and a comment block between the last comma and the column left it with `incomplete input` — which the test needed to build an older DB.

tests/test_server.py
```diff
@@ -402,10 +402,23 @@ def test_serving_a_db_older_than_a_table_says_so_and_says_to_re_import(tmp_path)
...
+    # on such a DB while the plain listing and `name =` answered 200.
+    conn = sqlite3.connect(settings.db_path)
+    conn.executescript(store.SCHEMA)  # the table comes back; the column is then dropped
+    conn.execute("ALTER TABLE gdrive_files DROP COLUMN title_fold")
+    conn.commit()
+    conn.close()
+    with pytest.raises(RuntimeError) as e:
+        with client_for(settings, reload=True):
+            pass
+    assert "gdrive_files.title_fold" in str(e.value)
```
khj809, line 425: Folding this into the existing test rather than adding a sibling is right, but the function is still named `..._older_than_a_table_...` while it now asserts both a table and a column. Suggest `test_serving_a_db_older_than_the_schema_says_so_and_says_to_re_import`, or whatever keeps the two halves in the name.

khj809 (commented):
One naming nit inline, non-blocking; the verification and sign-off are in the approval that follows.

khj809 (approved):

tests/test_server.py
nuri-yoo: Renamed in 625c583 to `test_serving_a_db_older_than_the_schema_says_so_and_says_to_re_import`, and the docstring now says which half is which — a table this build reads that the DB does not have, then a column one of its tables does not have. Nothing else referenced the old name (repo-wide grep: the definition only). Suite 2969 passed / 1 skipped, ruff and gen_docs clean.

PR #202, opened 2026-09-12
google: $.xgafv selects the error envelope on every route, and a value it does not take is refused before the route runs

### LINKED ISSUE #172: google: $.xgafv selects the error envelope, and every Google route ignores it
`$.xgafv` selects which error envelope a Google API answers with. Backlot declares it on no route and ignores it everywhere, so a client that asks for the v1 envelope silently gets the v2 one — and every Google endpoint's fidelity baseline carries a `missing_param` for it. On `google_drive` alone that is 5 acknowledged gaps on the Sheets read routes, and the same parameter is declared on the Drive, Docs and Slides methods beside them.

This is a **system parameter**, declared on every method of every Google discovery document. It is filed against the family rather than one source because closing it for a few routes would leave the rest inconsistent with them.

## Measured

Against the live Sheets v4, Docs v1 and Drive v3 APIs.

**What the value selects.** `2`, and an absent parameter, give the envelope Backlot already serves. `1` adds the legacy `errors[]` array; nothing else about the body changes, and a **success** response is byte-identical either way.

```
GET .../values/Data!B2?majorDimension=NOPE&$.xgafv=1
{"error": {"code": 400,
           "message": "Invalid value at 'major_dimension' (…sheets.v4.Dimension), \"NOPE\"",
           "errors": [{"message": "…same text…", "reason": "invalid"}],
           "status": "INVALID_ARGUMENT",
           "details": [{"@type": "…google.rpc.BadRequest", "fieldViolations": […]}]}}
```

**Which families it affects.** Exactly the ones Backlot already calls `EDITOR`:

| family | `errors[]` at `$.xgafv=1` | at `$.xgafv=2` |
|---|---|---|
| Sheets v4, Docs v1 | present | absent |
| Drive v3 | present | present |

So the change is not a new concept — it is making `errors.google._FAMILY_HAS_ERRORS[EDITOR]` a per-request answer instead of a constant `False`. Drive and Gmail are unaffected, which their `True` already says.

**An unknown value is refused**, with a message that names the parameter as a system one rather than a field:

```
$.xgafv=0 | 3 | NOPE  ->  400
{"error": {"code": 400,
           "message": "Invalid query parameters. Invalid value '3' for system query parameter : $.xgafv",
           "status": "INVALID_ARGUMENT"}}
```

**The `errors[]` entry is not uniform.** Measured across error kinds, which is the part an implementation has to get right rather than assume:

| error | entry keys |
|---|---|
| Sheets, bad enum value (400) | `message`, `reason` — **no `domain`** |
| Sheets, unparseable range (400) | `domain`, `message`, `reason` |
| Sheets, bad `fields` mask (400) | `domain`, `message`, `reason` |
| Sheets / Docs, not found (404) | `domain`, `message`, `reason` |
| Sheets, invalid token (401) | `domain`, `location`, `locationType`, `message`, `reason` |
| Sheets, no credentials (403) | `domain`, `message`, `reason` |

Backlot's `http_body` always writes `domain: "global"`, so the enum case is the one it would get wrong. Whether "a bad enum omits `domain`" generalises beyond that one error is not established — one more measurement pass over the remaining constructors would settle it before anything is written.

## Work

- `$.xgafv` accepted on the Google routes and validated: `1` or `2`, anything else the 400 above.
- `errors.google.http_body` renders the editor families with `errors[]` when the request asked for v1.
- `domain` emitted per the table rather than unconditionally.
- Regenerate the baselines: `backlot diff --source google_drive --update-baseline`, and the same for whichever other sources carry the parameter.

## Not in scope

The other four standard parameters left open beside this one, for reasons of their own: `key`, `access_token` and `oauth_token` are credential transports and belong to `backlot.auth`, not to a router; `uploadType` the real API *refuses* on a read, so the baseline's "the vendor accepts it" does not describe the vendor and the acknowledgement itself wants rewording.

nuri-yoo:
Closes #172. The Sheets measurements here were taken with the probe workbook #201 built, and the tests sit beside the ones it added; that branch has landed and this is on `main`.

**What the parameter does, measured.** Against the live Sheets v4, Docs v1 and Drive v3 APIs on 2026-09-12, and against Slides v1 and Gmail v1 on 2026-09-14 through the errors a request with no Authorization header reaches — that request carries no credential to reject, so Slides answers it rather than the quota-project 403 a user token gets, and Gmail answers it without a scope. What the parameter selects is the legacy `errors[]` array, and the three families answer it three ways. Docs, Sheets and Slides opt in: `1` adds the array, `2` and an absent parameter leave it off. **Gmail opts out**: the array is there unless the value is `2`. Drive carries it whatever the value says, measured on a `fields` refusal at `2` as well as anonymously. A 200 body is byte-identical under all of them. When the parameter is sent twice the last value wins, and it is that value each family's rule reads — measured on Sheets, `$.xgafv=2&$.xgafv=1` carries the array and `1&2` does not; measured on Gmail, `2&0` carries it, because a refused value is not a `2`, and `0&2` does not. That is also what Starlette's `QueryParams.get` answers, so nothing has to be written for it. Any other value (`0`, `3`, `NOPE`, `01`, the empty string) is refused with

```
400 {"error": {"code": 400, "message": "Invalid query parameters. Invalid value '3' for system query parameter : $.xgafv", "status": "INVALID_ARGUMENT"}}
```

and on Drive and Gmail the same with an `errors[]` entry of `badRequest`. The refusal comes first: measured ahead of a bad token, a missing credential and an unparseable range alike.

**The entry is not uniform, and the constructor decides it.** The issue measured six error kinds and left the rest open; this measured the remaining editor constructors Backlot raises. At `$.xgafv=1`:

| error | `errors[0]` |
|---|---|
| a typed value the proto layer refuses — enum (`majorDimension=NOPE`), bool (`includeGridData=maybe`), int32 (`startRowIndex: "x"`) | `message`, `reason: invalid` — no `domain` |
| an unparseable range, a range past the grid, a bad `fields` mask, `alt=media`, `alt=nope`, `dataFilter.filter must be specified.`, `Must specify at least one dataFilter.`, `No sheet with id`, a non-JSON body | `domain: global`, `reason: badRequest` |
| not found (Sheets, Docs) | `domain: global`, `reason: notFound` |
| bad token (Sheets, Docs, Drive) | `Invalid Credentials`, `domain: global`, `authError`, `location: Authorization`, `locationType: header` |
| no credentials, Sheets | `domain: global`, `reason: forbidden` |
| no credentials, Docs | `Login Required.`, `domain: global`, `reason: required`, `location: Authorization`, `locationType: header` |

So `errors.google` gains a `domain` attribute on `GoogleError` (default `global`, `None` for the typed refusal), `invalid_argument` now carries `badRequest` — the reason it never showed before, since no editor error rendered an entry — and the typed-value sites (`_a1_enum_error`, `_sheets_bool_value`, `_sheets_int32`) raise a new `invalid_field_value`, which is `invalid` with no `domain`. `bad_field_mask` moves to `badRequest` for the same reason. `missing_credentials` carries the short `Login Required.` at `location: Authorization` for all three OAuth-only APIs, which answer an anonymous request identically. The two editor 400s this PR did not measure keep whatever their constructor already renders: `Invalid gridRange` is `invalid_argument`, so `badRequest`, and an Office file read as a native document is `failed_precondition`, so `failedPrecondition`.

**Where it is read and validated, once.** `$.xgafv` is a system parameter — a top-level entry of a discovery document's `parameters`, which every method takes — so it is handled the same way rather than route by route: `errors.google.validate_system_parameters` is a dependency on the Google router, which is what puts the refusal ahead of everything a route does, and every family operation is sent a value it does not take to prove the dependency reaches it — the declaration is derived from the path, so it alone would not notice a family route served by some other router; `http_body` takes the request's query and renders the array when it says `1`, through the same `errors.http_body(path, exc, query)` every vendor is called with (Atlassian and GitHub accept and ignore the argument); and `openapi.google_system_parameters` declares it on every family operation after FastAPI builds the document — twenty-one operations, with the discovery document's own `V1 error format.` and `enum: ["1", "2"]` — and not on `/batch`, which is no family's. A route added later inherits all three.

**The baselines.** With the parameter declared, `backlot diff` stops reporting `missing_param … ?$.xgafv`; run live against the discovery documents, `--update-baseline` dropped exactly the 13 `google_drive` and 8 `gmail` entries that carried it and nothing else (`0 new` on both), and the two files' `measured` dates moved to 2026-09-12. A test holds that no Google baseline acknowledges the parameter again.

**What the tests hold.** Nine new tests in `tests/test_google.py`, with twelve parametrised cases between two of them, and one existing assertion widened (`about.get` now declares `fields` and `$.xgafv`). Each rule was checked by reverting it and watching the suite go red, counted in tests rather than in parametrised cases:

```
the router dependency is removed                   2: the refusal, and its place before auth
`1` never adds errors[]                            3
Gmail carries the array at `2` as Drive does       1
the family rule reads the first repeat, not last   1
the dependency is scoped to one family             1: the every-operation test
`domain` is always written                         1: the entry table
the typed refusal keeps `badRequest`               1: the entry table
the FIRST repeat wins                              1
the refusal sentence drifts by a space             2
the Docs missing-credential entry loses its short  1
the document declaration is skipped                2: about.get, and the every-operation test
the declaration lands on /batch too                1
the declared enum is dropped                       1
the refusal runs only for a credentialed request   1: its place before auth
```

The 21 baseline removals are held by `test_no_google_baseline_acknowledges_the_system_parameter_any_more`; a mutation there is a `git checkout` of the two JSON files, which that test reports by name.

**Not in this PR.** The rendering of the error body is #203's: real wraps it for `callback` (HTTP 200, `text/javascript`) and indents it whatever `prettyPrint` says, where the handler this PR hands the query to still writes one compact `application/json` line. Measured on all four families and filed rather than fixed here, because the same handler renders Atlassian's and GitHub's errors and neither is measured for it. The other four standard parameters the issue leaves open stay open, for its reasons: `key`, `access_token` and `oauth_token` are credential transports for `backlot.auth`, and `uploadType`'s acknowledgement wants rewording rather than a route. The `details` member all three OAuth-only APIs add to an anonymous request (`google.rpc.ErrorInfo` naming the service and method) is not rendered; it is there under `2` as well and is a separate gap.

**Verification.** At `114c159`, on `main` at `4c4a268`.

```
PYTHONPATH=. python -m pytest -rs                    2988 passed, 1 skipped
backlot diff --source google_drive                   200 divergence(s) · 200 acknowledged · 0 new
backlot diff --source gmail                          157 divergence(s) · 157 acknowledged · 0 new
```

The one skip is `llama_index.readers.hubspot`, an optional extra this branch does not touch. The two `backlot diff` lines are the live runs whose `--update-baseline` rewrote the baselines; before the rewrite each reported the same counts with `13` and `8` "acknowledged but no longer diverging". The measured bodies this PR is built on are the issue's and two more passes over the same probe workbook (`backlot #174 R1C1 probe`), which the constructors' docstrings and the tests' comments cite case by case.


backlot/errors/google.py
```diff
@@ -8,33 +8,56 @@
...
+Measured on Sheets and Docs at `$.xgafv=1`: a typed value the proto layer refuses — the enum, bool
+and int32 "Invalid value at '<field>' (<type>), \"<value>\"" messages — reports ``reason: invalid``
+and NO ``domain``; every other 400 measured — an unparseable range, a range past the grid, a bad
+``fields`` mask, an unsupported or unknown ``alt``, ``dataFilter.filter must be specified.``, ``Must
+specify at least one dataFilter.``, ``No sheet with id``, a non-JSON body — reports ``reason:
+badRequest`` with ``domain: global``, and the editor 400s not measured (an Office file read as a
+native document, ``Invalid gridRange``) take that entry too; a 404 is ``notFound``, a bad token
+``authError`` with ``location: Authorization``, an anonymous Sheets request ``forbidden``, and an
+anonymous Docs or Slides request ``required`` with the short ``Login Required.`` and the same
+location. Drive's and Gmail's entries are the shapes the tests beside them already pin.
```
khj809: **The Office-file 400 does not take the `badRequest` entry — and this section repeats what the six functions below it already say.**

**1. Lines 45-46 are wrong.** "the editor 400s not measured (an Office file read as a native document, ``Invalid gridRange``) take that entry too" holds for `Invalid gridRange`, which is raised through `invalid_argument` and so does render `badRequest`. It does not hold for the Office file: that one is `failed_precondition`, which this PR did not touch. Measured on this branch:

```
GET /sheets/v4/spreadsheets/<Budget Rollup.xlsx id>?$.xgafv=1

"errors": [{"message": "This operation is not supported for this document. The document must
                        not be an Office file.",
            "domain": "global", "reason": "failedPrecondition"}]
```

The same sentence also treats that case as unreachable from outside ("not reached from outside", in the PR body), but it is reachable over HTTP and `tests/test_google.py:1839` already reaches it. Nothing pins its entry at `$.xgafv=1`, so this docstring is the only statement about the shape and it points the wrong way.

**2. Compaction.** The first paragraph restates the `errors[]` column of the table 10 lines above, and then `has_errors_array`, `xgafv`, `validate_system_parameters` and `google_system_parameters` — each of which says the same thing in its own docstring below. The second restates `invalid_argument`, `invalid_field_value`, `bad_token`, `unregistered_caller` and `missing_credentials`. Keeping the measurements and pointing at the constructors leaves one place to update when a constructor changes.

The suggestion does both. Stating the Office entry at `failed_precondition` instead would work too (its docstring is outside this diff) — the point is that one of the two has to say `failedPrecondition`.

```suggestion
`errors[]` is what `$.xgafv` selects, and the middle column above is the whole rule
(:func:`has_errors_array`). It is a SYSTEM parameter — a top-level entry of a discovery document's
``parameters``, which every method takes — so it is validated once for the router
(:func:`validate_system_parameters`) and declared once for the document
(:func:`backlot.openapi.google_system_parameters`) rather than route by route. Measured: Drive
carries the array on a `fields` refusal at `2` as well as with no parameter; the LAST repeat is the
value every rule reads (`2&1` carries it on Sheets where `1&2` does not, `2&0` on Gmail — a refused
value is not a `2` — where `0&2` does not); a success body is the same under all of them; and a
value other than `1` or `2` is refused ahead of a bad token, a missing credential and an
unparseable range alike.

Inside `errors[]` the entry follows the constructor that raised it, and each one carries its own
measurement. Measured on Sheets and Docs at `$.xgafv=1`: a typed value the proto layer refuses is
``reason: invalid`` with NO ``domain`` (:func:`invalid_field_value`); every other measured 400 is
``badRequest`` under ``global`` (:func:`invalid_argument`, :func:`bad_field_mask`); a 404 is
``notFound``; a bad token ``authError`` at ``location: Authorization``; an anonymous Sheets request
``forbidden``; an anonymous Docs or Slides request ``required`` with the short ``Login Required.``.
The two editor 400s NOT measured keep whatever their constructor already renders — ``Invalid
gridRange`` is :func:`invalid_argument`, so ``badRequest``, but an Office file read as a native
document is :func:`failed_precondition`, so ``failedPrecondition``.
```

khj809 (commented):
Reviewed against #172. All four Work items in the issue are closed, the open question it left (whether "a bad enum omits `domain`" generalises) is settled by measuring the remaining constructors, and the design — one router dependency plus one document enrichment rather than 21 route declarations — is the right shape for a system parameter.

Verified locally at `c811f27`: full suite green (1 skip, the hubspot reader extra); `ruff check`, `ruff format --check` and `gen_docs.py --check` clean; `backlot diff --source google_drive` 200 / 200 acknowledged / 0 new and `--source gmail` 157 / 157 / 0 new against the live discovery documents; no `$.xgafv` left in any baseline; 13 + 8 = 21 operations carrying the declaration through to `/_meta/openapi/<source>`; and batch sub-requests inheriting the refusal while the outer `/batch` correctly ignores it.

Six comments: one measured mismatch between a docstring and the code, one question about the Gmail capture, one unacknowledged edge, and three on comment volume.

backlot/errors/google.py
```diff
@@ -210,8 +248,22 @@ def bad_token() -> GoogleError:
...
 
 
-def missing_credentials() -> GoogleError:
-    """No Authorization header, on an OAuth-only API (Gmail, Docs, Slides)."""
+def missing_credentials(path: str = "") -> GoogleError:
+    """No Authorization header, on an OAuth-only API (Gmail, Docs, Slides).
+
+    On the editor family the `errors[]` entry (shown at `$.xgafv=1`) is the short ``Login
+    Required.`` at ``location: Authorization`` — measured on Docs AND Slides, which answer it
+    identically. Gmail's entry is unmeasured, and keeps the long message with no location."""
```
khj809: **"Gmail's entry is unmeasured" contradicts this branch's own Gmail pass — and `short=` on line 265 makes the `GoogleError` docstring stale.**

**1.** The 2026-09-14 pass reached Gmail's anonymous 401 at `$.xgafv=1`: that is how the Gmail opt-out rule was established, and `test_gmail_carries_the_errors_array_unless_xgafv_is_2` asserts `errors[0]["reason"]` off it. You cannot observe whether `errors[]` is present without seeing what is inside it, so that capture carried the whole entry.

This matters beyond the parameter. Gmail renders `errors[]` by default, so if real's Gmail entry looks like Docs' and Slides' — the short `Login Required.` at `location: Authorization` — the mock is wrong on **every** anonymous Gmail request, not only at `$.xgafv=1`:

```
GET /gmail/v1/users/me/labels          (no Authorization header)

"errors": [{"message": "<the long MISSING_CREDENTIALS_MESSAGE>", "domain": "global",
            "reason": "required"}]
```

Could you re-read that capture? If the members were there, pin them the way Docs' and Slides' are pinned; if one genuinely went unread, the docstring should name which member and why, rather than calling the whole entry unmeasured.

**2.** `short="Login Required."` on line 265 is the second user of `short`, but the `GoogleError` docstring at :106 still says "``short`` is a distinct ``errors[0].message`` — only the bad-token 401 uses one". That line is outside this diff, so it needs a manual edit:

```diff
-    canonical code name. ``short`` is a distinct ``errors[0].message`` — only the bad-token 401 uses
-    one, where Google's top-level message is the long form and ``errors[0]`` says "Invalid
-    Credentials"."""
+    canonical code name. ``short`` is a distinct ``errors[0].message``, where Google's top-level
+    message is the long form: the bad-token 401 says "Invalid Credentials" and the editor family's
+    missing-credential 401 says "Login Required."."""
```

**3.** Minor: `path: str = ""` defaults to the non-editor shape, so a future Docs or Slides route calling `missing_credentials()` directly gets Gmail's entry silently. Only `no_credentials` calls it today, so making `path` required costs nothing.

The suggestion is the honest-wording version of (1), for the case where the member really was not read:

```suggestion
def missing_credentials(path: str) -> GoogleError:
    """No Authorization header, on an OAuth-only API (Gmail, Docs, Slides).

    On the editor family the `errors[]` entry (shown at `$.xgafv=1`) is the short ``Login
    Required.`` at ``location: Authorization`` — measured on Docs AND Slides, which answer it
    identically. Gmail's own entry was not read off the 2026-09-14 capture, and keeps the long
    message with no location until it is."""
```

backlot/routers/google.py
```diff
@@ -30,7 +30,8 @@
 from backlot.openapi import qp
 from backlot.pagination import decode_cursor, next_page_token
 
-router = APIRouter(tags=["google"])
+# `$.xgafv` is checked before any route runs — see `gerr.validate_system_parameters`.
+router = APIRouter(tags=["google"], dependencies=[Depends(gerr.validate_system_parameters)])
```
khj809: **A bad `$.xgafv` on a Google-family path with no route is a 404, not the refusal.**

A router dependency only runs once a route has matched, so the check is scoped to matched routes. Measured on this branch:

```
GET /sheets/v4/nope/nope?$.xgafv=3   ->   404 {"error": {"code": 404, "message": "Not Found"}}
```

Real validates system parameters in the gateway, ahead of method resolution, so it is expected to answer the 400 here too — but that is unmeasured, and "the refusal comes first" is the claim the whole design rests on. One probe against live Sheets settles it: if real refuses, the check belongs in middleware rather than in a router dependency; if it 404s, the current placement is exactly right and only wants saying.

Either way the comment should say what it covers:

```suggestion
# `$.xgafv` is checked before any route runs — see `gerr.validate_system_parameters`. Scoped to a
# MATCHED route: a family path with no route 404s, and real is unmeasured there.
router = APIRouter(tags=["google"], dependencies=[Depends(gerr.validate_system_parameters)])
```

backlot/openapi.py
```diff
@@ -181,6 +181,38 @@ def github_page_parameters(spec: dict, parameters: dict[str, tuple[int, str]]) -
...
+
+    for path, item in spec.get("paths", {}).items():
+        if gerr.family(path) is None:
+            continue
+        for method, op in item.items():
+            if method not in _METHODS:
+                continue
+            params = op.setdefault("parameters", [])
+            if any(p.get("name") == gerr.XGAFV and p.get("in") == "query" for p in params):
+                continue
```
khj809: **Two small things in this function.**

**1. The guard on line 202 carries no comment,** and its whole justification lives in the fourth paragraph of `test_every_google_operation_declares_the_system_parameter_and_checks_it` — another file, ~780 lines away. Someone simplifying this loop sees a guard that looks impossible to hit (the loop reads as if it runs once per document), deletes it, and every reader of the cached spec — `/_meta/openapi/<source>`, the MCP bridge — starts seeing `$.xgafv` twice on each of the 21 operations. One line here, and that test docstring can drop the paragraph.

**2. `from backlot.errors import google as gerr` on line 193 is function-local.** `backlot/openapi.py` imports nothing from `backlot` anywhere else — `github_page_parameters`' docstring states that as a rule ("Six routers import this module for `qp`, which is why it does not import that one") and honours it by taking `parameters` as an argument instead. Nothing forces the exception here: `backlot.errors.*` imports only stdlib and fastapi, so a module-scope import creates no cycle (checked). Either hoist it to the top of the file, or say in the docstring why it cannot be.

The suggestion only adds the comment — moving the import needs an edit at line 26 that this diff cannot reach.

```suggestion
    from backlot.errors import google as gerr

    for path, item in spec.get("paths", {}).items():
        if gerr.family(path) is None:
            continue
        for method, op in item.items():
            if method not in _METHODS:
                continue
            params = op.setdefault("parameters", [])
            # FastAPI caches the document and hands the same dict back by identity, so a second
            # call would append a second copy of the parameter.
            if any(p.get("name") == gerr.XGAFV and p.get("in") == "query" for p in params):
                continue
```

tests/test_google.py
```diff
@@ -770,6 +770,270 @@ def test_a_bad_token_is_unauthenticated_everywhere(client, path):
         assert e["errors"][0]["locationType"] == "header"
 
 
+# --- `$.xgafv`, the system parameter that selects the error envelope --------------------------
+#
+# Sheets shapes, measured on 2026-09-12 (#172). Which families show the array under which value is
+# one rule, stated once in `backlot.errors.google`; what this table adds is the entry INSIDE it,
+# which depends on the error rather than the family.
```
khj809: **The new test prose mostly restates the docstrings it is testing.**

This is the same duplication that already produced one stale sentence in this PR (`GoogleError`'s "only the bad-token 401 uses one"). Five spots, grouped here:

- **:775-777** — the same point as the `XGAFV_ENTRIES` header at :872-873 ("the entry is not uniform, and which constructor raised the error decides its shape"), 97 lines and five tests earlier, attached to no table. The suggestion drops it and keeps the divider.

- **:784-786** — near-verbatim `bad_system_parameter`'s docstring ("400 INVALID_ARGUMENT with this sentence, spacing included, and on Drive an `errors[]` entry of `badRequest` under `global`"). What the test knows and the constructor does not:
  ```python
      """Sheets carries no `errors[]` on its own refusal: the value that would have asked for one
      was the value refused."""
  ```

- **:851-856** — restates `has_errors_array`, then `xgafv`'s last-repeat rule, which `test_xgafv_1_puts_the_errors_array_on_an_editor_error_and_2_does_not` already pins on Sheets:
  ```python
      """Gmail is the family that opts OUT, measured on its anonymous 401 — the one Gmail error a
      request with no scope can reach — over `users/me/labels` and `users/me/messages`."""
  ```

- **:945-949** — restates `bad_token`, `unregistered_caller` and `missing_credentials` end to end:
  ```python
      """The three credential failures, which differ from each other rather than by family: a bad
      token, an anonymous Sheets request, an anonymous Docs or Slides one."""
  ```

- **:983-996** — separate comment below.

```suggestion
# --- `$.xgafv`, the system parameter that selects the error envelope --------------------------
```

tests/test_google.py
```diff
@@ -770,6 +770,270 @@ def test_a_bad_token_is_unauthenticated_everywhere(client, path):
...
+    Then the other direction: each operation is SENT a value the parameter does not take, and has
+    to refuse it. The declaration is derived from the path, so it would appear on a family route
+    served by some other router while the dependency that validates it did not run — a gap this
+    half sees and the declaration alone cannot. The path parameters are dummies: the refusal comes
+    before the route, so what an id would have resolved to never matters.
+
+    The document is asked for TWICE. The enrichment edits the schema FastAPI caches and hands back
+    by identity, so declaring the parameter has to be idempotent; without the guard that skips an
+    operation already carrying it, the second call appends a second copy and every later reader —
+    `/_meta/openapi/<source>`, the MCP bridge — sees the parameter twice."""
```
khj809: **14 lines of docstring, most of it `google_system_parameters`' own.**

Paragraph 1 restates that function's docstring almost clause for clause ("declared once here rather than on each of the twenty-one family routes, where a route added later would forget it and the fidelity diff would report the gap again"). Paragraph 3 is the only statement anywhere of why the idempotency guard exists — which is exactly why it should sit at the guard (see the comment on `backlot/openapi.py:202`), leaving the test to say that it asks twice and what that holds.

```suggestion
def test_every_google_operation_declares_the_system_parameter_and_checks_it(client):
    """Both directions, and the document asked for twice.

    Declared: every family operation carries it, `/batch` carries nothing. Checked: each operation
    is SENT a value the parameter does not take and has to refuse it — the declaration is derived
    from the path, so it would appear on a family route served by some other router while the
    dependency that validates it never ran, a gap only this half sees. The path parameters are
    dummies; the refusal comes before the route. The second fetch holds the enrichment idempotent,
    since it edits the schema FastAPI caches and hands back by identity."""
```

backlot/errors/google.py
nuri-yoo: Applied, and the Office half was a real mistake — thank you. Measured on this branch with an xlsx record: `GET /sheets/v4/spreadsheets/<id>?$.xgafv=1` answers `FAILED_PRECONDITION` with `reason: failedPrecondition`, where the same file through Docs answers `badRequest`. Both halves of that sentence were wrong: it is not `badRequest`, and it is not unreachable — `test_editor_apis_reject_an_office_file_of_their_own_family` reads an xlsx through Sheets over HTTP. The PR body said the same thing ("not reached from outside") and is rewritten. One edit on top of your text: the anonymous-credential clause now names all three OAuth-only APIs rather than Docs and Slides, for the reason in the thread below.

backlot/routers/google.py
nuri-yoo: Reworded, and the probe is in. Real does not answer a JSON envelope there at all — measured 2026-09-14, an unrouted family path comes back from Google's front end as HTML: 400 on `sheets.googleapis.com/v4/nope/nope` and `docs.googleapis.com/v1/nope`, 404 on `www.googleapis.com/drive/v3/nope/nope`, with `$.xgafv=1`, with `$.xgafv=3` and with no parameter alike. So neither branch of your question holds: there is no JSON shape to match, and the dependency's scope to matched routes is not contradicted by anything real does. The comment says that rather than calling it unmeasured.

backlot/openapi.py
nuri-yoo: Both taken. The comment is in at the guard (your suggestion, applied). On the import: you are right that nothing forces it — `backlot.errors.google` imports only `__future__`, `collections.abc` and `fastapi`, and `backlot/openapi.py` has no other `from backlot` line. I have left it function-local for now because hoisting it makes `openapi.py` import a package that six routers import `openapi` from, and the rule `github_page_parameters`' docstring states is about not creating that edge at all rather than about cycles specifically. If you read the rule as narrower than that, say so and I will hoist it.

tests/test_google.py
nuri-yoo: Applied for the divider. The other four are in `286b40a` and `7bef98a` as part of the two changes above, except `:851-856`, which I have kept: `test_gmail_carries_the_errors_array_unless_xgafv_is_2` is now the only place the `2&0` / `0&2` measurement is written down, since the module docstring you compacted points at `has_errors_array` for the rule rather than restating the repeats.

tests/test_google.py
nuri-yoo: Applied.

backlot/errors/google.py
nuri-yoo: Re-read, and you were right to ask. The 2026-09-14 capture did carry the whole entry, and it is Docs' and Slides' exactly: `{"message": "Login Required.", "domain": "global", "reason": "required", "location": "Authorization", "locationType": "header"}`, under a top-level message byte-identical to theirs. So the mock was wrong on every anonymous Gmail request, not only at `$.xgafv=1`, exactly as you said. Fixed in 286b40a rather than reworded: `missing_credentials` gives one answer to the three OAuth-only APIs and takes no path, since nothing is left to branch on. Your (3) dissolves with it. (2) is in 7bef98a — the `GoogleError` docstring names both users of `short` now. `test_the_credential_entries` asks Gmail with no parameter, where its entry shows, and Docs and Slides at `1`; reverting the fix fails it.

khj809 (commented):
Re-checked the whole branch at `7bef98a`. Every thread is addressed and the Gmail one turned out to be a real behaviour fix — an anonymous Gmail request now carries the same entry Docs and Slides do, confirmed at runtime, and dropping `path` from `missing_credentials` is the right shape once there is nothing left to branch on. Verified: 2987 passed / 1 skipped, `ruff check`, `ruff format --check` and `gen_docs.py --check` clean, `backlot diff --source google_drive` 200 / 200 / 0 new and `--source gmail` 157 / 157 / 0 new.

Three follow-ups, all small: the user-facing doc did not get either of the two corrections, one wrap is left over from an edit, and one docstring from the comment-volume thread was reported applied but is unchanged. Answer to the `openapi.py` import question is in that thread.

docs/supported-sources.md
```diff
@@ -205,6 +205,21 @@ because that is where Google puts them.
...
+out — the array is there unless `2` turns it off. Drive carries it whatever the value says. A
+success body is the same under all of them. A value other than `1` or `2` is refused before
+anything else is read, ahead of a bad token or an unparseable range, with real's sentence
+`Invalid query parameters. Invalid value '…' for system query parameter : $.xgafv`. Inside the
+array the entry follows the error: a typed value the proto layer refuses (an enum, a bool, an
+int32) is `reason: invalid` with no `domain`, everything else is `badRequest` under
+`domain: global`, and a missing credential on Docs and Slides is the short `Login Required.` at
+`location: Authorization`. Measured against the live Sheets, Docs and Drive APIs on 2026-09-12,
+and against Slides and Gmail on 2026-09-14 through the errors a request with no Authorization
+header reaches.
```
khj809, line 223: **This paragraph carries both of the claims that were corrected in `backlot/errors/google.py`, and neither correction reached it.**

Follow-up to the two threads on `errors/google.py`, filed here because the fix is in this file.

**1. Line 217-218, "everything else is `badRequest` under `domain: global`".** Same sentence the module docstring had, same exception: `failed_precondition` renders `failedPrecondition`. Confirmed on `7bef98a` — `GET /sheets/v4/spreadsheets/<xlsx id>?$.xgafv=1` answers `{"message": "…Office file.", "domain": "global", "reason": "failedPrecondition"}`.

**2. Line 218, "a missing credential on Docs and Slides".** True when written, not after `286b40a`. Gmail sends the identical entry now, and unlike the editor families it shows it by default — confirmed at runtime, an anonymous `GET /gmail/v1/users/me/labels` with no parameter carries `{"message": "Login Required.", "domain": "global", "reason": "required", "location": "Authorization", "locationType": "header"}`. This is the only user-facing sentence that now under-describes the mock; the 2026-09-14 date two lines below already backs the Gmail half.

```suggestion
array the entry follows the error: a typed value the proto layer refuses (an enum, a bool, an
int32) is `reason: invalid` and carries no `domain`; an Office file read as a native document is
`failedPrecondition` under `domain: global`; everything else is `badRequest` under the same domain;
and a missing credential on any of the three OAuth-only APIs — Gmail, Docs and Slides — is the
short `Login Required.` at `location: Authorization`, which Gmail shows by default where the editor
families show it only at `1`. Measured against the live Sheets, Docs and Drive APIs on 2026-09-12,
and against Slides and Gmail on 2026-09-14 through the errors a request with no Authorization
header reaches.
```

backlot/errors/google.py
```diff
@@ -8,33 +8,51 @@
...
+Inside `errors[]` the entry follows the constructor that raised it, and each one carries its own
+measurement. Measured on Sheets and Docs at `$.xgafv=1`: a typed value the proto layer refuses is
+``reason: invalid`` with NO ``domain`` (:func:`invalid_field_value`); every other measured 400 is
+``badRequest`` under ``global`` (:func:`invalid_argument`, :func:`bad_field_mask`); a 404 is
+``notFound``; a bad token ``authError`` at ``location: Authorization``; an anonymous Sheets request
+``forbidden``; an anonymous request on any of the three OAuth-only APIs ``required`` with the
+short ``Login Required.``.
+The two editor 400s NOT measured keep whatever their constructor already renders — ``Invalid
+gridRange`` is :func:`invalid_argument`, so ``badRequest``, but an Office file read as a native
+document is :func:`failed_precondition`, so ``failedPrecondition``.
```
khj809: **Leftover wrap from the edit: line 41 breaks mid-paragraph at 26 characters.**

The Office correction is confirmed on `7bef98a` — `GET /sheets/v4/spreadsheets/<xlsx id>?$.xgafv=1` answers `reason: failedPrecondition` under `FAILED_PRECONDITION`, which is what line 43-44 now says. The only thing left is cosmetic: replacing the Docs-and-Slides clause with "any of the three OAuth-only APIs" shortened line 40, so `short ``Login Required.``.` sits alone on line 41 and the next sentence starts a new line inside the same paragraph.

```suggestion
measurement. Measured on Sheets and Docs at `$.xgafv=1`: a typed value the proto layer refuses is
``reason: invalid`` with NO ``domain`` (:func:`invalid_field_value`); every other measured 400 is
``badRequest`` under ``global`` (:func:`invalid_argument`, :func:`bad_field_mask`); a 404 is
``notFound``; a bad token ``authError`` at ``location: Authorization``; an anonymous Sheets request
``forbidden``; an anonymous request on any of the three OAuth-only APIs ``required`` with the short
``Login Required.``. The two editor 400s NOT measured keep whatever their constructor already
renders — ``Invalid gridRange`` is :func:`invalid_argument`, so ``badRequest``, but an Office file
read as a native document is :func:`failed_precondition`, so ``failedPrecondition``.
```

tests/test_google.py
```diff
@@ -770,6 +770,270 @@ def test_a_bad_token_is_unauthenticated_everywhere(client, path):
...
+# --- `$.xgafv`, the system parameter that selects the error envelope --------------------------
+
+XGAFV_REFUSAL = "Invalid query parameters. Invalid value '{}' for system query parameter : $.xgafv"
+
+
+@pytest.mark.parametrize("value", ["0", "3", "NOPE", "01", ""])
+def test_xgafv_takes_1_or_2_and_refuses_the_rest_with_reals_sentence(client, admin_h, value):
+    """Measured on Sheets: 400 INVALID_ARGUMENT, this sentence, spacing and all, and no `errors[]`
+    — the value that would have asked for one was the value refused. Drive carries the array on
+    every error, this one included, as `badRequest`."""
```
khj809: **This is the one docstring from the earlier comment-volume thread that did not land.**

The reply there said the other four were in `286b40a` and `7bef98a` except `:851-856`; this one is unchanged too. It still restates `bad_system_parameter` at `backlot/errors/google.py:288-290` almost clause for clause:

- constructor — "400 INVALID_ARGUMENT with this sentence, spacing included, and on Drive an `errors[]` entry of ``badRequest`` under ``global``"
- test — "400 INVALID_ARGUMENT, this sentence, spacing and all… Drive carries the array on every error, this one included, as `badRequest`"

```suggestion
    """Sheets carries no `errors[]` on its own refusal: the value that would have asked for one
    was the value refused."""
```

backlot/openapi.py
khj809: Narrower, I read it. The rule `github_page_parameters` states is about the six routers that import `openapi` for `qp` — "which is why it does not import that one" names the routers package specifically, and `backlot.errors.google` is neither a router nor an importer of `openapi`, so hoisting adds no edge back into this module.

That said it is small either way, and the maintenance cost is yours. If you keep it function-local, put the reason you just gave into the docstring — otherwise the next reader re-derives the same question and may well answer it differently. Either resolution closes this for me.

docs/supported-sources.md
nuri-yoo: Applied in `1b7afc2` — verified by `git log -L 214,221:docs/supported-sources.md`, which now names that commit. You are right that this was the one user-facing sentence still carrying both corrections, and it is the failure mode I should have checked first: the code change made the exception, and the prose describing the old rule lived in a file outside the diff.

backlot/errors/google.py
nuri-yoo: Applied in `55eab07` (`git log -L 36,41:backlot/errors/google.py`). The short line was mine — replacing the Docs-and-Slides clause with the longer one pushed the wrap and I did not re-read the paragraph after the substitution.

tests/test_google.py
nuri-yoo: Applied in `c706cdc` (`git log -L 777,781:tests/test_google.py`), and my earlier reply was wrong: that docstring had not been touched since `4beda86`. Of the five spots, three landed (the divider and the declaration docstring through your suggestions, the credential one in `286b40a`) and one I kept deliberately; this fifth I claimed was in a commit without checking. Checking each location against `git log -L` before writing "applied" is the fix, and I have taken it as a rule rather than a one-off.

backlot/openapi.py
nuri-yoo: Hoisted, in `114c159` — and I withdraw the reason I gave. I read the rule as "`openapi.py` imports nothing from `backlot`", which the graph does not support: `import backlot.openapi` already loads `backlot` and `backlot.server` through the package `__init__`, and `backlot.errors.google` pulls only the other two error modules, neither this module nor a router. So the import at module scope adds no edge and nothing new at import time. The rule's own sentence now says which import it is about, so the question does not have to be re-derived.

khj809 (approved):
LGTM

PR #204, opened 2026-09-14
s3: the bucket listing answers ListObjects or ListObjectsV2, and reads the query parameters each of them takes (#178, #188, #192)

### LINKED ISSUE #178: s3: ListObjectsV2 ignores encoding-type, so a client that asked for url gets raw keys under no EncodingType element and a value real refuses is a 200
**Source and commit** — `s3`, at `be547e3` (on the branch for #169, PR #176).

ListObjectsV2 does not read `encoding-type`. `_list_objects_v2` reads `prefix`, `delimiter`, `start-after`, `max-keys` and `continuation-token`, and nothing else on the query string reaches the listing, so a request carrying `encoding-type=url` is answered as if it had not: the values come back as received and the response carries no `EncodingType` element. A value real refuses is a 200. Cyberduck puts `encoding-type=url` on every listing it issues (#169's trace: `GET /s3/acme-artifacts?encoding-type=url&max-keys=1000&prefix=&delimiter=%2F`), and so does boto3, which sets it on every `list_objects_v2` call the caller did not set it on (`botocore/handlers.py`, `set_list_objects_encoding_type_url`), so this is the query the two most common clients send, not an option somebody has to reach for.

|---|---|---|---|---|
| 1 | `?list-type=2&encoding-type=url` | `<EncodingType>url</EncodingType>` after `Delimiter`, before `IsTruncated` | no element | yes |
| 2 | `?list-type=2&encoding-type=url&prefix=run books/x+y%25z&delimiter=\|&start-after=k m` | `<Prefix>run+books/x+y%25z</Prefix>`, `<Delimiter>%7C</Delimiter>`, `<StartAfter>k+m</StartAfter>` | `run books/x y%z`, `\|`, `k m` | yes |
| 3 | `?list-type=2&encoding-type=url` on a key holding a space, `+`, `%` or a non-ASCII character | `Key` and `CommonPrefixes/Prefix` encoded the same way (spec only, below) | raw | yes |
| 4 | `?list-type=2&encoding-type=bogus`, `?list-type=2&encoding-type=` | 400 `InvalidArgument` "Invalid Encoding Method specified in Request", `ArgumentName` `encoding-type`, `ArgumentValue` as sent | 200, the listing | yes |
| 5 | `?list-type=2&prefix=a&prefix=b`, `?max-keys=1&max-keys=5` | the first value (`<Prefix>a</Prefix>`, `<MaxKeys>1</MaxKeys>`) | the last (`b`, `5`) | yes |

None of these is in `backlot/fidelity/baseline/s3.json`: the S3 comparison is a probe of which operations answer, not of which parameters an operation honours (`backlot/fidelity/s3_probe.py`), so a parameter the listing ignores is not a finding it can make.

**How you know** — measured against `s3.ap-northeast-2.amazonaws.com` on 2026-09-10, a general purpose bucket, path-style, SigV4, the query form-decoded and re-encoded with `quote` before signing, which is how `tests/test_s3.py`'s `_sign_get` builds one.

- Row 1: `?list-type=2&max-keys=2&encoding-type=url&delimiter=/&prefix=` answers `…<MaxKeys>2</MaxKeys><Delimiter>/</Delimiter><EncodingType>url</EncodingType><IsTruncated>true</IsTruncated>…`. `encoding-type=URL` is taken too and echoed `URL`.
- Row 2: the request in the table answers `<Prefix>run+books/x+y%25z</Prefix><StartAfter>k+m</StartAfter>…<Delimiter>%7C</Delimiter><EncodingType>url</EncodingType>`; the same request without `encoding-type` answers `<Prefix>run books/x y%z</Prefix><StartAfter>k m</StartAfter>…<Delimiter>|</Delimiter>` and no `EncodingType`. `prefix=한글/&encoding-type=url` answers `<Prefix>%ED%95%9C%EA%B8%80/</Prefix>`. The encoding is the one #176 measured on ListMultipartUploads and implements as `_url_encode`: letters, digits, `-_.*/` unchanged, space as `+`, the other bytes measured (`!'()~,;:@="<>|%` and UTF-8) as `%XX` in upper case.
- Row 3 is spec only. The API reference for ListObjectsV2 says of `EncodingType`: "If you specify the encoding-type request parameter, Amazon S3 includes this element in the response, and returns encoded key name values in the following response elements: Delimiter, Prefix, Key, and StartAfter." None of the four buckets this account can read holds a key with a character the encoding touches (5,818 keys scanned), so what real does to `Key` and to `CommonPrefixes/Prefix` was not observed, only read.
- Row 4: both values answer the 400 with the message and the two `Argument*` elements in the table; the same refusal ListMultipartUploads gives (#176).
- Row 5: `?list-type=2&max-keys=1&prefix=a&prefix=b` answers `<Prefix>a</Prefix>`, `?list-type=2&max-keys=1&max-keys=5` answers `<MaxKeys>1</MaxKeys>` and one key; the same on `?uploads&prefix=a&prefix=b`, `<Prefix>a</Prefix>`.

**Reproduction** — the bundled corpus, bucket `acme-artifacts`:

```python
import backlot, httpx
from urllib.parse import parse_qsl, quote, urlencode
from backlot.fidelity.s3_probe import _sign

with backlot.serve() as s:
    creds = httpx.get(f"{s.base_url}/_meta/users").json()
    ak, sk = creds["admin_s3_access_key_id"], creds["admin_s3_secret_access_key"]
    for q in (
        "list-type=2&max-keys=1&prefix=run books/x+y%25z&delimiter=|&start-after=k m&encoding-type=url",
        "list-type=2&max-keys=1&encoding-type=bogus",
        "list-type=2&max-keys=1&prefix=a&prefix=r",
        "list-type=2&max-keys=1&max-keys=5",
    ):
        url = f"{s.base_url}/s3/acme-artifacts?" + urlencode(parse_qsl(q, keep_blank_values=True), safe="-_.~", quote_via=quote)
        r = httpx.get(url, headers=_sign("GET", url, ak, sk))
        print(r.status_code, q, r.text.split("\n", 1)[1][:200])
```

At `be547e3` the first prints `200` with `<Prefix>run books/x y%z</Prefix>…<Delimiter>|</Delimiter><StartAfter>k m</StartAfter>` and no `EncodingType`; the second `200` with the listing; the third `<Prefix>r</Prefix>`; the fourth `<MaxKeys>5</MaxKeys>` and five keys. Row 3 cannot be shown on the bundled corpus: its 12 keys hold nothing the encoding touches, so a fixture with a key like `reports/Q3 2026.pdf` or `a+b.txt` is part of a fix.

**What it costs a client** — boto3 loses nothing today, and that is worth stating so the fix is not sized for it: `_decode_list_object` decodes only when the response says `EncodingType` is `url` and botocore itself set the parameter, so against Backlot it sees no element and leaves the raw keys alone, which is the right answer; against real it decodes, which is also right. The cost lands on a client that decodes because it asked, which the API reference's sentence above invites ("If you specify the encoding-type request parameter, Amazon S3 includes this element … and returns encoded key name values"): such a client turns a key `a+b.txt` into `a b.txt` and `100%.csv` into a decode error on Backlot, and gets both right on real. Whether Cyberduck is that client was not measured; on the corpus in #169's trace no key has such a character, so its listing rendered either way. Row 4 is the plain kind: a client sending a value real refuses gets a listing to parse instead of the error. Row 5 bites nobody observed; it is here because the listing reads `q.get`, which on Starlette's `QueryParams` is the last value. `?uploads` read the same way when this was filed and reads the first as of #176's `8fe3984`, through `_first` in `backlot/routers/s3.py`, so the listing's reads have a helper to go through.

**Where it lands** — `_list_objects_v2` in `backlot/routers/s3.py`. `_url_encode` (from #176) is the encoder, `_argument_error` the refusal, and the check goes where ListMultipartUploads has it: after the bucket is looked up (`?encoding-type=bogus` on a missing bucket is `NoSuchBucket` on real, measured for `?uploads`; the listing's own order was not measured and a fix should measure it). The element sits after `Delimiter` and before `IsTruncated` in real's `ListBucketResult`, which is where `_list_objects_v2` already emits `Delimiter`. `_P_BUCKET_GET` should declare `encoding-type`, unlike the ListMultipartUploads parameters #176 left undeclared: this one changes what comes back.

**What a fix has to settle**

- What `NextContinuationToken` and `ContinuationToken` do under `encoding-type=url`. The current API reference lists `Delimiter, Prefix, Key, and StartAfter` and not the tokens; botocore's comment on `decode_list_object_v2` lists `ContinuationToken` too and decodes `Delimiter`, `Prefix`, `StartAfter`, `Contents/Key` and `CommonPrefixes/Prefix`. Backlot's token is base64url with `=` padding, which the encoding would turn into `%3D`; real's opaque token carries `/`, `+` and `=` (`18FoPn2XX6d1wHFlrPPOihc5AJJSpHXicAWvb6AetIW/gfn3ePOb1gL+7OnoltZnCdq90b8Ov/7KZ7g/KFFid1Q==`, the `NextContinuationToken` real sent under `encoding-type=URL` in the row 1 measurement, with its `/`, `+` and `==` unencoded, which says `NextContinuationToken` is not encoded on real; `ContinuationToken`'s echo was not measured). The recommendation is to leave both tokens alone and encode the four the reference names plus `CommonPrefixes/Prefix`, which botocore decodes and which is a key prefix.
- How to measure row 3 without a write to a bucket somebody owns. A scratch bucket with three keys (`a b`, `a+b`, `한글/x`) is a one-off; if nobody has one, the fix ships row 3 as spec-backed with the docstring saying so, the way #176's `upload-id-marker` refusal is worded.
- Whether row 5 is fixed here or left. `_first` in `backlot/routers/s3.py` (#176, `8fe3984`) reads the first value and `?uploads` goes through it; leaving the listing on `q.get` keeps the two routes reading a repeated parameter differently. The recommendation is to fix it in the same change, since no client sending repeats was observed and the change is the reading of five parameters in one function.

### LINKED ISSUE #188: s3: a bare bucket GET answers the ListObjectsV2 body, so boto3's list_objects paginator cannot page
**Source and commit** — `s3`, at `50c0419`.

**How you know** — signed reads of a live bucket in `ap-northeast-2` on 2026-09-11, and the same requests against Backlot. `backlot/fidelity/baseline/s3.json` acknowledges the `?list-type=2` fallthrough with the note "a bare bucket GET is ListObjects, and `?list-type=2` selects its v2 form, so both correctly answer the same ListBucketResult". Measured, real answers two different bodies.

## 1. A bare bucket GET answers the ListObjectsV2 body, so ListObjects has no v1 pagination

**Request** — `GET /{bucket}?max-keys=1`, versus `GET /{bucket}?list-type=2&max-keys=1`.

| element | real v1 | real v2 | Backlot, both |
|---|---|---|---|
| `Marker` | present | absent | absent |
| `Owner` / `ID` | present | absent | absent |
| `KeyCount` | absent | `1` | `1` |
| `NextContinuationToken` | absent | present | present |

**Backlot serves** — one body for both, built by [`_list_objects_v2`](https://github.com/brekkylab/backlot/blob/50c0419/backlot/routers/s3.py#L351). `?marker=` is not read: `GET /{bucket}?marker=runbooks/a.md&max-keys=1` answers `runbooks/a.md` again.

**The real API serves** — the v1 shape for the bare GET, and it honours `?marker=`.

**Can a client tell?** Loudly, and only after it has already failed: botocore's own v1 paginator falls back to the last key as the next marker, Backlot ignores it, and the walk dies.

```
boto3 list_objects     -> PaginationError: The same next token was received twice: {'Marker': 'runbooks/a.md'}
boto3 list_objects_v2  -> pages=3 keys=['runbooks/a.md', 'runbooks/b.md', 'runbooks/c.md']
```

Silently, too: `list_objects(MaxKeys=1)` comes back with `['Contents', 'IsTruncated', 'MaxKeys', 'Name', 'Prefix']` — no `Marker`, and no `Owner` on any object — because botocore drops the v2 members the v1 output shape does not define.

**Reproduction** — three `s3` records sharing one `bucket`; `backlot serve`, then the boto3 paginator above.

**Why the probe does not see it** — `s3_probe.py` compares an operation's status and XML root element against the same path carrying no query, and both of these are `200 <ListBucketResult>`. ListObjects itself is never asked, being the bare form. Closing this needs the probe to compare child element sets, or ListObjects to get an asked case of its own; otherwise the fix cannot be verified by the measurement that is supposed to cover it.

### LINKED ISSUE #192: s3: a negative max-keys is a 500 or a listing with no page cap, where real refuses it with the two messages the max-uploads parser already speaks
**Source and commit** — `s3`, at `996eb24` (on the branch for #169, PR #176; `_list_objects_v2` is the same at `main` `50c0419`).

`_list_objects_v2` reads `max-keys` with `min(int(q.get("max-keys", _MAX_KEYS)), _MAX_KEYS)` and never rejects a negative, so `max_keys` goes negative and is then used both as a slice index and as the SQL LIMIT. Real parses `max-keys` the way it parses `max-uploads`: by value, as an int32, with one message for a value that parses but is below 0 and another for a value that does not parse as an int32 at all, each carrying `ArgumentName` and `ArgumentValue`. #176 writes that parser for `max-uploads` (`_max_uploads`), so as of that branch the two integer parameters on the same route sit at very different fidelity, and the less faithful one is the one every client sends. Found by khj809 reviewing #176 (https://github.com/brekkylab/backlot/pull/176#discussion_r3978444461); the real side is measured here.

|---|---|---|---|---|
| 1 | `?list-type=2&max-keys=-1` (also `-3`, `-2&prefix=design/`, and `?max-keys=-1` without `list-type`) | 400 `InvalidArgument` "Argument maxKeys must be an integer between 0 and 2147483647", `ArgumentName` `maxKeys`, `ArgumentValue` `-1` | 500, a bare `Internal Server Error` with no `<Error>` envelope | no |
| 2 | `?list-type=2&max-keys=-2` | the same 400 | 200 with `<MaxKeys>-2</MaxKeys><IsTruncated>true</IsTruncated>` and every visible key but the last two in `Contents`, after reading every ACL-visible row in the bucket | yes |
| 3 | `?list-type=2&max-keys=abc`, `= 5` (a leading space), `=2147483648`, `=-2147483649` | 400 `InvalidArgument` "Provided max-keys not an integer or within integer range", `ArgumentName` `max-keys`, `ArgumentValue` as sent | `abc`: 400 "max-keys must be an integer" with an empty `Resource` and no `ArgumentName` or `ArgumentValue`; ` 5`: 200, a page of 5; `2147483648`: 200, a page of 1000; `-2147483649`: 500 | the three 200s and the 500 |
| 4 | `?list-type=2&max-keys=` | 200, the default page, `<MaxKeys>1000</MaxKeys>` | 400 "max-keys must be an integer" | no |
| 5 | `?list-type=2&max-keys=0`, and `-0` | 200, `<KeyCount>0</KeyCount><MaxKeys>0</MaxKeys><IsTruncated>false</IsTruncated>` with keys in the bucket | 200 with `IsTruncated` true and no `NextContinuationToken` | yes |
| 6 | `?list-type=2&max-keys=1001` on a bucket holding more than 1000 keys | `<KeyCount>1000</KeyCount><MaxKeys>1001</MaxKeys><IsTruncated>true</IsTruncated>`: capped in what is served, echoed as sent | `<MaxKeys>1000</MaxKeys>` | yes |
| 7 | `?list-type=2&max-keys=1&max-keys=abc` | 200, a page of 1 (the first value) | 400 (the last), the reading #178 row 5 names | no |

None of these is in `backlot/fidelity/baseline/s3.json`: the S3 comparison is a probe of which operations answer, not of how an operation reads a parameter (`backlot/fidelity/s3_probe.py`).

**How you know** — measured against `s3.ap-northeast-2.amazonaws.com` on 2026-09-11, general purpose buckets, path-style, SigV4, the query form-decoded and re-encoded with `quote` before signing, which is how `tests/test_s3.py`'s `_sign_get` builds one.

- Rows 1 to 5 on a bucket holding 20 keys. `-1`, `-2147483648` and `?max-keys=-1` without `list-type` answer the `maxKeys` message; `abc`, ` 5`, `2147483648` and `-2147483649` answer the `max-keys` message with `ArgumentValue` as sent, the space included. `05` and `00000000005` answer `<MaxKeys>5</MaxKeys>` and five keys; `-0` and `0` answer `<KeyCount>0</KeyCount><MaxKeys>0</MaxKeys><IsTruncated>false</IsTruncated>`; `2147483647` is served with `<MaxKeys>2147483647</MaxKeys>`; empty answers `<MaxKeys>1000</MaxKeys>`.
- The two spellings are real's, not a slip in transcription: the range message and its `ArgumentName` say `maxKeys`, the other message and its `ArgumentName` say `max-keys`, on four requests across both forms of the listing. `max-uploads` spells both of its messages the same way (#176).
- Row 6 on a bucket holding more than 1000 keys: `max-keys=1001` answers `<KeyCount>1000</KeyCount><MaxKeys>1001</MaxKeys><IsTruncated>true</IsTruncated>`. This is not what `max-uploads` does, which echoes 1000 for 1001 (#176).
- Row 7: `?list-type=2&max-keys=1&max-keys=abc` answers `<MaxKeys>1</MaxKeys>` and one key; `abc&1` is the row 3 refusal naming `abc`.
- The order around the bucket lookup, on a bucket that does not exist: `?list-type=2&max-keys=abc` is the row 3 refusal, `?list-type=2&max-keys=-1` is `NoSuchBucket`. The same straddle `max-uploads` has (#176, `996eb24`): the parse before the lookup, the range after.
- The SQL side of row 2: "If the LIMIT expression evaluates to a negative value, then there is no upper bound on the number of rows returned" (SQLite, *SELECT*, "The LIMIT clause"). `limit = max_keys + 1` is `-1` for `-2`.

**Reproduction** — the bundled corpus, bucket `acme-artifacts`:

```python
import re, backlot, httpx
from urllib.parse import parse_qsl, quote, urlencode
from backlot.fidelity.s3_probe import _sign

with backlot.serve() as s:
    creds = httpx.get(f"{s.base_url}/_meta/users").json()
    ak, sk = creds["admin_s3_access_key_id"], creds["admin_s3_secret_access_key"]
    for q in ("list-type=2&max-keys=-1", "list-type=2&max-keys=-2", "list-type=2&max-keys=abc",
              "list-type=2&max-keys= 5", "list-type=2&max-keys=", "list-type=2&max-keys=0"):
        url = f"{s.base_url}/s3/acme-artifacts?" + urlencode(parse_qsl(q, keep_blank_values=True), safe="-_.~", quote_via=quote)
        r = httpx.get(url, headers=_sign("GET", url, ak, sk))
        print(r.status_code, q, re.findall(r"<(KeyCount|MaxKeys|IsTruncated|Message)>(.*?)</", r.text) or r.text)
```

At `996eb24` the first prints `500 Internal Server Error`; the second `200` with `KeyCount` 5, `MaxKeys` -2 and `IsTruncated` true, admin's seven keys less the last two; the third `400` with the message `max-keys must be an integer`; the fourth `200` with `KeyCount` 5 and `MaxKeys` 5; the fifth `400` with the same message; and the sixth `200` with `KeyCount` 0, `MaxKeys` 0 and `IsTruncated` true. Row 6 needs a bucket with more than 1000 keys, which the corpus is not.

**What it costs a client** — Row 1 is the loud one: the 500 is `overflow_row = rows[max_keys]` reading `rows[-1]` off an empty list, and it surfaces with no `<Error>` body, so boto3 retries it and then raises a generic 500 where real raises a `ClientError` naming the parameter. Row 2 is the quiet one and the worse: SQLite reads the negative LIMIT as no limit, so `store.list_s3_objects` walks every visible row in the bucket before `rows[:max_keys]` cuts the last two off and serves the rest, and the 1000-key cap and the keyset paging above it are gone for that request. Row 5 costs a client that stops paging on `IsTruncated`: real says the empty page is the last, Backlot says there is more and gives no token to fetch it with. Rows 3, 4 and 6 are echo and refusal shape.

**Where it lands** — `_list_objects_v2` in `backlot/routers/s3.py`, at the `min(int(...))`. `_max_uploads` is the parser this wants: give it the parameter name and its two messages as arguments and call it from both, and mind that the listing's two messages spell the name differently (`maxKeys`, `max-keys`) where `max-uploads` spells both the same. The parse sits where `_max_uploads` runs today, before `_bucket_visible`, and the range refusal after it, where `_list_multipart_uploads` has it. `_P_BUCKET_GET`'s description of `max-keys`, "keys per page, capped at 1000, which is also the default", stays true of what is served.

**What a fix has to settle**

- The echo past the cap. `_MAX_KEYS` clamps both the LIMIT and the `MaxKeys` element; real clamps the LIMIT and echoes the value sent (row 6). Separating the two is a one-line change and the recommendation, since it is what was measured and boto3 pages on `IsTruncated` and `NextContinuationToken`, not on `MaxKeys`.
- `IsTruncated` for `max-keys=0` (row 5). Real answers false with keys in the bucket; the comment above `store.list_s3_objects` in `_list_objects_v2` says the LIMIT stays 1 for `max_keys=0` "so IsTruncated is still computed", which is the opposite choice. The recommendation is real's answer, since a client that pages on `IsTruncated` is told there is no next page and is given no token for one either way.
- Whether the 500 in row 1 gets a regression test of its own or is covered by the refusal. Once a negative is refused, `rows[max_keys]` with a negative `max_keys` is unreachable; a test that the refusal happens is the guard.

nuri-yoo:
Closes #188, closes #178, closes #192. A bare bucket GET answers ListObjects and `list-type=2` answers ListObjectsV2; the listing reads `marker`, `start-after`, `continuation-token`, `encoding-type` and `max-keys`, each as its first value when it is sent twice, refusing what real refuses; and the S3 probe tells the two bodies apart, so ListObjectsV2 is no longer acknowledged as a fallthrough.

**The bucket GET is two listings, not one (#188).** `list-type=2` selects the V2 shape; the parameter absent and every other value select V1, which is what real answers for `list-type=1`, `list-type=0` and `list-type=bogus` alike. What the two carry, measured against a bucket in `ap-northeast-2` on 2026-09-14:

| element | V1 | V2 |
|---|---|---|
| `Marker` | always, empty when none was sent, echoed when one was | absent |
| `NextMarker` | only under a `delimiter`, naming the last entry of the page | absent |
| `KeyCount` | absent | present |
| `ContinuationToken` / `NextContinuationToken` | absent | echoed when sent / present when truncated |
| `StartAfter` | absent | echoed when sent and no continuation token displaced it |
| `Contents/Owner` | an `ID`, with no `DisplayName` in this region | absent |

Element order is real's on both: `Name`, `Prefix`, then V1's `Marker` and `NextMarker` or V2's `StartAfter`, `ContinuationToken`, `NextContinuationToken` and `KeyCount`, then `MaxKeys`, `Delimiter`, `EncodingType` and `IsTruncated`.

Two of those rows have a documented source and one contradicts it. `NextMarker` is the reference's own rule — "This element is returned only if you have the `delimiter` request parameter specified. If the response does not include the `NextMarker` element and it is truncated, you can use the value of the last `Key` element in the response as the `marker` parameter in the subsequent request" — and that fallback is what botocore's V1 paginator does, so sending a cursor where real sends none would make Backlot easier to page than the thing it stands in for. `Marker` goes the other way: the reference says it "is included in the response if it was sent with the request", and real sends the empty element either way, which is what is served here.

This is what the issue's client symptom was. botocore's V1 paginator falls back to the last key as the next `Marker`, the server ignored it and sent the same page again, and `list_objects` raised `PaginationError: The same next token was received twice` on the second page. It also dropped `Marker` and every `Owner`, because botocore keeps only the members the V1 output shape declares. Both paginators now walk the bundled corpus to the end, one key per page, and agree on the keys.

A `marker` whose own group has already been listed skips that whole group rather than walking back into it. Real answers `?delimiter=/&marker=docs/` and `?delimiter=/&marker=docs/a.txt` alike with what follows `docs/`, never `docs/` again — measured by paging a delimited listing to the end, one page each for `docs/`, `notes/` and a plain key, with no repeat — so a delimited V1 walk terminates. That is the bound V2 already reached through its group token.

A parameter belonging to the other listing is refused rather than ignored, before the bucket is looked up, with the message real sends and an `ArgumentName` and no `ArgumentValue` beside it:

| request | message |
|---|---|
| `?list-type=2&marker=…` | `Marker unsupported with REST.GET.BUCKET in list-type=2` |
| `?start-after=…` without `list-type=2` | `startAfter only supported in REST.GET.BUCKET with list-type=2` |
| `?continuation-token=…` without `list-type=2` | `continuation-token only supported in REST.GET.BUCKET with list-type=2` |

An empty value is refused like any other, a bucket that does not exist still takes the 400, and a V1 request carrying both V2 parameters is refused for `continuation-token`.

**`encoding-type` is read, echoed and applied (#178).** boto3 sets it on every `list_objects` and `list_objects_v2` the caller did not (`botocore/handlers.py`, `set_list_objects_encoding_type_url`) and Cyberduck puts it on every listing it issues, so this is the query the common clients send, and it was reaching nothing. Under `encoding-type=url` the response carries an `EncodingType` element and `Prefix`, `Delimiter`, `StartAfter`, `Marker`, `NextMarker`, every `Key` and every `CommonPrefixes/Prefix` come back encoded; the continuation tokens do not. The reference names four of those — "returns encoded key name values in the following response elements: `Delimiter`, `Prefix`, `Key`, and `StartAfter`" — and says nothing about V1's `Marker` and `NextMarker` or about the tokens, so those are measured rather than read: a marker sent as `a b.txt` is echoed `a+b.txt`, a `NextMarker` naming `100%.csv` comes back `100%25.csv`, and a token keeps its `/`, `+` and `=`. `URL` is taken like `url` and echoed as sent; any other value, empty included, is `Invalid Encoding Method specified in Request` after the bucket lookup.

The encoder is `_url_encode`, which #176 measured on ListMultipartUploads. What the issue left as spec-only, the effect on `Key` and `CommonPrefixes/Prefix`, is measured now, against a bucket created for it and deleted after:

| key stored | `Key` under `encoding-type=url` |
|---|---|
| `a b.txt` | `a+b.txt` |
| `a+b.txt` | `a%2Bb.txt` |
| `100%.csv` | `100%25.csv` |
| `한글/x.txt` | `%ED%95%9C%EA%B8%80/x.txt` |
| `run books/x.txt` | `run+books/x.txt` |

The first two rows are the pair that makes the encoding worth having: decoded they are the same string, and a client that decodes cannot tell them apart unless the server encoded. They also fix the order of the two steps — the listing sorts on the stored key and encodes after, which is why `a b.txt` comes back before `a+b.txt`, the reverse of what encoding first would give.

**A repeated parameter is read as its first value (#178).** The listing went through `QueryParams.get`, which is the last, where real reads the first and `?uploads` already did through `_first` (#176): `?prefix=a&prefix=zz.txt` lists under `a`, and `?max-keys=1&max-keys=abc` is a page of one rather than the refusal `abc` would be on its own. `list-type` is read that way too, so the first of two spellings picks the shape — `?list-type=1&list-type=2` answers V1 and the two the other way round answer V2.

**`max-keys` is parsed by value (#192).** It went through `int()` and `min()`, so a negative reached a slice index and a SQL `LIMIT`: `-1` was a bare 500 with no `<Error>` envelope, and `-2` was a 200 that walked every ACL-visible row in the bucket before trimming. It now takes the parser #176 wrote for `max-uploads`, which is the same parser real uses for both — `05` and `-0` and an empty value are served, a word and a leading space and a value outside an int32 are not.

Its two refusals spell the parameter differently, which is real's and not a transcription slip: a value that does not parse is `Provided max-keys not an integer or within integer range` under `ArgumentName` `max-keys`, and one that parses but is below zero is `Argument maxKeys must be an integer between 0 and 2147483647` under `maxKeys`. `max-uploads` spells both of its own the same way, so the range refusal takes its name as an argument.

The split is also a split around the bucket lookup, the same straddle `?uploads` has: the parse happens before it and the range after, which a bucket that does not exist tells apart. `encoding-type` is judged after the lookup and before the range. `MaxKeys` echoes the value as parsed and uncapped — real answers `max-keys=1001` with `<MaxKeys>1001</MaxKeys>` and at most 1000 keys — and `max-keys=0` is an empty page whose `IsTruncated` is false with keys in the bucket, where Backlot said true and gave no token to fetch the next page with.

**A prefix or marker at the last code point is a page, not a 500.** `store.key_successor` builds the half-open range a prefix filters on by incrementing the argument's last character, and the last code point has nothing above it, so `?prefix=a\U0010ffff` raised `ValueError` and went out as a 500 with no `<Error>` body. That is on `main` already; what this PR adds is a second route to it, since the listing's `marker` reaches the same helper through the CommonPrefixes group it resumes past. The helper now carries — a trailing run of that code point comes off and the character before it is incremented, which is still above every string starting with the argument — and returns `None` when the argument is nothing but that code point, which the prefix filter reads as "no upper bound" and the listing as "nothing sorts after this group". What real S3 answers for these is unmeasured, and nothing here claims otherwise; the claim is that the server answers.

**The probe tells the two bodies apart.** It judged an operation indistinguishable from the catch-all on its status and XML root element alone, and both listings are `200 <ListBucketResult>`, so ListObjectsV2 had to be acknowledged in the baseline to keep the comparison quiet — which meant the comparison that covers S3 could not see this fix regress. It now compares the set of direct child elements too: `KeyCount` against `Marker` separates them, and the acknowledgement is gone. A request the catch-all really does answer still matches on all three, which `?session` shows and which stays acknowledged, because real answers the listing there as well.

**What the tests hold.** Eight tests in `tests/test_s3.py`, two in `tests/test_store.py` and one in `tests/test_fidelity.py`. The listing tests run over a bucket of their own whose keys hold a space, a literal `+`, a `%` and a non-ASCII character, plus a folder whose name holds a space — the bundled corpus has none of these, so every key in it comes back the same encoded or not and it cannot tell the two apart. Each rule below, reverted in place, turns at least one of them red:

```
list-type is ignored and everything is V2
NextMarker is sent without a delimiter
a parameter of the other listing is ignored
encoding-type does not reach the keys
the range refusal spells the parameter max-keys
max-keys=0 reports IsTruncated true
V1 carries no per-object Owner
MaxKeys echoes the capped value
a marker inside a group does not skip it
NextMarker names the last raw key rather than the rolled-up prefix
an element moves in the body's order
the echoes of prefix, delimiter, start-after and marker are not encoded
a repeated parameter reads its last value, list-type included
the probe compares no child elements
key_successor increments the last character with no carry
```

**Not in this PR.** A malformed `continuation-token` is still read as the first page, where real refuses it with `The continuation token provided is incorrect`; that is unchanged by the split and is #205, where the refusal's place around the bucket lookup is measured. The `<Resource>` element Backlot puts in every S3 error body has no counterpart in real's `InvalidArgument` responses, which is router-wide and older than this change.

**Verification.** At `7000cb1`, rebased onto `main` at `ab8bae9`.

```
PYTHONPATH=. python -m pytest -rs                    2998 passed, 1 skipped
```

`backlot diff --source s3` against the live botocore model reports 41 divergences, 41 acknowledged and 0 new — one fewer than before, the ListObjectsV2 acknowledgement this removes, and nothing else reclassified. The one skip is `llama_index.readers.hubspot`, an optional extra this branch does not touch. Every measurement quoted above was taken on 2026-09-14 against `s3.ap-northeast-2.amazonaws.com`, path-style and SigV4, using a bucket created for it and deleted afterwards; the scripts are the same `parse_qsl` then `urlencode(quote)` form `tests/test_s3.py`'s `_sign_get` builds, so what reached the wire matches what the tests send. Against `backlot.serve()` on the bundled corpus, boto3's `list_objects` and `list_objects_v2` paginators both walk seven keys over seven pages at `PageSize=1`, and `list_objects` reports `Marker` and an `Owner` on every object; `s3fs` lists, walks and stats through the same server. On a 400-object corpus the two listings were walked to the end across ninety combinations — three callers with different ACL scopes, six prefix and delimiter shapes including a rolled-up group larger than a page, and five page sizes — and returned identical entry lists, with no entry repeated and every walk terminating. The element order was compared against real for six shapes, with every optional element present and absent.


backlot/routers/s3.py
```diff
@@ -497,43 +648,73 @@ def _list_objects_v2(request: Request, conn, bucket: str, visible) -> Response:
...
-            ts = r["updated_ts"] or r["created_ts"]
-            body.append(
-                f"<Contents><Key>{escape(val)}</Key>"
-                f"<LastModified>{synth.s3_iso(ts)}</LastModified>"
-                f"<ETag>{escape(synth.s3_etag(r['key'], r['content']))}</ETag>"
-                f"<Size>{len(r['content'].encode())}</Size>"
-                f"<StorageClass>{escape(r['subtype'] or 'STANDARD')}</StorageClass></Contents>"
-            )
+    if v2:
+        if start_after and not continuation:
```
khj809: **behaviour / blocker.** Truthiness, so `?list-type=2&start-after=` serves a 200 with no `StartAfter`; real sends it empty. Measured on a bucket holding the six `encoded-bucket` keys — real `...<Prefix></Prefix><StartAfter></StartAfter><KeyCount>6</KeyCount>...`, this `...<Prefix></Prefix><KeyCount>6</KeyCount>...`. The branch treats an empty value as sent everywhere else: `start-after=` is refused under V1, `marker=` echoes empty, `encoding-type=` is refused.

```suggestion
        if _first(q, "start-after", None) is not None and not continuation:
```

Worth asserting the empty echo beside the `Marker` one in `test_a_bare_bucket_get_is_list_objects_and_only_list_type_2_is_its_v2_form`, which already sets that state up.

khj809 (commented):
Two reviewer passes, both **block**. The behaviour pass reproduced the body's measurements live against `s3.ap-northeast-2.amazonaws.com` (path-style, SigV4) on two throwaway buckets, one populated with the six `encoded-bucket` fixture keys: an 80-query battery diffed against real came back byte-identical modulo tokens, timestamps, ETags and RequestId on **77 of 80** — the V1/V2 element split, the metadata order, `Contents` grouped before `CommonPrefixes`, `NextMarker` naming the last entry by key, all three cross-version refusals, both `max-keys` spellings and their straddle of the bucket lookup, the whole `encoding-type` table, first-value-on-repeat, and the refusal precedence. All nineteen rules the body lists were reverted in place and each turns at least one test red. The blockers below are the residue.

Inline: one behaviour blocker and four prose blockers, each with a suggestion. Two more that have no line in the diff to anchor to:

**Test coverage — `tests/test_s3.py:370`.** `test_s3_large_bucket_acl_scopes_listing` is the only test that sends a scoped token to a listing and it hardcodes `list-type=2`; grepping `big_bucket_tokens` shows no other use against a bucket GET. So no scoped token ever reaches the V1 body this PR adds — keys plus a new per-object `Owner`. Parametrize that test over `""` and `"list-type=2&"` so the admin/scoped pair is proved on both served bodies.

**PR body.** It uses none of `.github/pull_request_template.md`: `## What changed`, `## Why this is what the real API does` and `## Verification` are absent, the three Verification lines are replaced by a free-form code block that never names the install extras the 3072-passed/1-skipped count came from, and the three checkboxes are missing. Reflow the existing prose under the template's headings, put the pytest/ruff/gen_docs/`backlot diff` lines on the three Verification bullets with the `.[...]` extras named, and end at the checklist.

Notes worth folding in while you are here, not blocking: `s3_probe.py:176` restates the `KeyCount`-vs-`Marker` example the module docstring and `docs/fidelity.md` already give; four added test docstrings describe the deleted code rather than the measurement (`tests/test_s3.py:1363`, `:1619`, `:1649`, `tests/test_store.py:626`); `test_list_s3_objects_takes_the_lower_bound_alone_when_a_prefix_has_no_successor` duplicates the fixture and assertion form of the surrogate test directly above it; and the body's "The 2026-09-16 round re-measured..." sentence describes rounds of work rather than the finished state. Separately, empty `continuation-token` answers 200 where real is a 400 — unchanged from `main` and deferred to #205, but the V2-only gate is what makes it reachable, so it belongs in that issue's scope.

backlot/routers/s3.py
```diff
@@ -59,10 +73,11 @@
...
-# it, and advertising a parameter that changes nothing is worse than not offering it.
+# back, which is why `list-type`, `marker` and `encoding-type` are declared below: the first
+# chooses between the two listings, the second pages the V1 one, and the third changes how every
+# key comes back.
 # ListMultipartUploads' own parameters (`max-uploads`, `key-marker`, `upload-id-marker`,
-# `encoding-type`) are absent for a different reason. They are not inert: _max_uploads and
+# `encoding-type`) are absent for a different reason. They are not inert: _int32_param and
 # _list_multipart_uploads read all four, `max-uploads`, `key-marker` and `encoding-type` come back
 # echoed, and `max-uploads`, `encoding-type` and `upload-id-marker` can each turn the 200 into an
 # InvalidArgument. What none of them does is decide which uploads a caller gets, because there are
```
khj809: **prose / blocker.** Drift: this same diff declares `qp("encoding-type", ...)` in `_P_BUCKET_GET` (L109), and `_P_BUCKET_GET` is the parameter list for `GET /{bucket}`. The comment says a parameter is undeclared on the path where it is now declared.

```suggestion
# ListMultipartUploads' own `max-uploads`, `key-marker` and `upload-id-marker` are absent for a
# different reason. They are not inert: _int32_param and _list_multipart_uploads read them,
# `max-uploads` and `key-marker` come back echoed, and `max-uploads` and `upload-id-marker` can
# each turn the 200 into an InvalidArgument. What none of them does is decide which uploads a
# caller gets, because there are
```

(L84 onward stays as it is.)

backlot/routers/s3.py
```diff
@@ -497,43 +648,73 @@ def _list_objects_v2(request: Request, conn, bucket: str, visible) -> Response:
...
+        body.append(f"<Delimiter>{escape(enc(delimiter))}</Delimiter>")
+    if encoding_type is not None:
+        body.append(f"<EncodingType>{escape(encoding_type)}</EncodingType>")
+    body.append(f"<IsTruncated>{'true' if is_truncated else 'false'}</IsTruncated>")
+    # Every `Contents` first and every `CommonPrefixes` after, which is real's document order and
+    # not the key order `entries` is in: a page of `100%.csv`, `a b.txt`, `a+b.txt`, `run books/`
+    # and `zz.txt` comes back with `run books/` last, after `zz.txt`, on both listings (measured
+    # 2026-09-14 and 2026-09-16). `entries` keeps key order because that is what the cursors are
+    # cut from — `NextMarker` on that page is `zz.txt`, the last entry by key, not the element the
+    # body ends on.
```
khj809: **prose / blocker.** Third statement of the same rule with the same example keys and the same conclusion: the `_list_objects` docstring (L539-544) and the comment at L620-623 already carry it. Keep the docstring paragraph as the one place, and cut here to a pointer.

```suggestion
    # Every `Contents`, then every `CommonPrefixes` -- real's document order, not the key order
    # `entries` holds (see this function's docstring).
```

backlot/store.py
```diff
@@ -1157,12 +1157,38 @@ def list_documents(
...
+def key_successor(s: str) -> str | None:
+    """The smallest string greater than every string with prefix ``s``, so an S3 prefix becomes the
+    half-open range ``key >= s AND key < key_successor(s)``. The listing router also uses it to skip
+    a whole CommonPrefixes group in one bound.
+
+    Normally this increments the last character. Two ranges of code point have no usable character
+    one step up, and both are reachable from the wire, where ``?prefix=``, ``?delimiter=`` and
+    ``?marker=`` take whatever a client sends, and where each used to go out as a 500 with no
+    ``<Error>`` body. A step into the surrogate block goes over it instead. A trailing run of the
+    last code point has nothing above it at all, so the run comes off and the character before it
```
khj809: **prose / blocker.** "where each used to go out as a 500" narrates the pre-change behaviour in a production docstring. The docstring states what the helper returns now; the 500 is the PR body's material, where it already is.

```suggestion
    one step up, and both are reachable from the wire, where ``?prefix=``, ``?delimiter=`` and
    ``?marker=`` take whatever a client sends. A step into the surrogate block goes over it
    instead. A trailing run of the last code point has nothing above it at all, so the run comes
    off and the character before it
```

tests/test_s3.py
```diff
@@ -1173,3 +1257,439 @@ def test_presigned_unexpired_ok(path):
...
+def test_a_prefix_or_marker_no_character_steps_past_is_a_page_not_a_500(
+    big_bucket_client, big_bucket_settings, edge
+):
+    """Both parameters take any character a client sends, and two of them had no character the
+    key-range helper could step onto, so each went out as a 500 with no `<Error>` body for boto3 to
+    raise on. The last code point has nothing above it at all and raised `ValueError`; U+D7FF has
+    only the surrogate block, and `chr(ord(c) + 1)` landed on U+D800, which sqlite3 cannot bind
+    (`UnicodeEncodeError`). The listing's own `marker` reaches the same helper through the
+    CommonPrefixes group it resumes past, which is the route this PR opens. Neither is a listing
+    anyone wants; both are answered.
```
khj809: **prose / blocker.** A docstring that names "this PR" is provenance, not a measurement, and reads as stale the moment it merges.

```suggestion
    (`UnicodeEncodeError`). The listing's own `marker` reaches the same helper through the
    CommonPrefixes group it resumes past. Neither is a listing anyone wants; both are answered.
```

backlot/routers/s3.py
nuri-yoo: Applied, and measured before applying — `?list-type=2&start-after=` is `<StartAfter></StartAfter>` on real and the parameter absent gets no element (2026-09-17, bucket holding the six fixture keys). The `not continuation` half holds too: a token real issued alongside an empty `start-after=` comes back with `ContinuationToken` and no `StartAfter`.

I took the same behaviour one line earlier instead of reading the query twice — `start_after = _first(q, "start-after", None) if v2 else None`, which is how `continuation` right below it already reads. The echo is then `if start_after is not None and not continuation:`, and the lower bound stays on truthiness, so an empty value seeds nothing.

The empty echo is asserted beside the `Marker` one as suggested, together with the absent case. Reverting it to truthiness turns that test red.

6b6c013

backlot/routers/s3.py
nuri-yoo: Applied. The contradiction is inside the one comment block — L75 of the same diff says `encoding-type` is declared and L80 called it absent. Took the suggestion as written; `max-uploads` and `key-marker` are still the two that come back echoed and `max-uploads` and `upload-id-marker` the two that can turn the 200 into an InvalidArgument, so the sentence is true with `encoding-type` out of it.

e34c7ea

backlot/routers/s3.py
nuri-yoo: Applied as written. The docstring paragraph is the one place; the comment here and the one on `entries` both point at it now rather than restating the example keys.

e34c7ea

backlot/store.py
nuri-yoo: Applied as written.

e34c7ea

tests/test_s3.py
nuri-yoo: Applied as written.

e34c7ea

nuri-yoo:
All seven are in at `e34c7ea`, and nothing you raised turned out to be wrong when I checked it. The two without a line to anchor to, and the notes:

**Test coverage.** `grep -n big_bucket_tokens tests/test_s3.py` gives exactly what you said — three hits, all inside `test_s3_large_bucket_acl_scopes_listing`, which pinned `list-type=2`. It is parametrized over `""` and `"list-type=2&"` now, and asserts the scoped caller reads V1's per-object `Owner` and that V2's is absent, so `owner = ""` turns the `[v1]` case red rather than only the boto3 paginator test. `588a80f`.

**PR body.** You are right and it is worse than drift: `.github/pull_request_template.md` has been there since `1fdb177` (#86, 2026-08-26) and is in my own base, and the body had none of the three sections. Rewritten under them — the prose reflowed into `## What changed`, the measurements into `## Why this is what the real API does` as a `request → answer → date` block, and `## Verification` filled on the three bullets: `pytest` on `uv sync --all-extras --locked` plus `uv pip install --no-deps "llama-index-readers-hubspot<0.6"`, which is `3105 passed` with no skips, all nineteen `importorskip` targets in `tests/` resolving. The three checkboxes are there and checked with the evidence beside the ACL one.

The notes are folded in, all in `e34c7ea`: `_shape` drops the `KeyCount`/`Marker` example and keeps the pointer; the four test docstrings state the rule and the measurement instead of the code that was replaced; the surrogate `key_successor` test drops the `_s3_mini_db` block, since `test_a_prefix_or_marker_no_character_steps_past_is_a_page_not_a_500` already proves that bound reaching SQL over the wire; and the body's "2026-09-16 round" sentence is gone with the rest of the rewrite.

**The empty `continuation-token` is worth more than a scope note.** #205's body claimed the opposite as measured — "an empty `continuation-token=` answers page one with no echo, which is the same answer real gives it (measured — an empty value is not refused)". It does not. Real answers it with the same 400, the same message and `ArgumentName`, no `ArgumentValue`; on a bucket name that does not exist it is `NoSuchBucket`, so it sits in the same place around the lookup as `garbage` does. I checked the prepared URL to be sure the value was genuinely empty on the wire rather than a dropped parameter. #205 now carries that as its own row and says which sentence was wrong.

Since the page assembly moved after your pass, the checks are re-run at `e34c7ea` rather than quoted: twenty reverted rules each turn at least one test red, `backlot diff --source s3` is 41 · 41 acknowledged · 0 new, and the ninety-combination walk over a 400-object corpus — three ACL scopes, six prefix and delimiter shapes, five page sizes — still has the two listings returning identical entry lists with nothing repeated and every walk terminating.

tests/test_store.py
```diff
@@ -599,6 +599,48 @@ def test_list_s3_objects_prefix_no_like_wildcard_semantics(tmp_path):
...
+    """U+D7FF is one step below the surrogate block, which UTF-8 cannot encode and sqlite3 cannot
+    bind, so the step goes over it to U+E000.
+
+    That is still exact: a key sorting between the two would have to spell a surrogate, and a key
+    stored as UTF-8 cannot. `tests/test_s3.py` holds the same bound reaching SQL over the wire."""
+    assert store.key_successor("a\ud7ff") == "a\ue000"
+    assert store.key_successor("\ud7ff") == "\ue000"
+    assert store.key_successor("\ud7ff\U0010ffff") == "\ue000"
+    for prefix in ("a\ud7ff", "\ud7ff"):
+        assert prefix < prefix + "z" < store.key_successor(prefix)
```
khj809, line 633: **Question, not a blocker.** My earlier note was to fold the *last code point* case into an existing test, and that one is untouched — `test_list_s3_objects_takes_the_lower_bound_alone_when_a_prefix_has_no_successor` is still its own function at L636. What moved instead is this test, which lost its fixture and these two lines:

```python
    conn = _s3_mini_db(tmp_path)
    ...
        assert store.list_s3_objects(conn, "b", prefix=prefix) == []
```

So the docstring's claim that sqlite3 cannot bind a surrogate is now asserted nowhere at the store layer — U+D800 never reaches a `conn` in `tests/test_store.py`, and the pointer at L628 hands the whole binding half to `tests/test_s3.py`, where it arrives through the router. That is defensible: the wire path is the one a client takes, and a store-layer duplicate of it is the kind of second assertion the note was asking you to drop. It just is not what the note asked for, so I would rather hear it was deliberate than assume it.

If it was deliberate, nothing to do. If the SQL bind is worth keeping here, taking `tmp_path` back and restoring the one assertion inside the existing loop is enough — no new function.

nuri-yoo: Not deliberate in the way it reads — your note named the lower-bound test as the duplicate and I trimmed this one instead.

One correction, since it changes what was lost: those lines never asserted that a surrogate cannot bind. They bound `a\ud7ff` and `a\ue000`, both encodable; U+D800 only reaches a conn under the mutation. What went is where that mutation surfaces — reverting the skip still turns two tests red, but the store one now fails on equality at `test_store.py:629` where it used to raise `UnicodeEncodeError` out of `list_s3_objects`.

Rather than take the two lines back and leave two functions with the same shape, the SQL half is one test over both branches and the `key_successor` tests stay pure value assertions:

```python
def test_the_edge_bounds_key_successor_returns_reach_sql(tmp_path):
    conn = _s3_mini_db(tmp_path)
    for prefix in ("a\ud7ff", "\ud7ff", "\U0010ffff", "a\U0010ffff"):
        assert store.list_s3_objects(conn, "b", prefix=prefix) == []
    assert store.list_s3_objects(conn, "b", prefix="logs/") != []
```

One fixture, one `== []` form, both branches. Reverting the surrogate skip turns it red at `conn.execute`, and reverting the carry turns it red with `ValueError`, both at the store layer.

`2575021`. 3105 passed, ruff and `gen_docs --check` clean.

_(Edited: the escapes in the block above came through as literal code points the first time.)_

backlot/routers/s3.py
```diff
@@ -59,13 +73,14 @@
...
+# chooses between the two listings, the second pages the V1 one, and the third changes how every
+# key comes back.
+# ListMultipartUploads' own `max-uploads`, `key-marker` and `upload-id-marker` are absent for a
+# different reason. They are not inert: _int32_param and _list_multipart_uploads read them,
+# `max-uploads` and `key-marker` come back echoed, and `max-uploads` and `upload-id-marker` can
+# each turn the 200 into an InvalidArgument. What none of them does is decide which uploads a
+# caller gets, because there are
 # never any — all they shape is an echo of the caller's own input on a page that is always empty.
 # Declaring them would advertise a paging surface, a marker to resume from and a page size, over a
 # listing that never has a second page.
```
khj809: **nit, my fault.** L83 is 32 columns mid-paragraph — my earlier suggestion had to stop at the hunk boundary, and the short tail went in as written. `ruff` does not select E501 and the formatter does not reflow comment prose, so CI will never say anything about it.

```suggestion
# ListMultipartUploads' own `max-uploads`, `key-marker` and `upload-id-marker` are absent for a
# different reason. They are not inert: _int32_param and _list_multipart_uploads read them,
# `max-uploads` and `key-marker` come back echoed, and `max-uploads` and `upload-id-marker` can
# each turn the 200 into an InvalidArgument. What none of them does is decide which uploads a
# caller gets, because there are never any — all they shape is an echo of the caller's own input
# on a page that is always empty. Declaring them would advertise a paging surface, a marker to
# resume from and a page size, over a listing that never has a second page.
```

backlot/store.py
```diff
@@ -1157,12 +1157,37 @@ def list_documents(
...
+
+    Normally this increments the last character. Two ranges of code point have no usable character
+    one step up, and both are reachable from the wire, where ``?prefix=``, ``?delimiter=`` and
+    ``?marker=`` take whatever a client sends. A step into the surrogate block goes over it
+    instead. A trailing run of the last code point has nothing above it at all, so the run comes
+    off and the character before it is incremented — the result is still greater than every string starting with ``s``, since they
+    all share the smaller prefix. A string that is nothing but that code point has no successor,
+    and ``None`` says so: every string with that prefix sorts at the very end, which a caller reads
+    as "no upper bound" or "nothing follows" depending on which side it is bounding. Undefined for
+    an empty string; callers guard that case."""
```
khj809: **nit, my fault.** Same cause: L1181 is 132 columns. Reflowing the whole paragraph is safer than patching the one line, so this is the docstring end to end, every line at or under 100.

```suggestion
    Normally this increments the last character. Two ranges of code point have no usable character
    one step up, and both are reachable from the wire, where ``?prefix=``, ``?delimiter=`` and
    ``?marker=`` take whatever a client sends. A step into the surrogate block goes over it
    instead. A trailing run of the last code point has nothing above it at all, so the run comes
    off and the character before it is incremented — the result is still greater than every
    string starting with ``s``, since they all share the smaller prefix. A string that is nothing
    but that code point has no successor, and ``None`` says so: every string with that prefix
    sorts at the very end, which a caller reads as "no upper bound" or "nothing follows"
    depending on which side it is bounding. Undefined for an empty string; callers guard that
    case."""
```

khj809 (approved):
Approving. Every blocker is resolved in the code, not just in the replies, and the two rounds hold up under checking.

You were right to correct me on the store test, and the correction changes what the thread was about: those two lines bound `a\ud7ff` and `a\ue000`, both encodable — U+D800 only reaches a conn under the mutation, so nothing about a surrogate failing to bind was ever asserted there. I checked the three code points before writing this. `test_the_edge_bounds_key_successor_returns_reach_sql` is the better shape than the revert I offered: reverting the surrogate skip fails it with `UnicodeEncodeError: 'utf-8' codec can't encode character '\ud800'` at `store.py:1227`, and reverting the carry fails it with `ValueError: chr() arg not in range(0x110000)` at `store.py:1190` — both at the store layer, both in one function, with the `key_successor` tests left as pure value assertions.

Also re-checked at `2575021`: `ruff check`, `ruff format --check` and `gen_docs.py --check` clean, full suite green, and the empty `start-after` gate is load-bearing — reverting it to `if start_after and not continuation:` fails `test_a_bare_bucket_get_is_list_objects_and_only_list_type_2_is_its_v2_form` on `assert None == ''`. The parse change that carries it, `None` for absent against `""` for sent empty, is a better fix than the one-line gate I suggested.

Two nits inline, both artifacts of my own suggestions being applied literally at a hunk boundary — cosmetic wrapping, invisible to CI, and not worth another round on their own. Take them or leave them.

Issue #187, opened 2026-09-11
jira: POST search/jql reads jql, maxResults and nextPageToken from the query string, where real reads only the body


khj809:
**Source and commit** — `jira`, at `50c0419`.

**How you know** — live calls to `brekkylab.atlassian.net` on 2026-09-11, and the same requests against Backlot serving the corpus below. The six `extra_param` entries in `backlot/fidelity/baseline/jira.json` already record the parameters, read off Atlassian's v2 and v3 documents; what was not measured until now, and what the notes get wrong, is that real does not refuse these — it ignores them.

## 1. POST `search/jql` reads `jql`, `maxResults` and `nextPageToken` from the query string, where real reads only the body

**Request** — `POST /rest/api/{2,3}/search/jql`, parameters in the query string instead of the `SearchAndReconcileRequestBean` body.

| request | real | Backlot |
|---|---|---|
| `nextPageToken` in the body | page 2 | page 2 |
| `?nextPageToken=<page 2 token>`, body `{"jql": …, "maxResults": 1}` | **page 1 again** | page 2 |
| `?jql=…&maxResults=1`, body `{}` | **400**, unbounded JQL | 200, filtered |
| `?maxResults=1`, body `{"jql": …}` | **ignored**, all 4 issues | 1 issue |

**Backlot serves** — [`jira_search`](https://github.com/brekkylab/backlot/blob/50c0419/backlot/routers/atlassian.py#L373) builds `params` from `request.query_params` and lets the body update it, so on POST either placement works and the query string wins where the body is silent.

**The real API serves** — the query string is not read on POST at all. Both documents put these three in the request body for the POST form and in the query string for the GET form, and the live service follows the documents.

**Can a client tell?** No, and the `nextPageToken` row is the expensive one: a pager that puts its cursor in the query string walks the corpus against Backlot and re-reads page one forever against real Jira, with a 200 every time.

**Reproduction** — any corpus with two Jira issues in one project; `backlot serve`, then the four requests above.

**Closing it** — read the query string on GET and the body on POST. That drops all six `extra_param` entries from the Jira baseline.

khj809:
Fixed in #213, which splits `search/jql`'s GET query-string and POST body reading and adds the media-type, body-parse, page-token and no-`jql` refusals real gives.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #206, opened 2026-09-14
atlassian: a query parameter is converted, bounded and de-duplicated the way each product's own binder does, and a conversion failure carries that product's envelope (#180)

### LINKED ISSUE #180: atlassian: an integer query parameter is swallowed to its default when malformed, accepted when out of Java's int range, and read last-wins when repeated
**Source and commit** — `jira` and `confluence`, which share the helper, at `4c4a268`.

**How you know** — live calls to `brekkylab.atlassian.net` on 2026-09-09, 2026-09-10 and 2026-09-14, and the same requests against Backlot. All four divergences are in how a query parameter is *read*, so an issue with no comments and a site with one space settle them.

| # | Divergence | Silent? |
|---|---|---|
| 1 | a non-integer `maxResults` / `startAt` / `limit` / `start` is read as the default instead of refused | yes |
| 2 | an integer outside its parameter's Java width is accepted, where real refuses it | yes |
| 3 | a repeated parameter reads the last value, where real reads the first one (integers) or the comma-join (strings) | yes |
| 4 | Confluence accepts a negative `limit` / `start`, where real refuses it | yes |

Closing any of these means a second error envelope, because the two products answer a type-conversion failure differently and neither is the envelope `backlot/errors/atlassian.py::_body` emits today. That is why none of this is folded into #163.

```
GET /rest/api/3/issue/SCRUM-1/comment?maxResults=abc
  real   400  application/problem+json;charset=UTF-8
              {"type": "about:blank", "title": "Bad Request", "status": 400,
               "detail": "Failed to convert 'maxResults' with value: 'abc'",
               "instance": "/rest/api/3/issue/SCRUM-1/comment"}

GET /wiki/rest/api/space?limit=abc
  real   400  application/json
              {"statusCode": 400,
               "message": "org.springframework.web.method.annotation.MethodArgumentTypeMismatchException: Failed to convert value of type 'java.lang.String' to required type 'int'; nested exception is java.lang.NumberFormatException: For input string: \"abc\""}
```

Jira answers RFC 7807 with a `charset` on the media type and a short `detail`; Confluence answers its own two-key body carrying a raw Java exception string, on a bare `application/json`. Backlot answers `200` to both, and its one envelope emits five keys (`statusCode`, `message`, `reason`, `errorMessages`, `errors`) where each product sends a different subset of none of them.

## 1. A non-integer integer parameter is read as the default

**Request** — `?maxResults=abc`, `?startAt=abc` on any Jira route reading one through `_int`; `?limit=abc`, `?start=abc` on either Confluence listing.

**Backlot serves** — `200`, the first hundred comments or the first twenty-five spaces, as though the parameter had not been sent. [`_int`](https://github.com/brekkylab/backlot/blob/4c4a268/backlot/routers/atlassian.py#L1412) catches `ValueError`/`TypeError` and returns the default.

**The real API serves** — `400` on both products, in the two envelopes above, naming the parameter (Jira) or the value (Confluence) it could not convert.

**What real accepts** — an *empty* value is not a conversion failure on either product: `?maxResults=` is answered `200` with the default, which is what `_int`'s `v in (None, "")` guard already does. Beyond that the rule is narrower than "not a decimal integer", and the two products agree on every case: a leading `+` is read (`?maxResults=+3` is `3`); whitespace is *removed* rather than trimmed, so `?maxResults=%203%20` is `3` and `?maxResults=3%204` is **34**; the digits are Unicode's, so `?maxResults=%D9%A3` (ARABIC-INDIC DIGIT THREE) is `3`. `?maxResults=1.5` is `400`, and so is `?maxResults=1_0` — which is the one place Python's own `int()` disagrees with real, reading it as `10`.

**The one case the products do not share** is a value that is whitespace and nothing else. Jira reads `?maxResults=%20` as absent and answers `200` with the default; Confluence trims the same value to the empty string and fails to convert *that*, reporting `For input string: ""`. Backlot answers the default on both, so it matches Jira here and diverges from Confluence. The two also render the value they name differently when a conversion does fail: Jira's `detail` keeps it as it arrived (`?maxResults=%20abc%20` names `' abc '`), Confluence's `message` names the trimmed one (`"abc"`).

**Can a client tell?** No. A client that computes a page size and sends a malformed one — a float, a value that was `None` upstream — is silently served a different page size than it asked for, and finds out only against real.

**Scope** — router-wide, not one endpoint. `_int` reads `maxResults` on `search/jql`, `limit`/`start` on both Confluence listings, and `startAt`/`maxResults` on the comment read.

## 2. An integer outside its parameter's Java width is accepted

**Request** — `?maxResults=2147483648`, `?maxResults=-2147483649`, `?limit=2147483648`, `?startAt=9223372036854775808`.

**Backlot serves** — `200`. Python's `int` is unbounded, so the value converts and flows into the query.

**The real API serves** — `400`, the same conversion failure as #1 and with the same body, on both products.

**The width is per parameter, not per product**, and each was measured on its own rather than read off its neighbour. Jira's `maxResults` and both Confluence parameters are a Java `int`: `2147483647` and `-2147483648` are answered `200` and the values either side of them are refused. Jira's `startAt` is a Java `long` — `2147483648` and `9223372036854775807` are echoed back unchanged, and only `9223372036854775808` and `-9223372036854775809` are refused. On `search/jql`, `maxResults` is an `int` again, and the boundary is visible twice over: `2147483648` is the conversion failure while `2147483647` is a *different* `400`, in the plain envelope, refusing the page size on its own terms.

**Can a client tell?** No, and the value need not be absurd to reach this: a client that computes a page size from a timestamp or a byte count crosses 2^31 without looking unreasonable.

## 3. A repeated query parameter reads the last value

**Request** — `?startAt=3&startAt=5`, `?maxResults=1&maxResults=2`, `?limit=1&limit=2`, `?orderBy=bogus&orderBy=created`.

**Backlot serves** — the LAST value, for every parameter: Starlette's `QueryParams.get` returns it. So `startAt` is `5`, `maxResults` is `2`, and `orderBy` is `created`, which is valid, so the request is answered `200`.

**The real API serves** — it depends on the parameter's type, and the two halves disagree:

| request | real |
|---|---|
| `?startAt=5&startAt=abc` | `200`, `startAt: 5` |
| `?startAt=abc&startAt=5` | `400`, `detail: "Failed to convert 'startAt' with value: '[Ljava.lang.String;@5edeadf5'"` |
| `?startAt=3&startAt=5` | `200`, `startAt: 3` |
| `?limit=1&limit=2` (Confluence) | `200`, `limit: 1` |
| `?limit=abc&limit=2` (Confluence) | `400`, `For input string: "abc,2"`, converting a `java.lang.String[]` |
| `?orderBy=bogus&orderBy=created` | `400`, the message naming `bogus,created` |
| `?orderBy=created&orderBy=bogus` | `400`, the message naming `created,bogus` |
| `?jql=<a>&jql=<b>` | `400`, a JQL parse error at the character the comma lands on |

An integer parameter takes the FIRST value and ignores the rest; a string parameter is comma-joined and validated as one string. The two refusal messages show the same thing from the other side: when the first of the repeated values fails to convert, the value real reports is the whole array — rendered by Confluence as the comma-join and by Jira as a Java array's `toString`, whose trailing identity hash differs per request and is not reproducible.

**Can a client tell?** No. A client that builds a URL by appending rather than replacing — a retry layer adding `startAt` to a URL that already has one is the ordinary way this happens — is served the page real would have served for the *other* value, or a `200` where real refuses.

**Scope** — every `params.get(...)` in the Atlassian router: `jql`, `orderBy`, `expand`, `cql`, `spaceKey`, and every parameter `_int` reads. The same last-wins pattern is in the other routers, and whether each vendor shares Spring's rule is not measured, so this issue claims only the Atlassian ones.

## 4. Confluence accepts a negative `limit` or `start`

**Request** — `?limit=-1` or `?start=-1` on either Confluence listing.

**Backlot serves** — `200`. `_int(…, 25)` returns `-1` unclamped and it reaches the store, where SQLite reads a negative `LIMIT` as no limit at all — so the answer is the whole collection, which is the opposite of what the parameter asked for.

**The real API serves** — `400`, `{"statusCode": 400, "message": "java.lang.IllegalArgumentException: limit cannot be less than zero"}`, and the same naming `start` for that one. Zero is fine on both: `?limit=0` is answered `200` with an empty page.

**Not the same on Jira**, which is why this is its own row: Jira CLAMPS a negative instead of refusing it. `?startAt=-5` is answered `200` with `startAt: 0`, and `?maxResults=-1` and `?maxResults=0` are both answered `200` with `maxResults: 1`. Backlot's Jira comment route already clamps to the same floors; only Confluence is wrong.

**Reproduction** — any corpus with a Jira issue and a Confluence page; `backlot serve`, then the requests above against Backlot instead of Atlassian.

## Not in this issue

The `orderBy` value grammar itself — one leading `+`/`-`, whitespace ignored, field matched case-insensitively — is measured and served, see #165.

The POST form of `search/jql` reads these parameters out of a JSON body, where real uses a different parser again with its own rules: `{"maxResults": "abc"}` is `400 {"errorMessages": ["Invalid request payload. Refer to the REST API documentation and try again."]}` on a bare `application/json`, and `{"maxResults": 1.5}` is *accepted* and truncated to `1`, where the query string refuses the same value. That belongs with #187, which separates the body from the query string in the first place; this issue is about the query string only.

khj809:
Closes #180.

A query parameter on an Atlassian route is read the way the product serving it reads one: converted under that product's own rules, bounded to the Java width its own parameter has, and taken from the first of repeated values rather than the last. A value that will not convert is refused with the body and media type real answers, which is a different body per product.

## What each product does, measured

Live calls to `brekkylab.atlassian.net` on 2026-09-14 and 2026-09-15, on Jira's comment read and on Confluence's `content` listing, which are the two routes these rules apply to. None of these is what Backlot did.

**Conversion.** An empty value reads as absent and takes the default. A leading `+` is read. Whitespace is *removed* rather than trimmed, so `?maxResults=3%204` is thirty-four. The digits are Unicode's — `?maxResults=%D9%A3` is `3`. `1.5` is refused, and so is `1_0`. Python's own `int()` disagrees with real in three places, all handled here: it reads `1_0` as `10`, it refuses the internal whitespace real removes, and `str.isspace()` counts the three non-breaking spaces (U+00A0, U+2007, U+202F) as whitespace where Java's `Character.isWhitespace` excludes them — a value holding one is a 400 on both products wherever it sits, and `int()` would strip a leading or trailing one itself.

The products part on a value that cleans away to nothing, and they part twice. Jira reads `?maxResults=%20` as absent and answers 200 with the default; Confluence cleans the same value to the empty string and fails to convert *that*, reporting `For input string: ""`. And a first value that is empty is the default on Jira (`?startAt=&startAt=5`) but, on Confluence, only when the parameter does not repeat: `?limit=&limit=5` is a 400 naming `",5"`, the array with an empty element in front, where a lone `?limit=` is the default.

They also render the value they name differently. Jira's `detail` keeps it as it arrived — `?maxResults=%20abc%20` names `' abc '`. Confluence's `message` names it with the whitespace removed, the same cleaning it converts by, so `?limit=a%20b` names `"ab"`; and the cleaning is per value with the join after it, which is the only order that gives `?limit=%20abc%20&limit=2` the measured `"abc,2"` rather than `"abc ,2"`.

**Width, per parameter rather than per product.** Jira's `maxResults` and both Confluence parameters are a Java `int`; Jira's `startAt` is a Java `long`, echoing `9223372036854775807` back unchanged and refusing only past that. On `search/jql` the boundary is visible twice over: `2147483648` is the conversion failure, while `2147483647` is a *different* 400 in the plain envelope, refusing the page size on its own terms.

**Repetition.** An integer parameter takes the first value and ignores the rest; a string parameter is comma-joined and validated as one string. Starlette's `QueryParams.get` returns the last, so a client that appends to a URL rather than replacing in it — a retry layer adding `startAt` to a URL that already carries one — was served real's *other* page under a 200, and `?orderBy=bogus&orderBy=created` was answered 200 where real refuses it in either order.

**Negatives, and which refusal wins.** Confluence refuses a negative `limit` or `start`; Jira clamps one to its floor and answers 200. Unclamped, `-1` reached SQLite, which reads a negative `LIMIT` as no limit at all, so `?limit=-1` was answered with the whole collection. When both parameters are wrong at once the order is measured: conversion runs for both before either negative check, so `?limit=-1&start=abc` is the conversion failure about `abc`; and between two negatives it is `start` that real names.

## Two envelopes

Jira answers a conversion failure as RFC 7807 on `application/problem+json;charset=UTF-8`; Confluence answers two keys carrying the Spring exception's own `toString` on a bare `application/json`. Neither is the five-key body `errors/atlassian.py::_body` emits, and that body stays where it is for every other refusal — these carry their own on an `AtlassianError`, the way a `GoogleError` already carries its own fields, and `http_body` serves it as it stands. The media type rides on the exception rather than through `errors.json_media_type`, because Jira's plain 400s and its problem+json 400 reach that hook with the same path and the same status.

What real reports for a repeated parameter it cannot convert is the whole array rather than the one value it tried: Confluence renders it as the comma-join, Jira as a Java array's `toString`, whose trailing identity hash differs per request. The shape is reproduced with a Python `id` standing in — equally arbitrary, and not a promise.

## POST `search/jql` keeps the read it has

These rules are the GET path's. `POST search/jql` still merges the query string under the body and reads both leniently through `_int`, because real does not read the query string on that method **at all**: `POST /rest/api/3/search/jql?maxResults=abc` is a 200 whatever the body says. Applying the GET rules there would answer 400 where real answers 200 — one divergence traded for another. The body has a third parser again, Jackson, which takes `{"maxResults": 1.5}` as `1` where the query string refuses it and answers `{"maxResults": "abc"}` with a body-wide message naming no parameter. #187 is where the query string stops being read on POST, and where the body's own rules get measured.

## Verification

Beyond the suite, a 170-case matrix was driven at real Atlassian and at Backlot and the two compared on how each *read* the parameter — status, media type, error body, and the echoed `startAt`/`maxResults`/`limit`/`start` on a 200, never the payload, since the two serve different corpora. All 170 identical.

The matrix runs `GET /rest/api/3/issue/{key}/comment` and `GET /wiki/rest/api/content`, which is where the rules apply; `search/jql` and the CQL search are not in it, and their own reads are described above and in #212 and #216. Per parameter it covers: empty, whitespace-only, a leading `+` both raw and percent-encoded, surrounding and internal whitespace, each of the three non-breaking spaces and three of the characters Java does remove, in each of four placements on each parameter, a Unicode digit, an underscore, a float, each side of `2**31` and `2**63`, negative, zero, the parameter repeated in both orders and with an empty or unconvertible element in front, both page parameters wrong at once, and a bad integer beside a bad `orderBy` in both orders.

`backlot diff --source confluence` is clean. `--source jira` reports 4 new `missing_operation` gaps on `forge/panel/action/bulk/status` and `workflows/copy` — vendor-spec drift on routes this branch does not touch, left unacknowledged rather than folded into this baseline.

## Noticed, not fixed

Three divergences turned up alongside this work and are filed rather than folded in.

#207 and #219 — four Confluence listings read neither `limit` nor `start`: `space`, and `child/page`, `child/comment` and `label`, which bind `limit` as an int on real and page. They are the listings these rules do not reach, left out deliberately — refusing a malformed `limit` on a route that then ignores a valid one is half a reproduction, and closing them means paging the routes.

#208 — real Jira sends `application/json;charset=UTF-8` on every response where Backlot sends a bare `application/json`, and Confluence does not share it. It is the class `errors.json_media_type` exists for, but it is every Jira response rather than these parameters.

#212 — `search/jql` serves a `maxResults` outside real's 1–5000 rather than refusing it, so `?maxResults=-1` still answers the whole collection at this head. The "Width" section above describes real's second refusal on that route without Backlot reproducing it; that reproduction is #212's, not this PR's, which changes only how the value is read.

Three more were filed from the review of this PR and are not for it either: #216 (the CQL search's bodiless 404, which is why that route keeps its own read), #217 (`limit` capped at 1000 and `start` above 100000 refused on `content`), and #218 (`spaceKey` and `type` naming nothing real knows).


backlot/routers/atlassian.py
```diff
@@ -473,9 +486,11 @@ async def jira_issue_comments(key: str, request: Request):
             status_code=404,
             detail="Issue does not exist or you do not have permission to see it.",
         )
-    start = max(0, _int(params.get("startAt"), 0))
+    # `startAt` is the one parameter measured to take a Java LONG rather than an int.
+    start = max(0, _int_param(request, "startAt", 0, width=JAVA_LONG))
```
nuri-yoo, line 490: Measured: the conversion failure outranks the 404 and the `orderBy` check. `GET /rest/api/3/issue/NOPE-99999/comment?maxResults=abc` is 400 `application/problem+json` naming `maxResults` where the same key with no parameter is 404; `?startAt=abc` the same; `?maxResults=abc&orderBy=bogus` is the problem+json about `maxResults`, not the `orderBy` envelope; and with no credential at all `?maxResults=abc` is still the 400. Backlot at `0c7e4b0` answers 404 to the first two and the `orderBy` envelope to the third in either order.

That is the ordering the comment above this block already argues for `orderBy` ("BEFORE the issue is resolved"), and Spring's binder fires before the handler does, so it holds for the integer parameters at least as strongly. Reading `startAt`/`maxResults` beside `raw_order`, before `auth.conn`, reproduces all four measurements; a test with an unknown key and `?maxResults=abc` expecting 400 pins it.

backlot/errors/atlassian.py
```diff
@@ -42,6 +53,86 @@ def owns(path: str) -> bool:
...
+    a REPEATED parameter is the whole array rather than the one value it tried: Confluence renders
+    it as the comma-join, Jira as a Java array's ``toString``. The identity hash in Jira's differs
+    per request on real, so a Python ``id`` stands in — equally arbitrary, equally not a promise.
+
+    The products also disagree on whether the value they name is the one that ARRIVED: Jira echoes
+    it whitespace and all (`?maxResults=%20abc%20` names `' abc '`), Confluence names the trimmed
+    one (the same value names `"abc"`, and a whitespace-only value names `""`).
+    """
+    if path.startswith(WIKI):
+        shown = ",".join(values).strip()
```
nuri-yoo, line 95: Measured on `/wiki/rest/api/content`: `?limit=%20abc%20&limit=2` is `For input string: "abc,2"`, `?limit=abc&limit=%202%20` is `"abc,2"`, `?limit=%20&limit=2` is `",2"`. This line gives `"abc ,2"` and `"abc, 2"` for the first two, because the strip runs on the join rather than on each value.

The single-value rule is also not a trim: `?limit=a%20b` is `For input string: "ab"`, so Confluence removes whitespace inside the value and names what is left, the same removal `_int_param` does before converting. One cleaning applied per value and then joined matches every case measured, including the whitespace-only `""`. The docstring's "names the trimmed one" would then read "names the value with its whitespace removed". The whitespace set has to be Java's, see the comment on `_int_param`.

backlot/routers/atlassian.py
```diff
@@ -1410,7 +1423,106 @@ def _confluence_page(conn, request: Request, row, expand: str) -> dict:
...
+    The FIRST value when the parameter repeats, not the last: `?startAt=3&startAt=5` is 3 on both
+    products, where Starlette's ``QueryParams.get`` returns 5. A client that appends to a URL
+    rather than replacing in it — a retry layer adding `startAt` to a URL that already carries one
+    is the ordinary way — was being served real's other page under a 200.
+    """
+    values = request.query_params.getlist(name)
+    if not values or values[0] == "":
+        return default
+    raw = values[0]
+    cleaned = "".join(raw.split())
```
nuri-yoo, line 1483: Measured: `?maxResults=3%C2%A04` (U+00A0, no-break space) is 400 on both products, Jira naming `'3\xa04'` and Confluence `"3\xa04"`, where `3%E2%80%834` (U+2003), `3%0A4` and `3%094` are 34 on both. Backlot answers 34 to the no-break space too.

`str.split()` treats U+00A0 as whitespace; Java's `Character.isWhitespace` is documented to exclude the three non-breaking spaces (U+00A0, U+2007, U+202F) and to include the other Unicode space separators, which is what the four measurements show. Something like

```python
cleaned = "".join(ch for ch in raw if not ch.isspace() or ch in "\xa0\u2007\u202f")
```

reproduces it, and the same filter is what `integer_conversion_failure` should apply per value (comment there). A parametrize case for the no-break space on each product pins it; today nothing in the suite distinguishes the two whitespace sets.

backlot/routers/atlassian.py
```diff
@@ -1410,7 +1423,106 @@ def _confluence_page(conn, request: Request, row, expand: str) -> dict:
...
+) -> int:
+    """One query parameter as an int, refused the way real refuses it.
+
+    The FIRST value when the parameter repeats, not the last: `?startAt=3&startAt=5` is 3 on both
+    products, where Starlette's ``QueryParams.get`` returns 5. A client that appends to a URL
+    rather than replacing in it — a retry layer adding `startAt` to a URL that already carries one
+    is the ordinary way — was being served real's other page under a 200.
+    """
+    values = request.query_params.getlist(name)
+    if not values or values[0] == "":
```
nuri-yoo, line 1480: Measured: the products part here as well. Jira `?startAt=&startAt=5` is 200 with the default (`startAt: 0`), as the test below asserts. Confluence `?limit=&limit=5` is 400, `For input string: ",5"`, the array shape with an empty first element. Only a lone `?limit=` is the default on Confluence, which is the case the PR measured.

So the early return is right for Jira and for a single empty value, and wrong for Confluence when the parameter repeats. Gating it on `len(values) == 1` for the wiki prefix (or letting the wiki side fall through to the conversion, which then names `",5"` through the per-value cleaning) matches all three. The mixed repetitions the tests assert for Jira (`5&abc` → 5, `&5` → default) I re-measured and they hold.

backlot/routers/atlassian.py
```diff
@@ -1058,8 +1073,7 @@ async def confluence_cql_search(request: Request):
     want_type = mt.group(1) if mt else None
     ml = re.search(r'label\s*(?:=|in)\s*"?([^")\s]+)"?', cql)
     want_label = ml.group(1) if ml else None
-    limit = _int(request.query_params.get("limit"), 25)
-    start = _int(request.query_params.get("start"), 0)
+    limit, start = _confluence_page_params(request)
```
nuri-yoo, line 1076: This route is not bound the way `content` is, and the measurement in the body ("both Confluence listings that page") holds for the negatives only. On `/wiki/rest/api/search?cql=type%3Dpage&…`: `limit=abc`, `limit=1.5`, `limit=1_0`, `limit=2147483648` and `start=abc` are all **404 with `content-length: 0` and no `content-type`**, the status JAX-RS specifies for a `@QueryParam` that fails to convert; `limit=%20` and `limit=` are 200 with the default; `limit=-1` and `start=-1` are the same 400 `IllegalArgumentException` body as `content`, which the branch reproduces. `limit=2147483647` is echoed back where `content` caps at 1000.

Before this PR the route answered 200 with the default to `abc`; at `0c7e4b0` it answers `content`'s Spring 400. Filed as #216 with the responses. Whether it is taken here or left on its previous read with a pointer is your call; the negative check is right for both routes and should stay shared either way.

backlot/routers/atlassian.py
```diff
@@ -374,21 +374,35 @@ async def jira_search(request: Request):
...
-            params.update(await request.json())
+            parsed = await request.json()
         except Exception:
-            pass
-    jql = str(params.get("jql", ""))
+            parsed = None
+        if isinstance(parsed, dict):
+            params.update(parsed)
+        jql = str(params.get("jql", ""))
+        limit = _int(params.get("maxResults"), default_size)
```
nuri-yoo, line 393: The unbounded `maxResults` is on the GET side of this `if` as well as here: at `0c7e4b0`, `GET search/jql?jql=project = payments&maxResults=-1` answers every issue the caller can see (12/12 on a 12-issue corpus), `maxResults=2147483647` the same, and this branch's `-1` in the body the same. The body's "Width" paragraph describes real's second refusal on `2147483647` without saying Backlot answers 200 to it. #212 (yours, filed after this PR opened) is exactly that gap, so the "Noticed, not fixed" section wants it as a third entry rather than the fix landing here.

backlot/routers/atlassian.py
```diff
@@ -1410,7 +1423,106 @@ def _confluence_page(conn, request: Request, row, expand: str) -> dict:
...
         return int(v) if v not in (None, "") else default
     except (ValueError, TypeError):
         return default
+
+
+# What real converts a query parameter with. Measured on brekkylab.atlassian.net, 2026-09-14,
+# against Jira's comment read and Confluence's space listing, which agree on every case:
+#
+#   ?maxResults=          the default, as though unsent -- an empty value is not a failure
+#   ?maxResults=+3        3, so a leading sign is read
```
nuri-yoo, line 1444: Nit: on the wire a raw `+` decodes to a space, so `?maxResults=+3` reaches the binder as ` 3` and reads 3 by whitespace removal, not by sign reading. Measured both: `maxResults=+3` and `maxResults=%2B3` are 3 on real. The test uses `%2B3`, which is the spelling that shows the sign is read; the example here could use it too.

backlot/routers/atlassian.py
```diff
@@ -1410,7 +1423,106 @@ def _confluence_page(conn, request: Request, row, expand: str) -> dict:
...
+    except (ValueError, TypeError):
+        raise errors_atlassian.integer_conversion_failure(request.url.path, name, values) from None
+    if not width[0] <= n <= width[1]:
+        # Python's int is unbounded, so a page size computed from a timestamp or a byte count
+        # flowed into the query where real answered 400.
+        raise errors_atlassian.integer_conversion_failure(request.url.path, name, values)
+    return n
+
+
+def _str_param(request: Request, name: str, default: str | None = None) -> str | None:
```
nuri-yoo, line 1499: The join transfers to Confluence, measured: `/wiki/rest/api/content?spaceKey=NOPE1&spaceKey=NOPE2` is 404 `No space with key : NOPE1,NOPE2`, so `spaceKey` is read as the comma-join there as well. Two things the docstring could carry: the measurement set is Jira's `orderBy` and `jql` plus Confluence's `spaceKey`; and the repeated `jql` case is reproduced only as far as the join, since real refuses it with the JQL parser's 400 (`Expecting either 'OR' or 'AND' but got ','. (line 1, character 16)`) and Backlot's `_project_from_jql` answers 200 with the project. That is the parser's leniency, not this PR's, but a reader of "is refused with a parse error" should not expect the refusal here. (A single unknown `spaceKey` is a 404 on real and an empty page here, filed as #218, separate from repetition.)

backlot/routers/atlassian.py
```diff
@@ -1410,7 +1423,106 @@ def _confluence_page(conn, request: Request, row, expand: str) -> dict:
...
+    Measured: `?orderBy=bogus&orderBy=created` is refused naming `bogus,created`, in either order,
+    and a repeated `jql` is refused with a parse error at the character the comma lands on. So a
+    repeated string is one value with a comma in it, and reading the last alone answers 200 to
+    whichever ordering happens to put a valid spelling second.
+    """
+    values = request.query_params.getlist(name)
+    return ",".join(values) if values else default
+
+
+def _confluence_page_params(request: Request) -> tuple[int, int]:
```
nuri-yoo, line 1511: Two bounds real applies on `content` after conversion belong in this helper once they are taken, so noting them here: `limit` is capped at 1000 (`?limit=1001` echoes `limit: 1000`; the CQL search echoes `1001`), and `?start=100001` is a 400 `BadRequestException: Start of this size is no longer supported…` with a `data` block the negative 400 does not carry (`100000` is 200; found by bisection; `space` and the CQL search accept it). Filed as #217, not for this PR. The `child/page`, `child/comment` and `label` listings also bind `limit` as an int (400 on `abc`) and page — #219, a sibling of #207, which makes "the one Confluence listing these rules do not reach" in the body four.

backlot/main.py
```diff
@@ -156,6 +156,15 @@ async def _http_exception_handler(request: Request, exc: StarletteHTTPException)
     body = errors.http_body(request.url.path, exc, request.query_params)
     if body is None:
         body = {"detail": exc.detail}
+    # A vendor may answer one refusal under a media type of its own — Jira's type-conversion 400 is
+    # RFC 7807 on `application/problem+json`, where its other 400s are plain JSON. The exception
+    # carries it, because the path and status this is reached by are the same for both and
+    # `errors.json_media_type` sees only those two.
+    media_type = getattr(exc, "media_type", None)
```
nuri-yoo, line 163: Nit: Starlette's `Response.__init__` keeps the class media type when `media_type` is `None` (`responses.py`, `if media_type is not None: self.media_type = media_type`), so the branch collapses to one return with `media_type=getattr(exc, "media_type", None)`. Separately, `errors/__init__.py`'s module docstring lists what an envelope module provides and is where a reader will look for where Atlassian's `problem+json` comes from; the `media_type` attribute on the exception is not mentioned there, and the `json_media_type` bullet still reads "Only GitHub has one so far".

nuri-yoo (commented):
Reviewed in a worktree at `0c7e4b0` against `main` at `ab8bae9`. Full suite 3025 passed, 1 skipped (`llama_index.readers.hubspot` not installed); `ruff check`, `ruff format --check`, `scripts/gen_docs.py --check` all exit 0; `backlot diff --source confluence` is clean and `--source jira` reports exactly the four `missing_operation` rows the body names. Every claim below that says "measured" was measured against `brekkylab.atlassian.net` on 2026-09-15 with my own account (profile language set to English), and the same request run against Backlot at `0c7e4b0`; the raw responses are kept if you want any of them.

Five are worth fixing before merge, all inline: the conversion failure has to come before the 404 and the `orderBy` check on the comment route; the Confluence repeated-value rendering strips the join rather than each value; the whitespace set is Python's where real's is Java's (no-break space is a 400 on both products); an empty first value in a repeated Confluence parameter is a 400 on real; and the CQL search route is not bound by Spring at all, so `_confluence_page_params` gives it an envelope real never sends (#216). The rest are nits and test coverage.

The body needs four edits to describe the head. "Two divergences turned up" is three with #212, which also belongs under "Noticed, not fixed" since `GET search/jql?maxResults=-1` still answers the whole collection at this head. "The one Confluence listing these rules do not reach" is four: `child/page`, `child/comment` and `label` bind `limit` as an int and page on real (#219). "`1_0`, which is the one case Python's own `int()` disagrees on" is two, the code comment already says so (internal whitespace is the other). And "All 49 identical" wants its case list stated, because `search/jql?maxResults=2147483647` and the CQL route's `limit=abc` cannot have been in it.

Two things outside the diff: `confluence_child_pages` (line 1183) is the one `expand` read still on `request.query_params.get`, so a repeated `expand` is last-wins on that route alone; and `errors/__init__.py`'s contract docstring does not mention the `media_type` attribute `backlot.main` now reads off the exception.

Measured and matching, so you need not re-measure them: the first-value rule for repeated integers including `startAt=5&startAt=abc` → 5 and `startAt=&startAt=5` → default on Jira and `limit=5&limit=abc` → 5 on Confluence; a raw `+3` and `%2B3` both 3; every shared conversion case on Confluence including both int boundaries; `search/jql` reading `maxResults=` and `maxResults=%20` as the default and ignoring `startAt=abc`, `maxresults=abc` and `MAXRESULTS=abc`; `instance` carrying the path without the query, on `/rest/api/2` as well; and the no-credential ordering, Jira 400 before anonymous and Confluence 403 before binding.

Filed alongside, none of them for this PR: #216 (CQL search's bodiless 404 and whitespace-only default), #217 (`limit` capped at 1000 on `content`/`space`, `start` above 100000 refused on `content`, `limit=0`'s self-pointing `next`), #218 (`spaceKey` naming no space is 404, `type` naming no type is 501), #219 (`child/page`, `child/comment`, `label` ignore `limit`/`start`).

backlot/errors/atlassian.py
```diff
@@ -42,6 +53,86 @@ def owns(path: str) -> bool:
...
+                "statusCode": 400,
+                "message": (
+                    "org.springframework.web.method.annotation."
+                    "MethodArgumentTypeMismatchException: Failed to convert value of type "
+                    f"'{java_type}' to required type 'int'; nested exception is "
+                    f'java.lang.NumberFormatException: For input string: "{shown}"'
+                ),
+            },
+        )
+    shown = values[0] if len(values) == 1 else f"[Ljava.lang.String;@{id(values):x}"
```
nuri-yoo, line 109: Nit: over six requests the rendering was distinct each time, so "differs per request" holds. Two smaller things: CPython's `id` is a heap address and comes out nine hex digits here (`[Ljava.lang.String;@109619f00`), where Java's `Integer.toHexString(hashCode())` is at most eight; `id(values) & 0xFFFFFFFF` gives the Java width and stops the response carrying an address.

backlot/routers/atlassian.py
```diff
@@ -1410,7 +1423,106 @@ def _confluence_page(conn, request: Request, row, expand: str) -> dict:
...
+        return default
+    raw = values[0]
+    cleaned = "".join(raw.split())
+    if cleaned == "" and not request.url.path.startswith(errors_atlassian.WIKI):
+        return default  # Jira alone reads a whitespace-only value as absent
+    try:
+        if "_" in cleaned:
+            raise ValueError(cleaned)
+        n = int(cleaned)
+    except (ValueError, TypeError):
```
nuri-yoo, line 1490: Nit: `TypeError` is unreachable here, `getlist` returns `list[str]`. And the "is this Confluence" decision now lives twice, on the `WIKI` prefix here (line 1484) and in `integer_conversion_failure`; one helper taking the path would keep the two from drifting, and #216 suggests the split is by route rather than by product anyway.

tests/test_atlassian.py
```diff
@@ -1049,3 +1049,226 @@ def test_jira_comment_paging_is_declared_so_a_client_can_discover_it(paged):
...
+
+def test_jira_search_refuses_an_integer_parameter_it_cannot_convert(client, admin_h):
+    r = client.get("/atlassian/rest/api/3/search/jql?maxResults=abc", headers=admin_h)
+    assert r.status_code == 400, r.text
+    assert r.json()["detail"] == "Failed to convert 'maxResults' with value: 'abc'"
+    assert r.json()["instance"] == "/rest/api/3/search/jql"
+
+
+@pytest.mark.parametrize("param", ["limit", "start"])
+def test_confluence_refuses_an_integer_parameter_it_cannot_convert(client, admin_h, param):
```
nuri-yoo, line 1177: The Confluence side of the shared conversion rules is unpinned. Mutation: giving both `_int_param` calls in `_confluence_page_params` `width=JAVA_LONG` leaves `tests/test_atlassian.py` at 117 passed. I re-measured the shared cases on `/wiki/rest/api/content` and they all match Backlot — `%2B3` → 3, `%203%20` → 3, `3%204` → 34, `%D9%A3` → 3, `1_0` → 400, `1.5` → 400, `2147483647` → 200, `2147483648` → 400, `-2147483649` → 400, `9223372036854775807` → 400 — so the code is right and a parametrize mirroring `test_jira_takes_the_integer_spellings_the_real_api_takes` and the width cases of `test_jira_refuses_an_integer_parameter_it_cannot_convert` on this route is all that is missing.

khj809:
Re-measured every claim against `brekkylab.atlassian.net` before touching anything; all five blockers and all four body corrections hold, and the head is `fd60f2a`. The 73-case real-vs-Backlot comparison is identical on both routes, `pytest` is 3050 passed 1 skipped, `ruff check`/`format` clean, `backlot diff --source confluence` clean.

**Taken, all five.**

The binder outranking the 404 and the sort check: `startAt` and `maxResults` are now read beside `raw_order` and before it, above `auth.conn`. Your ordering measurement extends one step further than the comment says — `?startAt=abc&maxResults=xyz` names `startAt` in *either* URL order, so the precedence is the handler signature's rather than the query string's, and `startAt` is read first for that reason. Six parametrize cases, each run against both an existing key and an unknown one.

The Confluence rendering: cleaning is now per value with the join after it. Your three cases all pass, and the single-value rule is indeed removal rather than a trim — `?limit=a%20b` names `"ab"`, which the old strip-the-join code got wrong in a second way I had not seen.

The whitespace set: Java's, via a shared `strip_java_whitespace`. I measured all three non-breaking spaces rather than just U+00A0 — U+2007 and U+202F are 400 too, and U+2003, LF and TAB are 34 — so the set is exactly the documented one. `_int_param` and the Confluence rendering share the function, so they cannot drift.

The empty first value: gated on `len(values) == 1` for Confluence, which leaves Jira's `?startAt=&startAt=5` on the default and gives Confluence's `?limit=&limit=5` the `",5"` conversion failure.

The CQL search: left on its previous lenient read with a comment pointing at #216, and the negative check stays shared — I confirmed `?limit=-1` and `?start=-1` are the same `IllegalArgumentException` 400 there. Taking the bodiless 404 here would mean reproducing a second framework's conversion contract on measurements #216 already holds, so it belongs there. This was the same mistake as the POST one: a rule measured on one route applied to a route that isn't bound by the same thing.

**Nits, all taken.** `media_type=getattr(...)` collapsed to one return — you're right that Starlette keeps the class media type on `None`. `id(values) & 0xFFFFFFFF`, so it is Java's width and not a live address. The unreachable `TypeError` is gone. The `WIKI` check is now one `errors_atlassian.is_confluence(path)` used by both sides. The comment's `+3` is `%2B3`. `_str_param`'s docstring carries `spaceKey` and the caveat that only the join is reproduced, since `_project_from_jql` is lenient where real's JQL parser is not.

**Both out-of-diff items taken.** `confluence_child_pages` now reads `expand` through `_str_param`, so no route is left last-wins. `errors/__init__.py` documents the `media_type` attribute and no longer says GitHub is the only implementer.

**Coverage.** Your mutation was right that the Confluence side was unpinned — `width=JAVA_LONG` on both reads passed. Added parametrized spellings and width cases on `content` mirroring the Jira ones, plus the whitespace-set, empty-first-value, per-value-join and CQL cases.

**Body.** All four edits made, and two more the fixes forced: "both Confluence listings that page" is now `content` alone, since the CQL route no longer shares the read; and the "trimmed" wording is "whitespace removed" throughout. #212 is a third entry under "Noticed, not fixed", #207 and #219 are one entry naming four listings, "the one case Python's `int()` disagrees on" is three (underscore, internal whitespace, the non-breaking spaces), and the 73-case list is stated along with which routes it does and does not run.

Not taken, by your own framing: #217's caps and #218 belong to their issues.

backlot/routers/atlassian.py
```diff
@@ -1410,7 +1438,131 @@ def _confluence_page(conn, request: Request, row, expand: str) -> dict:
...
+    values = request.query_params.getlist(name)
+    if not values:
+        return default
+    confluence = errors_atlassian.is_confluence(request.url.path)
+    if values[0] == "" and (len(values) == 1 or not confluence):
+        return default
+    raw = values[0]
+    cleaned = errors_atlassian.strip_java_whitespace(raw)
+    if cleaned == "" and not confluence:
+        return default  # Jira alone reads a whitespace-only value as absent
```
nuri-yoo: `strip_java_whitespace` keeps U+00A0, U+2007 and U+202F in `cleaned`, and then `int()` removes them itself wherever they lead or trail: `int('\xa03')` is 3. Only the interior spelling reaches `ValueError` — the one spelling my comment measured.

Measured again over `maxResults` and `startAt` on the comment read and `limit` and `start` on `content`, three characters by four spellings, 48 cases: real refuses all 48, naming the value with the non-breaking space kept and the ordinary whitespace removed. This head refuses the 12 interior ones and converts the other 36 — `?limit=%C2%A03` is a 200, `?maxResults=3%C2%A0` falls through to the 404 on an unknown key and to a 200 on a known one, where real is `Failed to convert 'maxResults' with value: '3\xa0'`. The control characters (U+2003, space, tab) agree on all 48 either way.

The extended snippet:

```python
cleaned = errors_atlassian.strip_java_whitespace(raw)
if cleaned == "" and not confluence:
    return default  # Jira alone reads a whitespace-only value as absent
if any(ch.isspace() for ch in cleaned):
    # Only the three non-breaking spaces survive the cleaning, and Java throws on one wherever it
    # sits, where Python's own int() strips a leading or trailing one before converting.
    raise errors_atlassian.integer_conversion_failure(request.url.path, name, values) from None
```

With it the 48 match real down to the body: Confluence names `"\xa03"`, `"3\xa0"` and `"\xa03\xa0"`, Jira echoes `'\xa03'` on `application/problem+json;charset=UTF-8`, and the suite is `3050 passed, 1 skipped`. The rendering needed no change — real names `?limit=%20%C2%A03%20` as `"\xa03"`, which is what the per-value cleaning already gives.

A spelling axis on `test_atlassian_reads_javas_whitespace_set_not_pythons` — `{s}3`, `3{s}` and `{s}3{s}` beside `3{s}4` — pins it.

backlot/routers/atlassian.py
```diff
@@ -1410,7 +1438,131 @@ def _confluence_page(conn, request: Request, row, expand: str) -> dict:
...
+#   ?maxResults=1.5       400
+#
+# Python's `int` agrees with all of it but the underscore and the internal whitespace, so those two
+# are handled here and the rest is left to `int` rather than restated as a pattern that would then
+# have to be kept in step with it.
+#
+# The products part on ONE case, a value that is whitespace and nothing else. Jira reads it as
+# absent and answers 200 with the default (`?maxResults=%20` and `?maxResults=%09` both); Confluence
+# trims it to the empty string and fails to convert THAT, reporting `For input string: ""`. A
+# genuinely empty `?limit=` is the default on both, so it is the whitespace, not the emptiness,
```
nuri-yoo: Two sentences here stayed behind the fix. `int` disagrees with real on the non-breaking spaces too, which is what `strip_java_whitespace` exists for, so "the underscore and the internal whitespace" is two of the three the body now counts. And "The products part on ONE case" predates `_int_param`'s docstring naming the second one, the empty first value of a repeated parameter on Confluence. The `trims it` two lines down is the wording the fix replaced elsewhere — for a whitespace-only value trim and removal land on the same empty string, so it reads right and describes the wrong operation.

nuri-yoo (approved):
Approving. I re-ran the checks at `fd60f2a` rather than reading the reply for them, and re-measured at `brekkylab.atlassian.net`.

What holds: the binder outranks both the 404 and the sort check — an unknown key with `?maxResults=abc` is the conversion 400 where the key alone is the 404, and `?startAt=abc&maxResults=xyz` names `startAt` in either URL order, which is the signature ordering you describe. The Confluence rendering cleans per value and joins after, `?limit=a%20b` included. `?limit=&limit=5` is the `",5"` 400 where Jira's `?startAt=&startAt=5` takes the default, and real answers `?start=&start=5` the same way, so the rule holds for the pair rather than for `limit` alone. The CQL route is back on its own read with the negative check shared, and its conversion failure is the bodiless 404 #216 describes. Locally `3050 passed, 1 skipped`, and the mutation behind the coverage note — `width=JAVA_LONG` on `_confluence_page_params` — now fails two cases where it passed before.

One thing did not survive a wider measurement, and it is my snippet that was short rather than your reading of it: the whitespace set is Java's now, but only for a non-breaking space that sits between digits. The inline comment carries the extended snippet and the 48-case grid. It is one guard line plus a spelling axis on the test, and the suite stays at `3050 passed, 1 skipped` with it, so it fits this PR or the commit right behind it — with the note that the body's and the test docstring's "a value holding one is a 400 on both products" is not true at this head until it lands.

khj809:
Taken at `5149389`, and your snippet was right — I re-measured before applying it rather than reading it off the comment, and real refuses all twelve non-breaking spellings on `maxResults`, `startAt`, `limit` and `start` alike.

The guard is the one you wrote. The diagnosis is worth restating because it is not the one the earlier comment implied: `strip_java_whitespace` was already correct: it kept the non-breaking spaces in `cleaned` exactly as Java does. What undid it was the line after, `int(cleaned)`, which strips leading and trailing Unicode whitespace of its own accord, so the character survived the cleaning and was then thrown away by the converter. Only the interior placement had no way out. That is the third distinct place Python's `int()` disagrees with Java's, after the underscore and the internal whitespace, and it is now the one the comment above `_int_param` counts.

Also taken: the two stale sentences. "the underscore and the internal whitespace" is three, and "The products part on ONE case" is two — the second being the empty first value of a repeated parameter, which `_int_param`'s docstring had already named while the block above it had not. The "trims it" wording is "cleans it" now. Thank you for catching that the right-reading sentence described the wrong operation.

Verification: the real-vs-Backlot comparison now carries the whole grid — six characters by four placements on both integer parameters of both routes, plus `?limit=%20%C2%A03%20`, which real names `"\xa03"` and the per-value cleaning already gave. 170/170 identical. Suite `3050 passed, 1 skipped`, `ruff` clean.

The body now says "wherever it sits" rather than leaving the placement open, and the verification section counts 170 with the placement axis stated, so the claim you flagged as not-yet-true is true at this head.

Issue #212, opened 2026-09-15
jira: search/jql serves a maxResults outside 1-5000, a body field it cannot coerce, and an unknown body field, where real refuses all three


khj809:
**Source and commit** — `jira`, at `a3e679d`.

**How you know** — live calls to `brekkylab.atlassian.net` on 2026-09-15, and the same requests against Backlot.

| # | Divergence | Silent? |
|---|---|---|
| 1 | `maxResults` outside 1–5000 is served, where real refuses it — and a negative one serves the WHOLE collection | yes |
| 2 | a body field of a type the vendor will not coerce is read as absent, where real refuses the request | yes |
| 3 | an unknown body field is ignored, where real refuses the request | yes |

## 1. `maxResults` outside 1–5000

**Request** — `?maxResults=0`, `?maxResults=-1`, `?maxResults=5001` on the GET form, and the same values in the POST body.

**Backlot serves** — `200` to all of them. `?maxResults=5001` serves the page. `?maxResults=0` serves an empty one. `?maxResults=-1` serves **every issue the caller can see**: unclamped, `-1` reaches SQLite, which reads a negative `LIMIT` as no limit at all — the same trap #180 closed for Confluence's `limit`, which `search/jql` was never clamped against.

**The real API serves** — `400` for `0`, `-1` and `5001`, in the classic `{errorMessages, errors}` envelope, with a sentence naming the 1–5000 range. `1` and `5000` are answered `200`, so the bounds are inclusive and measured on both sides. The rule is the same on both methods, so it belongs to the operation rather than to either parser.

This is also what makes `?maxResults=2147483647` two different refusals on real, which #180 noted without explaining: `2147483648` is the type-conversion `400` in RFC 7807, while `2147483647` converts fine and is then refused by THIS range check, in the plain envelope.

**Can a client tell?** No, and the negative case is the dangerous one: a client computing a page size that underflows asks for one page and is handed the entire corpus under a `200`.

## 2. A body field the vendor will not coerce

**Request** — `POST search/jql` with `{"maxResults": "abc"}`, `{"maxResults": true}`, or `{"jql": 123}`.

**Backlot serves** — `200`. `_int` returns the default for `"abc"`, and `int(True)` is `1`, so `{"maxResults": true}` quietly becomes a one-issue page. `{"jql": 123}` is stringified to `"123"` and searched as if it were JQL.

[...]

PR #213, opened 2026-09-15
jira: search/jql reads its parameters from the query string on GET and from the body on POST, and refuses the media type, the body and the page token the way real does (#187, #191)

### LINKED ISSUE #187: jira: POST search/jql reads jql, maxResults and nextPageToken from the query string, where real reads only the body

### LINKED ISSUE #191: jira: POST search/jql answers a request with no body, and reads a malformed nextPageToken as offset zero
**Source and commit** — `jira`, at `50c0419`.

**How you know** — live calls to `brekkylab.atlassian.net` on 2026-09-11, and the same requests against Backlot.

| # | Divergence | Silent? |
|---|---|---|
| 1 | a POST with no body is answered as a search, where real refuses the media type | no |
| 2 | a malformed `nextPageToken` is read as offset zero, where real refuses it | yes |

## 1. POST `search/jql` with no body is answered

**Request** — `POST /rest/api/3/search/jql?jql=…&maxResults=1`, no `Content-Type` and no body.

**Backlot serves** — `200` and a page of issues: [`jira_search`](https://github.com/brekkylab/backlot/blob/50c0419/backlot/routers/atlassian.py#L373) catches the `await request.json()` failure and falls through to whatever the query string held.

**The real API serves** — `415 application/problem+json`, `{"title": "Unsupported Media Type", "detail": "Content-Type 'null' is not supported.", "instance": "/rest/api/3/search/jql"}`.

## 2. A malformed `nextPageToken` is read as offset zero

**Request** — `POST /rest/api/3/search/jql`, body `{"jql": …, "maxResults": 1, "nextPageToken": "BOGUS"}`.

**Backlot serves** — `200` and page one. `pagination.decode_cursor` returns zero for anything it cannot decode.

**The real API serves** — `400`, naming the token as invalid or expired.

**Can a client tell?** Not for #2. A client that truncates, caches or otherwise corrupts a cursor is served page one under a 200 and reads the same page for as long as it keeps going.

**Reproduction** — any corpus with two Jira issues in one project; `backlot serve`, then the two requests above.

**Relation to #180** — the same family of leniency, but neither is covered there: #180 is scoped to `_int` and to repeated query parameters, and `nextPageToken` is read through `decode_cursor`, while a missing request body is not a parameter at all. Fold this in if the envelope work lands together.

khj809:
Closes #187, closes #191.

## What changed

`search/jql` takes one set of parameters in two places, and real reads exactly one of them per method: the query string on GET, a `SearchAndReconcileRequestBean` body on POST. Backlot read the query string on both and let the body override it. It also answered a POST carrying no body at all, read a page token it could not decode as offset zero, and answered a `jql` absent from both places as the unfiltered corpus.

GET and POST are one operation each in the served spec — the GET declaring the three query parameters, the POST declaring the body — so a generated client cannot keep sending a POST cursor in the query string without seeing why it never pages. That is what the six `extra_param` entries dropped from `backlot/fidelity/baseline/jira.json` were acknowledging; they are removed rather than re-acknowledged, and nothing else in that file moves.

The split is made in the document, by `openapi.jira_search_placement`, rather than by giving each method its own route: `openapi_extra` is per route, and a path served by two single-method routes can only advertise one of them in `Allow`, where real names both. One route keeps `PUT /rest/api/3/search/jql` at `Allow: POST, GET`. What that 405's BODY should be is a separate question and still Starlette's here; #245 is where the Atlassian one gets measured.

Three refusals are added: the media type before the bytes, three body-parse sentences (with the boundaries a JSON parser draws, not the ones Backlot drew before), and an undecodable page token or a missing `jql` — the token check first, ahead of the JQL's own project resolution.

### The placement, per method

Measured on `brekkylab.atlassian.net`, 2026-09-15:

| request | real | Backlot before |
|---|---|---|
| `nextPageToken` in the POST body | page 2 | page 2 |
| `?nextPageToken=<page 2>`, body without one | **page 1 again** | page 2 |
| `?maxResults=1`, body without one | **ignored**, every issue | 1 issue |
| `?jql=…&maxResults=1`, body `{}` | **400**, unbounded JQL | 200, filtered |

The `nextPageToken` row is the expensive one: a pager that puts its cursor in the query string of a POST walked the corpus here and re-reads page one forever against real, with a 200 every time.

**The media type, before the bytes.** A POST whose `Content-Type` is not `application/json` is `415 application/problem+json;charset=UTF-8` — the same five-key RFC 7807 shape #206 introduced for the conversion failure. A request carrying no header at all is named `'null'` in the `detail`, the literal string. The check is on the header alone: a perfectly good JSON body sent without one is refused identically. Matching is case-insensitive and ignores parameters, so `APPLICATION/JSON` and `application/json; charset=utf-8` are both read, while `*/*` and `application/xml` are not.

**The body, three sentences, and the boundaries are not a JSON parser's.** Backlot caught every parse failure with one `except` and fell through to whatever the query string held, which is how a POST with no body was answered as a search. Real sorts it into: nothing to read (a body of zero LENGTH, or a literal `null`); bytes that do not parse; and JSON that parses to anything but an object (`[]`, `5`, `"x"`, `true`). Two boundaries had to be measured rather than assumed — a body that is only whitespace is the not-an-object sentence rather than either of the other two, and bytes trailing a complete value are *ignored* (`{"jql": …} junk` is a 200), so the reader takes the first value and lets the rest go. Each refusal arrives as `errorMessages` alone — no `errors`, unlike the token and jql refusals below — so all three carry their own body on an `AtlassianError`.

**The page token.** An undecodable `nextPageToken` is `400` in the classic `{errorMessages, errors}` envelope, on both methods alike, and ahead of the JQL's own project resolution: measured 2026-09-16, `?jql=project+%3D+ZZZNOPE999&nextPageToken=BOGUS` (a project matching nothing, plus a bad token) still draws the 400 on real, not the unresolved project's empty `200`. Read as offset zero instead, a client that truncated or cached a cursor was served page one under a 200 for as long as it kept following it. Real localises this sentence — the account it was measured against answers in Korean — so the envelope is reproduced and the wording is Backlot's own, as it already is for the `orderBy` refusal.

**A `jql` that is null, too.** A JSON `null` is the parameter unsent: `{"jql": null}` draws the same refusal as `{}` on real, where reading it through a bare `str()` makes it the string `"None"`, which restricts nothing and reaches the whole visible corpus.

**No `jql` at all.** Measured 2026-09-16, on both methods: real refuses with `400 {"errorMessages": ["무제한 JQL 쿼리는 여기에서 허용되지 않습니다..."], "errors": {}}` ("unbounded JQL queries are not allowed here") rather than answering the unfiltered corpus — including a POST whose query string carries a `jql` the body (the only place POST reads one) omits, which is exactly the fourth row of the table above. Backlot's own crawl helper, SDK example and SDK test, which listed everything by sending no `jql` at all, now send `project is not EMPTY`. An `ORDER BY` is not a search restriction — `?jql=order by created` draws this same refusal on real (measured 2026-09-16) — so the replacement has to be a clause that restricts; this one is a 200 there and still selects everything here, because `_project_from_jql` filters only on a `project =`.

## Not in this change

`_int` stays lenient for values read out of the body, because the body's parser is Jackson and it genuinely is lenient: `{"maxResults": 1.5}` is accepted as `1` where the query string refuses the same value, and `{"maxResults": "5"}` is coerced. The rest of what Jackson does to that body — `maxResults` has a measured 1–5000 range, unknown fields are rejected, `"abc"` and `true` are refused where `"5"` is not — is field-level validation rather than the structural parse this change reproduces, and it is filed as #212 rather than half-implemented here. That issue also covers a `maxResults` of `-1` on this endpoint, which SQLite reads as no limit at all and answers with the whole collection.

## Verification

- **Tests** — `pytest` on a `.[all]` install: 3080 passed, 1 skipped. The one skip is `llama_index.readers.hubspot`, which is not installed.
- **Optional surfaces that ran rather than skipped** — `.[mcp]`, `.[official-sdk]` and the rest of `.[all]` were installed and ran.
- **Lint** — `ruff check . && ruff format --check .`: clean. `scripts/gen_docs.py --check`: clean.
- **Fidelity** — `backlot diff --source jira` reports the same four `missing_operation` gaps as `main`; `--source confluence` is clean.

The new assertions were also run against the base commit with only the test file copied over: every one that asserts changed behaviour fails there and the guard tests pass, so nothing in the suite is passing for the wrong reason.

The bridge was diffed against the base at both layers. Its operation LIST is unchanged — `name_operations` collapses the two methods back to `jira_search` before `dedupe_operations` picks the GET. The TOOL is not: built through FastMCP 4.0.3 over the served spec, `jira_search` goes from six arguments to three. `jql`, `maxResults` and `nextPageToken` each appeared twice, once from the GET's `requestBody` and once suffixed `__query` from its parameters, and the first three were a JSON body FastMCP would put on a GET that the handler never reads. Declaring each placement on the method that reads it deletes them.

`backlot diff --source jira` reports the same 4 `missing_operation` gaps as #206 — vendor-spec drift on `forge/panel/action/bulk/status` and `workflows/copy`, routes neither branch touches — and no longer reports the 6 resolved entries. `--source confluence` is clean.

- [x] A new endpoint serving corpus content is ACL-scoped, proved for both an admin and a scoped token (n/a — no new endpoint; `jira_search` still computes `visible_ids` unchanged)


backlot/routers/atlassian.py
```diff
@@ -70,8 +71,12 @@ class ConfluencePage(_ALoose):
...
 
 
-_X_JIRA_SEARCH = {
-    "parameters": [qp("jql"), qp("maxResults", "integer"), qp("nextPageToken")],
+# One operation per METHOD, because the two take the same three parameters in different places:
+# the GET form in the query string, the POST form in a `SearchAndReconcileRequestBean` body. Both
+# of Atlassian's documents split them this way, down to the schema name, and the live service
+# follows.
+_X_JIRA_SEARCH_GET = {"parameters": [qp("jql"), qp("maxResults", "integer"), qp("nextPageToken")]}
+_X_JIRA_SEARCH_POST = {
```
nuri-yoo: The served spec says this body is optional and the handler does not: a POST with no body is the 415, because the missing header refuses it first, and one that carries the header with an empty body is `No content to map to Object due to end of input`. Since the reason for splitting the operations is that a generated client can see where each parameter goes, `"required": True` beside `content` is the other half of that.

backlot/routers/atlassian.py
```diff
@@ -360,49 +365,57 @@ async def jira_project_role(key: str, role_id: int, request: Request):
...
         jql = _str_param(request, "jql") or ""
         limit = _int_param(request, "maxResults", default_size)
         token = _str_param(request, "nextPageToken")
+    # An undecodable token is refused before the project clause is resolved (measured
+    # 2026-09-16; see test_jira_search_refuses_a_page_token_it_cannot_decode).
+    offset = decode_cursor_or_none(None if token is None else str(token))
+    if offset is None:
+        raise errors_atlassian.bad_page_token()
+    # No `jql` at all is refused rather than answered as the unfiltered corpus (see
+    # test_jira_search_refuses_no_jql_at_all).
```
nuri-yoo: Three lines up the token comment carries `(measured 2026-09-16; ...)` and this one carries no date, though both facts were measured the same day and the same way. Naming the test is fine and has precedent (`backlot/store.py:103` does it beside the rule it proves); it is the date that the checklist line about comments stating what was measured is asking for.

backlot/routers/atlassian.py
```diff
@@ -1530,6 +1543,49 @@ def _int_param(
...
+    first value and letting the rest go.
+    """
+    content_type = request.headers.get("content-type")
+    if (content_type or "").split(";")[0].strip().lower() != "application/json":
+        raise errors_atlassian.unsupported_media_type(request.url.path, content_type)
+    raw = await request.body()
+    if not raw:
+        raise errors_atlassian.body_not_read(errors_atlassian.BODY_EMPTY)
+    try:
+        parsed, _end = json.JSONDecoder().raw_decode(raw.decode("utf-8", "replace").lstrip())
```
nuri-yoo: Two boundaries here are wider than the parser this reproduces. Both measured against the site today:

```
POST  " " + {"jql": ...}   real 400 There was an error parsing JSON. Check that your request body is valid.    here 200
POST  {"jql": "...\xff"}        real 400 Invalid request payload. Refer to the REST API documentation and try again. here 200
```

`str.lstrip()` strips every Unicode space where JSON's whitespace is the four ASCII ones, so a NBSP in front of the object is skipped here and refused there, and `decode("utf-8", "replace")` repairs malformed bytes into U+FFFD, so a body real will not decode parses. `lstrip(" \t\n\r")` and a strict decode whose `UnicodeDecodeError` maps to `BODY_NOT_AN_OBJECT`, which is what real answers for those bytes, put both back inside the three sentences above.

tests/test_atlassian.py
```diff
@@ -1385,15 +1392,219 @@ def test_confluence_refuses_the_parameter_real_names_when_both_are_wrong(
...
+        (None, "null"),
+        ("text/plain", "text/plain"),
+        ("application/xml", "application/xml"),
+        ("*/*", "*/*"),
+    ],
+)
+def test_jira_search_post_refuses_a_media_type_it_does_not_read(
+    client, admin_h, content_type, named
+):
+    """Measured: each content type below is refused or read as shown."""
```
nuri-yoo: The table this points at is the `parametrize` above the function rather than below it, and nothing in this test is read: all four cases are the 415, and the read cases are the next test's. The sentence 88c9755 replaced stated a measurement ("the header decides before the bytes are looked at"); this one states less than the assertions already do.

tests/_helpers.py
```diff
@@ -432,7 +432,9 @@ def crawl_github_repo(client, headers, org, repo):
 def crawl_jira(client, headers):
     out, token = [], None
     while True:
-        p = {"maxResults": 6}
+        # real refuses an empty `jql` outright, so a full crawl needs a jql that restricts
+        # nothing -- any project-free, text-free clause does, `order by created` included.
```
nuri-yoo: Measured on the site today, on both methods: `order by created` is `400 Unbounded JQL queries are not allowed here. Please add a search restriction to your query.`, the same refusal as no `jql` at all, because an `ORDER BY` is not a search restriction. So this comment states a vendor rule that real contradicts, and the crawl below it is a request only Backlot answers.

`project is not EMPTY`, `created >= "1970-01-01"` and `created > -5200w` are each a 200 there, and each leaves the crawl unfiltered here for the same reason this one does: `_project_from_jql` looks for a `project =`.

examples/using-official-sdk/jira.py
```diff
@@ -82,7 +82,9 @@
         print(f"authenticating as {username} → responses are ACL-filtered to that user")
     jira = Jira(url=f"{s.base_url}/atlassian", username=username, password=password)
 
-    issues = jira.get("rest/api/3/search/jql", params={"maxResults": 5})["issues"]
+    issues = jira.get("rest/api/3/search/jql", params={"jql": "order by created", "maxResults": 5})[
```
nuri-yoo: Same query as the crawl helper, and this is the copy that leaves the repo: run against a real site it is a 400 where Backlot serves a page, which is the shape of divergence the rest of this change removes. `{"jql": "project is not EMPTY", "maxResults": 5}` is a 200 on both, checked through this same `jira.get` call.

tests/test_sdk.py
```diff
@@ -450,7 +450,7 @@ def jira():
     from atlassian import Jira
 
     j = Jira(url=f"{BASE}/atlassian", username="svc@x", password=ADMIN)
-    res = j.get("rest/api/3/search/jql", params={"maxResults": 50})
+    res = j.get("rest/api/3/search/jql", params={"jql": "order by created", "maxResults": 50})
```
nuri-yoo: Same substitution as the example and the crawl helper. Worth keeping all three on one query so the SDK check demonstrates what the example tells a reader to run.

nuri-yoo (changes requested):
Reviewed at `11d6748` against `main` at `612f888`. I re-measured the claims in the body against `brekkylab.atlassian.net` today (2026-09-16) with that site's own credentials, ran the suite and the gates here, and diffed the bridged spec against the base.

The structural work reproduces real everywhere I could check it: each of the eight media types I sent, all three body sentences, the trailing-bytes rule, the token refusal on both methods and ahead of project resolution, and each of the four rows in the placement table. Four more token shapes agree as well (an empty one and a `null` are page one; a repeated one and a numeric one are the 400). `backlot diff --source jira` reports only the four `missing_operation` gaps #228 is taking. The full suite here is 3076 passed, 1 skipped (`llama_index.readers.hubspot` is not installed), `ruff check`, `ruff format --check` and `scripts/gen_docs.py --check` are clean, and the two `tests/test_mcp.py` failures the body reports do not reproduce on this machine.

Four things want a change before this lands.

## 1. `order by created` is refused by real, so the crawl helper, the SDK example and `test_sdk.py` now send a query only Backlot answers

Measured today, both methods:

```
GET  ?jql=order by created            400  Unbounded JQL queries are not allowed here. Please add a search restriction to your query.
POST body {"jql":"order by created"}  400  the same sentence, the same envelope
GET  ?jql=ORDER BY created DESC       400
```

An `ORDER BY` is not a search restriction, so the replacement draws the identical refusal this change exists to reproduce. `examples/using-official-sdk/jira.py` is the one that costs something: a reader copies it, points it at their own site, and gets a 400 where Backlot served a page.

Three clauses do restrict and still select everything. Each is a 200 on real and a drop-in here, checked through the same official-SDK call on a two-project corpus, which answers with both issues exactly as today:

```
project is not EMPTY
created >= "1970-01-01"
created > -5200w
```

`project is not EMPTY` reads closest to what the call means, and `_project_from_jql` needs a `project =` before it filters, so the crawl stays unfiltered.

## 2. `{"jql": null}` is answered with the unfiltered corpus

`str(body.get("jql", ""))` turns a JSON null into the string `"None"`, which passes the new `if not jql.strip()`, matches no project and no text clause, and reaches the whole visible corpus.

```
real   POST {"jql": null}   400  Unbounded JQL queries are not allowed here...
here   POST {"jql": null}   200  every issue the caller can see
```

That is the hole this change closes, reachable with one null. The two neighbouring shapes answer differently on real, so they can be told apart rather than folded together: `{"jql": 5}` is coerced to `"5"` and then refused by the JQL parser, and `{"jql": ["order by created"]}` is `Invalid request payload. Refer to the REST API documentation and try again.`, which is the sentence `BODY_NOT_AN_OBJECT` already carries. The null belongs here; the other two are #212's field-level half and could go into its body with those measurements.

## 3. Splitting one route into two narrowed the `Allow` header

```
main   PUT /rest/api/3/search/jql   405  Allow: POST, GET
here   PUT /rest/api/3/search/jql   405  Allow: POST
real   PUT /rest/api/3/search/jql   405  Allow: POST, GET
```

The same on `/rest/api/2`, and on DELETE and PATCH. Starlette fills the header from the single route that partially matched, so a path served by two single-method routes can only ever advertise one of them. The Slack GET+POST routes, untouched here, still send `POST, GET`. #233 row 3 is about this header ("real Jira sends the methods it accepts") and #245 is where the Atlassian 405 is being rewritten, so it is worth settling here whether that fix reads one route's methods or the union of the routes on a path.

## 4. The `_int` docstring attributes the 1-5000 range to Jackson

`backlot/routers/atlassian.py:1459` says "What Jackson refuses (`"abc"`, `true`, a range outside 1-5000) is a body-wide message naming no parameter". Measured today, the range is none of those three things:

```
body maxResults "abc" / true    400  {"errorMessages":["Invalid request payload. Refer to the REST API documentation and try again."]}
body maxResults 0 / -1 / 5001   400  {"errorMessages":["The max results parameter has to be between 1 and 5,000."],"errors":{}}
GET  ?maxResults=0 / -1 / 5001  400  the same body, to the character
```

It is refused on the query string too, so it is the operation's own check rather than either parser's; it names the parameter; and it carries `errors`, where the two type refusals carry `errorMessages` alone. #212 already draws that line ("The rule is the same on both methods, so it belongs to the operation rather than to either parser"), so the sentence here contradicts the issue it defers to. The rest of that docstring holds: `1.5` really does arrive as `1` (one issue and a token), `"5"` as `5`, and the query string answers `Failed to convert 'maxResults' with value: '1.5'`.

## One claim the body understates

"The MCP bridge's operation list was diffed against the base as well ... leaves it byte-identical" is true of the operation list and not of the tool. Built through FastMCP 4.0.3 over `/_meta/openapi/atlassian`, `jira_search` goes from six arguments to three: `jql`, `maxResults` and `nextPageToken` each appeared twice, once from the GET's `requestBody` and once suffixed `__query` from its parameters, and the first three were a JSON body FastMCP would put on a GET that the handler never read. Deleting them is worth claiming rather than describing as unchanged.

Seven smaller things are inline.

## What else I checked against real, and it matches

- Media type: no header, `text/plain`, `*/*` and `application/xml` are the 415 with `'null'` and each name in `detail`; `application/json`, `APPLICATION/JSON` and both charset spellings are read.
- Body: zero length is `No content to map to Object due to end of input`; whitespace alone, `[]`, `5`, `"x"` and `true` are `Invalid request payload...`; `{` is the parse error; bytes after a complete value are ignored.
- Token: refused on both methods, ahead of an unresolvable project, and ahead of the unbounded-JQL refusal, which is the order here. An empty token and a `null` one are page one; a repeated one, which `_str_param` comma-joins, and a numeric one are the 400.
- Placement: all four rows of the table, re-measured on the same site.
- `/rest/api/2` and `/rest/api/3` answer identically, the `instance` in the 415 included.
- Credentials: a Bearer value that is not a Connect token is the 403 ahead of the media type on real as it is here, and a Basic pair that does not resolve is served anonymously and reaches the 415 and the 400 with `X-Seraph-LoginReason: AUTHENTICATED_FAILED` on both, again as here.

khj809:
Thank you — I re-measured every claim here against the same site before acting on it, and each one reproduced. All four are addressed in `097e25e`, three of them as findings and the `Allow` one narrowly. Details below, including where I took a different route than the one suggested.

## 1. `order by created` — replaced in all three places

Confirmed on both methods: it is the same unbounded-JQL refusal as no `jql` at all, and `ORDER BY created DESC` with it. Your three replacements are each a 200 there; I took `project is not EMPTY` for the reason you gave — it reads closest to what the call means, and `_project_from_jql` needs a `project =` before it filters, so the crawl stays unfiltered.

Changed in `tests/_helpers.py`, `examples/using-official-sdk/jira.py` and `tests/test_sdk.py`, and the comment above the crawl no longer states a vendor rule real contradicts.

## 2. `{"jql": null}` — fixed

Reproduced exactly: 200 and the whole visible corpus, through `str(None)`. It is now `"" if body.get("jql") is None else str(body["jql"])`, so a null is the parameter unsent and draws the refusal `{}` draws.

The two beside it stay out, as you suggested, and I have added your measurements of them to #212's body — `{"jql": 5}` coerced then refused by the JQL parser, `{"jql": ["order by created"]}` the `Invalid request payload` sentence `BODY_NOT_AN_OBJECT` already carries, applied to a field rather than to the whole body.

## 3. `Allow` — the regression is gone, and not by way of a 405 mechanism

Reproduced, on both versions and on DELETE and PATCH as well.

I first wrote the union-over-matching-routes fix in the 405 handler and backed it out: this FastAPI keeps an included router as an `_IncludedRouter` whose routes are not reachable through `.routes`, only through `effective_candidates` / `original_router`, and a header that depends on walking those is a FastAPI bump away from silently narrowing again.

So I removed the cause instead. `search/jql` is one route serving both methods again, and the per-method split moved into the served document — `openapi.jira_search_placement`, beside the `github_page_parameters` and `google_system_parameters` post-processors `main.py` already runs. `PUT` is back to `Allow: POST, GET` on both versions, the document still declares the parameters on the GET and the body on the POST, and the six `extra_param` entries stay dropped.

That leaves the 405's BODY untouched and still Starlette's, which I take to be #245's to measure and reproduce. Happy to be told otherwise if you would rather that land here.

## 4. The `_int` docstring — corrected

You are right that the range is the operation's own: refused on the query string too, to the character; it names the parameter; it carries `errors` where the two type refusals carry `errorMessages` alone. The sentence now says that and no longer attributes it to Jackson, so it agrees with #212 instead of contradicting it. The rest of the docstring I re-measured and kept.

## The body reader's leading bytes — fixed

Both reproduced. `str.lstrip()` is now `lstrip(" \t\n\r")`, and the decode is strict with `UnicodeDecodeError` mapping to `BODY_NOT_AN_OBJECT`, which is what real answers for those bytes. Trailing bytes after a complete value are still ignored, so both boundaries are now measured rather than inherited from a JSON reader's defaults.

## The MCP claim — restated

You are right, and the correction survives the change above: the operation list is unchanged, the tool is not. Measured on this head against `main` through FastMCP 4.0.3, `jira_search` goes from six arguments to three — `jql__query`, `maxResults__query`, `nextPageToken__query` and the three body ones, down to the three the GET reads. The body now claims the deletion rather than describing the surface as unchanged.

## The smaller inline ones

The `requestBody` is `"required": True`; the no-`jql` comment carries its measurement date like the token comment three lines up; and the media-type test's docstring states the measurement again ("the header decides before the bytes are looked at") rather than restating its assertions.

## Verification after the change

The real-vs-Backlot matrix grew to 24 cases with `{"jql": null}`, the NBSP-prefixed body and the malformed-UTF-8 body in it, and is 24/24 identical. Suite is 3080 passed, 1 skipped; `ruff check` and `ruff format --check` clean; `backlot diff --source jira` still reports only the four `missing_operation` gaps.

tests/test_atlassian.py
```diff
@@ -393,10 +395,13 @@ def test_jira_search_filtered_by_project(client, admin_h):
...
-    # no project clause at all -> unfiltered (same three issues here, since payments is the
-    # only Jira project in the SAMPLE corpus -- the earlier assertions are what prove filtering,
-    # not this equality)
-    unfiltered = client.get("/atlassian/rest/api/3/search/jql", headers=admin_h).json()
+    # a jql with no project clause at all -> unfiltered (same three issues here, since payments
+    # is the only Jira project in the SAMPLE corpus -- the earlier assertions are what prove
+    # filtering, not this equality). An empty jql is refused instead of read as this same
+    # unfiltered case -- see test_jira_search_refuses_no_jql_at_all.
+    unfiltered = client.get(
+        "/atlassian/rest/api/3/search/jql", headers=admin_h, params={"jql": "order by created"}
```
nuri-yoo: `order by created` is still here, and in `test_atlassian_responses_unchanged_by_enrichment` below. My review named the crawl helper, the example and `test_sdk.py`, so this is a gap in what I asked for rather than in what you did, and nothing fails: both are Backlot-only assertions.

What it leaves is the comment right above this call offering the query as the example of "a jql with no project clause at all", when that request is one real refuses — so a reader meets the discredited spelling and the replacement in the same file. `project is not EMPTY` reads the same here.

tests/test_atlassian.py
```diff
@@ -604,7 +609,9 @@ def test_atlassian_serverinfo_has_typed_response_schema(client):
 
 
 def test_atlassian_responses_unchanged_by_enrichment(client, admin_h):
-    search = client.get("/atlassian/rest/api/3/search/jql", headers=admin_h).json()
+    search = client.get(
+        "/atlassian/rest/api/3/search/jql", headers=admin_h, params={"jql": "order by created"}
```
nuri-yoo: The second of the two. Same substitution and same reason as the call in `test_jira_search_filtered_by_project`.

tests/test_atlassian.py
```diff
@@ -1385,15 +1392,273 @@ def test_confluence_refuses_the_parameter_real_names_when_both_are_wrong(
...
+        "Unbounded JQL queries are not allowed here. Add a search restriction to the query."
+    ]
+
+
+@pytest.mark.parametrize(
+    "raw,message",
+    [
+        # JSON's whitespace is the four ASCII ones, so a non-breaking space in front is not skipped
+        (
+            " {}",
```
nuri-yoo: This is a real U+00A0 in the source, and the only raw one in the repo: `backlot/errors/atlassian.py:68` writes it as `"   "`, and the two comments that discuss the same character (`backlot/routers/atlassian.py:1534`, and the docstring at line 1265 of this file) write `\xa0`. The one value a reader has to see to understand the case is the one character they cannot see.

It fails loudly rather than silently if it is ever normalised to a space — the body would parse and draw the unbounded refusal instead — so this is consistency rather than risk. `" {}"` matches what the rest of the repo does with it.

tests/test_atlassian.py
```diff
@@ -1385,15 +1392,273 @@ def test_confluence_refuses_the_parameter_real_names_when_both_are_wrong(
...
+    POST cursor in the query string and never sees why it does not page."""
+    paths = client.app.openapi()["paths"]["/atlassian/rest/api/3/search/jql"]
+    assert {p["name"] for p in paths["get"]["parameters"]} == {
+        "jql",
+        "maxResults",
+        "nextPageToken",
+    }
+    assert "requestBody" not in paths["get"]
+    assert "parameters" not in paths["post"] or paths["post"]["parameters"] == []
+    assert "requestBody" in paths["post"]
```
nuri-yoo: `required` is pinned nowhere: this line holds the requestBody's presence and the lines around it hold the placement. Since the reason for declaring it is that the handler refuses a POST carrying no body at all, `assert paths["post"]["requestBody"]["required"] is True` here would keep the document and the handler saying the same thing.

nuri-yoo (approved):
Approving. I re-ran each of the four at `097e25e` rather than reading the diff for them, and they hold, with the behaviour around them unchanged.

```
Allow              PUT, DELETE and PATCH on /rest/api/2 and /rest/api/3: 405 with Allow: POST, GET
{"jql": null}      400 unbounded, where 11d6748 answered 200 and the whole visible corpus
leading bytes      NBSP is the parse error, malformed UTF-8 the not-an-object sentence
still true beside  leading ASCII whitespace still reaches the object, trailing bytes still ignored,
                   the three sentences unchanged across seven bodies
served document    GET has the three parameters and no requestBody, POST has requestBody.required
                   and no parameters, on both versions
idempotent         app.openapi() twice, and again after clearing the cache: identical
bridge             jira_search is three arguments
gates              3080 passed 1 skipped · ruff check and format clean · gen_docs --check clean ·
                   backlot diff --source jira at the same four gaps
guards             the four new cases all fail at 11d6748 with only the test file copied over
```

Taking the route back and splitting the placements in the document is better than the union I suggested: it does not depend on walking `_IncludedRouter`, and it restores the two duplicate operationIds the split had removed, so the paragraph in `backlot/openapi.py` that names `search/jql` among the routes serving more than one method reads true again. That paragraph was in my working notes and did not make it into the review; the change made it moot either way.

One note that is not inline because it is not in a file: the **Verification** bullet still reads "the same two pre-existing failures on `tests/test_mcp.py`", where your comment above reports 3080 passed, 1 skipped, which is what I measure here as well. Worth a line before this merges, since the body is where a reader looks for the last run.

Four notes are inline. None of them blocks.

khj809:
Thank you. All five are in `4f85f6d`, and none needed a judgement call.

**The Verification bullet** was the stale one you caught — it described an earlier run on a sandbox that could not reach PyPI, and had survived several rounds of edits to the sections around it. It now reads 3080 passed, 1 skipped with the skip named (`llama_index.readers.hubspot`, not installed), and I added the `gen_docs --check` and `backlot diff` lines so the last run is stated in one place rather than half in the body and half in a comment.

**The two remaining `order by created` calls** — `test_jira_search_filtered_by_project` and `test_atlassian_responses_unchanged_by_enrichment` — are `project is not EMPTY` now. You are right that the comment was the cost: it offered the discredited spelling as the example of "a jql with no project clause at all", so a reader met it and its replacement in the same file. That comment now says the clause still has to restrict, and names the empty jql and the bare `ORDER BY` as the same refusal.

**The raw U+00A0** is `"\xa0{}"`. It was the only one in the repo, as you say; `backlot/errors/atlassian.py` and both comments that discuss the character already write it escaped, and the one value a reader has to see was the one they could not.

**`required`** is pinned: `assert paths["post"]["requestBody"]["required"] is True`, replacing the presence-only assertion, so the document and the handler's refusal of a body-less POST are held together in the same test.

Suite is 3080 passed, 1 skipped, and `ruff check` / `ruff format --check` clean after the change.

On the `openapi.py` paragraph naming `search/jql` among the multi-method routes — glad that reads true again. It was not something I checked when I split the routes, and it would have gone stale silently.

PR #220, opened 2026-09-15
google: an error body is indented and names its charset, and a callback answers it at 200 as a script

### LINKED ISSUE #203: google: an error body is never wrapped for callback and never indented, where real answers JSONP at 200 and always indents
**Source and commit** — `google_drive` and `gmail`, at `9c2e96a` (`main`).

Every Google error body Backlot serves is rendered by one line in `main._http_exception_handler` — `JSONResponse(status_code=…, content=body)` — which reaches none of the three things real does to the same body. A success on the Sheets routes already does all three, in `routers.google._sheets_respond`, so the divergence is between Backlot's own two paths as much as against the vendor.

| # | What real does to an error body | Backlot | Silent? |
|---|---|---|---|
| 1 | `callback=cb` wraps it: **HTTP 200**, `text/javascript; charset=UTF-8`, `// API callback\ncb({…});` | the error status, `application/json`, unwrapped | yes |
| 2 | indents it two spaces and ends it with a newline, whatever `prettyPrint` says | one compact line, no trailing newline | yes |
| 3 | sends `application/json; charset=UTF-8` | `application/json` | yes |

Silent in all three: `backlot diff` builds its findings from the discovery document's operations and parameters (`fidelity.operations`), so no comparison it runs reads a response body, and every one of these is a 200-vs-400 or byte-level difference inside one.

**How you know** — measured against `sheets.googleapis.com`, `docs.googleapis.com`, `gmail.googleapis.com` and `www.googleapis.com/drive/v3` on 2026-09-14.

- `callback` on an error, all four families: Sheets (a bad range), Drive (a bad `fields`), Docs and Gmail (no `Authorization` header) each answered **200** with `text/javascript; charset=UTF-8` and the error object inside `// API callback\ncb(…);`. The same four without `callback` answered their real status with `application/json; charset=UTF-8`.
- Indentation is not `prettyPrint`'s to turn off on an error: the same bad range with no parameter, with `prettyPrint=false` and with `prettyPrint=true` came back two-space indented all three times, 114 bytes ending `"NOT_FOUND"\n  }\n}\n`.
- Slides is measured for the anonymous error only; its authenticated reads answer a quota-project 403 to this token. Whether an error body differs by family beyond these three properties was not measured — every family answered the same shape.

```python
import httpx
from backlot import serve

with serve() as s:
    tok = httpx.get(f"{s.base_url}/_meta/users").json()["admin_token"]
    h = {"Authorization": f"Bearer {tok}"}
    sid = httpx.get(f"{s.base_url}/drive/v3/files", headers=h,
                    params={"q": "mimeType='application/vnd.google-apps.spreadsheet'"}).json()["files"][0]["id"]
    ok = httpx.get(f"{s.base_url}/sheets/v4/spreadsheets/{sid}/values/Sheet1!A1?callback=cb", headers=h)
    err = httpx.get(f"{s.base_url}/sheets/v4/spreadsheets/{sid}/values/NOPE!!?callback=cb", headers=h)
    print(ok.status_code, ok.headers["content-type"], repr(ok.text[:18]))
    print(err.status_code, err.headers["content-type"], repr(err.text[:18]))
```

The first spreadsheet the corpus answers with is `Pricing Sheet 2026`, whose one sheet is `Sheet1`. The success prints `200 text/javascript; charset=UTF-8 '// API callback\ncb({'` and the error prints `400 application/json '{"error":{"code":400'`. Real answers the second one the way it answers the first.

**What it costs a client** — JSONP is the case where the status code is the whole point. A browser loading the response through a `<script>` tag cannot read a body that did not arrive as JavaScript, and cannot see a status at all, so against real a failed call runs `cb({"error": …})` and the page handles it, while against Backlot the script tag fires `onerror` and the callback never runs: the client's error branch is the one path its tests cannot exercise against Backlot. The indentation and the charset cost less but cost it everywhere — a client that records a response body byte-for-byte, or one that sniffs the charset off the header, sees a different body from Backlot on every Google error.

**Where it lands** — `backlot.main._http_exception_handler`, which is where the body from `backlot.errors.http_body(path, exc, query)` becomes a response; the query is already passed in for `$.xgafv` (#172), so `callback` is readable at the same seam. The shape to follow is `routers.google._sheets_respond`, which already renders the success side to the byte: compact means no space after `:` or `,` and no trailing newline, indented means two spaces and one, and a `callback` wraps the text as `// API callback\n{cb}({text});` under `text/javascript; charset=UTF-8`.

**What a fix has to settle**

- Whether the handler is the right place, or whether Google's rendering belongs behind `errors.google` with the handler asking for a `Response` rather than a dict. The handler serves Atlassian and GitHub too, and neither vendor's error rendering has been measured for `callback` or indentation, so a change made in the handler has to leave those two exactly as they are.
- What a JSONP error's status is for a family real was not measured on with a credential — the measurement here is one error per family, and all four agreed on 200, so the simplest reading is that the wrap always wins. Worth one more request per family before it is written down as a rule.
- Whether an unparseable or hostile `callback` value is refused. Real was not asked; Backlot's success path accepts any string and interpolates it, which a fix should not extend to errors without measuring first.
- Recommendation: take 2 and 3 first — they need no query, apply to every Google error, and are a two-line change with a byte-for-byte test. Leave 1 until the JSONP status is measured a second time per family.

nuri-yoo:
Closes #203. Every Google error body now reaches the wire the way real writes one: two spaces deep with a trailing newline, `application/json; charset=UTF-8`, `<` and `>` escaped, and — where the request carries a `callback` — HTTP 200 with `text/javascript; charset=UTF-8` and the error inside `// API callback\ncb({…}\n);`.

| What real does to an error body | Before | Now |
|---|---|---|
| `callback=cb` wraps it: 200, `text/javascript; charset=UTF-8` | the error status, `application/json`, unwrapped | the same 200 script, byte for byte |
| indents it two spaces and ends it with a newline, whatever `prettyPrint` says | one compact line, no trailing newline | indented, and `prettyPrint` reaches the success path only |
| sends `application/json; charset=UTF-8` | `application/json` | the charset |

Measured 2026-09-15 against `sheets.googleapis.com`, `docs.googleapis.com`, `gmail.googleapis.com`, `slides.googleapis.com` and `www.googleapis.com/drive/v3`, authenticated where a credential reaches the family and anonymous where it does not, over a 400, a 401, a 403 and a 404. The issue left the JSONP status open pending one more request per family — all five answer 200, on every one of those statuses, with a credential and without, so the wrap always wins. `prettyPrint=true`, `false` and no parameter each gave a byte-identical indented error on all five.

The issue also left open whether an unparseable `callback` is refused. It is, and real names its own rule in the refusal: `Invalid JSONP callback name: 'a b'; only alphabet, number, '_', '$', '.', '[' and ']' are allowed.` A character-by-character sweep on Sheets — `cb<CH>x` for each ASCII punctuation mark and for space, tab, newline and `é`, plus `a<ZWSP>b` — accepts exactly those seven classes and refuses everything sent outside them, so "alphabet" is ASCII letters and nothing wider. Position and length do not matter: `.cb`, `cb.`, `1bad`, `$` and a 2,000-character name are all accepted. The refusal still arrives wrapped, through the very name it refuses, with `<` and `>` escaped there as they are in the body — so a page that asked for a callback gets a script that calls something and the error object reaches its error branch rather than `onerror`.

Three parameters decide the answer between them, and the order is measured. `$.xgafv` goes first: `callback=a b&$.xgafv=9` answers the `$.xgafv` sentence, wrapped through the name the other check would have refused. The `callback` refusal goes next, ahead of a bad token, a missing credential, an unparseable range and a mistyped `fields` mask. An `alt` other than `json` takes the request out of the JSONP path altogether — `alt=media` and `alt=zzz` answer their own 400 as `application/json` and unwrapped even when `callback` is itself unparseable — and an empty `callback=` is no callback at all.

`backlot.main._http_exception_handler` now asks the envelope for a whole `Response` rather than a dict, through a `rendered(status_code, body, query, headers)` that a vendor module may or may not define. Google defines it; Atlassian and GitHub do not, and neither vendor's error rendering has been measured for `callback` or indentation, so both keep exactly the `JSONResponse` they had, GitHub's own charset included. The bytes themselves live in `errors.google.respond`, which `routers.google._sheets_respond` now renders successes through as well: a success and a failure on the same route come back through one serializer on real, and the divergence this closes was between Backlot's own two paths as much as against the vendor.

`callback` stays declared on the Sheets routes rather than router-wide beside `$.xgafv`. Sheets is the only family whose SUCCESS is wrapped — the other four honour `callback` on their errors now and on nothing else — and `openapi.qp` declares only what Backlot honours. Closing that gap means rendering every Drive, Gmail, Docs and Slides success through the same serializer, which needs a measured success body per family; this token reads Sheets, Docs and Drive but has no Gmail scope and no quota project for Slides, so it is left for its own issue rather than written down from the two families that were measured.

None of this is anything `backlot diff` can report: for a Google source it compares `app.openapi()` against the discovery document, operations and parameters only, so no comparison it runs reads a response body.

## Verification

- `pytest -rs` with all extras installed: `3071 passed, 1 skipped`. The skip is `tests/test_llamaindex.py:121`, `llama_index.readers.hubspot` not importable — `llama-index-readers-hubspot` is deliberately absent from the `llamaindex` extra, and `pyproject.toml` says why, so it skips wherever the suite runs.
- `ruff check .` and `ruff format --check .` clean; `python scripts/gen_docs.py --check` clean.
- Thirteen mutations of the new rules, each run against one test on a clean tree: dropping the trailing newline, `indent=4`, dropping the charset, answering JSONP at the real status, removing the `<`/`>` escape, treating an empty `callback` as one, removing the `alt` guard, widening and narrowing the name pattern, swapping the `$.xgafv`/`callback` order, skipping the render seam, exempting Drive from the check, and rendering GitHub through Google's serializer. All thirteen were caught.


backlot/errors/google.py
```diff
@@ -297,21 +304,119 @@ def bad_system_parameter(name: str, value: str) -> GoogleError:
...
+    render where real answers a 400 are a gap of their own.
+
+    Both are read as :func:`first_repeat`, not off ``QueryParams.get``: real answers a repeated
+    `callback` through the first name and a repeated `alt` through the first format.
+    """
+    if request.method != "GET":
+        return None
+    query = request.query_params
+    alt = first_repeat(query, ALT)
+    if alt is not None and alt != "json":
```
khj809: `alt` is case-insensitive on real, and an empty `alt=` is no value at all — this guard suppresses the wrap on both.

Measured 2026-09-17, anonymous on all five families and authenticated on Sheets:

```
?alt=JSON&callback=cb   drive · gmail · docs · slides · sheets   200 text/javascript; charset=UTF-8   wrapped
?alt=Json&callback=cb   drive · gmail · docs · slides · sheets   200 text/javascript; charset=UTF-8   wrapped
?alt=&callback=cb       sheets success · drive 403               200 text/javascript; charset=UTF-8   wrapped
?alt=                   sheets success                           200 application/json; charset=UTF-8  served
```

Backlot answers every one of those unwrapped at the error status — `/docs/v1/documents/x?callback=cb&alt=JSON` is a plain 401 here where real is a 200 script. Casefolding and reading the empty value as absent covers both:

```python
alt = (first_repeat(query, ALT) or "").casefold()
if alt and alt != "json":
    return None
```

The same comparison in `_sheets_respond` reads the same way and is wrong the same way, though that half predates this PR: `alt=JSON` and `alt=` are 400s here where real serves the 200, and `alt=MEDIA` lands on the generic sentence where real answers `Unsupported alt type "media" for non byte stream request.` with the value lowercased. `test_an_alt_the_api_cannot_render_takes_the_request_out_of_the_jsonp_path` sends only `media` and `zzz`, so neither half is covered.

backlot/errors/google.py
```diff
@@ -372,3 +477,94 @@ def validation_body(path: str, errors) -> None:
...
+
+
+def _escaped(text: str) -> str:
+    """``<`` and ``>`` as their `\\u003c` and `\\u003e` escapes, which is what real's serializer
+    writes.
+
+    Measured on both sides of the same API: a Sheets cell holding ``<b>&'x`` followed by two
+    non-ASCII letters comes back with the brackets escaped and the letters raw, and an error
+    message echoing an unparseable range spelled ``<b>&'x`` does the same. So `&`, `'` and
+    non-ASCII stay as they are and only the two angle brackets move -- the escape that keeps a
```
khj809: The rule is wider than the two angle brackets, and the wrapper needs more of it than the body does.

Measured 2026-09-17 on Sheets, `?callback=cb<CH>x`, reading the name real writes into `// API callback\n…(`:

```
escaped:  \t   \n   \r   \\   \"   \u003c   \u200b (ZWSP)   \u202e (RLO)   \u2028
raw:      '    é    가    \u00a0 (NBSP)    😀
```

It reaches the body as much as the wrapper: an unparseable range carrying U+200B comes back from real as `"Unable to parse range: a\u200bb!!"`, where `json.dumps(…, ensure_ascii=False)` emits the character raw. So the serializer escapes the C0 controls and the Unicode format / line-separator characters on top of what JSON requires — `é`, `가`, NBSP and an emoji are what actually stay as they are, not non-ASCII as a class.

`test_the_characters_a_callback_name_may_not_hold` parametrises `"\t"`, `"cb\n"` and ZWSP, and `_jsonp` builds its expected prefix with the `<` / `>` replacement alone, so the test pins Backlot's literal tab, newline and ZWSP as the answer. Not a security problem — `<` is escaped, so `</script>` cannot be formed, and the name sits in identifier position — but it is a byte divergence in the one path this PR measures to the byte.

Either widen the escape to that rule, or narrow the sentence to what was measured: `&`, `'` and non-ASCII letters.

tests/test_google.py
```diff
@@ -1057,6 +1058,395 @@ def test_a_missing_header_differs_by_family(client, path, code, status):
...
+    indented with no parameter, with `false` and with `true` alike, byte for byte. A SUCCESS under
+    `false` is compact, so the parameter is the success path's alone — which is why the success is
+    read here too, as the half of the pair that does change."""
+    params = {} if pretty is None else {"prettyPrint": pretty}
+    bad = _values(base, admin_h, sheet_id, "NOPE!!", **params)
+    assert bad.status_code == 400, bad.text
+    assert bad.text == json.dumps(bad.json(), ensure_ascii=False, indent=2) + "\n"
+    ok = _values(base, admin_h, sheet_id, "Sheet1!A1", **params)
+    compact = pretty == "false"
+    assert ok.text.endswith("}") if compact else ok.text.endswith("}\n")
```
khj809: These three lines restate what `test_pretty_print_is_on_by_default_and_only_false_turns_it_off` (line 4418) already holds, and could come out.

That test pins both spellings of a Sheets success to the byte — `{"range":"Summary!A1"` with no trailing newline under `prettyPrint=false`, `"range": "Summary!A1"` ending `\n` without it — so re-reading `endswith("}")` / `endswith("}\n")` here adds no case. What is this PR's is the half above: an ERROR is indented whatever the parameter says, and the `[None, "false", "true", "NOPE"]` sweep over `bad` carries that on its own.

The same trim fits `test_a_callback_wraps_a_success_and_an_error_through_the_same_serializer` below: `ok_cb.text == f"// API callback\ncb({ok.text});"` is `test_callback_wraps_the_body_as_jsonp` (line 4445) again, and the claim that is new — a success and an error rendered through one serializer — is carried by the `bad_cb` line and the two status codes agreeing, which hold with `ok_cb` read once rather than compared twice.

Suggestion rather than a request: both tests are correct as they stand, and dropping the duplicated halves only shortens what a reader has to hold to see which assertion is the measurement.

khj809 (commented):
Reviewed by re-measuring every claim in the body against the live APIs on 2026-09-17 — Sheets, Docs, Drive, Gmail and Slides, anonymous and authenticated.

Everything the body states holds. The JSONP 200 and its `text/javascript; charset=UTF-8`, the two-space indent and trailing newline under `prettyPrint` in all three spellings, the charset on the plain type, the refusal sentence verbatim and wrapped through the name it refuses, `$.xgafv` ahead of `callback`, `callback` and `alt` read from the first repeat against `$.xgafv`'s last, a second repeat left unvalidated, an empty `callback=` as no callback, a POST ignoring the parameter while still honouring `$.xgafv` and `prettyPrint`, and an unrouted family path answered as HTML — 400 on Sheets, Docs and Slides, 404 on Drive and Gmail — all reproduce. The served document is byte-identical to the merge base, `backlot diff` is 0 new on `google_drive` and `gmail`, and the suite, ruff and `gen_docs.py --check` are green at `6427c42`.

Two divergences found, both narrow, neither blocking the shape of the change: the `alt` guard reads a value real reads case-insensitively, and the escape rule is wider than `<` and `>`. Comments inline, with the measurements. One suggestion on test duplication.

Worth refreshing the verification block before merge — it pins `1f56813` on `main` at `a319735`, and the head has taken two merges from `main` since (the suite is 3176 passed / 1 skipped now, not 3145).

backlot/errors/google.py
nuri-yoo: Confirmed both halves and fixed them in `114d3ce`.

Re-measured 2026-09-17, anonymous on all five families and authenticated on Sheets: `alt=JSON`, `alt=Json` and `alt=` each come back 200 `text/javascript; charset=UTF-8` wrapped, and authenticated `alt=JSON` and `alt=` each serve the 200 success. Backlot answered every one of them unwrapped at the error status, exactly as you read it.

One thing a plain casefold would have got wrong, so it is worth recording: the two refusal sentences quote different spellings. `alt=MEDIA` and `alt=Media` answer `Unsupported alt type "media" for non byte stream request.` with the format lowercased, where `alt=ZZZ` answers `Invalid value "ZZZ" for query parameter 'alt'` through the spelling that arrived. So `errors.google.alt_format` returns the folded value and `_sheets_respond` reads the parameter again for the generic sentence.

The `_sheets_respond` half is in the same commit, and it turned out to be three sites rather than two. Drive's download reads `alt` too, and it was reading it twice over: measured against `www.googleapis.com/drive/v3/files/<id>` with a credential, `alt=MEDIA` and `alt=Media` hand back the file, and `alt=media&alt=json` downloads where `alt=json&alt=media` answers the metadata. Backlot handed back metadata on both spellings and read the last repeat. All three now go through `alt_format`.

Four tests cover it: the wrap on all five families over `["json", "JSON", "Json", ""]`, the Sheets success over the same four, the five refusal spellings, and the Drive download. That last one also sends both spellings at a file only the admin can see, so the 404 a scoped token gets is pinned beside the 403 the admin gets.

backlot/errors/google.py
nuri-yoo: Confirmed, and the rule is wider than either of us wrote, so I widened the escape rather than narrow the sentence. `5c61790`.

Rather than sample it I swept the whole space through the probe you found: `?alt=<value>` echoes the value through the same serializer, so 1,112,063 codepoints went through it in 1,013 requests — every one a query string carries but the surrogates and U+0000, which the front end hands back as the literal `%00` rather than decoding. 209 come back escaped:

```
U+0001-U+001F  U+0022  U+003C  U+003E  U+005C  U+007F-U+009F  U+00AD
U+0600-U+0603  U+06DD  U+070F  U+17B4-U+17B5  U+200B-U+200F   U+2028-U+202E
U+2060-U+2064  U+206A-U+206F  U+FEFF  U+FFF9-U+FFFB  U+1D173-U+1D17A
U+E0001        U+E0020-U+E007F
```

The one correction to your reading: it is not the format category plus the separators as a category test would return them today. U+17B4 and U+17B5 are escaped and have been Mn since Unicode 4.1, while U+061C, U+0604 and U+180E are not escaped and are each Cf now. U+1D173-U+1D17A and the tag block are escaped as surrogate pairs, where U+110BD, U+13430 and U+1BCA0 — astral format characters too — are not. It is Unicode 4.0's format category, so the ranges are data in the code and in the test rather than a lookup.

`json.dumps` already writes U+0000-U+001F, `"` and `\` the way real does, short forms included, so `_escaped` adds the other 145 and `_escaped_name` runs a name through `json.dumps` first for the half a body has already had. Astral goes out as the pair real spells it with.

Your body-side observation reproduces without a credential, which is cheaper than the range parse: `?alt=zz<ZWSP>z` answers `Invalid value "zz\u200bz" for query parameter 'alt'` anonymously.

`test_the_characters_the_serializer_escapes` holds the ranges as literal data and checks the edges of each one plus the characters the rule is easy to guess wrong. `_jsonp` now escapes through the same function, and `test_a_callback_name_reaches_the_wrapper_escaped` pins the eight literal byte forms so the helper is not checking the code against itself.

tests/test_google.py
nuri-yoo: Took the first half in `808702a` and left the second, with a note saying why.

The three lines are gone. The docstring went with them — it ended `which is why the success is read here too, as the half of the pair that does change`, which stopped being true the moment the lines came out, so it now points at `test_pretty_print_is_on_by_default_and_only_false_turns_it_off` for that half instead.

I kept `ok_cb.text == f"// API callback\ncb({ok.text});"`. `test_callback_wraps_the_body_as_jsonp` asserts `startswith("// API callback\ncb({")` and `endswith(");")` and nothing between them, so it does not hold that a wrapped success is its own unwrapped body byte for byte — the error side has that in `test_a_callback_answers_an_error_at_200_as_a_script`, and the success side has it only here. The docstring now says which line is the measurement, so the next reader does not have to re-derive that. Happy to drop it if you read 4445 as covering it and I am splitting hairs.

backlot/errors/google.py
```diff
@@ -372,3 +498,159 @@ def validation_body(path: str, errors) -> None:
...
+
+# --- how a Google body reaches the wire ---------------------------------------------------------
+
+
+# The characters real's serializer writes as `\\uXXXX` rather than as themselves, beyond the ones
+# JSON requires of every serializer. Measured 2026-09-17 by sending each of the 1,112,063
+# codepoints a query string carries -- every one but the surrogates and U+0000, which the front end
+# hands back as the literal `%00` rather than decoding -- through the `alt` echo on Sheets in 1,013
+# requests, and reading which came back escaped. 209 do, and these are the 145 of them that
+# ``json.dumps`` leaves alone.
```
khj809: `_ALSO_ESCAPED` holds 176 of them, not 145 — the code is right and the number beside it is not.

209 is the whole set and the test pins it. What `json.dumps(…, ensure_ascii=False)` writes on its own, among the codepoints the sweep could reach, is U+0001-U+001F, `"` and `\\` — 33 of them, since U+0000 is the one the front end never decodes. 209 − 33 = 176, which is what this tuple spells:

```python
>>> sum(hi - lo + 1 for lo, hi in _ALSO_ESCAPED)
176
>>> {cp for cp in ALL_209 if json.dumps(chr(cp), ensure_ascii=False)[1:-1] == chr(cp)} == set(_ALSO_ESCAPED_expanded)
True
```

`test_the_characters_the_serializer_escapes` asserts `len(escaped) == 209` against `MEASURED_ESCAPES`, so the total is held; nothing reads the 145, which is why it can sit here wrong. The same number is in `5c61790`'s message, where it cannot be corrected now.

Worth saying while I am here: I re-ran the sweep independently before reading this — the whole BMP, the astral ranges this tuple names, and 8,000 astral codepoints at random, 71,765 in all through the same `alt` echo. 209 came back escaped and the set matches this one codepoint for codepoint, U+17B4/U+17B5 in and U+061C, U+0604, U+0605, U+180E out, U+1D173 and U+E0001 in as surrogate pairs and U+110BD, U+13430, U+1BCA0 out. The rule reproduces; only the arithmetic in the sentence does not.

docs/supported-sources.md
```diff
@@ -222,6 +222,23 @@ families show it only at `1`. Measured against the live Sheets, Docs and Drive A
...
+lets a page loading the answer through a `<script>` element reach its error branch rather than
+`onerror`. A name that cannot be a JavaScript one is refused with real's own sentence — `only
+alphabet, number, '_', '$', '.', '[' and ']' are allowed` — ahead of a bad token, a missing
+credential, an unparseable range and a mistyped `fields` mask, though `$.xgafv` is refused ahead of
+it and an `alt` other than `json` suppresses the wrap altogether. An empty `callback=` is no
+callback; a repeated one is answered through the first name where `$.xgafv` is answered through the
+last; and a POST ignores the parameter outright, as real does, since JSONP is what a `<script>`
+element fetches and a `<script>` element issues a GET. A SUCCESS body is wrapped and indented on
+the `/sheets/v4` routes only; the other four families honour `callback` on their errors and not yet
+on their 200s. Measured against the live Sheets, Docs, Drive, Gmail and Slides APIs on 2026-09-15.
```
khj809: This paragraph still describes the `alt` rule `114d3ce` replaced, and its date line predates two of the three measurements it now carries.

`an alt other than json suppresses the wrap altogether` was true when it was written and reads wrong now: `alt=JSON`, `alt=Json` and `alt=` are each *other than* `json` as a string and each is the JSON the default spells, so a reader who takes the sentence literally gets the opposite of what `alt_format` does. `an alt naming a format other than json` — or naming the fold outright — would carry the fix.

The closing `Measured … on 2026-09-15` covers the wrap, the indent and the charset, but the 209-character escape set and the case-insensitive `alt` were both measured 2026-09-17, and the sentence about them sits above this line.

The same `An alt other than json takes the request out of the JSONP path altogether` is in the PR body's **Five things decide the answer between them** paragraph, one section below **Which alt still means JSON**, which says the opposite about the same three spellings. Worth the same edit so the two do not disagree.

khj809 (commented):
Re-reviewed at `25cb1e9`. All three comments are addressed, and two of them went further than what I asked.

I re-measured the new claims rather than reading them. `alt=MEDIA` and `alt=Media` answer `Unsupported alt type "media"` with the format lowercased where `alt=ZZZ` and `alt=Zzz` answer through the spelling that arrived; Drive's download hands back the bytes on `alt=MEDIA`, `alt=Media` and `alt=media&alt=json`, and the metadata on `alt=json&alt=media`. Both halves reproduce, and the download site is one I had missed.

The escape sweep I checked independently rather than sampling: the whole BMP, the astral ranges `_ALSO_ESCAPED` names, and 8,000 astral codepoints at random — 71,765 through the same `alt` echo. 209 come back escaped and the set matches this branch codepoint for codepoint, the six that a category test gets wrong in either direction included. The wrapper is byte-identical to real on every spelling I compared, surrogate pairs included. Holding the ranges as data in the test rather than importing them from `gerr` is the right call for a rule that cannot be derived.

On the third: you are right and I was not. `test_callback_wraps_the_body_as_jsonp` reads a prefix and a suffix, so it does not hold a wrapped success against its own unwrapped body — that assertion exists only where you kept it.

Gates at this head: `3241 passed, 1 skipped` (the hubspot reader, as `pyproject.toml` describes), `ruff check` and `ruff format --check` clean over 141 files, `gen_docs.py --check` exit 0, `backlot diff --source google_drive` 0 new.

Two things left, both prose over correct code — a count in a comment and a paragraph the `alt` fix outran. Inline.

backlot/errors/google.py
nuri-yoo: Confirmed and fixed in `9e91029`. `sum(hi - lo + 1 for lo, hi in _ALSO_ESCAPED)` is 176, and 209 minus the 33 `json.dumps` writes itself — U+0001-U+001F, `"` and `\` — is the same 176.

The comment now carries that subtraction rather than only its result, so the number has something to be checked against on the page. You are right that nothing read the 145: `test_the_characters_the_serializer_escapes` pins 209 and the ranges, and the split between what `json.dumps` does and what `_escaped` adds is an implementation detail no assertion touches. That is exactly the shape of sentence I should have counted before writing, and I did not.

`5c61790`'s message keeps the wrong number, as you say. The same number was in my reply above; I edited it to 176 and left a line saying so, since a comment can be corrected where a message in a pushed commit cannot.

Thanks for re-running the sweep rather than reading mine. Two independent runs agreeing on the six codepoints a category test gets wrong in either direction is a better guarantee than the run itself.

docs/supported-sources.md
nuri-yoo: Both halves confirmed, fixed in `9e91029`.

The sentence now reads `an alt naming a format other than json suppresses the wrap altogether — the format is matched without regard to case and an empty alt= names none, so alt=JSON, alt=Json and alt= each ask for the JSON the default serves rather than for a format of their own`. Your reading is the one a reader would take: `JSON` is other than `json` as a string, and the paragraph was telling them the opposite of what `alt_format` does.

The same sentence was in `jsonp_callback`'s docstring, which you did not flag — `So is any alt other than json`. It now says `naming a format` too and points at `alt_format` for which spellings name it. A repo-wide grep for the phrasing finds no third copy.

The date line now names the three days and what each one covered: the wrap, the indent and the charset on 2026-09-15, the suppression across the four non-Sheets families on 2026-09-16, and the escape set and the case-insensitive `alt` on 2026-09-17.

The PR body's **Five things decide the answer between them** carried the same sentence and has the same edit, so it no longer disagrees with **Which `alt` still means JSON** two paragraphs above it.

The whole commit is prose — `git diff 25cb1e9..9e91029 -- tests/` is empty, so the delta against the revision you verified holds no assertions.

khj809 (approved):
Approving at `ffdc512`. Everything raised across the two rounds is closed, and the three properties #203 opened reproduce against the live APIs.

**What I verified rather than read.** Re-measured on 2026-09-17 against `sheets.googleapis.com`, `docs.googleapis.com`, `gmail.googleapis.com`, `slides.googleapis.com` and `www.googleapis.com/drive/v3`, anonymous on all five and authenticated where a credential reaches one: the JSONP 200 and its `text/javascript; charset=UTF-8` on a 400, a 401, a 403 and a 404 alike; the two-space indent and trailing newline under `prettyPrint` in all three spellings where a success under `false` is compact; the charset on the plain type; the refusal sentence verbatim, arriving wrapped through the name it refuses; `$.xgafv` ahead of `callback` and wrapped through the name the other check would have taken; `callback` and `alt` read from the first repeat against `$.xgafv`'s last, with a second repeat never validated; an empty `callback=` as no callback; a POST ignoring the parameter while still honouring `$.xgafv` and `prettyPrint`; and a family path with no route answered from real's front end as HTML, 400 on Sheets, Docs and Slides and 404 on Drive and Gmail.

**The escape set, independently.** I swept it myself before reading yours — the whole BMP, the astral ranges `_ALSO_ESCAPED` names, and 8,000 astral codepoints at random, 71,765 through the same `alt` echo. 209 come back escaped and the set matches this branch codepoint for codepoint, including every one a category test gets wrong in either direction: U+17B4 and U+17B5 in, U+061C, U+0604, U+0605 and U+180E out, U+1D173 and U+E0001 in as surrogate pairs where U+110BD, U+13430 and U+1BCA0 stay raw. Feeding that swept set through `_escaped_name` at this head leaves nothing unescaped, and the wrapper is byte-identical to real on every spelling I compared it against. Holding the ranges as data in the test rather than importing them from `gerr` is what makes that check mean something.

**The `alt` fold.** `alt=MEDIA` and `alt=Media` answer `Unsupported alt type "media"` with the format lowercased where `alt=ZZZ` and `alt=Zzz` answer through the spelling that arrived, so returning the folded value and reading the parameter again for the generic sentence is the right split. Drive's download reproduces too: the bytes on `alt=MEDIA`, `alt=Media` and `alt=media&alt=json`, the metadata on `alt=json&alt=media`. That third read site and the ACL test beside it were yours, not mine — the right instinct, since upper-casing a value that now reaches corpus content has to reach the same ACL.

**Gates at this head.** `3248 passed, 1 skipped` — the hubspot reader `pyproject.toml` explains — `ruff check` and `ruff format --check` clean over 141 files, `python scripts/gen_docs.py --check` exit 0, `backlot diff --source google_drive` 200 divergence(s) · 200 acknowledged · 0 new, and the served document byte-identical to the merge base.

**On the three rounds.** Two of the fixes went past what I asked: the `alt` comment led you to a third read site I had not found, and the escape comment led you to a sweep rather than the narrowed sentence I offered as the cheaper option. The one place I was wrong — reading `test_callback_wraps_the_body_as_jsonp` as covering a wrapped success byte for byte — you answered with the assertion it actually makes rather than taking the note. The 145 and the stale `alt` sentence were prose over correct code and are both corrected, the docstring copy I missed included.

Nothing outstanding. The six follow-ups the body names (#221, #222, #223, #224, #225, #237) all exist and carry their own measurements, and #276 records what a Google `HEAD` does, which this PR is right not to touch.

PR #241, opened 2026-09-16
s3: a HEAD on a sub-resource names the methods Backlot actually serves in Allow

### LINKED ISSUE #230: s3: a HEAD carrying a sub-resource is a 405 with no Allow, where real names the methods that sub-resource takes
Found while measuring for #169.

**Source and commit** — `s3`, `main` at `90b90dd`.

`_head_refusal` in `backlot/routers/s3.py` answers 405 with an empty `application/xml` body and no headers. Real answers the same 405 and the same empty body, and adds an `Allow` naming the methods that sub-resource takes. The docstring's reason for sending none, "Backlot takes none of them", stopped being true when `?location` and `?uploads` became GET reads.

|---|---|---|---|---|
| 1 | `HEAD /{bucket}?location` | 405, `Allow: GET` | 405, no `Allow` | no |
| 2 | `HEAD /{bucket}?uploads` | 405, `Allow: POST, GET` | 405, no `Allow` | no |
| 3 | `HEAD /{bucket}?versioning`, `?acl` | 405, `Allow: GET, PUT` | 405, no `Allow` | no |
| 4 | `HEAD /{bucket}?tagging`, `?policy` | 405, `Allow: DELETE, GET, PUT` | 405, no `Allow` | no |
| 5 | `HEAD /{bucket}?versions` | 405, `Allow: GET` | 405, no `Allow` | no |
| 6 | `HEAD /{key}?acl` | 405, `Allow: GET, PUT` | 405, no `Allow` | no |
| 7 | `HEAD /{key}?tagging` | 405, `Allow: DELETE, GET, PUT` | 405, no `Allow` | no |
| 8 | `HEAD /{key}?uploadId=abc123` | 405, `Allow: DELETE, POST, GET` | 405, no `Allow` | no |
| 9 | `HEAD /{bucket}?acl&versioning` | 400, no `Allow` | 400, no `Allow` | — no divergence, listed so a fix does not add one |

**How you know** — measured 2026-09-15 against `s3.ap-northeast-2.amazonaws.com`, a general purpose bucket, path-style, SigV4 with an `x-amz-content-sha256` header. `Content-Type` is `application/xml` on every row, which Backlot already matches; `Allow` is the whole difference. The methods named are the sub-resource's own on real S3: `PUT` and `DELETE` appear for sub-resources Backlot serves no method of at all, so a fix has to decide what Backlot's `Allow` should say rather than copy real's verbatim.

**Reproduction** — the bundled corpus, any bucket; sign a `HEAD` the way `backlot.fidelity.s3_probe._sign` does and read the response headers.

**Can a client tell?** RFC 9110 makes `Allow` mandatory on a 405, so a client that reads it to decide what to try next gets nothing where real gives it a list. Not silent: the status and the body already agree with real, so nothing is misparsed.

**What a fix has to settle** — what Backlot should name. Real's list describes real's surface; Backlot serves `GET` for `?location` and `?uploads` and no method for anything else, so `Allow: GET` for those two and none for the rest is one answer, and copying real's list is another that advertises methods Backlot would refuse. Either way `_head_refusal`'s docstring needs the sentence that is now untrue replaced by what was decided.

khj809:
Closes #230.

## What changed

`_head_refusal` in `backlot/routers/s3.py` answered every `HEAD` carrying a sub-resource selector with a 405 and no headers at all. Real S3 sends the same 405 and the same empty body, but adds an `Allow`. What an `Allow` names is the resource answering rather than S3 — "a list of the target resource's currently supported methods" (RFC 9110, [Section 15.5.6](https://www.rfc-editor.org/rfc/rfc9110.html#section-15.5.6)) — so `_head_refusal` now sends `Allow: GET` for exactly the selectors the same path answers on a `GET`, and no `Allow` for a selector whose `GET` is a 501. Real's own list for those includes `PUT` and `DELETE` on several; naming a method Backlot still answers 501 on would be as false a claim as the missing header was.

The set is `_BUCKET_GETS`, read both by `bucket_get`'s 501 and by the header, so the two cannot name different things; and it reaches `_head_refusal` from the caller, because the selectors a path serves differ between a bucket's path and a key's. Backlot serves no object sub-resource, so an object's 405 names no method — and it keeps naming none if a bucket-named selector like `location` ever joins `_OBJECT_SELECTORS` to close #267.

Sending no header at all leaves Section 15.5.6's MUST unmet. [Section 10.2.1](https://www.rfc-editor.org/rfc/rfc9110.html#section-10.2.1) offers the alternative — "An empty Allow field value indicates that the resource allows no methods" — and an empty value does survive this stack, the client reading `Allow: ''` rather than no header. Real S3 sends an empty `Allow` on none of these rows, so it is chosen against: a value real never sends would be a divergence where the absent header is a gap. The docstring records the choice.

The 400 for two selectors at once is unchanged: real sends no `Allow` there either (measured), and a test now pins that.

Measured 2026-09-16 against a general purpose bucket in `us-east-1` and again 2026-09-17 in `ap-northeast-2` (path-style, SigV4, credentials in the environment), reproducing the issue's table with the same request shapes:

| request | real `Allow` | Backlot before | Backlot now |
|---|---|---|---|
| `HEAD /{bucket}?location` | `GET` | none | `GET` |
| `HEAD /{bucket}?uploads` | `POST, GET` | none | `GET` |
| `HEAD /{bucket}?versioning`, `?acl` | `GET, PUT` | none | none |
| `HEAD /{bucket}?tagging`, `?policy` | `DELETE, GET, PUT` | none | none |
| `HEAD /{bucket}?versions` | `GET` | none | none |
| `HEAD /{key}?acl` | `GET, PUT` | none | none |
| `HEAD /{key}?tagging` | `DELETE, GET, PUT` | none | none |
| `HEAD /{key}?uploadId=abc123` | `DELETE, POST, GET` | none | none |
| `HEAD /{bucket}?acl&versioning` | 400, no `Allow` | 400, no `Allow` | 400, no `Allow` (unchanged) |

A bucket that does not exist answers the same way on every row — `HEAD /{missing}?location` is 405 `Allow: GET` on real — so the header travels with the refusal that comes before the lookup, which is a row in the test.

Status codes and the empty `application/xml` body matched real on every row already and are unchanged, and `Allow` is the divergence the issue measured. It is not the whole of what real sends: real's 405 also carries `x-amz-request-id` and `x-amz-id-2` on each of these rows and Backlot carries neither, which is #269 rather than this PR.

## Verification

- **Tests** — `pytest` on an all-extras sync: 3104 passed, 1 skipped.
- **Optional surfaces that ran rather than skipped** — `tests/test_s3.py`, `tests/test_sdk.py`, `tests/test_mcp.py` (boto3 installed via `official-sdk`).
- **Mutation** — dropping the header, widening `_BUCKET_GETS` with `versioning`, and adding an `Allow` to the two-selector 400 each turn `tests/test_s3.py` red.
- **Lint and docs** — `ruff check . && ruff format --check .` clean; `python scripts/gen_docs.py --check` exit 0.
- `backlot diff --source s3`: 42 divergence(s) · 42 acknowledged · 0 new. The `s3` probe reads no response header, so the `Allow` is outside the diff's view — `docs/fidelity.md`'s "Served, not declared" section now carries it, beside GitHub's `HEAD` and the `x-ratelimit-*` headers, with the tests named as the record.

- [x] A new endpoint serving corpus content is ACL-scoped, proved for both an admin and a scoped token — N/A, no new endpoint; an existing refusal gained a header


backlot/routers/s3.py
```diff
@@ -243,10 +243,15 @@ def _head_refusal(selected: list[str]) -> Response:
...
-    takes none of them, so it sends none.
+    (measured). Real S3 also sends an ``Allow`` naming every method that sub-resource takes, GET and
+    PUT and DELETE among them (measured); Backlot names only the one it actually serves, ``GET`` for
+    ``location`` and ``uploads``, and sends none for a sub-resource it does not implement at all —
+    naming a method that still answers 501 would be as false as naming real's PUT or DELETE.
     """
-    return Response(status_code=400 if len(selected) > 1 else 405, media_type="application/xml")
+    if len(selected) > 1:
+        return Response(status_code=400, media_type="application/xml")
+    headers = {"Allow": "GET"} if selected[0] in ("location", "uploads") else None
```
nuri-yoo: `selected[0] in ("location", "uploads")` keys the header on the selector alone, but `_head_refusal` answers for both shapes: `head_bucket` passes it `_BUCKET_SELECTORS` and `object_get` passes `_OBJECT_SELECTORS`. That is correct today only because `_OBJECT_SELECTORS` happens to contain neither name.

It happens to contain neither because of a gap I measured while reading this: real refuses fifteen bucket-named selectors at a key path, `?location` and `?uploads` among them, where Backlot answers the object (#267). The moment either name joins `_OBJECT_SELECTORS` to close that, this line begins sending `Allow: GET` for `HEAD /{bucket}/{key}?location`, where Backlot serves nothing at all — the false claim the docstring above is written to avoid. Nothing would go red either: the test lists no key-path `?location` row.

The served set is also written twice — `bucket_get` decides it with `selected == ["location"]` and `selected == ["uploads"]`, and nothing ties the two spellings together. Keying on the resource as well as the selector fixes the first half; for the second, an invariant beats a second literal list, since it survives a future GetBucketVersioning without being edited: for each selector, `Allow: GET` is present exactly when the same path's `GET` is not a 501.

backlot/routers/s3.py
```diff
@@ -243,10 +243,15 @@ def _head_refusal(selected: list[str]) -> Response:
     No sub-resource has a HEAD form: real S3 answers ``HEAD /{bucket}?versioning`` and
     ``HEAD /{key}?acl`` 405 with an empty ``application/xml`` body, whether or not the bucket or the
     key exists, and two selectors at once with the conflict's 400 and the same empty body
-    (measured). Real S3 also sends an ``Allow`` naming the methods that sub-resource takes; Backlot
-    takes none of them, so it sends none.
+    (measured). Real S3 also sends an ``Allow`` naming every method that sub-resource takes, GET and
+    PUT and DELETE among them (measured); Backlot names only the one it actually serves, ``GET`` for
+    ``location`` and ``uploads``, and sends none for a sub-resource it does not implement at all —
+    naming a method that still answers 501 would be as false as naming real's PUT or DELETE.
```
nuri-yoo: Two things about the new sentence.

"naming **every** method that sub-resource takes" is wider than the measurement behind it: what was measured is what real's `Allow` says, not that the list is exhaustive of the methods that sub-resource takes. #230 and the sentence being replaced both said "the methods", which is the same fact without the universal.

The decision itself has an external source worth citing right here, since AGENTS.md wants one beside a claim about the right shape. RFC 9110 §15.5.6: "The origin server MUST generate an Allow header field in a 405 response containing a list of the target resource's currently supported methods." That is this PR's argument in the spec's own words — real's list describes real's resource, not this one's.

The same paragraph is also the one question the body leaves open. On the sub-resources that get no header at all the MUST is unmet, and §10.2.1 names the option the body's table does not consider: "An empty Allow field value indicates that the resource allows no methods." I checked that an empty value survives this stack — with `headers={"Allow": ""}` the client reads `Allow: ''` rather than no header at all. Sending it or not is a judgement call; the issue raised RFC 9110 explicitly, so the docstring should say which was chosen and why.

tests/test_s3.py
```diff
@@ -593,21 +593,24 @@ def test_what_does_not_exist_is_reported_before_the_subresource_except_for_list_
     assert err.code == 501 and b"NotImplemented" in err.read()
 
 
-def test_head_with_a_subresource_is_405_and_a_bare_head_still_answers(live_server):
+def test_head_with_a_subresource_is_405_and_names_what_backlot_serves_in_allow(live_server):
```
nuri-yoo: The rename drops "and a bare head still answers", and the loop that clause named is still in the function — `for path in ("/s3/eng-artifacts", OBJECT_PATH)` at the end, asserting 200. The name now covers the first half only.

tests/test_s3.py
```diff
@@ -593,21 +593,24 @@ def test_what_does_not_exist_is_reported_before_the_subresource_except_for_list_
...
     token = settings.admin_token
-    for path in (
-        "/s3/eng-artifacts?versioning",
-        "/s3/eng-artifacts?uploads",  # served on a GET, and still no HEAD form
-        f"{OBJECT_PATH}?acl",
-        f"{OBJECT_PATH}?uploadId=x",
+    # See `_head_refusal` for which selectors get `Allow: GET`.
+    for path, allow in (
+        ("/s3/eng-artifacts?location", "GET"),
+        ("/s3/eng-artifacts?uploads", "GET"),  # served on a GET, and still no HEAD form
```
nuri-yoo: `?location` on the line above is served on a `GET` too, so a comment on this row alone reads as a distinction between the two that does not exist. Moving it up, or attaching it to the pair, keeps it true.

tests/test_s3.py
```diff
@@ -593,21 +593,24 @@ def test_what_does_not_exist_is_reported_before_the_subresource_except_for_list_
...
+        ("/s3/eng-artifacts?versioning", None),
+        (f"{OBJECT_PATH}?acl", None),
+        (f"{OBJECT_PATH}?uploadId=x", None),
         # Before the bucket or the key is looked up, as on real S3.
-        "/s3/no-such-bucket?versioning",
-        "/s3/eng-artifacts/no/such.md?acl",
+        ("/s3/no-such-bucket?versioning", None),
```
nuri-yoo: Worth one more row: `("/s3/no-such-bucket?location", "GET")`. The two missing-target rows pin that the refusal comes before the lookup, but with `None` on both they do not pin that the *header* does too. Real answers 405 `Allow: GET` for a bucket that does not exist (measured against `s3.ap-northeast-2.amazonaws.com`, 2026-09-17) and Backlot already agrees, so it is a free row.

nuri-yoo (commented):
Reviewed at `1642674`, with the vendor side re-measured rather than read off the issue. Nothing here blocks: the change is the right call, and the test pins it in both directions. Two things to fix, and three follow-ups filed.

**What I checked.**

- Reproduced all nine rows of the issue's table independently, against `s3.ap-northeast-2.amazonaws.com` on 2026-09-17 (path-style, SigV4 with `x-amz-content-sha256`). Every row matches what the body has, including the two-selector 400 carrying no `Allow`, and a bucket that does not exist still answering `Allow: GET` for `?location`.
- The boundary this PR draws is right as well: `?session`, `?Location`, `?list-type=2`, `?foo=bar` and `?x-id=GetObject` are 200 on real exactly as they are here; `?uploads&max-uploads=abc` is a 405 on real, so refusing before the parameter is parsed matches; `?location&x-id=…` and `?uploads&session` keep the header.
- `tests/test_s3.py` 72 passed. `ruff check .` and `ruff format --check` clean. `scripts/gen_docs.py --check` exit 0. `backlot diff --source s3` prints "42 divergence(s) · 42 acknowledged · 0 new", so the body's line holds, and leaving the baseline alone is right: `s3_probe` never reads a response header, so a hand-written `acknowledged` entry would come back on the next run as "acknowledged but no longer diverging".
- Mutation in both directions: removing the header, and widening the tuple to `versioning`, each turns the test red.
- Nothing can regress through an SDK: botocore declares `HeadBucket` and `HeadObject` as its only `HEAD` operations and neither carries a selector, so no boto3 or s3fs path reaches this code.
- Naming only what Backlot serves also matches what the repo already does elsewhere — `docs/supported-sources.md` says a repository "carries a URL template for each sub-resource Backlot actually serves, and none for the ones it doesn't".

**One claim to scope.** "Status codes and the empty `application/xml` body already matched real on every row and are unchanged; `Allow` was the whole divergence" is wider than the measurement under it. Real's 405 also carries `x-amz-request-id` and `x-amz-id-2` on every row, and Backlot's carries neither — its only request id anywhere is the one `object_get` builds. Not this PR's doing (that is #269), but the sentence closes the surface as if it had been measured whole, and the next reader will take it that way. Naming those two, or scoping the sentence to what the issue measured, keeps it true.

**Two things I could not anchor inline.**

- `test_head_with_two_subresources_is_the_conflicts_400_with_an_empty_body` asserts nothing about `Allow`. Row 9 of the issue exists so that a fix does not add one on that path, and the body says the path is unchanged; one `assert err.headers.get("Allow") is None` is what keeps that true tomorrow rather than only today.
- `docs/fidelity.md`'s "Served, not declared" section collects this exact class — a fact the schema diff cannot see, each with a line saying where the record lives ("The tests are the record here … not the baseline"). The `Allow` real sends and Backlot deliberately does not is a third entry for it. Optional.

**Filed from this review**, each measured on the same day, none of them in this PR's scope: #267 (a bucket sub-resource selector at a key path is answered with the object, which is the one that interacts with `_head_refusal` — see the inline note), #268 (five unserved methods on a bucket path share one FastAPI JSON 405 naming `HEAD` in `Allow`, where real answers each of them in its own words: `PATCH` is its own 405 with `Allow: HEAD, DELETE, POST, GET, PUT`, `POST` is a 412, `DELETE` is a 204), #269 (request ids).

One housekeeping thing: the body ends with a `claude.ai/code/session_…` link. Nobody but its author can open it, and none of the last twelve merged PRs here carries one.

backlot/routers/s3.py
khj809: Both halves fixed. `_head_refusal` now takes the served set from the caller — `head_bucket` passes `_BUCKET_GETS`, `object_get` passes nothing — so the header is keyed on the resource, not on the selector name. With `location` added to `_OBJECT_SELECTORS` by hand, the object path answers no `Allow` and the bucket path still answers `GET`, so #267 can land without this line becoming a false claim.

`_BUCKET_GETS` is the single spelling: `bucket_get`'s refusal is now `selected[0] not in _BUCKET_GETS` rather than a trailing `if selected:` after the two `==` branches, so the 501 and the `Allow` read the same set. Widening it to `versioning` without a branch beside it turns three tests red rather than passing quietly.

backlot/routers/s3.py
khj809: Both taken. "every method" is back to "the methods", and §15.5.6 is quoted in the docstring and the body — it is the argument in one line, so thank you.

On the empty value: kept as no header, and the docstring now says so and why. Real sends an empty `Allow` on none of these rows (re-measured 2026-09-17, ap-northeast-2), so `Allow: ""` would be a value real never sends, where the absent header is a gap of the kind `docs/fidelity.md` already records. §10.2.1 and your finding that it survives the stack are both in the docstring so the choice is legible rather than implicit.

tests/test_s3.py
khj809: Renamed to `test_head_with_a_subresource_names_what_a_get_serves_and_a_bare_head_still_answers` — both halves back.

tests/test_s3.py
khj809: Right. The `?versioning` and object rows moved out of this table entirely (they are asserted beside their own 501 now, see below), so the pair that is left is `?location` and `?uploads` and the clause covers them together above the table.

tests/test_s3.py
khj809: Added, with your measurement and its date in the comment beside it — my own run against `ap-northeast-2` on 2026-09-17 agrees, 405 `Allow: GET` for `?location` on a bucket that does not exist.

The unserved rows went the other way: `assert head.headers.get("Allow") is None` is folded into `test_an_unimplemented_bucket_subresource_...` and `test_an_unimplemented_object_subresource_...`, which already parametrize over exactly the selectors whose GET is a 501. That is your invariant as a test — the header check now runs over all 25 bucket and 8 object selectors, driven by the same list that proves the 501, instead of two hand-picked rows.

khj809:
Pushed as d959d9e. Everything above is in, plus the two things you could not anchor inline and the housekeeping.

- **The scoped claim.** The body no longer closes the surface. It now says `Allow` is the divergence *the issue measured*, names `x-amz-request-id` and `x-amz-id-2` as what real also sends and Backlot does not, and points at #269. Confirmed on my own run: every 405 row carries both.
- **The two-selector 400.** `assert err.headers.get("Allow") is None` is in `test_head_with_two_subresources_is_the_conflicts_400_with_an_empty_body`. Adding an `Allow` to that path turns it red.
- **`docs/fidelity.md`.** Third entry written under "Served, not declared", beside GitHub's `HEAD` and the `x-ratelimit-*` headers, with your reasoning about why a hand-written acknowledgement would come back as "acknowledged but no longer diverging" and `tests/test_s3.py` named as the record.
- **The session link** is gone from the body.

Re-measured 2026-09-17 against `s3.ap-northeast-2.amazonaws.com` before changing anything: all nine rows, the missing-bucket `?location`, and the empty-`Allow` question all land where you said.

Full suite on an all-extras sync: 3058 passed, 1 skipped — the two `test_mcp.py` failures the body used to disclaim were the sandboxed TLS fetch and pass here. `ruff`, `gen_docs.py --check` and `backlot diff --source s3` unchanged and clean.

khj809:
Rebased onto main (`7d9f1ce`) — the bucket-listing change in #204 had landed and conflicted on `bucket_get`'s tail. The resolution keeps main's `_list_objects(..., v2=v2, max_keys=max_keys)` and the `resource` local, with the `_BUCKET_GETS` refusal in the one place it now lives. CI green on 3.11–3.14, and the three mutations still go red (`tests/test_s3.py`: 84 passed, full suite 3104 passed / 1 skipped).

backlot/routers/s3.py
```diff
@@ -282,16 +286,29 @@ def _conflict(selected: list[str], resource: str) -> Response:
     )
 
 
-def _head_refusal(selected: list[str]) -> Response:
+def _head_refusal(selected: list[str], served_on_get: frozenset[str] = frozenset()) -> Response:
```
nuri-yoo: nit, not blocking. The default is what keeps the object path honest, so the decision now lives at the call sites and the signature does not ask for it. I confirmed your measurement here: with `location` in `_OBJECT_SELECTORS`, `HEAD /{bucket}/{key}?location` is 405 with no `Allow` while `HEAD /{bucket}?location` still answers `Allow: GET`.

What is left is the other direction. Passing `_BUCKET_GETS` from `object_get` would turn no test red today, because no bucket-named selector is in `_OBJECT_SELECTORS` yet, so the false `Allow: GET` this parameter exists to prevent would arrive silently in exactly the commit that closes #267. Making the argument required, with `object_get` passing `frozenset()` in so many words, puts the choice where the resource is known and costs one word.

nuri-yoo (approved):
Approving. All nine points from the first round are in the code rather than only in the replies, and I re-ran the claims I could not take on faith at `7d9f1ce`:

```
tests/test_s3.py                                     84 passed
ruff check .                                         exit 0
ruff format --check .                                exit 0
PYTHONPATH=$PWD python scripts/gen_docs.py --check   exit 0
_BUCKET_GETS widened with "versioning"               3 failed
"location" in _OBJECT_SELECTORS, then a HEAD on it   the key: 405, no Allow
                                                     the bucket: 405, Allow: GET
```

That last row is the finding from the first round actually closed, and not by the selector list happening to be short. Folding `Allow is None` into the two parametrized tests is also more than the two rows I asked for: the header is now checked over all 25 bucket and 8 object selectors, driven by the list that proves the 501.

One nit inline, and one here, neither blocking: the Verification block still reads `3058 passed, 1 skipped`, which is the run before the rebase. The last run is the `3104 passed, 1 skipped` in your comment.

backlot/routers/s3.py
khj809: Taken — `served_on_get` is required now, and `object_get` passes `frozenset()` in so many words with the reason beside it: a key's path serves no sub-resource on a GET, so its 405 names no method even for a selector a bucket's path does serve. The commit that closes #267 now has to state the choice rather than inherit a default.

The stale count is fixed too: the Verification block reads `3104 passed, 1 skipped`.

PR #244, opened 2026-09-16
s3: a HEAD 405 on a sub-resource carries an Allow naming the methods Backlot takes

### LINKED ISSUE #230: s3: a HEAD carrying a sub-resource is a 405 with no Allow, where real names the methods that sub-resource takes

khj809:
Closes #230.

## What changed

`_head_refusal` in `backlot/routers/s3.py` answered a `HEAD` carrying a sub-resource selector with an empty `405` and no `Allow` header, for every selector alike. Real S3 sends an `Allow` naming the methods that sub-resource takes. Backlot now sends `Allow: GET` for the two bucket selectors it answers on a GET — `location` (GetBucketLocation) and `uploads` (the always-empty ListMultipartUploads page) — and no `Allow` for every other selector, bucket or object, since Backlot takes none of their methods. The two-selector conflict's `400` is unchanged: no `Allow` there either, on either side.

Measured against `s3.amazonaws.com` (the `backlot-measurement` bucket, which real S3 places in `us-east-1` regardless of the `AWS_DEFAULT_REGION` this environment sets), path-style, SigV4, 2026-09-16:

```
HEAD /?location             -> 405, Allow: GET
HEAD /?uploads               -> 405, Allow: POST, GET
HEAD /?versioning            -> 405, Allow: GET, PUT
HEAD /?acl                   -> 405, Allow: GET, PUT
HEAD /?tagging                -> 405, Allow: DELETE, GET, PUT
HEAD /?policy                 -> 405, Allow: DELETE, GET, PUT
HEAD /?versions                -> 405, Allow: GET
HEAD /?acl&versioning           -> 400, no Allow
```

This reproduces every bucket-level row of #230's table exactly. The bucket-level `acl` and `tagging` rows also match the object-level rows the issue measured on 2026-09-15 (`GET, PUT` and `DELETE, GET, PUT` respectively), which is expected since ACL and tagging take the same methods on a bucket or an object; the object-level rows themselves could not be re-measured here because the credential in this environment holds no `s3:PutObject` on the measurement bucket, so no object exists to address.

Backlot's own `Allow` is not a copy of real's list. Real's `Allow: POST, GET` on `?uploads` names `CreateMultipartUpload`, which Backlot does not implement; sending it would advertise a write Backlot refuses. Backlot names only the methods it itself takes for a selector — `GET` for `location` and `uploads`, nothing for the rest — which is the reading `_head_refusal`'s own docstring already commits to ("Backlot takes none of them, so it sends none"; that sentence is now true only for the selectors it still takes none of).

## Verification

- **Tests** — `pytest` on a `.[all]` install (`uv sync --all-extras --locked`): 3054 passed, 3 skipped.
- **Optional surfaces that ran rather than skipped** — `tests/test_s3.py`, `tests/test_sdk.py`, `tests/test_mcp.py` (all but two pre-existing, environment-only failures below), and the reader, Google and mirage tests in `tests/test_integrations.py`. Skips: `tests/test_llamaindex.py:121` (`llama_index.readers.hubspot`, not published for the pinned llama-index) and two `tests/test_mcp.py` cases (`test_mcp_notion_acl_enforced`, `test_mcp_s3_lists_objects`) deselected — both fail identically on `main` with no change of mine applied, from `uvx` unable to fetch its MCP server package through this sandbox's TLS-intercepting proxy (`SELF_SIGNED_CERT_IN_CHAIN` / `invalid peer certificate: UnknownIssuer` against `pypi.org`/`registry.npmjs.org`), unrelated to S3 or this fix.
- **Fidelity** — `backlot diff --source s3 --json`: 42 compared, 0 new, 0 resolved — the probe checks status and XML shape, not headers, so it is unaffected either way.

- [x] A new endpoint serving corpus content is ACL-scoped, proved for both an admin and a scoped token — not applicable, no new endpoint; this changes a header on an existing refusal

https://claude.ai/code/session_01DoHFq2cG5QbLEV9BwY8upA

---
_Generated by [Claude Code](https://claude.ai/code/session_01DoHFq2cG5QbLEV9BwY8upA)_


khj809:
Duplicate of #241, which closes #230 and was opened first. A second run took the issue while #241 was already open; the survey rule that allowed it is corrected in #242.

Issue #225, opened 2026-09-15
jira: every body is application/json, where real names a charset the Confluence routes do not


nuri-yoo:
**Source and commit** — `jira`, at `463d180` (the branch for #203).

Every Jira body Backlot serves is `application/json`. Real Jira sends `application/json;charset=UTF-8` — with no space after the semicolon — on every route and status measured. Confluence, served by the same router, really does send the bare `application/json`, so this is Jira's alone and the seam that already draws that kind of line is `backlot.errors.github.json_media_type`.

| # | Response | Real | Backlot | Silent? |
|---|---|---|---|---|
| 1 | Jira 200 (`myself`, `serverInfo`, `project/search`) | `application/json;charset=UTF-8` | `application/json` | yes |
| 2 | Jira 400 / 401 / 404 | `application/json;charset=UTF-8` | `application/json` | yes |
| 3 | Confluence 200 and 404 | `application/json` | `application/json` | — correct |

Silent in both: `backlot diff` compares operations and parameters against the vendor's OpenAPI description, and a response header is neither.

**How you know** — measured against a live Jira Cloud site on 2026-09-15 with an API token.

- `rest/api/3/myself` (200), `rest/api/3/search/jql?jql=NOT+A+JQL` (400), `rest/api/3/issue/NOPE-999999` (404), the same issue path under `rest/api/2` (404), and `rest/api/3/myself` with no credential (401) each answered `application/json;charset=UTF-8`. The spelling is exact and unusual: no space after the `;`, and `UTF-8` upper-case — where GitHub's, already modelled here, is `application/json; charset=utf-8` with a space and lower-case.
- Every one of those bodies was compact, on one line, with no trailing newline, which is what Backlot already sends.
- Confluence's own REST — `wiki/rest/api/space`, `wiki/rest/api/space/NOPESUCHSPACE` (404), `wiki/rest/api/search`, `wiki/rest/api/content/999999999` (404), which are the Confluence routes Backlot serves — answered the bare `application/json`, compact, on all four. So the divergence does not extend to them, and a fix must not give them a charset.
- The 401 body on Jira is `Client must be authenticated to access this resource.` as plain text under that JSON content-type, which is real being inconsistent rather than something to copy; this issue claims nothing about it.

**Reproduction** — the bundled corpus, repo `pipeline`:

```python
import httpx
from backlot import serve

with serve() as s:
    tok = httpx.get(f"{s.base_url}/_meta/users").json()["admin_token"]
[...]

PR #247, opened 2026-09-16
fix(jira): send application/json;charset=UTF-8 on JSON bodies

HarshRajSinghania:
Closes #225.

## Summary
Jira Cloud answers JSON with `application/json;charset=UTF-8` (no space after the semicolon, `UTF-8` upper-case). Confluence, served by the same Atlassian router under `/wiki`, answers the bare `application/json`.

## Motivation
#225 measured that split against a live Jira Cloud site. Backlot previously sent `application/json` for both products. Clients or fixtures that compare the `content-type` string lock in a value real Jira never sends. The existing `errors.json_media_type` / `vendor_json_media_type` seam is how GitHub already names its charset.

## Implementation
- Add `json_media_type` to `backlot.errors.atlassian`, using the existing `is_confluence` predicate.
- Jira paths return `application/json;charset=UTF-8`.
- Confluence paths return `None`, so FastAPI's default is left alone.
- The middleware only rewrites a response whose type is exactly `application/json`, so Jira's RFC 7807 `application/problem+json;charset=UTF-8` bodies are unchanged.

## Testing
Ran:
- `pytest tests/test_atlassian.py::test_jira_json_bodies_name_a_charset_that_confluence_does_not tests/test_atlassian.py::test_confluence_refuses_an_integer_parameter_it_cannot_convert tests/test_github.py::test_github_json_carries_the_charset_real_sends_except_on_code_search` — 4 passed
- `ruff check` on the touched Python files — passed
- `ruff format --check` on the new atlassian code — passed
- `python scripts/gen_docs.py --check` — exit 0

The full suite was not run here; the change is confined to the Atlassian media-type hook plus the GitHub test that previously asserted Jira stayed on the bare type.


backlot/errors/atlassian.py
```diff
@@ -60,6 +60,30 @@ def is_confluence(path: str) -> bool:
...
+    ``application/json``, so a Jira RFC 7807 body on ``application/problem+json`` is
+    untouched.
+
+    ``status_code`` is unused: the charset does not vary by status on the routes this
+    server serves. It stays in the signature so the dispatch in ``errors.json_media_type``
+    can treat every envelope alike.
+    """
+    if is_confluence(path):
+        return None
+    return JIRA_JSON_MEDIA_TYPE
```
khj809, line 84: Based on our measurement (2026-09-16), Jira's 403 response does not append the charset on content-type.

```
$ curl -si -H 'Authorization: Bearer zzz' https://<site>.atlassian.net/rest/api/3/serverInfo
    HTTP/2 403
    content-type: application/json

    {"error": "Failed to parse Connect Session Auth Token"}
```

The fix would be quite simple:

```suggestion
    if is_confluence(path) or status_code == 403:
        return None
    return JIRA_JSON_MEDIA_TYPE
```

That also makes the docstring's "`status_code` is unused" false, which is what currently hides this case.

docs/supported-sources.md
```diff
@@ -251,6 +251,9 @@ than there being a set per type.
 `search/jql`, `issue/{key}`, `issue/{key}/comment`, `field` and `serverInfo` are served under
 `rest/api/2` as well as `/3`.
 
+Jira JSON responses use `application/json;charset=UTF-8` (no space after the semicolon).
```
khj809, line 254: According to the above measurement, would be better to rephrase it as below:

```suggestion
A JSON body is `application/json;charset=UTF-8` — no space after the semicolon, `UTF-8`
upper-case — as real's is on every route and status measured, except where real answers a
different type altogether: a type-conversion failure is RFC 7807 on
`application/problem+json;charset=UTF-8`, and the gateway's 403 for a bearer it cannot read as
a Connect token is the bare `application/json`.
```
```

File: docs/supported-sources.md [hunk 3]
```diff
@@ -251,6 +251,9 @@ than there being a set per type.
...
+Confluence JSON responses under `/atlassian/wiki` stay `application/json`.
```
khj809, line 255: Could you move the Confluence-related sentence under `### Confluence` section at L35?

khj809 (commented):
Hey @HarshRajSinghania, thanks for your first contribution!

I checked this against a live Jira and Confluence site, and the core split holds up — the charset on Jira, the bare type on Confluence, and the exact spelling. I left a few notes below; the first one is a regression I think is worth closing before this merges.

We'd also love to hear what you're using Backlot for and what you'd like to see added — GitHub issues or Discord, whichever you prefer!

khj809:
Due to the inactivity in this PR, we decided to supersede it with #284. Sorry for the quick decision, but you're still counted as our contributor. Thanks again, and we'd be glad to see you on the next one!

Issue #197, opened 2026-09-11
slack: `edited.user` is a Slack id the corpus has to invent, the value #164 stopped asking for in a reaction


nuri-yoo:
**Source and commit** — `slack`, at `b30ef32` (on the branch for #164).

#164 replaced the Slack user ids in `reactions.users` with addresses, on the ground that a `synth.slack_user_id` value is one no corpus author can compute. The same class of value is still hand-written in `edited`: the schema types it as a bare object ("`{user, ts}` if the message was edited", on a root message and on a reply alike), and `_message` passes the stored object through at `backlot/routers/slack.py:1094-1096`. The bundled sample does exactly that, twice: `"edited": {"user": "UC20FA2B1C0", "ts": "1770311280.000000"}` on its first Slack message and on a reply of its second. That id is `synth.slack_user_id("ava@acme.com")`, so on the sample the value was computed rather than invented; a BYO author has no such function to hand.

Three records validate today and each serves a different thing:

| # | Corpus writes | Import | Served `edited.user` | `users.info` on it |
|---|---|---|---|---|
| 1 | `{"user": "UC20FA2B1C0", "ts": …}` (the sample) | accepted | the id, verbatim | `ok`, `ava@acme.com` |
| 2 | `{"user": "ava@acme.com", "ts": …}` | accepted | the address, verbatim | `user_not_found` |
| 3 | `{"anything": 1}` | accepted | `{"anything": 1}` | — |

Row 2 is the record an author writes after reading #164's `reactions` shape, and it serves a value that fails Slack's own `defs_user_id` pattern (`^[UW][A-Z0-9]{2,}$`). Row 3 is the open object the reaction item was closed against in #164 (ff29c7a). `files` is typed the same way ("Slack files objects", a bare array) and is the same class of gap; this issue is about `edited` because the sample exercises it.

**How you know** — spec only; no live workspace was called. Slack's `message` event reference says "If the message has been edited after posting, it will include an `edited` property, which includes the user ID of the editor and the timestamp for when the edit occurred", with the example `"edited": {"user": "U123ABC456", "ts": "1355517536.000001"}`. The id's type is `defs_user_id` in Slack's published OpenAPI, the definition #164's schema description already cites for `reactions.users`. The three rows above were measured against this branch on 2026-09-11.

**Reproduction** — the sample corpus, its first Slack record rewritten the way #164 teaches for reactions:

```python
import json, re, backlot, requests

[...]

Issue #256, opened 2026-09-17
slack: a `files` entry can carry the same invented ids `edited` and `reactions.users` did


khj809:
Found while measuring for #197.

**Source and commit** — `slack`, at `46f3b43` (`main`).

**How you know** — Slack's own OpenAPI spec (`slack_web_openapi_v2.json`), fetched live 2026-09-17: `objs_file.editor` and `objs_file.last_editor` are typed `defs_user_id` (`^[UW][A-Z0-9]{2,}$`) — the same pattern `reactions.users` (#164) and `edited.user` (#197) were already found to require, and a value a corpus author has no way to invent. `objs_file.user` (the file's owner) is left untyped in the spec, but every file Slack itself serves populates it with a user id too.

`backlot/schemas/slack.schema.json`'s `files` property is a bare array with no `items` schema at all (`"description": "File attachments (Slack files objects)."`), and `_message` in `backlot/routers/slack.py` passes whatever a corpus writes straight through. Nothing types or renders `user`/`editor`/`last_editor` the way `_reactions` and `_edited` now render their own id fields from an address.

**Reproduction** — no bundled corpus states any of these three fields today (`examples/bring-your-own-corpus/sample_corpus.jsonl`'s two `files` entries carry only `id`/`name`/`mimetype`/`title`), so nothing currently served is wrong. The gap is that a record which does state one imports and serves it completely unchecked:

```python
from backlot.validation import record_errors
rec = {
    "source_type": "slack", "channel": "incidents", "content": "x",
    "author_email": "a@x.com", "created": "2026-03-01T09:00:00Z",
    "files": [{"id": "F1", "name": "x", "user": "U01", "editor": "not-a-real-id-at-all"}],
}
record_errors(rec)  # -> [] — both the invented id and the garbage string are accepted
```

**Can a client tell?** A `200` with a wrong or unvalidatable `user`/`editor`/`last_editor` is the expensive case: import raises nothing, `--dry-run` names nothing, and a client that resolves the id via `users.info` or validates the response against Slack's spec finds out only once it runs against the real workspace.

brekkybot:
loop: claimed by run https://claude.ai/code/session_01PY4QTQTTPdQn4HNtUFH3jn

---
_Generated by [Claude Code](https://claude.ai/code)_

brekkybot:
Closed by #291: `files[].user` now renders through `synth.slack_user_id`, confirmed live against `brekkylab.slack.com`.

---
[...]

PR #257, opened 2026-09-17
slack: an edited message's `user` is rendered from the address the corpus writes, matching `reactions.users`

khj809:
Closes #197.

## What changed

`edited` on a Slack message (root or reply) now declares its shape instead of accepting any object. The schema requires `user` — an address, not an id — and `ts` in Slack's own `defs_ts` shape (`^\d{10}\.\d{6}$`, so a bare JSON number and a short `"1770311280.0"` are both refused), and refuses any other key. `_message` in `backlot/routers/slack.py` renders `edited.user` from that address with `synth.slack_user_id`, the same way `_reactions` already renders `reactions.users` — a corpus names the editor, the response mints the id Slack's own spec types it as.

Two rules a schema cannot state, because each reads a second field of the same record, are checked in `backlot/importer/byo.py` before any row is written: `edited.user` is the message's own author, and `edited.ts` names a later SECOND than that message's own `created` (a root's own clock, or a reply's own — not the root's).

The two sample-corpus records that carried a hand-computed id (`synth.slack_user_id("ava@acme.com")`, written by hand as `UC20FA2B1C0`) now write `ava@acme.com` instead, so no invented user id is left anywhere the way #164 already made true of `reactions`.

| request | real | Backlot |
|---|---|---|
| every live `user`/`parent_user_id`/`reply_users` value read off `brekkylab.slack.com` today (`conversations.history`/`conversations.replies`, via this session's own `SLACK_USER_TOKEN`) | a `defs_user_id` (`^[UW][A-Z0-9]{2,}$`), e.g. `U08QJADP89Y` — never an address | same, everywhere Backlot renders one |
| Slack's own OpenAPI spec (`slack_web_openapi_v2.json`), fetched live 2026-09-17 | `defs_user_id` pattern `^[UW][A-Z0-9]{2,}$`, `defs_ts` pattern `^\d{10}\.\d{6}$`; no definition in the spec types `edited` at all (spec-only field, matched by #197's own finding) | schema requires `edited.user` to be an address, same class as `reactions.users`, and `edited.ts` to match `defs_ts` |
| `chat.update`, the only Web API method that gives a message an `edited` block (a human editing in the client is the other path — the same page notes the `(edited)` label "renders only on messages edited by humans") | "Only messages posted by the authenticated user are able to be updated using this method"; anything else answers `cant_update_message` ("Authenticated user does not have permission to update this message") | the importer refuses an `edited.user` that is not the record's own `author_email` |
| Slack's message event reference (prose, since the spec doesn't type it) | "includes the user ID of the editor and the timestamp for when the edit occurred", example `"edited": {"user": "U123ABC456", "ts": "1355517536.000001"}` on a message whose own `user` is the same `U123ABC456` | `_edited` renders exactly that shape from the address, for an editor who is the author |

The later-second rule is the served `ts`'s own doing rather than a vendor statement: `synth.slack_fmt_ts` takes the six-digit fraction from a hash of the message, so a corpus author writing `edited.ts` cannot tell whether a fraction inside the message's own second falls before or after it. Only a later second is ordered after the message from outside, which is what the check, the schema's two `ts` descriptions and the refusal message all say.

This session's Slack credential (`SLACK_USER_TOKEN`, read-only: `identify, channels:history, channels:read, search:read, users:read`, no `chat:write`) was used to scan every channel's full history on the real `brekkylab` workspace for a message that already carries `edited` — none exists, and the credential cannot create one to observe first-hand. The fix instead follows #164's already-established precedent for the identical shape: `reactions.users` was fixed the same way, on the same `defs_user_id` evidence, without a live-created reaction either — real's own answer for the reactor ids didn't change, only the corpus's input contract did, which is exactly this fix's shape too. `reactions` reaches the opposite answer on who may appear, and for a reason: `reactions.add` works for anyone, so a reactor who is not the author is a state real Slack holds, while an editor who is not the author is one it refuses.

`tests/test_schema.py` covers each of the four schema rules — old id spelling refused, both keys required, unknown keys refused, `ts` shape refused — on a root and on a reply, since `backlot/schemas/` carries no `$ref` and the two blocks are hand-kept copies. `tests/test_importer_byo.py` covers both import checks, each on a root and on a reply. `tests/test_slack.py::test_slack_edited_renders_the_editors_id` covers the rendered response on a root and on a reply, which `_message` serves for history, replies and search hits alike.

A related gap in `files` — Slack's own spec types `editor`/`last_editor` the same `defs_user_id` way, and Backlot's `files` accepts anything unchecked — is outside this issue (no bundled corpus states one) and filed as #256. The sample corpus's two `files` entries state ids of their own (`FD09F114C47` and `F0A12B34C56`, both of which `defs_file_id`'s `^[F][A-Z0-9]{8,}$` accepts and `_message` passes through verbatim); Backlot mints no file ids and serves no `files.info`, so nothing contradicts them, and #256 is where they are answered.

## Verification

- **Tests** — `pytest -rs` (the repo's own settings) at `881ab8a` on a `uv sync --all-extras --locked` install: 3288 passed, 1 skipped, exit 0.
- **Optional surfaces that ran rather than skipped** — the one skip is `tests/test_llamaindex.py::test_hubspot`, which `importorskip`s a reader `pyproject.toml` deliberately leaves out of the extras (it pins `hubspot-api-client<9`); every other optional surface ran, the two MCP tests included.
- **Every rule is pinned** — deleting each of the four schema rules from the root block and again from the reply block turns a test red, eight for eight; neutralizing the author check, the later-second check, and each of the two call sites does too.

- [x] A new endpoint serving corpus content is ACL-scoped, proved for both an admin and a scoped token — n/a, no new endpoint; `edited` is a field on a message endpoint already ACL-scoped


backlot/importer/byo.py
```diff
@@ -306,6 +306,24 @@ def _thread_seconds(where, root_sec, replies):
     return out
 
 
+def _check_edited_ts(where, edited, created_sec):
+    """An `edited.ts` checked against its OWN message's `created`, before any row is written.
+
+    Beyond what the schema states: real Slack's edit time is always later than the message it
+    edited, so one at or before `created` is not an edit history a corpus author could have
+    observed.
+    """
```
nuri-yoo: `edited.user` can name someone who is not the message's author, and nothing refuses it. Slack cannot produce that message: `chat.update` states "Only messages posted by the authenticated user are able to be updated using this method" and answers anything else `cant_update_message` ("Authenticated user does not have permission to update this message"); the help centre's editing/deletion permissions give owners and admins deletion of a member's message and no editing of one. Slack's own message-event example carries the rule in its values -- `user` and `edited.user` are both `U123ABC456`.

Measured on this head, a record whose author is `ava@x.com` and whose `edited.user` is `mallory@x.com` imports clean and is served:

```
served message user : U65BF17E86F
served edited       : {'user': 'U597CC93604', 'ts': '1772614801.000000'}
users.info(U65BF17E86F) -> ok=True   ava@x.com
users.info(U597CC93604) -> ok=False  user_not_found
```

This PR added the `ts` pattern on exactly this reasoning -- real Slack never sends that shape -- so the one rule the vendor writes down belongs here too. `_reactions` reaches the opposite answer for `reactions.users` with a reason: `reactions.add` works for anyone, so a reactor who resolves to nobody is a real state. An editor has no such reason; the editor is the author, and the author always resolves.

This function already holds `rec` (and `rep` at the reply call site), so the check is a sibling of the one above it, and the sample corpus already satisfies it -- both of its `edited` entries name the record's own author.

It also decides `tests/test_slack.py:1470`, which makes editor `bo@x.com` on a message authored by `ava@x.com` the canonical rendered example. If the rule lands, that fixture becomes the author; if you read the vendor differently, the test is where the reason to keep a different editor should be stated.

backlot/schemas/slack.schema.json
```diff
@@ -185,7 +202,24 @@
...
+                "pattern": "^\\d{10}\\.\\d{6}$",
+                "description": "When it was last edited, as a Slack ts (`defs_ts`: `^\\d{10}\\.\\d{6}$` — every live `ts` this project has read is a quoted string in this shape, never a bare number). Stated rather than derived — an edit time is information the author holds, not a value the response can work out from something else — and checked at import to fall after this reply's own `created`."
+              }
+            },
+            "required": [
+              "user",
+              "ts"
+            ],
+            "additionalProperties": false,
+            "description": "Present if the reply was edited. Slack's message event reference: \"includes the user ID of the editor and the timestamp for when the edit occurred\"."
```
nuri-yoo: Three of this block's four rules are unpinned. Deleting each from the REPLY copy alone and running `tests/test_schema.py tests/test_importer_byo.py tests/test_slack.py` (444 tests, the three files that touch `edited`):

```
reply: remove ts pattern               -> 444 passed
reply: remove additionalProperties     -> 444 passed
reply: remove required                 -> 444 passed
reply: remove user format: email       -> 1 failed, 443 passed
root:  remove ts pattern               -> 1 failed, 443 passed
root:  remove additionalProperties     -> 1 failed, 443 passed
root:  remove required                 -> 1 failed, 443 passed
root:  remove user format: email       -> 1 failed, 443 passed
```

`backlot/schemas/` carries no `$ref` anywhere (0 across the eleven files), so this block and the root's are a pair kept in step by hand -- `reactions` is byte-identical between the two today. A test is the only thing that notices when the copies drift, and only one of the four rules has one.

`test_slack_edited_refuses_the_id_a_corpus_cannot_know` is the one that does, and extending the other three the same way is the same two lines each. The PR body's "old id spelling refused, both keys required, unknown keys refused, on a root and on a reply" is true of the first item alone.

backlot/importer/byo.py
```diff
@@ -306,6 +306,24 @@ def _thread_seconds(where, root_sec, replies):
...
+    if not edited:
+        return
+    ts = edited["ts"]
+    sec = _epoch(ts)
+    if sec is None or sec <= created_sec:
+        raise SystemExit(
+            f"{where}: edited.ts must be after this message's own created "
+            f"(got {ts!r}, created at {created_sec})"
+        )
```
nuri-yoo: The rule implemented here is a whole later second, but the three sentences that describe it all say "after `created`". `_seconds` truncates with `int(n)`, so every `edited.ts` inside the message's own second is refused whatever its fraction, and the refusal reads as a contradiction:

```
line 1: edited.ts must be after this message's own created (got '1772355600.500000', created at 1772355600)
```

`1772355600.500000` is after `1772355600`, so a corpus author who wrote a typo fix half a second later is told something that is not true of the value they wrote.

The rule itself holds up, and for a reason none of the three sentences gives: the served `ts` carries a hashed micro-fraction (`1770310800.146538` for the sample root), so a corpus author cannot know whether `.500000` lands before or after the message's own ts, and only a later second is safe from outside. Worth saying that here, in the schema's two `ts` descriptions, and in the message -- "must name a later second than this message's own created" refuses the same values and explains itself.

Separately, `sec is None` is unreachable. `record_errors` runs at the top of the same method (`byo.py:1763`) and the pattern refuses anything `_epoch` cannot read, so the branch can only be belt-and-braces -- in which case a parse failure currently borrows the ordering message.

tests/test_slack.py
```diff
@@ -1446,3 +1446,35 @@ def test_slack_reaction_ids_and_count_are_derived_from_the_addresses(tmp_path):
     ]
     # Slack's own `defs_user_id`, so an id Backlot mints is one the vendor's spec would accept.
     assert all(re.fullmatch(r"[UW][A-Z0-9]{2,}", u) for r in reactions for u in r["users"])
+
+
+def test_slack_edited_renders_the_editors_id(tmp_path):
```
nuri-yoo: A reply's `edited` is never rendered in a test, and it is the one place this PR stops at the root: the schema tests reach a reply, the importer tests reach a reply, this one does not.

The path does work -- serving the shipped sample corpus over HTTP, `conversations.replies` answers the second record's reply with `edited={'user': 'UC20FA2B1C0', 'ts': '1770746880.000000'}` -- so this is about holding it, not fixing it. `_message` is the single renderer for history, replies and search hits, which is what makes one more assertion cheap.

nuri-yoo (changes requested):
The design is right and the vendor attributions hold. I fetched `slack_web_openapi_v2.json` myself: `defs_ts` is `^\d{10}\.\d{6}$` and `defs_user_id` is `^[UW][A-Z0-9]{2,}$`, exactly as the two schema descriptions quote them, and `"edited"` occurs zero times in the whole spec, so "no definition in the spec types `edited` at all" is accurate too. Serving the shipped sample corpus is byte-identical to before the change, since `synth.slack_user_id("ava@acme.com")` is the `UC20FA2B1C0` the corpus used to state by hand. Four inline comments, two of which I think should land before this merges.

Two notes on the body itself.

**"no invented id is left anywhere"** is wider than the corpus. The same reply record this diff edits keeps `"files": [{"id": "FD09F114C47", ...}]`, and `defs_file_id` in the fetched spec is `^[F][A-Z0-9]{8,}$`, which that value matches; `_message` passes `files` through untouched, so it is served verbatim. It is not harmful the way a user id is -- Backlot mints no file ids and serves no `files.info`, so nothing can contradict it -- but the sentence claims a range the corpus does not meet, and #256, which the body's own last paragraph cites, is the open issue about ids in `files`. "no invented user id" is the claim the diff supports.

**The `ts` shape rule is missing from the body.** A whole commit (`366a076`) and `test_slack_edited_ts_refuses_a_bare_number` are about it, and "What changed" stops at "`user` (an address) and `ts` are both required, and any other key is refused". `ts` must now also match `^\d{10}\.\d{6}$`, so a bare JSON number and `"1770311280.0"` are both refused, and the vendor table has no row for it either.

**Verification on my side**, at `b2fa158`, on a `--all-extras` install:

```
pytest -o addopts="" -q -rs -n 8   -> 3097 passed, 0 skipped
ruff check .                       -> exit 0, All checks passed!
ruff format --check .              -> exit 0, 141 files already formatted
```

`tests/test_mcp.py::test_mcp_notion_acl_enforced` and `::test_mcp_s3_lists_objects`, which the body reports as environment failures, both pass here. Reverting `_edited`, the root `_check_edited_ts` call and the reply `_check_edited_ts` call each turns a test red, so the first checkbox is real. I could not reproduce the live-workspace scan -- this machine has no Slack credential -- so everything above rests on the spec and the published reference.

backlot/importer/byo.py
khj809: Confirmed and implemented. `_check_edited` now refuses an `edited.user` that is not the record's own author, on the root and on the reply. `tests/test_slack.py`'s rendered example makes the editor the author, and `test_byo_slack_edited_user_must_be_the_messages_own_author` covers both positions — it states the `reactions` contrast too, since `reactions.add` works for anyone and an editor has no such reason.

backlot/schemas/slack.schema.json
khj809: Fixed. The four `edited` schema tests are parametrized over root and reply through one `_edited_record` helper. Re-ran your mutation check on the new tests: all eight deletions (both copies × ts pattern / additionalProperties / required / user format) turn a test red.

backlot/importer/byo.py
khj809: Both fixed. The rule now says what it does — "must name a later second than this message's own created" — in the refusal message, the check's docstring and the schema's two `ts` descriptions, each carrying your reason: `synth.slack_fmt_ts` takes the six-digit fraction from a hash, so a corpus author cannot order a fraction inside the message's own second against it. The test's same-second case is now `.999999` so the second is visibly the rule. `sec is None` is gone: the parse goes through `_epoch_field`, which reports its own failure instead of borrowing the ordering message.

tests/test_slack.py
khj809: Added. `test_slack_edited_renders_the_editors_id` now carries a reply with its own `edited` and asserts `_message` on both rows — the root from `list_slack_top_level`, the reply from `slack_thread`.

nuri-yoo (approved):
All four are addressed, and I re-ran the checks rather than reading the diff for them. Approving; one line of the body is worth fixing before this goes in, and it does not block.

**The editor rule.** `_check_edited` refuses a non-author editor on a root and on a reply. Neutralising the check turns two tests red, and a sharper mutation holds too: passing the ROOT's `author` at the reply call site turns three red, so the reply really is measured against its own author rather than the thread's. The sample corpus satisfies it on both of its `edited` records.

**The reply schema.** All eight deletions are caught now — root and reply, times `ts` pattern, `additionalProperties`, `required` and `user`'s `format` — where three of the reply four survived before. The `_edited_record` helper is the right shape for it given `backlot/schemas/` has no `$ref`.

**The later-second rule.** The refusal message, the check's docstring and both schema `ts` descriptions now say what the code does and why, and `<=` to `<` is caught. The parse failure reports itself through `_epoch_field` instead of borrowing the ordering message. The reply test's docstring is also correct now in a way the old one was not: the fixture's `edited.ts` is after the root and before the reply, which is exactly what the new sentence describes.

**The render test** reads a reply row as well as a root, and reverting `_edited` is caught.

## One thing to fix

The replacement for "no invented id is left anywhere" undercounts:

> The sample corpus's **one** `files` entry states an id of its own (`FD09F114C47`, ...)

There are two, and both are hand-written ids in Slack's own shape:

```
record  5  reply 3 : FD09F114C47    defs_file_id ^[F][A-Z0-9]{8,}$ -> matches
record 34  root    : F0A12B34C56    defs_file_id ^[F][A-Z0-9]{8,}$ -> matches
```

Both were already there at `b2fa158`, so the `main` merge did not introduce the second one, and #256 — which this same paragraph cites — says "two `files` entries" in its own reproduction. My own comment named only the one in the record the diff touched, so the count was missing on my side too. "one" to "two", with both ids listed, settles it.

## Not blocking

**`## Verification` predates the head.** Its numbers were measured at `08f6931`: that revision collects 3103 tests, which is the body's 3102 passed plus 1 skipped. The head is the `main` merge `d2b216d`, which collects 3289. Worth one re-run so the block is the last result at head, since the template's checklist asks for exactly that.

**`if author and edited["user"] != author` cannot take the false branch.** `author_email` is required on a root and on a reply, and `format: email` refuses `""` (measured against the same checker the validator uses). That is the same reachability argument that retired `sec is None` in this commit, so the guard is a candidate for the same treatment.

**"`chat.update` is the only way a message acquires an `edited` block"** is wider than the sources. Editing in the Slack client gives a message an `edited` block too, and `chat.update`'s own page says the `(edited)` label "renders only on messages edited by humans". The rule survives either path — the help centre's editing and deletion permissions give owners and admins deletion of a member's message and no editing of one — so "the only Web API method" carries the same weight with a source behind it. The docstring and the vendor table's row label both say it.

**A suggestion for the render test.** Nothing asserts `_message(row)["user"] == _message(row)["edited"]["user"]`. That identity is what Slack's own example shows and what this body quotes, and it is the form the rule takes on the surface a client reads, so one line would pin it where the client sees it.

## Checks on my side

At `d2b216d`, with the repo's own pytest settings:

```
pytest -rs                 -> 3289 passed, 0 skipped, exit 0
ruff check .               -> exit 0, All checks passed!
ruff format --check .      -> exit 0, 141 files already formatted
14 mutations               -> 14 of 14 caught, working tree clean afterwards
```

The 0 skips are my environment rather than a wider run than yours: this machine happens to have `llama-index-readers-hubspot` installed, which `pyproject.toml` deliberately keeps out of the extras, so a clean `--all-extras` install skips `test_llamaindex.py::test_hubspot` exactly as the body says. The mutations are the twelve the body lists plus two more — reverting `_edited`'s rendering, and passing the root's author at the reply call site.

khj809:
Thanks — the `files` count was wrong on my side and is fixed in the body: two entries, `FD09F114C47` (record 5, reply 3) and `F0A12B34C56` (record 34, root), both matching `defs_file_id`'s `^[F][A-Z0-9]{8,}$`, both answered by #256.

All three non-blocking items are in `881ab8a`.

**The guard.** `author and` is gone. Measured the reachability the same way you did: `author_email` is required on a root and on a reply, and the validator answers `''` with `[author_email]: '' is not a 'email'`, so the false branch had nothing to reach. A comment states that where the check reads it.

**"the only way".** Narrowed to "the only Web API method". `chat.update`'s own page carries the other path in the line you quoted — "The (edited) label renders only on messages edited by humans" — so the docstring and the vendor table's row label now say a human editing in the client sets an `edited` block too, and that the rule reads the same either way. I left the help centre out: the editing-and-deletion-permissions article 404s from here, so I can't stand behind restating it.

**The identity.** `test_slack_edited_renders_the_editors_id` now asserts `_message(row)["edited"]["user"] == _message(row)["user"]` on the root and on the reply, next to the `defs_user_id` shape check.

**Verification at `881ab8a`**, `uv sync --all-extras --locked`:

```
pytest -rs                 -> 3288 passed, 1 skipped, exit 0
ruff check .               -> exit 0, All checks passed!
ruff format --check .      -> exit 0, 141 files already formatted
```

The one skip is `test_llamaindex.py::test_hubspot`, the reader `pyproject.toml` keeps out of the extras — same difference you diagnosed. Re-ran your sharper mutation too: passing the root's `author` at the reply call site turns 3 red, neutralizing the author check turns 2 red, and reverting `_edited` turns 1 red.

Issue #198, opened 2026-09-11
slack: no member is ever deactivated, so a workspace with an offboarded person cannot be stated


nuri-yoo:
**Source and commit** — `slack`, at `b30ef32` (on the branch for #164; `main` at `bef7320` is the same).

Every member `users.list` and `users.info` serve carries `"deleted": False`, a constant at `backlot/routers/slack.py:318` in `_user_obj`. Nothing in the BYO contract can set it: the slack schema does not name the key, a roster entry (`departments` or `contacts` under `--roster`) states whether a person holds a token and nothing else about their account, and outside `backlot/fidelity/baseline/` the word `deleted` occurs in `backlot/`, `docs/` and `examples/` on that one line. grf53 noticed on #164 that the sample has no `deleted: True` user; no corpus can have one.

**How you know** — spec only. Slack's user object reference: "`deleted` Boolean — This user has been deactivated when the value of this field is true. Otherwise the value is false, or the field may not appear at all." `users.list`: "This method returns a list of all users in the workspace. This includes both invited users and deleted/deactivated users." A deactivated person's token stops working: the `users.list` and `users.info` references both list `account_inactive`, "Authentication token is for a deleted user or workspace". Measured on Backlot on 2026-09-11: the bundled corpus answers `users.list` with 10 members and `{False}` as the set of `deleted` values; the BYO sample, 9 members and the same set.

**Reproduction** — the bundled corpus:

```python
import backlot, requests

with backlot.serve() as m:
    h = {"Authorization": f"Bearer {m.token}"}
    members = requests.post(f"{m.base_url}/slack/api/users.list", headers=h).json()["members"]
    print(len(members), {u["deleted"] for u in members})
```

Prints `10 {False}`. No record and no roster entry changes the second value.

**What it costs a client** — a directory sync, or a "who is still here" filter that reads `deleted` to drop offboarded accounts, sees a workspace nobody ever left. A corpus modelling a real company, where the messages of someone since deactivated stay in the channel history, cannot say that their author is gone, so a pipeline that attributes an answer to current staff has no signal to attribute it to a former one instead. And an ACL evaluation cannot include the case real handles most sharply: a deactivated person's own token answering `account_inactive` rather than content.

[...]

Issue #258, opened 2026-09-17
slack: a deactivated member's users.list/.info object still carries the active-only fields


khj809:
Found while measuring for #198.

**Source and commit** — `slack`, at `main` (the branch fixing #198's `deleted`/`is_forgotten` gap does not touch this).

**How you know** — live against a real workspace, 2026-09-17, `users:read` scope, comparing two non-bot members: one active (`U05EEFY9473`), one deactivated (`U05DV7X8X1C`).

Active member's top-level keys: `id, name, is_bot, updated, is_app_user, team_id, deleted, color, is_email_confirmed, real_name, tz, tz_label, tz_offset, is_admin, is_owner, is_primary_owner, is_restricted, is_ultra_restricted, has_2fa, who_can_share_contact_card, profile`.

Deactivated member's top-level keys: `id, name, is_bot, updated, is_app_user, team_id, deleted, is_forgotten, profile`. Real drops `color`, `is_email_confirmed`, `real_name`, `tz`, `tz_label`, `tz_offset`, `is_admin`, `is_owner`, `is_primary_owner`, `is_restricted`, `is_ultra_restricted`, `has_2fa`, `who_can_share_contact_card` entirely — not `false`/empty, absent — and gains `is_forgotten`. The nested `profile` is almost identical between the two; the only difference measured is `huddle_state_expiration_ts`, present active, absent deactivated.

## 1. `_user_obj` in `backlot/routers/slack.py:307` serves every field for a deactivated member

**Request** — `users.info` (or a `users.list` page) for a member `slack_deactivated_users` names (see #198).

**Backlot serves** — the same full object an active member gets: `is_admin`, `is_owner`, `is_primary_owner`, `is_restricted`, `is_ultra_restricted`, `has_2fa`, `tz`/`tz_label`/`tz_offset`, `color`, plus (once #198 lands) `deleted: true` layered on top. Not `is_forgotten`: #198 measured it as a separate state (GDPR erasure, `true` on 5 of that workspace's 9 deactivated members and absent on the other 4), and a roster states no such thing for it to be derived from, so it stays unserved.

**The real API serves** — the trimmed object above: those admin/ownership/timezone/contact fields are gone, not defaulted.

[...]

PR #259, opened 2026-09-17
slack: a deactivated member is deleted, dropped from conversations.members, and account_inactive on their own token

khj809:
Closes #198.

## What changed

A roster entry can now carry `deactivated: true` (`backlot.importer.byo.load_roster`), Slack's own offboarded-member state, read as a strict boolean — the readers around it take a value in any shape, but `bool("false")` is `True`, so a quoted value would deactivate the person it states is active. On import the person's email lands in a new `slack_deactivated_users` table (`backlot/store.py`) and their token is written to `tokens.yaml` exactly as before: deactivation never drops the credential, it changes what the credential is answered with. A roster is authoritative on every load, not just the first — re-importing with a roster that no longer states it un-deactivates the person, the same way the roster already replaces the whole principal set.

`backlot/routers/slack.py` acts on it in two places: `_user_obj` answers `deleted: true`, and `_caller_or_error` answers `account_inactive` for a resolved token whose person is deactivated, before the caller reaches a handler. In `backlot/store.py` every path that says who is in a channel excludes them — `slack_private_channel_members`, `slack_channel_member_emails`, `slack_channel_has_author`, `count_slack_channel_members` and `slack_channel_member_counts` — so `conversations.members`, `num_members` and `is_member` cannot answer one question two ways, while `conversations.history` keeps every message they wrote. `slack_membership_violations`, the import-time private-channel consistency check, excludes them too: somebody in no channel's membership cannot reach the state it refuses, which is being served by `conversations.members` while `conversations.info` answers them `channel_not_found`. The refusal is Slack's alone, since the state is Slack's rather than an org-wide suspension: the same token still reads every other source, asserted against `/gmail/v1/users/me/profile`.

The roster's new key is documented where the roster is: `backlot/schemas/README.md` and the Slack section of `docs/supported-sources.md`.

Measured live against a real workspace on 2026-09-17 (`users:read`, `channels:read`, `channels:history`, `search:read`):

| request | real | Backlot before this PR |
|---|---|---|
| `users.list` | `deleted` present on all 19 members, `true` on 9 of them | always `false` (a constant in `_user_obj`) |
| `users.info` on a deactivated member | `deleted: true` | the same constant `false` |
| `conversations.members` on all 8 channels this token can read | none of the 9 deactivated members appears in any of them | no such concept — a deactivated member's messages made them a member like anyone else |
| `is_forgotten` across the same 19 | `true` on 5 of the 9 deactivated, absent on the other 4, absent on all 10 active | n/a — no such key ever served |

The membership row is consistent with removal rather than evidence of it, and the comment at `store.py` says so: none of the 9 has ever spoken in any of the 8 channels (10 visible messages in all, no channel reporting `has_more`), so their absence is equally the answer real gives for somebody who was never in the channel — and since Backlot derives a public channel's membership from speaking, the only row a filter can drop there is the one state the workspace does not contain. Its private channels are outside the token's scopes as well (`types=private_channel` answers `missing_scope`, which needs `groups:read`). What carries the behaviour is Slack's own sentence for it: "They'll be removed from all channels, signed out of your workspace(s) on all devices, and unable to sign back in", in the same breath as "nor are their messages or files deleted" (Slack, "Deactivate a member's account") — which is both halves of what this serves.

`deleted` is served unconditionally, which is what that workspace answers 19 for 19, where the reference allows either: "This user has been deactivated when the value of this field is true. Otherwise the value is false, or the field may not appear at all."

`is_forgotten` is deliberately not served, and it is not an undocumented or unmodellable key: Slack's reference gives it as "Whether the user has been GDPR-forgotten", and `slack_web_openapi_v2.json` — the document `docs/fidelity.md` names as Slack's comparison source — declares `definitions.objs_user.items[*].properties.is_forgotten` a boolean. GDPR erasure is a separate state an admin triggers, which is why 5 of the 9 carry it and 4 do not. A roster states no GDPR erasure, so the field has nothing here to be derived from: an acknowledged gap in what a corpus can express, not a coin flip between two spellings of deactivation.

`account_inactive` is the vendor's error for a deactivated member's own token, and its spec carries it unqualified: of the 12 methods Backlot serves, the 8 that declare an error enum at all list it (`api.test` and `search.messages` declare none, and `search.all`/`search.files` are absent from the spec). The method reference's prose is narrower than its own enum — it ends "when using a bot token", where Backlot's per-person tokens are user tokens — so the enums carry the argument. This sandbox has no deactivated member's token to call live with, so the shape is asserted from the spec rather than from a live transcript — the standard `docs/fidelity.md`'s coverage table already applies to Slack elsewhere.

This resolves #198's `needs-maintainer` decision (`/decision 1`, "serve, with refusal"): a roster field for it, the token kept and refused rather than dropped, `deleted` on the user object, and dropped from `conversations.members`. The decision's mention of `is_forgotten` assumed it tracked `deleted` 1:1, which the measurement does not bear out.

A narrower divergence turned up while measuring — a deactivated member's user object still carries every active-only field (`is_admin`, `tz`, `has_2fa`, … 13 of them across all 19 members) that real drops entirely for one — filed separately as #258 rather than folded in here, since #198's decision scoped this to `deleted` and membership.

## Verification

- **Tests** — `pytest -q` on an `--all-extras` install: 3291 passed, 1 skipped, nothing failed — the `mcp`, SDK and reader cases run rather than skip on that install. `ruff check . && ruff format --check .`: clean. `python scripts/gen_docs.py`: already current.
- **Each behaviour is pinned on its own.** Reverted one at a time on an otherwise clean tree, every one of them turns a test red:

| reverted alone | red |
|---|---|
| `_user_obj`'s `deleted` | `test_slack_deactivation_changes_every_slack_answer_about_a_member_and_nothing_else` |
| `_caller_or_error`'s `account_inactive` | same |
| `slack_private_channel_members` | same |
| `slack_channel_member_emails` | same |
| `slack_channel_has_author` | same |
| `count_slack_channel_members` | same (it is read before the warm cache lands, so the test clears the cache and asks `conversations.info` again) |
| `slack_channel_member_counts` | same |
| `slack_membership_violations` | `test_byo_roster_deactivated_reaches_the_db_tokens_yaml_and_the_membership_check` |
| `deactivated:`'s strict boolean | `test_byo_roster_deactivated_is_a_boolean_that_unions_or_is_refused` |

- [x] A new endpoint serving corpus content is ACL-scoped, proved for both an admin and a scoped token — n/a, no new endpoint; the tests exercise an admin token and both people's own tokens


backlot/routers/slack.py
```diff
@@ -310,12 +316,19 @@ def _user_obj(conn, email: str) -> dict:
     parts = display.split()
     updated = synth.epoch("user:" + email)
     is_bot = not u and email.split("@")[0].endswith("bot")  # display-only "*bot" speakers
-    return {
+    # A roster entry's `deactivated: true`. Measured against a live workspace on 2026-09-17: a
+    # deactivated member's users.list/.info answers "deleted": true 9-for-9. `is_forgotten` (an
+    # undocumented key, present true or absent, never false) does NOT track `deleted` — of the
```
nuri-yoo: `is_forgotten` is documented, and it is not a variant of `deleted`.

Slack's own reference, the same page this PR quotes for `deleted`, has it: "`is_forgotten` Boolean — Whether the user has been GDPR-forgotten." And `slack_web_openapi_v2.json`, which `docs/fidelity.md` names as Slack's comparison document, declares `definitions.objs_user.items[*].properties.is_forgotten` as `boolean`.

I reproduced the 5/4 split on the same workspace, so that number is right. Its cause is named by the vendor: GDPR erasure is a separate state an admin triggers, and 5 of the 9 happen to be in both states. The response even carries the distinction, since the deactivated members come back in three different key-count shapes (8, 9, 10) against 20/21 for an active one.

Leaving it unserved stays right. "no signal in either the roster or the corpus a BYO import carries that predicts which" is the part to replace: a roster states no GDPR-erasure state, so there is nothing for the field to be derived from. That reads as a gap a reader can act on, where the current sentence teaches that a documented field is undocumented and unmodellable.

backlot/store.py
```diff
@@ -3203,7 +3211,11 @@ def slack_private_channel_members(conn, channel) -> list[str] | None:
             members.add(pid)
         elif ptype == "group":
             members.update(r["id"] for r in group_members(conn, pid))
-    return sorted(members)
+    # A deactivated member is dropped from conversations.members (measured against a live
+    # workspace on 2026-09-17: none of 9 deactivated members appeared in any of 8 readable
+    # channels) — their messages stay in history unchanged, only membership is affected.
```
nuri-yoo: This measurement does not separate the two explanations it is offered for.

I measured the same workspace with my own token. None of the 9 has ever spoken in any of the 8 channels: 10 visible messages in total and no channel reported `has_more`, so it is not a paging artefact.

```
ch  members  num_members  deactivated_in_members  msgs  deactivated_authors
c1  8        8            0                       6     0
c2  8        8            0                       2     0
c3  3        3            0                       0     0
c4  8        8            0                       2     0
c5  3        3            0                       0     0
c6  3        3            0                       0     0
c7  3        3            0                       0     0
c8  5        5            0                       0     0
```

Absence from `conversations.members` is the answer real gives for anyone who was never in the channel. So it does not distinguish "real drops a deactivated member" from "these 9 were never in these 8 channels".

It matters most right here. Backlot derives a public channel's membership from speaking, so the only row a filter can remove there is a deactivated person who spoke, which is the one state never observed. Private channels are out of the measurement as well: `types=private_channel` needs `groups:read`, which is not among the four scopes this PR names, so the path this very line changes has neither a live observation nor a test.

Modelling it is still right. The comment should state what was observed rather than presenting the absence as the finding.

Two smaller things on the same function. Its docstring says an empty list means "a private channel nobody may read (`"readers": []`, or a grant to a group with no members)", and there is now a third way in: every grantee deactivated. I built that case and it serves `is_private: true` with `num_members: 0` and history intact. And reverting just this line leaves the whole suite green (`2939 passed, 37 skipped`), so nothing pins it.

backlot/store.py
```diff
@@ -3226,6 +3238,7 @@ def slack_channel_member_emails(conn, channel, limit=100, offset=0) -> list[str]
         r[0]
         for r in conn.execute(
             "SELECT DISTINCT author_email FROM slack_messages WHERE channel = ? "
+            "AND author_email NOT IN (SELECT email FROM slack_deactivated_users) "
```
nuri-yoo: `slack_channel_has_author` (`store.py:3282`) is a membership path and is not filtered, so the body's "every membership path in `backlot/store.py` ... excludes a deactivated person" is not every one.

Its docstring claims parity with this function by name: "which is being a member of a PUBLIC one, the same set `slack_channel_member_emails` pages there, asked about one person". After this change that is false for a deactivated person, since this function excludes them and that one answers `True`.

Nothing observable diverges yet, because `_is_member` asks it only about the caller and a deactivated caller is already refused in `_caller_or_error`. That leaves the invariant resting on an unrelated refusal with nothing recording the dependency, which is the shape that breaks quietly later. Either filter it too, or say in its docstring why it does not need to be.

backlot/store.py
```diff
@@ -3310,14 +3327,25 @@ def count_slack_channel_members(conn, channel) -> int:
     if members is not None:
         return len(members)
     return conn.execute(
-        "SELECT COUNT(DISTINCT author_email) FROM slack_messages WHERE channel = ?", (channel,)
+        "SELECT COUNT(DISTINCT author_email) FROM slack_messages WHERE channel = ? "
+        "AND author_email NOT IN (SELECT email FROM slack_deactivated_users)",
```
nuri-yoo: Reverting this filter alone leaves the suite green (`2939 passed, 37 skipped`), where the same revert on `slack_channel_member_counts` fails the new test.

This is the path taken before the warm cache lands (`main.py:107` fills it on a background thread), so `num_members` is only proven on the cached side. A test that reads `conversations.info?include_num_members=true` against a freshly opened client, before the cache is warm, would separate the two.

backlot/store.py
```diff
@@ -3291,7 +3306,9 @@ def slack_channel_member_counts(conn) -> dict[str, int]:
     counts = {
         r[0]: r[1]
         for r in conn.execute(
-            "SELECT channel, COUNT(DISTINCT author_email) FROM slack_messages GROUP BY channel"
+            "SELECT channel, COUNT(DISTINCT author_email) FROM slack_messages "
+            "WHERE author_email NOT IN (SELECT email FROM slack_deactivated_users) "
```
nuri-yoo: The two measurements in this docstring now describe a different query. "~1.9s on the biggest channel measured" and "this is 12.2s once" were taken before this PR added a per-row `NOT IN` probe to that exact scan.

The reasoning survives, for what it is worth. I checked the plans and the shape holds:

```
counts  SCAN slack_messages USING COVERING INDEX idx_slack_channel_author
        USING INDEX sqlite_autoindex_slack_deactivated_users_1 FOR IN-OPERATOR
paging  SEARCH slack_messages USING COVERING INDEX idx_slack_channel_author (channel=?)
        USING INDEX sqlite_autoindex_slack_deactivated_users_1 FOR IN-OPERATOR
```

Either re-measure the two numbers on this head or drop them for the plan, since as they stand they are timings of code that no longer exists.

backlot/routers/slack.py
```diff
@@ -169,12 +169,18 @@ def _caller_or_error(request: Request) -> tuple[Caller | None, dict | None]:
...
+    the caller it would otherwise resolve to; this is the one place that refusal is drawn.
     """
     token = auth.slack_token(request)
     caller = auth.acl(request).resolve(token)
-    if caller is not None:
-        return caller, None
-    return None, _err("invalid_auth" if token else "not_authed")
+    if caller is None:
+        return None, _err("invalid_auth" if token else "not_authed")
+    if not caller.is_admin and store.slack_is_deactivated(auth.conn(request), caller.email):
```
nuri-yoo: `not caller.is_admin` cannot change an answer. An admin `Caller` is only ever constructed with `email=None` (`acl.py:46` and `acl.py:81`), so `slack_deactivated_users` can never hold its email. I dropped the clause and the suite stayed green.

Fine to keep as a deliberate skip of one query per admin request, but then say that, since as written it reads as a carve-out protecting an admin who could otherwise be refused.

backlot/importer/byo.py
```diff
@@ -698,6 +699,13 @@ def load_roster(path) -> dict:
     ``contacts`` are people a corpus names who are not accounts — they own and read documents but
     cannot authenticate, the distinction ``tokens.yaml`` draws.
 
+    An entry's ``deactivated: true`` is Slack's own offboarded-member state — the person keeps
+    their token in ``tokens.yaml`` (real answers a deactivated member's own calls with
+    ``account_inactive`` rather than dropping the credential), but ``users.list``/``.info`` answer
+    them ``deleted: true`` and ``conversations.members`` drops them. No other vendor Backlot
+    serves has this concept, so it is read here and acted on only in ``backlot.routers.slack``
+    and ``backlot.store``.
```
nuri-yoo: "acted on only in `backlot.routers.slack` and `backlot.store`" is the right scope for #198's decision, and I confirmed it holds: a deactivated person's token still answers 200 on `/gmail/v1/users/me/profile`.

It lives only in prose, though. Moving this check into shared auth later would widen it to every vendor with nothing going red. One assertion that a deactivated token still succeeds on a non-Slack route would pin the choice where a reader can find it.

backlot/importer/byo.py
```diff
@@ -689,6 +689,7 @@ def load_roster(path) -> dict:
             - {name: Ava Chen, email: ava.chen@redwoodinference.com}
             - {name: Bo Ryu, email: bo.ryu@redwoodinference.com,
                groups: [proj-checkout-rework, res-emea-support]}
+            - {name: Cy Ito, email: cy.ito@redwoodinference.com, deactivated: true}
```
nuri-yoo: `backlot/schemas/README.md` carries this same YAML block as the roster's documented shape (lines 202-212), and it did not get the new key. This docstring's copy now has three department entries where the README's has two.

Nothing catches the drift: `scripts/gen_docs.py` does not write that file, and `tests/test_schema.py::test_every_record_the_docs_show_would_load` only validates the JSONL record snippets in it. `docs/supported-sources.md`'s Slack table also documents `conversations.members` and `users.list` with no mention of the state that now changes both.

backlot/importer/byo.py
```diff
@@ -767,6 +792,7 @@ def _merge(email: str, name: str, groups: list[str], token: bool, *, stated: boo
                 _groups(p, slugify(dept) or None),
                 True,
                 stated=bool(p.get("name")),
+                deactivated=bool(p.get("deactivated")),
```
nuri-yoo: `deactivated: "false"` deactivates the person, and so does `"no"`:

```
deactivated: true      -> True
deactivated: "false"   -> True
deactivated: "no"      -> True
deactivated: 0         -> False
```

Consistent with how tolerant the readers around it are, and `_slugs` says why that tolerance is deliberate for the other fields. This is the roster's first boolean, though, and tolerance that is harmless for a name or a group list inverts a stated meaning here. `p.get("deactivated") is True`, or refusing a non-boolean, keeps a quoted value from meaning its opposite.

nuri-yoo (changes requested):
Reviewed at `aafdeb7`. I measured against a real workspace with a token of my own (`users:read`, `channels:read`, `channels:history`, `search:read`) rather than reading the numbers off this body, and ran each filter reverted on its own against the full suite.

The behaviour is right end to end, and the official `slack_sdk` reads it. What blocks is the evidence two of the comments claim for themselves, and that three of the five store filters no test can tell apart.

## What reproduced exactly

```
users.list total 19    deleted true 9    deleted false 10    "deleted" key absent 0
is_forgotten  among the 9 deleted : true 5, absent 4
              among the 10 active : absent 10
public channels 8      visible messages across all 8: 10      channels reporting has_more: 0
```

So `deleted` 9-for-9 holds, and the 5/4 `is_forgotten` split holds. `deleted` is also present on all 19, which is worth keeping: Slack's reference says of it "Otherwise the value is false, or the field may not appear at all", and serving it unconditionally matches what this workspace answers.

## The membership row of the table is not discriminating

```
ch  members  num_members  deactivated_in_members  msgs  deactivated_authors
c1  8        8            0                       6     0
c2  8        8            0                       2     0
c3  3        3            0                       0     0
c4  8        8            0                       2     0
c5  3        3            0                       0     0
c6  3        3            0                       0     0
c7  3        3            0                       0     0
c8  5        5            0                       0     0

deactivated people who authored a visible message anywhere: 0 of 9
```

None of the 9 has ever spoken in any of the 8 channels. The whole visible history is 10 messages and no channel reported `has_more`, so this is not a paging artefact. Their absence from `conversations.members` is the same answer real gives for somebody who was never in the channel, so it does not separate "real removes a deactivated member from a channel" from "these 9 were never in these channels".

That gap lands exactly where this PR acts. Backlot derives a public channel's membership from speaking, so the only row the new filter can ever remove there is a deactivated person who spoke, and that is the one state the measurement never contains. The table reads as though the absence were the finding.

Private channels are outside the measurement too: `types=private_channel` needs `groups:read`, which is not among the four scopes named above. `slack_private_channel_members` is therefore the one path with neither a live observation nor a test behind it.

I am not asking for a different decision. Real does remove deactivated members from channels, and modelling it is right. I am asking the body and the comment at `store.py:3214` to say what was measured, so the next reader does not inherit an absence as a positive.

## `is_forgotten` is documented, and it does not mean deactivated

Slack's own reference, the same page this body quotes for `deleted`, carries it: "`is_forgotten` Boolean — Whether the user has been GDPR-forgotten." `slack_web_openapi_v2.json`, the document `docs/fidelity.md` names as Slack's comparison source, declares `definitions.objs_user.items[*].properties.is_forgotten` as `boolean`.

GDPR erasure is a separate state an admin triggers, not a variant of deactivation, which is why 5 of 9 carry it. The response says so itself:

```
top-level key count   deactivated: 8 / 9 / 10 (three shapes)    active: 20 / 21
present on active members and on none of the 9 deactivated (13 keys):
  color  has_2fa  is_admin  is_email_confirmed  is_owner  is_primary_owner
  is_restricted  is_ultra_restricted  real_name  tz  tz_label  tz_offset
  who_can_share_contact_card
present on deactivated and on no active member: is_forgotten
```

Not serving it stays correct. The reason has to change: it is not that no signal predicts which, it is that a roster states no GDPR-erasure state, so the field has nothing to be derived from.

## "every membership path" is not every one

`slack_channel_has_author` (`store.py:3282`) answers membership by its own docstring, "which is being a member of a PUBLIC one, the same set `slack_channel_member_emails` pages there", and it is unfiltered. After this PR that sentence is false for a deactivated person.

Nothing observable diverges today, because `_is_member` asks it only about the caller and a deactivated caller is refused in `_caller_or_error` first. That makes the invariant rest on an unrelated refusal, with nothing recording the dependency. Filter it as well, or say in its docstring why it does not need filtering.

## Three of the five filters survive reverting

Each reverted alone, on a clean tree, whole suite (`.[dev]`, baseline `2939 passed, 37 skipped`):

```
slack_channel_member_emails     filter removed -> 1 failed        caught
slack_channel_member_counts     filter removed -> 1 failed        caught
slack_private_channel_members   filter removed -> 2939 passed     survives
count_slack_channel_members     filter removed -> 2939 passed     survives
slack_membership_violations     filter removed -> 2939 passed     survives
not caller.is_admin             dropped        -> 2939 passed     survives
```

The first checkbox is true as written, since reverting the three files together does fail three tests. Per function the discrimination is not there. The new test builds a public channel only, so the private path, the cold `num_members` path before the warm cache lands, and the import-time violation check are all unpinned.

## What I verified as correct

Built a corpus whose private channel grants all three people and whose public channel has the deactivated person as its only speaker, then served it. The private channel drops them from `members`, answers `num_members` 2, and keeps both messages in history. The public channel answers `num_members` 0 with its one message intact. A scoped non-admin token gets identical answers to the admin token on all three changed bodies.

The official `slack_sdk` parses it: `users_list` gives `deleted=True` with no `is_forgotten` key, `conversations_members` omits them, and `auth_test` on their own token raises `SlackApiError: account_inactive`.

`account_inactive` is the right error and the spec backs it: it is in the error enum of all eight served methods that declare one. The documented sentence is worth quoting whole, though, since this body says the shape is asserted from the reference rather than measured. It ends "when using a bot token" on `auth.test`, `users.list`, `users.info`, `conversations.members` and `conversations.history` alike, and Backlot's per-person tokens are user tokens. The spec's enums carry no such qualifier, so let them carry the argument.

A DB built at the base commit and served by this head refuses at startup with "was built by an older Backlot and has no slack_deactivated_users. Re-import the corpus", and `missing_schema` returns exactly `['slack_deactivated_users']`.

Query plans keep their shape. Paging stays `SEARCH slack_messages USING COVERING INDEX idx_slack_channel_author (channel=?)` with `USING INDEX sqlite_autoindex_slack_deactivated_users_1 FOR IN-OPERATOR` beside it, and the counts pass stays a covering-index scan.

Every authenticated Slack route reaches `_caller_or_error`, the two search methods through `_messages_block`. Only `api.test` does not, which is what real answers for it, so the refusal has no bypass.

## One thing outside this PR

#258's key list matches my own sweep exactly, 13 keys, measured across all 19 members rather than two. One sentence in it has gone stale: it says `deleted: true` and `is_forgotten: true` will be "layered on top" once #198 lands, and this PR decided not to serve `is_forgotten`. Worth correcting there while it is open.

backlot/routers/slack.py
khj809: Right, and the comment was teaching the opposite. Rewritten: it now names the reference's own line for `is_forgotten` ("Whether the user has been GDPR-forgotten") and the `boolean` declaration in `slack_web_openapi_v2.json`, and gives the reason as a roster stating no GDPR erasure for the field to be derived from. Same correction in the body.

backlot/store.py
khj809: Reproduced your table with a token of my own — 8 channels, 10 visible messages, no `has_more`, 0 of the 9 having spoken anywhere — so the absence really does not separate the two readings. The comment now says that outright and rests the behaviour on Slack's own sentence instead: "They'll be removed from all channels ... nor are their messages or files deleted" (help article "Deactivate a member's account"), which carries both halves. It also records that `types=private_channel` answers `missing_scope` without `groups:read`, so no live observation reaches that path. The docstring's empty-list sentence now lists every grantee deactivated as the third way in, and a test builds that corpus.

backlot/store.py
khj809: Filtered it. The docstring says deactivation is excluded for the same reason it is in `slack_channel_member_emails` — a channel's membership and one person's place in it are one fact — so the invariant no longer rests on the refusal in `_caller_or_error`. Reverting this line alone is now red.

backlot/store.py
khj809: Pinned. The test joins the warm-up thread, clears `app.state.channel_members`, and asks `conversations.info?include_num_members=true` again on both channels, asserting the cold answer equals the cached one. Reverting this filter alone is now red.

backlot/store.py
khj809: Dropped them for a measurement on this head: 5.6M messages, 36 channels, biggest 19,543 distinct authors — this pass 6.6s, the same counts one channel at a time 6.6s. The docstring now says the cache exists to pay that once at startup rather than per request, and carries the plan shape you quoted.

backlot/routers/slack.py
khj809: Dropped. `Caller(email=None)` for an admin makes the clause unreachable either way, and the docstring now says an admin or anonymous caller has no email to be one of these people.

backlot/importer/byo.py
khj809: Pinned: the served test asserts `/gmail/v1/users/me/profile` still answers 200 on the deactivated person's token, next to the Slack calls that refuse it. `docs/supported-sources.md` says it in prose too.

backlot/importer/byo.py
khj809: Added there — the YAML block carries the key, and a paragraph below it states the vendor shape. The Slack section of `docs/supported-sources.md` gained the same paragraph, since `conversations.members` and `users.list` are documented there.

backlot/importer/byo.py
khj809: Made it strict: `deactivated:` is a boolean or nothing, and anything else is refused with the address and the value. Reading `"false"` as True is the one case where the tolerance around it would invert a stated meaning rather than widen it. Test added, and reverting the strictness is red.

backlot/store.py
```diff
@@ -3236,7 +3212,17 @@ def slack_private_channel_members(conn, channel) -> list[str] | None:
             members.add(pid)
         elif ptype == "group":
             members.update(r["id"] for r in group_members(conn, pid))
-    return sorted(members)
+    # Slack drops a deactivated member from every channel and keeps what they wrote: "They'll be
+    # removed from all channels ... nor are their messages or files deleted" (Slack, "Deactivate a
```
nuri-yoo: The elision reverses the source order. On the help article the clause about messages comes first: "People aren't notified when their accounts are deactivated, nor are their messages or files deleted. They'll be removed from all channels, signed out of your workspace(s) on all devices, and unable to sign back in."

So `"They'll be removed from all channels ... nor are their messages or files deleted"` reads as though the second clause followed the first, and a reader who opens the article to check finds it above instead. The PR body gets this right with "in the same breath as", which asserts nothing about order.

Both fragments are verbatim and the citation carries exactly the two halves this serves, so it is only the joining `...`. Quoting them in the article's own order, or joining with a semicolon rather than an ellipsis, would fix it.

tests/test_slack.py
```diff
@@ -1446,3 +1447,122 @@ def test_slack_reaction_ids_and_count_are_derived_from_the_addresses(tmp_path):
     ]
     # Slack's own `defs_user_id`, so an id Backlot mints is one the vendor's spec would accept.
     assert all(re.fullmatch(r"[UW][A-Z0-9]{2,}", u) for r in reactions for u in r["users"])
+
+
+def test_slack_deactivated_member_is_deleted_and_dropped_from_membership(tmp_path):
```
nuri-yoo: The docstring grew to cover the new ground and the name did not. It now also pins three things the name does not reach: that a non-Slack route still answers 200 on the same token, that the cold count and the warm cache agree, and `is_member` through `slack_channel_has_author` after the client closes.

No new test function was added anywhere in this PR — `tests/test_slack.py` has 45 and `tests/test_importer_byo.py` 146 at both `aafdeb7` and `65ab116`, so all 277 added test lines went into the three existing functions. That is the right place for them, since each is one behaviour seen from several surfaces. It does mean the names are now the least accurate description of what each covers, and the names are what a future reader greps.

tests/test_importer_byo.py
```diff
@@ -2206,6 +2206,162 @@ def test_byo_roster_duplicate_entries_union_their_groups(tmp_path):
     assert len(bo["groups"]) == 5  # deduplicated, so no membership row is doubled
 
 
+def test_byo_roster_deactivated_is_parsed_and_unions_true(tmp_path):
```
nuri-yoo: Same as the sibling test: this one now also pins that a non-boolean is refused, which `_is_parsed_and_unions_true` does not suggest. `_reaches_the_db_and_tokens_yaml` below has the same shape — it now carries the `slack_membership_violations` control pair, which is neither the DB nor `tokens.yaml`.

backlot/importer/byo.py
```diff
@@ -725,6 +733,22 @@ def _primary(raw) -> str | None:
...
+        any shape at all.
+
+        Their tolerance widens what a field accepts; here it would invert what one says. ``bool()``
+        reads the string ``"false"`` as True, and ``"no"`` too, so a quoted value would deactivate
+        the person it states is active. YAML spells the value both ways unquoted already, which
+        leaves nothing for a string to express."""
+        raw = entry.get("deactivated")
+        if raw is None or isinstance(raw, bool):
+            return bool(raw)
+        raise SystemExit(
```
nuri-yoo: This is the roster's first validation, and `--dry-run` does not reach it. Measured on this head with a roster carrying `deactivated: "false"`:

```
backlot import c.jsonl --roster roster.yaml --dry-run   exit 0   "OK: 1 records valid."
backlot import c.jsonl --roster roster.yaml             exit 1   roster entry 'a@acme.com': `deactivated: 'false'` is not a boolean. …
```

`load`'s `dry_run` branch returns before `_load_records`, so the roster is never read there. Arguably out of scope, since `--dry-run` has never validated the sidecar and `docs/corpus.md` describes it in terms of records. It is worth a thought only because the command whose job is to say "this would load" now says so about a roster that will not.

The message itself is good — it names the address and the value it rejected.

nuri-yoo (approved):
Approving at `65ab116`. All four blocking items are addressed, and I checked each by running rather than from the replies.

## The mutation table reproduces, 9 for 9

Reverted one at a time on a clean tree, whole suite (`.[dev]`, baseline `2939 passed, 37 skipped`). CAUGHT means `returncode == 1` with `failed` in the output:

```
_user_obj deleted                  rc1  CAUGHT  test_slack_deactivated_member_is_deleted_and_dropped_from_membership
_caller_or_error account_inactive  rc1  CAUGHT  same
slack_private_channel_members      rc1  CAUGHT  same          (survived before)
slack_channel_member_emails        rc1  CAUGHT  same
slack_channel_has_author           rc1  CAUGHT  same          (new filter)
count_slack_channel_members        rc1  CAUGHT  same          (survived before)
slack_channel_member_counts        rc1  CAUGHT  same
slack_membership_violations        rc1  CAUGHT  test_byo_roster_deactivated_reaches_the_db_and_tokens_yaml  (survived before)
deactivated strict boolean         rc1  CAUGHT  test_byo_roster_deactivated_is_parsed_and_unions_true
tree clean check: (clean)
```

The three that survived before are pinned by the right kind of test each: a private channel for the grant-derived path, `warm_thread.join()` then clearing the cache and asking `conversations.info` again for the cold count, and a control pair for the violation check — the same corpus refuses with the person active and loads with them deactivated, leaving the private membership as one person and the message still in the channel.

## The vendor claims hold

The help-article sentence this now rests on is verbatim: "People aren't notified when their accounts are deactivated, nor are their messages or files deleted. They'll be removed from all channels, signed out of your workspace(s) on all devices, and unable to sign back in." Both halves of what this serves are in it.

`is_forgotten` is documented as I read it too, "Whether the user has been GDPR-forgotten" on the user-object reference and a `boolean` in `slack_web_openapi_v2.json`, and the reason now given (a roster states no GDPR erasure) is the one the vendor's own wording supports.

I recounted `account_inactive`: of the 12 methods served, the 8 that declare an error enum all list it, `api.test` and `search.messages` declare none, and `search.all`/`search.files` are absent from the spec. 8 + 2 + 2 = 12, and the `when using a bot token` qualifier is now in the quote.

`slack_channel_has_author`'s docstring still claims a seek, and it is one: `SEARCH slack_messages USING COVERING INDEX idx_slack_channel_author (channel=? AND author_email=?)` with the deactivated set entering as `USING INDEX sqlite_autoindex_slack_deactivated_users_1 FOR IN-OPERATOR`.

## Gates

`ruff check` and `ruff format --check` both exit 0 (141 files), `scripts/gen_docs.py --check` exits 0, CI is 6 for 6. The roster key reached both places it is documented, and #258's sentence about `is_forgotten` landing on top is corrected.

## Non-blocking notes

Four small things, three of them inline. The fourth is here: in **Why this is what the real API does**, the last sentence of the `account_inactive` paragraph has lost a connective — "This sandbox has no deactivated member's token to call live with, the standard `docs/fidelity.md`'s coverage table already applies to Slack elsewhere."

None of them changes behaviour, so none should hold this up. Worth knowing before merge: this branch is BEHIND `main` (base `571da34`, `main` now `10baa592`), so it needs an update either way.

backlot/store.py
khj809: Fixed in `c0daa89`: both sentences now in the article's own order, with the ellipsis only trailing — "People aren't notified when their accounts are deactivated, nor are their messages or files deleted. They'll be removed from all channels ...".

tests/test_slack.py
khj809: Renamed in `e901ff7` to `test_slack_deactivation_changes_every_slack_answer_about_a_member_and_nothing_else` — the second half is the non-Slack 200 and the messages left in history.

tests/test_importer_byo.py
khj809: Renamed in `e901ff7`: `..._is_a_boolean_that_unions_or_is_refused` and `..._reaches_the_db_tokens_yaml_and_the_membership_check`, with both docstrings extended to the ground they had grown into.

backlot/importer/byo.py
khj809: Left as is, on your reading of it: `--dry-run` has never read the sidecar, and widening it to the roster is a change to what that command means rather than to this fix. Happy to file it separately if it should be one.

PR #283, opened 2026-09-18
google: the batch endpoint is compared against the batchPath its documents declare

### LINKED ISSUE #171: google: batch endpoint needs its own comparison kind, the way S3 has a probe

khj809:
Closes #171. Supersedes #200.

## What changed

`backlot diff` now compares Google's batch endpoint. `GoogleDiscoveryComparison` gains `batch_mount` — the batch routes a source speaks for, as `ProbeComparison.mount` already answers the same question for a probe — and `google_discovery_diff.batch_divergences` compares each document's `batchPath` against them. `/batch` leaves `UNCOMPARED` and becomes the fourth bucket a served path can land in: under a document's mount, named as a source's batch route, probe-compared, or declared uncompared with a reason. `batch_mount` defaults to empty and the check is skipped when it is, so the suite refuses a Google source that names no batch route — a third one registered without it would otherwise have every document's `batchPath` go unread.

Three kinds are reported, and the vocabulary follows the existing reading rule that `missing_*` is reliable while `extra_*` means undocumented by the vendor rather than proof of a bug.

| Kind | Severity | Fires when | What drift it catches |
|---|---|---|---|
| `extra_batch_api` | breaking | a document declares no `batchPath` | Google drops batch for one API while Backlot goes on answering it |
| `missing_batch_path` | gap | a declared value no route the source mounts answers | a value moves to a third shape |
| `extra_batch_route` | breaking | a mounted route no document of that source selects | Drive moves to its own host, leaving `/batch/{api}/{version}` served for nobody |

A `missing_batch_path` finding is identified by the document and the value together — `drive:v3 v3/batch`, not `drive:v3`. A gap is acknowledged by identity alone (`Baseline.unacknowledged` reads a gap's detail for nothing), so keyed on the document, the entry `--update-baseline` wrote for one move would answer for the next one too. The path diffs have that property for free, since there the moved path is the identity. Two of the three are breaking, so `--update-baseline` cannot acknowledge them and a scheduled run stays red until someone writes the note by hand — the existing rule for a breaking finding, not a new one.

`COMPARISONS` still holds exactly the `source_type`s Backlot serves, so `test_the_registry_names_exactly_the_source_types_backlot_serves` stays `set(store.SOURCE_TABLE)` with no exception spelled into it. `cli.FIDELITY_SOURCES`, the CI matrix, the dispatcher's isinstance guard, the `Comparison` union and the four comparison kinds are untouched, and there is no new baseline file: `gmail.json` and `google_drive.json` read the same documents they always did and acknowledge nothing new, because today's measurement diverges nowhere. `docs/fidelity.md` gains a section on how the field is compared, and its coverage paragraph now counts four buckets.

Every discovery document names a batch endpoint in its top-level `batchPath`, and none declares it under `resources`. Measured 2026-09-17 over the five documents the Gmail and Drive comparisons already read:

| Document | `batchPath` | Host (`rootUrl`) | Backlot answers it at |
|---|---|---|---|
| `gmail:v1` | `batch` | `gmail.googleapis.com` | `/batch` |
| `docs:v1` | `batch` | `docs.googleapis.com` | `/batch` |
| `sheets:v4` | `batch` | `sheets.googleapis.com` | `/batch` |
| `slides:v1` | `batch` | `slides.googleapis.com` | `/batch` |
| `drive:v3` | `batch/drive/v3` | `www.googleapis.com` | `/batch/{api}/{version}` |

Walking `resources` on all five yields 168 operations and not one whose `servicePath`-joined path is a `batchPath` value, so a path diff has nothing on the vendor's side to pair Backlot's two batch routes with. The methods that carry the word are a different thing and are compared as operations already: measured 2026-09-17 there are eleven of them — `messages.batchModify` and `messages.batchDelete` on Gmail, `documents.batchUpdate` on Docs, seven `batch*` methods on Sheets (six under `values`, plus `spreadsheets.batchUpdate`), `presentations.batchUpdate` on Slides, and none on Drive.

Each API answers batch at the root of its own host; Drive alone sits on the shared `www.googleapis.com`, and the `drive/v3` is what discriminates it there. Backlot collapses those hosts onto one origin, so `/batch` stands in for four APIs at once and belongs to no single source's `Spec.mount`.

That is why the endpoint is mounted rather than given a comparison kind of its own. What drifts is a document, and a finding names the document — `extra_batch_api docs:v1`, `missing_batch_path drive:v3 v3/batch` — so gmail's source reports about `gmail:v1` and google_drive's about its four, and the shared route puts no finding in two baselines. What is left over is the reverse direction, which no single document can answer, and that is what `batch_mount` carries: without it, Drive moving to its own host would leave every document declaring `batch`, every declared value answered, and `/batch/{api}/{version}` standing for nobody behind a green check. `batch_mount` is deliberately not a `Spec.mount` — a mount over these paths would hand them to the path diff, which would report both routes as surface Backlot invented — and it sits on `GoogleDiscoveryComparison` rather than on `Spec`, which is shared with the six OpenAPI sources.

A finding is identified by what the document calls itself, not by the URL it was fetched from: measured 2026-09-17, `www.googleapis.com/discovery/v1/apis/gmail/v1/rest` and `gmail.googleapis.com/$discovery/rest?version=v1` both answer the document that calls itself `gmail:v1` — same revision 20260907, same 217,687 bytes, equal once parsed, differing only in key order, so repointing the registry from one to the other must not read as a new divergence. `document_id` falls back to `name:version` and then to the URL for a document that says neither.

This costs no extra vendor round trips. A discovery document carries both contracts, so `GoogleDiscoveryComparison.divergences` memoizes this run's fetches by URL — the same `seen` memo `OpenAPIComparison` already uses for HubSpot's index — and hands the batch check the documents the path diffs fetched. Measured 2026-09-17, those five documents total 1.3 MB, the largest of them 369 KB. Taking the vendor's side from each source's own `specs` is also what keeps the two in step: a registry that repoints a document repoints this check with it, and a sixth Google API cannot arrive with its `batchPath` unchecked.

Nothing about what the batch endpoint *does* is compared — the multipart envelope, the `Content-ID` pairing, the outer credential applying to a sub-request that carries none. Those are behaviour, held by `tests/test_google.py`, and reaching them would need the probe kind rather than this one. Nor is the endpoint asked of Google, which would need a credential.

One thing found while measuring and left alone: `POST /batch` declares `api` and `version` as query parameters in Backlot's own `/openapi.json`, because the handler serves both routes and FastAPI reads the two defaults as query parameters on the route whose template does not name them. No discovery document declares any parameter for `batchPath`, and what real Google does with `?api=&version=` on `gmail.googleapis.com/batch` is unmeasured here, so this comparison — which is over paths — says nothing about it either way.

## Verification

At `1cca62f`, on `main` at `1a62604`.

- **Tests** — `pytest` on a `uv sync --all-extras --locked` install: 3302 passed, 1 skipped. The skip is the HubSpot reader, which no extra carries (`CONTRIBUTING.md`).
- **Optional surfaces that ran rather than skipped** — `tests/test_sdk.py` 5, `tests/test_s3.py` 84, `tests/test_mcp.py` 43, `tests/test_llamaindex.py` 12, `tests/test_integrations.py` 22.
- **Lint** — `ruff check . && ruff format --check .`: All checks passed!, 141 files already formatted. `python scripts/gen_docs.py --check` exits 0.

Live against Google, so these are the measurement above rather than a fixture: `backlot diff --source gmail` answers `157 divergence(s) · 157 acknowledged · 0 new` and `--source google_drive` `200 · 200 · 0 new`.

What the report looks like when it fires, with Drive's declared value patched to `batch` and everything else the real command:

```
 breaking  extra_batch_route    /batch/{}/{}
           Backlot serves it for this source, and no document of it declares a batchPath the route answers
```

Nothing is reported under `gmail`, which is the point of naming the routes per source. Gmail moving to `batch/gmail/v1` is the mirror: `extra_batch_route /batch` and `missing_batch_path gmail:v1 batch/gmail/v1` under gmail, nothing under google_drive. Docs dropping its `batchPath` reports only `extra_batch_api docs:v1`, because `/batch` is still selected by Sheets and Slides. Docs moving to `batch/docs/v1` reports nothing, and that is correct rather than a miss: `routers/google.py` binds `api` and `version` without reading them, so Backlot does answer that path.

- [x] A test fails without this change — ten do, reproduced by copying only this branch's `tests/test_fidelity.py` onto a worktree at `1a62604`. Each rule is held by reverting it: relaxing `answers()`'s segment count (6), removing the `extra_batch_route` loop (6), keying the gap on the document alone (5), `batch_path` not canonicalising (1), identifying a document by its URL (5), dropping either `document_id` fallback (1 each), `google_drive` losing its `batch_mount` (2) or naming a route Backlot does not serve (3), a Google source carrying an empty `batch_mount` (2), `/batch` going back into `UNCOMPARED` (1), the batch check never running (1), and each document being fetched again for it (1).
- [x] A new endpoint serving corpus content is ACL-scoped, proved for both an admin and a scoped token — n/a, no endpoint: the diff touches no router and no corpus read.


backlot/fidelity/comparisons.py
```diff
@@ -165,9 +175,21 @@ def endpoints(self) -> tuple[str, ...]:
...
-        )
+        # One memo across this run's specs, so the batch check reads the documents the path diffs
+        # already fetched rather than asking the vendor for them a second time: measured
+        # 2026-09-17, the five documents these two sources read total 1.3 MB, the largest of them
+        # 369 KB.
+        fetched: dict[str, dict] = {}
+        per_spec = [
+            google_discovery_diff.divergences(s, timeout=timeout, seen=fetched) for s in self.specs
+        ]
+        if self.batch_mount:
```
nuri-yoo: The default `()` turns this into a silent opt-out: a third Google source constructed without `batch_mount` never has its documents' `batchPath` read, whereas `batch_divergences((), docs)` would report each declared value as `missing_batch_path`. Measured with a made-up `calendar:v3` declaring `batch/calendar/v3` and `fetch_json` stubbed: `divergences()` returns no batch finding. Not blocking; a test that every `GoogleDiscoveryComparison` in `COMPARISONS` has a non-empty `batch_mount` would close it, since the test above pins only the two that exist.

backlot/fidelity/google_discovery_diff.py
```diff
@@ -73,15 +79,126 @@ class SpecTarget(Protocol):
...
+        segment == "{}" or segment == other for segment, other in zip(ours, theirs)
+    )
+
+
+def batch_divergences(
+    batch_mount: Iterable[str], documents: Iterable[tuple[str, Mapping[str, Any]]]
+) -> list[Finding]:
+    """Backlot's batch routes against the ``batchPath`` each document declares.
+
+    Per SOURCE, not per document, which is what separates this from the path diff above. Measured
```
nuri-yoo: Positional: the path diff's entry point `divergences` is defined below this function (line 185), and only its reader `from_google_discovery` is above. "the path diff" alone, or "`divergences` below", reads true from either end of the file.

docs/fidelity.md
```diff
@@ -136,6 +136,48 @@ children are what keep the two listings apart: `?list-type=2` and a bare bucket
...
+| `drive:v3` | `batch/drive/v3` | `www.googleapis.com` | `/batch/{api}/{version}` |
+
+Each API answers batch at the root of its own host; Drive alone sits on the shared
+`www.googleapis.com`, and the `drive/v3` is what discriminates it there. Backlot collapses those
+hosts onto one origin, so one served route stands in for several documents.
+
+That is why the endpoint is not a mount. A mount selects paths for the path diff, which pairs
+operations — and no document declares this one, so both routes would report as surface Backlot
+invented. The methods that carry the word are a different thing and are compared as operations
+already: `messages.batchModify` on Gmail, `documents.batchUpdate` on Docs, seven `batch*` methods
```
nuri-yoo: This reads as the full list and is two short. Walking `resources` on the five documents today, eleven method ids carry the word: Gmail `messages.batchDelete` and `messages.batchModify`, Docs `documents.batchUpdate`, Sheets seven, Slides `presentations.batchUpdate`, Drive none. The PR body has all eleven; the page drops Gmail's `batchDelete` and Slides' `batchUpdate`.

nuri-yoo (approved):
Approving. Nothing here blocks; the notes below are prose and one design question, none of which needs the code to change before merge.

**What I ran at `0d88a35`, on `main` at `1a62604`.** `uv sync --all-extras --locked` then `pytest`: 3303 passed, 1 skipped (the HubSpot reader). `ruff check .`, `ruff format --check .` and `PYTHONPATH=$PWD python scripts/gen_docs.py --check` all exit 0. Live, `backlot diff --source gmail` answers `157 · 157 acknowledged · 0 new` and `--source google_drive` `200 · 200 · 0 new`.

**The five documents, fetched today with a cache-busting query.** `gmail:v1` rev 20260907 `batch`, `drive:v3` rev 20260913 `batch/drive/v3`, `docs:v1` rev 20260909 `batch`, `sheets:v4` rev 20260915 `batch`, `slides:v1` rev 20260914 `batch`; `rootUrl`s as the table says. `from_google_discovery` yields 168 operations over the five, and no method's `path` equals a `batchPath` value. The five total 1,330,060 bytes, the largest (Sheets) 378,032. All of that matches the body and the comments.

**The mutations.** I applied each of the twelve reversions the first checkbox lists to this worktree and ran `tests/test_fidelity.py` after each: segment count 6, `extra_batch_route` loop 6, gap keyed on the document 5, `batch_path` not canonical 1, identity by URL 5, each `document_id` fallback 1, `google_drive` losing `batch_mount` 2, naming an unserved route 3, `/batch` back in `UNCOMPARED` 1, the guard never running 1, the batch check calling `fetch_json` itself 1. Every count is the body's.

**Where the body drifts from what I measured.**

- "returns a document byte-identical to the one `gmail.googleapis.com/$discovery/rest?version=v1` answers": not byte-identical. Both answers are 217,687 bytes, revision 20260907, and `json.loads` of the two compares equal, but the key order differs (a line diff of the two texts, 4,524 lines each, changes 7,506 lines). The docstring on `document_id` says it the right way, "both answer the document that calls itself `gmail:v1`"; the body should say the same.
- "seven `values.batch*` methods on Sheets": six are under `values` (`batchGet`, `batchGetByDataFilter`, `batchUpdate`, `batchUpdateByDataFilter`, `batchClear`, `batchClearByDataFilter`); the seventh is `spreadsheets.batchUpdate`. `docs/fidelity.md` says "seven `batch*` methods", which is right.
- "nine do": copying this branch's `tests/test_fidelity.py` onto a worktree at `1a626040` fails ten (the eight parametrized cases, the read-once test, and `test_a_google_source_is_compared_against_every_api_it_is_served_through`).
- `## Verification` has no `At <head sha>, on main at <base sha>.` line, and its 3283 was counted before the merge commit; at `0d88a35` the suite is 3303.

**One design question, not a request.** `if self.batch_mount:` makes the field's default a silent opt-out: a `GoogleDiscoveryComparison` constructed without `batch_mount` reports nothing about `batchPath`, while `batch_divergences((), docs)` on the same document reports `missing_batch_path`. I checked with a made-up `calendar:v3` declaring `batch/calendar/v3`: the comparison's `divergences()` returns no batch finding. So "a sixth Google API cannot arrive with its `batchPath` unchecked" holds when it arrives as a `Spec` of `gmail` or `google_drive`, and not when it arrives as a third Google source. Cheapest closure is a test that every `GoogleDiscoveryComparison` in `COMPARISONS` has a non-empty `batch_mount`; `test_a_google_source_is_compared_against_every_api_it_is_served_through` pins the two that exist today and says nothing about a third. Fine as a follow-up.

docs/fidelity.md
```diff
@@ -136,6 +136,48 @@ children are what keep the two listings apart: `?list-type=2` and a bare bucket
 Requests are signed with [`backlot.sigv4`](../backlot/sigv4.py) — the module that verifies them —
 so the probe adds no dependency and a change to signing breaks both sides at once.
 
+## Google's batch endpoint is a field, not an operation
+
+Every discovery document names a batch endpoint in its top-level `batchPath`, and none declares it
```
nuri-yoo: This section says it, but the overview at lines 18–19 still states the rule without its exception: "the request surface — which operations exist, and which query parameters each accepts." After this PR a Google document is also read for `batchPath`, which is neither an operation nor a parameter, so a reader of the overview learns the rule as if it had none. One clause up there would do it — "and, for Google, the batch endpoint its `batchPath` names". Anchored here because those lines are outside the diff.

nuri-yoo (approved):
Second pass, approving again. I went back over what the first review did not cover: every place in the repository that enumerates finding kinds or coverage buckets (`operations.py`, `graphql_diff.py`, `s3_probe.py`, README, AGENTS.md, CONTRIBUTING.md, `fidelity.yml`), the `ProbeComparison.mount` analogy the new comment leans on, the document order `_batch_docs` assumes against `google_drive`'s `specs`, and how a `Finding.path` carrying a space and a colon (`drive:v3 batch/v3`) survives the CLI renderer, the JSON emitter and the baseline key. None of those needs a change: the kind enumerations are each comparison's own, the analogy holds (both answer "which served paths does this comparison speak for", and only the consumer differs), the order is drive, docs, sheets, slides on both sides, and the path is never split.

Two things did turn up, both small.

**The overview in `docs/fidelity.md` still states the rule without its exception** — inline below, anchored on the new section because the overview lines are outside the diff.

**#200 stays open on merge.** "Supersedes #200" is not a closing keyword, and #200 currently carries only merge commits on top of the approach this PR replaces. Worth closing it by hand when this lands, so #171 does not have two pull requests open against it.

khj809:
Thanks — all of it is addressed at `1cca62f`, and I reproduced each measurement before changing anything rather than taking the correction on trust.

**Where the body drifted.** All four were mine, and the body now says what I measure.

- *byte-identical* — it is not. The two URLs answer the same 217,687 bytes at revision 20260907 and `json.loads` of the two compares equal, but `list(...)` of the two does not: the key order differs. The body now says "both answer the document that calls itself `gmail:v1` — same revision, same size, equal once parsed, differing only in key order", which is the docstring's phrasing.
- *seven `values.batch*` on Sheets* — seven `batch*`, six of them under `values`, the seventh `spreadsheets.batchUpdate`. Corrected, and folded into the full count below.
- *nine* — ten. Copying this branch's `tests/test_fidelity.py` onto a detached worktree at `1a62604` fails the eight parametrized cases, the read-once test, and `test_a_google_source_is_compared_against_every_api_it_is_served_through`. I folded the old standalone identity test into the parametrize after first writing that number and did not re-count. The checkbox now says ten and names the base commit rather than "main".
- *no `At <sha>` line, and 3283 counted pre-merge* — both fixed; `## Verification` opens with `At 1cca62f, on main at 1a62604.`

One thing I could not reconcile: at `0d88a35` I get **3302 passed, 1 skipped**, not 3303, on `uv sync --all-extras --locked` with the same single skip (the HubSpot reader). Two runs, same number. I left my own figure in the body rather than yours; if you still see 3303 it is worth finding the one test we disagree about before this merges.

**The two documentation notes.** Both were right.

- `docs/fidelity.md:159` listed three of the methods as if it were the list. Walking `resources` on the five today gives eleven: Gmail `messages.batchModify` and `messages.batchDelete`, Docs `documents.batchUpdate`, Sheets seven, Slides `presentations.batchUpdate`, Drive none. The page now names all of them and says Drive has none, which is the fact that makes the point.
- The overview at lines 18–19 now carries the exception — "which operations exist, which query parameters each accepts, and, for Google, the batch endpoint its `batchPath` names" — so the rule is not stated without it 120 lines above the section that qualifies it.

**The positional reference.** Fixed in both places: `batch_divergences`'s docstring and the module docstring said "the path diff above" when `divergences` is below it at line 185 and only `from_google_discovery` is above. Both now say "the path diff", which reads true from either end.

**`batch_mount`.** This is the one where I went a different way than the obvious fix, so it is worth saying why.

The first thing I tried was removing the default, so a `GoogleDiscoveryComparison` cannot be constructed without naming its batch routes. It half works. Measured:

- `credentials` carries a default, so a no-default `batch_mount` has to be `field(kw_only=True)` or move ahead of it.
- There are three construction sites. Two are the registry; the third is `test_a_finding_two_documents_both_declare_is_recorded_once`, which builds either class through `getattr(comparisons, registry)(...)`, so it needs a conditional kwarg for the Google half alone. With that, the suite is green.
- But it refuses only an *omission*. `batch_mount=()` passed explicitly still skips the whole check, because the guard is `if self.batch_mount:`.

So what landed instead is your suggestion, as one assertion inside `test_every_mount_selects_something_backlot_serves` — the test that already walks every comparison and already reads `batch_mount`:

```python
if isinstance(comparison, comparisons.GoogleDiscoveryComparison):
    assert comparison.batch_mount, f"{name}: a Google source names no batch route"
```

It refuses both cases — I checked by deleting gmail's `batch_mount` and, separately, by setting it to `()`; two tests go red each time — and it guards the registry, which is what actually ships and what the nightly runs, rather than guarding construction. It also costs no `kw_only` on the dataclass and no branch in a fixture shared with `OpenAPIComparison`.

What it does not settle is your sharper point: `()` produces *silence* where `batch_divergences((), docs)` on the same documents produces a `missing_batch_path` per declared value. Silence is the wrong answer — a source Backlot serves no batch route for is a gap, acknowledgeable like any other, not an absence of a question. The fix is to drop the `if self.batch_mount:` guard entirely, and that is more than a one-liner: the fixture above patches `google_discovery_diff.divergences` wholesale, so the run's memo never fills and the unguarded call raises `KeyError: 'https://a.invalid/s.json'`. That wants the fixture rethought, which I would rather do on its own than inside this PR. I will open it as a follow-up unless you would rather it blocked here.

**#200** is closed, with a comment pointing at this one. #171 stays open until this lands.

Issue #263, opened 2026-09-17
gen_docs.py renders from the installed backlot and writes to its own checkout, so in a git worktree it silently rewrites the generated blocks from another revision


nuri-yoo:
**What you ran** — the documented regeneration command, from a `git worktree` of this repo, with the package installed editable from the primary checkout. The command is given the same way in `CONTRIBUTING.md:147`, `AGENTS.md:40`, `skills/backlot/SKILL.md:180` and `.claude/skills/backlot-loop/SKILL.md:190`.

```bash
git worktree add ../wt <branch>
cd ../wt
python scripts/gen_docs.py
```

**What happened** — exit 0, two lines of output, and both generated blocks rewritten from the other checkout's code. On a worktree of the branch behind #245, while the primary checkout sat on an unrelated branch:

```
wrote docs/supported-sources.md
wrote skills/backlot/SKILL.md
exit 0
```

```diff
-| `confluence`   | … | 9  | … A Confluence page or blogpost.
+| `confluence`   | … | 10 | … A Confluence page or blogpost.
-| `github`       | … | 33 | … A GitHub issue, pull request, file, or the repository itself.
+| `github`       | … | 31 | … A GitHub issue or pull request.
-| `google_drive` | … | 13 | … A Google Drive file.
+| `google_drive` | … | 11 | … A Google Drive file.
```

The confluence row is that branch's own change undone. The github and google_drive rows are older still: counts and a record summary belonging to whatever revision the primary checkout happens to be on. Neither of the two output lines says a table was rewritten with content the current tree never produced.

**Where it is** — `scripts/gen_docs.py:42` fixes the output at the tree the file lives in, `REPO = Path(__file__).resolve().parent.parent`, while the input comes from `from backlot.main import app` at line 38, which `sys.path` resolves. Run as a script, `sys.path[0]` is `<tree>/scripts`, which holds no `backlot` package, so the import falls through to the installed one:

```
sys.path[0] = /…/wt/scripts
backlot     = /Users/…/backlot/backlot/__init__.py
cwd         = /…/wt
```

The rendering and the file it is written to therefore come from two different revisions. This is not a question of which branch the primary checkout is on — the two revisions being identical is the only case that is right, so a worktree of `origin/main` is wrong in the same way while the primary checkout is elsewhere.

[...]

PR #284, opened 2026-09-18
jira: a JSON body names the charset real sends, which Confluence and the gateway's 403 do not (#208, #225)

### LINKED ISSUE #208: jira: every response carries application/json where real carries application/json;charset=UTF-8, which Confluence does not
**Source and commit** — `jira`, at `4c4a268`.

**How you know** — live calls to `brekkylab.atlassian.net` on 2026-09-14, over every Jira route Backlot serves and the Confluence routes beside them.

| # | Divergence | Silent? |
|---|---|---|
| 1 | a Jira response carries `application/json` where real carries `application/json;charset=UTF-8` | yes |

## Jira's `content-type` carries a charset

**Request** — any Jira route.

**Backlot serves** — `application/json`, FastAPI's `JSONResponse` default.

**The real API serves** — `application/json;charset=UTF-8`, on all nine routes Backlot serves (`serverInfo` at both versions, `project/search`, `field`, `issueLinkType`, `search/jql`, `issue/{key}`, `issue/{key}/comment`) and on the statuses beside their 200s — the 404 for an issue that does not exist, and the 200s answered with no credential at all. The one Jira body measured NOT to carry it is the type-conversion 400, which is `application/problem+json;charset=UTF-8` instead — a different type, same charset, and served already since #180.

**Confluence does not share it**, which is what makes this Jira's rather than Atlassian's: `/wiki/rest/api/space`, `/wiki/rest/api/content` and the space 404 are all a bare `application/json`, matching what Backlot already sends. So the fix is per-product, not per-prefix.

**Can a client tell?** Not from the body — JSON is UTF-8 by definition and no byte changes. What it changes is the header a client or a recorded fixture compares as a string, and a contract test asserting real's `content-type` passes against Jira and fails here.

**Reproduction** — any corpus with a Jira issue; `backlot serve`, then read the `content-type` of any `/atlassian/rest/api/3/…` response.

## Closing it

`backlot/errors/__init__.py` already has the hook: `json_media_type(path, status_code)`, which GitHub implements (`backlot/errors/github.py`) and the `vendor_json_media_type` middleware applies. `backlot/errors/atlassian.py` does not, so it returns the default for every path. The work is that function plus the Jira/Confluence split above.

khj809:
Closes #208. Closes #225. Supersedes #247.

## What changed

Every JSON body Backlot serves on a Jira path carries `application/json;charset=UTF-8`, the exact spelling real uses: no space after the semicolon, `UTF-8` upper-case, which is not GitHub's `application/json; charset=utf-8`. Two Jira answers stay as they were, because real sends them under another type: the gateway's 403 for a `Bearer` it cannot read as a Connect token is the bare `application/json`, and the RFC 7807 refusals (type-conversion 400, 405, 415) are `application/problem+json;charset=UTF-8`. Every Confluence body under `/wiki` stays the bare `application/json`.

The rule lives in `backlot.errors.atlassian.json_media_type`, the hook `backlot.errors.json_media_type` already dispatches to and `main.vendor_json_media_type` already applies: the charset for a Jira path, `None` for a `/wiki` path or a 403. The 403 is selected by status alone because the Connect-token refusal is the only 403 this server answers on a Jira path. The middleware rewrites only a response that is exactly `application/json`, so the RFC 7807 bodies on `AtlassianError` are untouched, and the OpenAPI document still keys the JSON body as `application/json`, as Jira's own description does, so `backlot diff` is unchanged. `docs/supported-sources.md` states the Jira rule with its two exceptions under Jira and the bare type under Confluence.

Measured on brekkylab.atlassian.net with an API token, 2026-09-15 and again 2026-09-18:

| Response | Real `content-type` |
|---|---|
| Jira 200 (`serverInfo` with and without a credential, `issueLinkType`, `project/search`, `field`), 404 (`issue/NOPE-999999` under `/2` and `/3`), 400 (`comment?orderBy=bogus`, an unparseable JQL) | `application/json;charset=UTF-8` |
| Jira 403 for `Bearer usr-nope` on `serverInfo`, `field`, `project/search` | `application/json` |
| Jira `search/jql?maxResults=abc` 400, `POST serverInfo` 405 | `application/problem+json;charset=UTF-8` |
| Confluence 200 (`space`, `search`), 404 (`space/{key}`, `content/{id}`), 400 (`space?limit=abc`), 403 (no credential), 405 | `application/json` |
| `api.atlassian.com/ex/jira/{cloudId}/rest/api/3/…` 200 and 404 | `application/json;charset=UTF-8` |

Before this change Backlot answered the bare `application/json` on every row of the first and last groups, so a client decoding by the header's charset, or a fixture comparing the header as a string, saw a value real Jira never sends.

## Verification

- **Tests** — `pytest` on a `.[all]` install: 3294 passed, 1 skipped (`tests/test_llamaindex.py:127`, `llama_index.readers.hubspot` is not installed; unrelated to this change).
- **Optional surfaces that ran rather than skipped** — mcp, official-sdk, llamaindex (all but the HubSpot reader), mirage, fsspec, S3.

- [x] A test fails without this change — `test_jira_json_carries_the_charset_real_sends_and_confluence_does_not` fails on `main` at its first row, and the GitHub and Google cross-vendor assertions that pinned Jira to the bare type fail with it.
- [ ] A new endpoint serving corpus content is ACL-scoped, proved for both an admin and a scoped token — no new endpoint.


tests/test_atlassian.py
```diff
@@ -1282,6 +1282,44 @@ def test_confluence_refuses_an_integer_parameter_it_cannot_convert(client, admin
     }
 
 
+def test_jira_json_carries_the_charset_real_sends_and_confluence_does_not(client, admin_h):
+    """Measured on a live Atlassian Cloud site, 2026-09-15 and 2026-09-18: Jira answers
+    `JIRA_JSON_MEDIA_TYPE` on its 200s, a 404 under either mount and a plain 400, and the bare
+    `application/json` on the one 403 this server answers, the gateway's for a bearer it cannot read.
+    Confluence answers the bare type on every route served. The RFC 7807 refusals keep
+    `application/problem+json`, which the middleware never touches."""
+    jira, bare = errors_atlassian.JIRA_JSON_MEDIA_TYPE, "application/json"
```
nuri-yoo: **[note]** This test reads the spelling from `JIRA_JSON_MEDIA_TYPE`, so it cannot catch the constant being reformatted into GitHub's `application/json; charset=utf-8` — mutation-checked: with the constant spelled that way this test passes and only the cross-vendor asides in `test_github.py` (line 3542) and `test_google.py` (line 1615) fail. #225 asked for the exact spelling pinned here ("since it is unusual enough that a reformat would fix it into GitHub's"), and this is the file a reader of the Jira rule opens.

```suggestion
    jira, bare = "application/json;charset=UTF-8", "application/json"
    assert errors_atlassian.JIRA_JSON_MEDIA_TYPE == jira
```

backlot/errors/atlassian.py
```diff
@@ -61,6 +61,33 @@ def is_confluence(path: str) -> bool:
...
+
+def json_media_type(path: str, status_code: int) -> str | None:
+    """The `content-type` real puts on a JSON body answered at ``path`` with ``status_code``, or
+    ``None`` to keep FastAPI's bare `application/json`.
+
+    Jira names the charset above on every JSON status measured but one: the gateway's 403 for a
+    bearer it cannot read as a Connect token (:func:`connect_token_body`) is the bare type, measured
+    2026-09-16 and 2026-09-18 on `serverInfo`, `field` and `project/search`. That 403 is the only one
+    this server answers on a Jira path, so the status alone selects it. Confluence answers the bare
+    type on every route this server serves (measured 2026-09-15 and 2026-09-18), so `/wiki` is
```
nuri-yoo: **[note]** "the bare type on every route this server serves" is a claim about routes where the measurement is about statuses: on the same routes real Confluence answers the 401 for an unparseable Basic as `text/html;charset=utf-8` (Tomcat's page, as the `CONFLUENCE_UNAUTHORIZED` comment above says), measured today, and Backlot serves that 401 as the bare JSON envelope. The docs paragraph gets this right by naming the statuses ("the 200s, the 404s, the 400, 403 and 405 measured"); this sentence and the test docstring at `tests/test_atlassian.py:1289` ("on every route served") do not. Something like: "Confluence answers the bare type on every JSON body measured — the 200s, the 404s, the 400, 403 and 405 (2026-09-15 and 2026-09-18); its 401 is Tomcat's HTML page, which the envelope here does not reproduce."

backlot/main.py
```diff
@@ -443,8 +443,9 @@ async def vendor_json_media_type(request: Request, call_next):
     """Put the vendor's own `content-type` on a JSON body, where one is measured.
 
     Real GitHub answers `application/json; charset=utf-8` on every JSON response but code search's
-    (see ``backlot.errors.github.json_media_type``); FastAPI's ``JSONResponse`` answers
-    `application/json`. A middleware rather than a ``default_response_class`` on the router, because
+    (see ``backlot.errors.github.json_media_type``) and Jira `application/json;charset=UTF-8` on all
+    but its gateway's 403 (``backlot.errors.atlassian.json_media_type``); FastAPI's ``JSONResponse``
+    answers `application/json`. A middleware rather than a ``default_response_class`` on the router, because
```
nuri-yoo: **[note]** The reflow left this line at 108 characters where the rest of the docstring wraps at 100 (ruff does not check docstring width here, so CI is quiet). Also the sentence now says Jira carries the charset "on all but its gateway's 403", while `errors.atlassian.json_media_type`'s docstring counts the RFC 7807 trio as the other exceptions; from this middleware's seat they never arrive as `application/json`, so it is true as read, but "on every `application/json` body but its gateway's 403" would say why.

nuri-yoo (approved):
Reviewed at `7b74046` in its own worktree, on `main` at `1a62604`. Every content-type the body and the docstrings claim reproduces against brekkylab.atlassian.net today, the rule holds on the Jira answers the test does not visit, and reverting each half of the predicate turns the new test red. Nothing here blocks; four notes inline and in this body, one of them a body edit worth making before merge.

**Re-measured on real, 2026-09-18.** Every row in the body's table came back as stated. Beyond the table, the Jira answers Backlot serves that the new test does not visit are `application/json;charset=UTF-8` on real and on this head alike: the 200s of `field`, `issueLinkType` and a bounded `search/jql`; the 404 of `project/{key}/role` for an unknown key; the 400s of an unbounded JQL, a bad `nextPageToken`, and the three `body_not_read` sentences; and the 403 for `Bearer usr-nope` on `issue/{key}` is bare like its three measured siblings. The 415 and the 405 are `application/problem+json;charset=UTF-8` on both.

```
real                                             head
200 field / issueLinkType / search/jql (bounded)  application/json;charset=UTF-8   same
404 project/NOPEXX/role                           application/json;charset=UTF-8   same
400 unbounded jql, bad nextPageToken              application/json;charset=UTF-8   same
400 POST body "", "[]", "{"                       application/json;charset=UTF-8   same
415 POST text/plain, 405 POST serverInfo          application/problem+json;charset=UTF-8   same
403 Bearer usr-nope on issue/NOPE-999999          application/json                 same
```

**Mutations**, each from a copy and restored, one node id per run, caught means exit 1:

```
                                        test_jira_json_carries…  conf int-param  conf 405  gh cross-vendor  google callback
drop `or status_code == 403`            caught                   —               —         —                —
drop `is_confluence(path) or`           caught                   caught          caught    —                —
spell constant "application/json; charset=utf-8"   survived      —               —         caught           caught
return None always (main's behaviour)   caught                   —               —         caught           caught
```

The third row is the inline note on the test: the Atlassian test reads the spelling from the constant it is meant to pin, so a reformat to GitHub's spelling is caught only by the two cross-vendor asides in `test_github.py` and `test_google.py` (the whole suite under that mutation: 2 failed, 3292 passed).

**Suite** on this head in a worktree, `.[all]` env: 3294 passed, 1 failed — `test_generated_docs_are_current`, which is the worktree-imports-the-editable-install trap of #263 (with `PYTHONPATH=$PWD`, `gen_docs.py --check` exits 0). `ruff check` and `ruff format --check` clean.

**Body.** #208 (2026-09-14) is the same gap as #225, measured a day earlier over the routes Backlot serves; this PR closes only #225. Please add `Closes #208.` so the merge closes both, and the title then takes the `(#208, #225)` suffix the repo uses for a PR closing two. #247, which this supersedes, is still open under its author's name — worth closing when this merges, with a pointer here.

**Outside this PR, two things the measurement turned up.** A Jira path no route matches (`/rest/api/3/nopesuchroute`) is RFC 7807 on real, `application/problem+json;charset=UTF-8` with `{"type":"about:blank","title":"Not Found","status":404,"detail":"No endpoint …"}`; Backlot answers the shared `statusCode` envelope, now under `application/json;charset=UTF-8`. The Confluence counterpart (`/wiki/rest/api/nopesuchroute`) is `application/xml` on real. Both are body gaps this PR neither opened nor claims; filed as #285. And the 401 that #225 lists as `application/json;charset=UTF-8` — `rest/api/3/myself` with no credential — is `text/html;charset=UTF-8` today, so the body's table is right to leave it out.

khj809:
Thanks for the re-measurement and the mutation table. All four notes are in `10f64c3`:

- The Atlassian test now pins `application/json;charset=UTF-8` as a literal and asserts the constant equals it, so a reformat into GitHub's spelling fails in this file, not only in the two cross-vendor asides.
- The Confluence sentence in `json_media_type`'s docstring and the test docstring names the statuses measured (200, 404, 400, 403, 405) rather than "every route", and notes that its 401 is Tomcat's HTML page, which the envelope does not reproduce.
- The middleware docstring in `main.py` is rewrapped at 100 and says "every `application/json` body but its gateway's 403".
- The body now closes #208 as well as #225, and the title carries the `(#208, #225)` suffix.

PR #286, opened 2026-09-18
jira: search/jql refuses a maxResults outside 1-5000, a body field the vendor cannot coerce, and an unknown body field, the way real does

### LINKED ISSUE #212: jira: search/jql serves a maxResults outside 1-5000, a body field it cannot coerce, and an unknown body field, where real refuses all three

brekkybot:
Closes #212.

## What changed

`search/jql` (`GET` and `POST`, `/rest/api/2` and `/rest/api/3`) now refuses three requests it used to serve a `200` to. A `maxResults` outside 1-5000 is refused on both methods — where a negative value used to reach SQLite as an unclamped `LIMIT` and serve the whole visible collection. A `POST` body field Jackson will not coerce into `maxResults` (a non-numeral string, a boolean) is refused with the same body-wide sentence a body it cannot deserialize at all gets, rather than falling back to a default or reading `int(True)` as `1`. A `POST` body carrying a property the vendor's fixed bean does not declare — an unknown field, or `jql` or `nextPageToken` shaped as a list or an object — is refused the same way, rather than the property being silently ignored.

The five fields real's bean also takes but Backlot does not act on (`fields`, `fieldsByKeys`, `expand`, `properties`, `reconcileIssues`) are kept off that unknown-property refusal, so a client using them is not newly broken by it.

The range check runs after the page-token decode and before the unbounded-JQL refusal, matching the precedence measured on real for a request that is wrong on more than one count at a time. It applies only to a `maxResults` the caller actually sent, on either method: Backlot's own `default_page_size` is never a value the vendor validated, so a deployment that misconfigures it outside 1-5000 must not turn a request that sent no `maxResults` at all into a refusal.

All measured 2026-09-18 against `brekkylab.atlassian.net`, live calls with `email:api_token` Basic auth, on a project with four issues (`jql=project = SCRUM`).

**1. `maxResults` outside 1-5000**

| Request | Real | Backlot before |
|---|---|---|
| `?maxResults=0` / `-1` / `5001` (GET) | `400`, `{"errorMessages":["…1과…5,000…"],"errors":{}}` | `200` |
| `{"maxResults": 0 \| -1 \| 5001}` (POST) | same `400` | `200` |
| `?maxResults=1` / `5000` | `200` | `200` |
| `{"maxResults": null}` | same `400` (Jackson reads a null int field as `0`, which is not absence — a request that omits the key entirely is unaffected) | `200`, default page size |

```
$ curl -u $EMAIL:$TOKEN '.../rest/api/3/search/jql?jql=project+%3D+SCRUM&maxResults=-1'
{"errorMessages":["최대 결과 매개 변수는 1에서 5,000 사이여야 합니다."],"errors":{}}
```
The sentence real gives is localised to the account's language, so it is not reproduced verbatim — Backlot states the rule in English, the same choice already made for the neighbouring `unbounded_jql` and `bad_page_token` refusals.

Precedence, also measured: a bad `nextPageToken` wins over an out-of-range `maxResults`; an out-of-range `maxResults` wins over the unbounded-JQL refusal (no `jql` at all).

**2. A body field the vendor will not coerce**

| Body | Real | Backlot before |
|---|---|---|
| `{"maxResults": "abc"}` | `400 Invalid request payload…` | `200`, default page size |
| `{"maxResults": true}` | same `400` | `200`, one-issue page (`int(True) == 1`) |
| `{"maxResults": "5"}` | `200`, page of 5 | `200`, page of 5 (unchanged) |
| `{"maxResults": 1.5}` | `200`, page of 1 (`int()` truncates) | `200`, page of 1 (unchanged) |

**3. An unknown body field**

| Body | Real | Backlot before |
|---|---|---|
| `{"jql": "…", "bogus": 1}` | `400 Invalid request payload…` | `200`, field ignored |
| `{"jql": "…", "startAt": 0}` | same `400` (`startAt` belongs to the older `search` op) | `200`, field ignored |
| `{"jql": ["order by created"]}` | same `400` | `200`, `str()`-coerced and searched |
| `{"jql": "…", "nextPageToken": [...]}` / `{...}` | same `400` | `200`, `str()`-coerced and answered as a bad page token |
| `{"jql": "…", "fields": [...]}` / `fieldsByKeys` / `expand` / `properties` / `reconcileIssues` | `200` (bean fields Backlot does not act on) | `200` (unchanged — kept off the new refusal) |

`{"jql": 5}` and `{"jql": true}` are coerced to a string and reach real's JQL parser, which then fails on JQL grounds (`JQL 쿼리 오류: …`) rather than the invalid-payload sentence — that is the parser's own leniency `_str_param`'s docstring already notes predates this issue, and is out of scope here. `nextPageToken` splits the same way: a scalar (an int, a bool) coerces to a string and reaches the token decoder instead, drawing the `bad_page_token` refusal rather than the invalid-payload one.

## Verification

- **Tests** — `pytest` on a `uv sync --all-extras --locked` install: full suite green except the two pre-existing `tests/test_mcp.py` failures (`test_mcp_notion_acl_enforced`, `test_mcp_s3_lists_objects`) that also fail on `main` in this sandbox — a `pypi.org` TLS fetch the network policy here does not allow, unrelated to this change.
- **Optional surfaces that ran rather than skipped** — the full `tests/test_atlassian.py` module, all HTTP-level.

- [ ] A new endpoint serving corpus content is ACL-scoped, proved for both an admin and a scoped token — not applicable, no new endpoint


tests/test_atlassian.py
```diff
@@ -1605,6 +1605,123 @@ def test_jira_search_reads_a_null_jql_as_one_that_was_not_sent(client, admin_h):
...
+@pytest.mark.parametrize("value", ["abc", True, False])
+def test_jira_search_post_refuses_a_max_results_the_vendor_will_not_coerce(client, admin_h, value):
+    """Measured 2026-09-18: Jackson refuses a non-numeral string or a boolean with the body-wide
+    sentence, naming no parameter."""
+    r = _search_post(client, admin_h, jql="project = payments", maxResults=value)
+    assert r.status_code == 400, r.text
+    assert r.json() == {"errorMessages": [errors_atlassian.BODY_NOT_AN_OBJECT]}
+
+
+@pytest.mark.parametrize("value,want_first", [("5", True), (1.5, True)])
```
khj809: `want_first` is never read in the body, and the only assertion is `status_code == 200`, so the truncation this test's own docstring claims ("`1.5` truncates to `1`") is not encoded anywhere. Change the float branch of `_jira_search_max_results` to fall back to `default` and this test still passes — which is the regression it exists to catch.

The difference is observable on the fixture corpus. Measured on this branch, `POST /rest/api/3/search/jql` with `jql=project = payments`:

| body `maxResults` | status | `len(issues)` |
|---|---|---|
| `1.5` | 200 | 1 |
| `"5"` | 200 | 3 (the whole project) |

So `assert len(r.json()["issues"]) == 1` for `1.5` pins it. Either give `want_first` that job — an expected page length rather than a bare `True`, which both cases carry identically and nothing reads — or drop the parameter.

backlot/routers/atlassian.py
```diff
@@ -374,6 +374,59 @@ async def jira_project_role(key: str, role_id: int, request: Request):
...
+    is NOT absence here as it is for `jql`'s own null handling in this endpoint: Jackson reads a
+    null int field as `0`, which then fails the range check like any other out-of-bounds value
+    (measured 2026-09-18, `{"maxResults": null}` draws the same refusal as `{"maxResults": 0}`).
+
+    A digit string and a float are read fine — `"5"` as `5`, `1.5` truncated to `1` — where a
+    non-numeral string and a boolean are refused with the body-wide sentence a body Jackson cannot
+    deserialize at all gets (:data:`errors_atlassian.BODY_NOT_AN_OBJECT`): `bool` is checked before
+    `int`/`float` because Python's `int` is their common base class.
+    """
+    if "maxResults" not in body:
```
khj809: An absent `maxResults` returns `default`, and `default` is then subject to the same range check at line 475. `default_page_size` is a plain `int` with no bounds (`backlot/config.py:65`), so a deployment that sets it to `0` or above `5000` answers `400` to every `search/jql` — including requests that send no `maxResults` at all, which real never refuses, since its own default is always in range.

Either exempt the default from the range check (it is Backlot's value, not the caller's, and the docstring here already says it is "never subjected to that range" — the code does subject it), or bound the setting with `Field(ge=1, le=5000)`.

Related: the pagination comment at `backlot/config.py:60` says "the Jira search caps nowhere", whose contrast is that a sent `maxResults` is unbounded there. That stops being true with this PR.

backlot/routers/atlassian.py
```diff
@@ -393,10 +446,22 @@ async def jira_search(request: Request):
...
     if request.method == "POST":
         body = await _jira_search_body(request)
+        if not _JIRA_SEARCH_BODY_KEYS.issuperset(body):
+            # The body deserializes into a fixed bean, and a property it does not declare is a
+            # failure, not surplus (measured 2026-09-18, `{"bogus": 1}` and `{"startAt": 0}` —
+            # `startAt` belongs to the older `search` operation, not this one). Same sentence a
+            # body Jackson cannot deserialize at all gets, below.
+            raise errors_atlassian.body_not_read(errors_atlassian.BODY_NOT_AN_OBJECT)
+        raw_jql = body.get("jql")
+        if isinstance(raw_jql, (list, dict)):
```
khj809: `jql` shaped as a list is refused here, but `nextPageToken` shaped as a list or an object still falls through to `str(token)` at line 472 and comes back as the `bad_page_token` refusal instead. The bean argument that makes `{"jql": [...]}` an invalid payload predicts the same answer for `{"nextPageToken": [...]}` — but that is a prediction, and a measurement never carries to the neighbouring parameter, so please measure it rather than infer it. If it is out of scope for this PR, say so in the body; right now the asymmetry reads as deliberate without saying why.

backlot/errors/atlassian.py
```diff
@@ -249,6 +249,24 @@ def unbounded_jql() -> AtlassianError:
     )
 
 
+def max_results_out_of_range() -> AtlassianError:
+    """Jira's 400 for `search/jql`'s `maxResults` outside 1-5000, on both methods and both
+    placements — the query string and the POST body. Measured 2026-09-18 against
+    brekkylab.atlassian.net: `0`, `-1` and `5001` are all refused, `1` and `5000` both answered
```
khj809: Drop the site name. A measurement comment carries the vendor, the endpoint and the date — `Measured 2026-09-18 against Jira Cloud` — and never the tenant it happened to be measured on. The tenant is an internal account: it tells a reader nothing reproducible (they have no access to it), and it puts our own site's hostname in a published package.

Same edit at `backlot/routers/atlassian.py:378` and `tests/test_atlassian.py:1613`. The review body has where the rule should be written down so this stops recurring.

khj809 (commented):
The measurements and their precedence are well covered — the token/range/unbounded-JQL order is pinned by tests, and keeping the five bean fields Backlot does not act on off the unknown-property refusal is the right call. `pytest -k search` is green and ruff is clean on this branch here too. Five things before this merges; the last one is a repeat offence rather than anything specific to this change.

**`_int`'s docstring now describes a caller that no longer exists** — `backlot/routers/atlassian.py:1690`.

This PR takes `search/jql`'s body off `_int`. Its only remaining callers are Confluence's CQL **query** parameters at `backlot/routers/atlassian.py:1317-1318`, but all three paragraphs of the docstring are about the POST body: "A value already read out of a JSON request body", "what parses the POST `search/jql` body on real is Jackson rather than Spring's parameter binding", and "The 1-5000 range is NOT Jackson's ... so it belongs to the operation rather than to either parser" — reasoning for a range that now lives in `_jira_search_max_results` instead. #212 framed the fix as making `_int` itself strict; a separate function is fine, but then the docstring has to be rewritten around the callers it actually has. The sentence worth keeping (the range belongs to the operation, not to either parser) belongs next to `_JIRA_SEARCH_MAX_RESULTS_RANGE`.

**Where the site-name rule should live.** The inline note below is the third time a measurement comment has named the tenant, and there are 25 such lines already in `backlot/` and `tests/` (`backlot/auth.py`, `backlot/main.py`, `backlot/errors/atlassian.py`, `backlot/routers/atlassian.py`, `tests/test_auth.py`, `tests/test_atlassian.py`), so telling one author once will not hold. Three places make it mechanical:

1. `AGENTS.md`, under "Documentation rules" — one bullet: a measurement comment names the vendor, the endpoint and the date, never the site it was measured on. It has to be that section specifically, because that is the section `.claude/agents/prose-reviewer.md` is told to enforce.
2. `tests/test_docs.py` — the same shape as `_COUNT_RE` / `test_no_prose_states_the_source_count`, over `_scanned_files(".py", ".md")`: a regex for a tenant host with `ecosystem.atlassian.net` and the other vendor-published sample sites allowlisted. That test would fail today on the 25 existing lines, so it lands with the sweep, not before it.
3. `.claude/agents/prose-reviewer.md` — a numbered item under "What you check", so a run catches it at review time rather than at CI time.

Worth an issue for the sweep plus the test; happy to open one.

tests/test_atlassian.py
brekkybot: Fixed — `want_first` is now `want_len` and asserts `len(r.json()["issues"])`: 3 for `"5"` (the whole three-issue project), 1 for `1.5`. Confirmed it fails against the pre-fix assertion (bare `status_code == 200`) by reverting locally and re-running.

---

backlot/errors/atlassian.py
brekkybot: Dropped the tenant at all three sites named in the review (here, `backlot/routers/atlassian.py:378`, `tests/test_atlassian.py:1613`) — each now reads "Measured 2026-09-18 against Jira Cloud". Left the other 25 pre-existing lines alone; filed #287 for the sweep, the `test_docs.py` check and the `prose-reviewer.md` line, per the review body.

---

backlot/routers/atlassian.py
brekkybot: Confirmed and fixed by exempting the default instead of bounding the setting: `maxResults`'s absence (GET: no parameter, or one that's empty/whitespace after Java's own stripping, matching `_int_param`'s own "absent" cases; POST: key not in the body) now skips the range check entirely, so `default_page_size` is never subjected to it. Added `test_jira_search_default_page_size_is_never_checked_against_the_range`, which fails on the pre-fix code with `BACKLOT_DEFAULT_PAGE_SIZE=0`. Also updated the `config.py` comment you flagged as related — it now says the Jira search refuses a sent out-of-range value rather than "caps nowhere".

---

backlot/routers/atlassian.py
brekkybot: Measured rather than inferred: `{"nextPageToken": [...]}` and `{"nextPageToken": {...}}` both draw the same `Invalid request payload…` refusal as a listed `jql` (an int or a bool coerces to a string and reaches the token decoder instead, same as `jql`'s scalar case). Added the same list/dict guard for `nextPageToken` and two tests (`test_jira_search_post_refuses_a_next_page_token_shaped_as_a_list_or_object`).

---

khj809:
All five are addressed, and I verified the two behavioural ones rather than taking the replies at their word — mutating `max_results_sent` out of the range check fails all three parameter cases of the new default test, dropping the `raw_token` half of the shape guard fails both new token cases, and making the float branch return the default fails `test_jira_search_post_coerces_a_max_results_the_vendor_coerces[1.5-1]`. `tests/test_atlassian.py` is 234 passed and ruff is clean on `378037cb` here. Exempting the default rather than bounding the setting is the better of the two options I offered: a value the vendor never validated should not be subjected to the vendor's rule, and it leaves the deployment's own knob alone.

One thing left: **the body no longer describes the finished state.** Two behaviours that exist in the code now are absent from it, so a reader of the PR gets a different endpoint than the one the branch serves.

1. `nextPageToken` shaped as a list or an object is refused. "What changed" lists the third refusal as "a property the vendor's fixed bean does not declare — an unknown field, or `jql` shaped as a list", and the §3 table has a row for `{"jql": ["order by created"]}` but none for the token. The measurement is now a pair, so it should read as one: the paragraph naming both fields, and a table row alongside the `jql` one — `{"jql": "…", "nextPageToken": [...]}` / `{...}` → `400` / `200`, `str()`-coerced to `"[...]"` and refused as a bad page token. The trailing paragraph about `{"jql": 5}` and `{"jql": true}` reaching the JQL parser wants the token's scalar case beside it too, since it is the same split.

2. Backlot's own `default_page_size` is never checked against the range. The precedence paragraph ends at "matching the precedence measured on real for a request that is wrong on more than one count at a time" — one more sentence there covers it: the range applies only to a `maxResults` the caller actually sent, on either method, because the default is Backlot's setting rather than a value real ever validated. Without it, the §1 row `{"maxResults": null}` → `400` reads as though any absence is refused.

Please fold these into the sections that already exist rather than appending a section for this round — the body describes the endpoint, not the passes that produced it.

khj809 (approved):

Issue #254, opened 2026-09-16
github: an authenticated caller's 404 for a path no route matches carries the ratelimit headers and version echo real omits


khj809:
Found while measuring for #250.

**Source and commit** — `github`, `main` at `70a0b68`.

**How you know** — a maintainer's re-measurement against api.github.com on 2026-09-16, with a valid token: `GET /github/nonexistent-route-zz` (a path no route matches at all) answers 404 with neither the five `x-ratelimit-*` headers nor `x-github-api-version-selected`, the same as the trailing-slash 404 #250 fixes; `GET /repos/psf/ghost-zz-9876` (a path whose route DOES match, on a repository that does not exist) answers 404 WITH both. Backlot answers `/github/nonexistent-route-zz` with both today.

**Reproduction** — the bundled corpus, over `backlot.serve()`, with a valid admin token:

```python
import backlot, requests

with backlot.serve() as m:
    h = {"Authorization": f"Bearer {m.token}"}
    r = requests.get(f"{m.base_url}/github/nonexistent-route-zz", headers=h)
    print(r.status_code, sorted(k for k in r.headers if k.lower().startswith("x-ratelimit") or k.lower() == "x-github-api-version-selected"))
```

prints `404 ['x-github-api-version-selected', 'x-ratelimit-limit', 'x-ratelimit-remaining', 'x-ratelimit-reset', 'x-ratelimit-resource', 'x-ratelimit-used']`, where real prints `404 []`.

## 1. An authenticated 404 for a path with no matching route carries headers real omits

**Request** — `GET /github/<path no router declares>`, valid token.

**Backlot serves** — Starlette's default 404, which still runs through `report_github_rate_limit` and `echo_github_api_version` — both middleware that add their headers to any response whose path starts with `/github`, matched route or not.

**The real API serves** — 404 with neither the ratelimit headers nor the version echo, unlike a 404 for a route that DID match on a resource that does not exist (`/repos/{org}/ghost-zz-9876`), which carries both.

**Can a client tell?** Only if it looks for the headers' absence specifically — the status code alone is the same 404 either way, so a client that only checks `status_code` before treating a 404 as "not found" never notices; one that paces by `x-ratelimit-remaining` sees an extra, wrongly-present data point on a request that should not have counted through this path at all.

brekkybot:
loop: claimed by run https://claude.ai/code/session_01TFjWdRRQzEGotmvdQvrLb1

---
[...]

Issue #235, opened 2026-09-16
github: the search windows are an hour where real measures 60 seconds, and six more rate-limit answers differ from real a week after #175


khj809:
Found while re-measuring #168 against `main` after #175 merged.

**Source and commit** — `github`, `main` at `90b90dd` (`report_github_rate_limit` in `backlot/main.py`, `GET /github/rate_limit`).

**How you know** — every cell measured against api.github.com on 2026-09-15, unauthenticated with `curl -4` and authenticated with the `gh` CLI's token, within a few minutes of each other; each Backlot cell measured on `main` the same day. #175's measurements are dated 2026-09-09/10, so a row could be a GitHub-side change in the intervening week rather than a misreading; the window lengths are not that kind of row, and the others are listed so each can be re-measured on its own.

| # | Divergence | Silent? |
|---|---|---|
| 1 | `search` and `code_search` windows are an hour, where real's are 60 seconds | yes |
| 2 | an anonymous caller gets a `code_search` window of its own at limit 10, where real shares `core` at 60 | yes |
| 3 | a bad bearer's 401 carries the five headers and is counted, where real sends none and counts nothing | yes |
| 4 | a path that routes nowhere, asked by a resolved credential, carries the five headers and is counted, where real sends none | yes |
| 5 | `/rate_limit` serialises `rate` before `resources`, where real answers `resources` first | no |
| 6 | the version header on `/rate_limit` applies to every caller, where real reads it only for a credential that resolves | no |
| 7 | an `Authorization` real cannot parse is put on the address's window, where real gives each distinct value a window of its own | yes |

**The real API serves**

`core` is hourly and the two search resources are per minute, measured on each rather than carried from one to the next. Three anonymous `/search/issues` in the same second answered `used` 1, 2, 3 against one `reset` 59 seconds out, and the same request 65 seconds later answered `used: 1` against a fresh window. An authenticated `/search/code` did the same at `limit: 10`. `core`'s `reset` held at 1789487150 across a session spanning several minutes and read 3474 seconds out when `/rate_limit` was asked. A window that has not been opened reports `reset` as now plus its length and moves forward with each read.

[...]

PR #292, opened 2026-09-21
github: the rate-limit answers key on what the request carried, and the search windows measure a minute

khj809:
Finishes #235: rows 1, 2, 3, 4, 6 and 7, measured from a machine outside the session proxy that left them open there. Closes #254, which is row 4 under its own number. Stacked on #289 (row 5); merging this brings that commit with it.

## What changed

`core` keeps its hour and the two search resources get their own minute, so `RATE_LIMITS` carries a window length beside the two limits and `RateLimitWindows` reads it per resource instead of from one constant.

A caller with no credential has no `code_search` window: `rate_limit_window` maps that resource to `core` for them, so `/rate_limit` reports `core`'s four numbers under both names and an anonymous request that counts against `core` moves both.

`serves_rate_limit_headers` is the new gate on the five `x-ratelimit-*` headers, the count and the version echo. It answers no to a credential that arrived and did not resolve, and to a path no route matches asked by any caller that sent an `Authorization` header at all. `report_github_rate_limit` and `echo_github_api_version` ask it, `_matches_a_route` tells it whether a route matched, and `refuse_a_trailing_slash_on_github` asks it in place of the `auth.bearer_token(request) is None` it used to draw the same line with — which gave an unparseable credential headers real does not give it.

`rate_limit_caller` keys an `Authorization` real cannot parse to a window of its own per address, apart from that address's bare-anonymous one and shared by every unparseable value from it.

`honours_api_version` answers no for `/rate_limit` asked with no `Authorization` header, which takes that request out of the version 400, out of the echo and onto `2022-11-28`'s payload; `_version` returns the default wherever real does not read the header.

`/rate_limit` lists `resources` in the order real lists them in, which differs between the two callers.

Measured against api.github.com on 2026-09-21, anonymously with no `Authorization` header, with a token, with a bad bearer under both schemes, and with unparseable values (`Basic …`, `Digest …`, scheme-less), each on cache-busted urls — an anonymous public read served from the edge otherwise answers a stale body, which is what made an anonymous version 400 look like a 200 on the first pass.

| # | Request | Backlot (before) | Real |
|---|---|---|---|
| 1 | `search` / `code_search` window | `reset` 3600s out | `reset` 60s out; `used` 1, 2, 3 in one window and 1 again 65s later |
| 2 | anonymous `/rate_limit` `code_search` | its own window at `limit: 10` | `core`'s four numbers, `limit: 60`, and one anonymous core call moves both |
| 3 | bad bearer, `Bearer` or `token` | five headers + echo, counted | no `x-ratelimit-*`, no echo, `core` `used: 13` before and after |
| 4 | `/zzz-unrouted-path` with a token | five headers + echo | no `x-ratelimit-*`, no echo; the same path with no `Authorization` carries them and counts (`used` 32, 33, 34) |
| 6 | anonymous `/rate_limit` pinning `2026-03-10` / `1999-01-01` | `resources` alone + echo / 400 | `resources` and `rate`, no echo / 200 — where a token and an unparseable `Basic` both get `resources` alone + echo / 400 |
| 7 | `Basic Zm9vOmJhcg==` on a repo | counted with the address's anonymous window | its own window: `used` 10–15 across two values real had not seen, while bare anonymous calls interleaved read 23, 24, 25 |

Two of these correct what #235 recorded. Its row 7 reads "each distinct value its own window"; alternating two fresh values ran one counter, 10, 11, 12, 13, 14, 15, across `Basic`, `Digest` and scheme-less alike, so what a value buys is separation from the address's anonymous window, not a window per value. Its row 6 says the version header is read for "a credential that resolves"; an unparseable `Basic` gets it read too, so the line is the header's presence, and the same line decides the five on an unrouted 404.

Real lists `resources` as `code_search`, `core`, `graphql`, `integration_manifest`, `search` for a caller with no credential and `core`, `search`, …, `code_search` for a token, so the two answers differ in order as well as in membership.

## Verification

- **Tests** — `pytest -q tests/test_github.py`: passes. The full suite passes but for `tests/test_docs.py::test_no_prose_states_the_source_count` and `::test_no_prose_states_the_bundled_corpus_size`, which fail on this checkout with the change reverted too: they scan the working tree and pick up untracked local files.
- **Without the change** — the rate-limit test fails on the parent commit at the first search-window assertion.

- [ ] A new endpoint serving corpus content is ACL-scoped, proved for both an admin and a scoped token


khj809:
Consolidated into #289, which now carries every row of #235 rather than row 5 alone.

